PR #19, opened 2022-07-21
Restablish last roi extractors pull requests

h-mayorquin:
This should re-esablish two PRs already merged that were lost in the transition from nwb-conversion-tools (https://github.com/catalystneuro/nwb-conversion-tools/pull/601 and https://github.com/catalystneuro/nwb-conversion-tools/pull/602)


src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -25,7 +26,7 @@
 from hdmf.backends.hdf5.h5_utils import H5DataIO
 
 from ..nwb_helpers import get_default_nwbfile_metadata, make_nwbfile_from_metadata, make_or_load_nwbfile
-from ...utils import (
+from neuroconv.utils import (
```
CodyCBakerPhD, line 29: Always use relative paths for internal source code, please

h-mayorquin: Isn't the standard recommendation the opposite?

h-mayorquin: Yes, this is where I read it:

https://peps.python.org/pep-0008/

And I have been wondering for a while why we don't follow this. Where are you getting your convention from?

CodyCBakerPhD: That PEP8 is a bit outdated, it was a big problem in earlier Python versions due to bugginess of relative imports. See https://bugs.python.org/issue25294 for extended discussion and cases where (with more modern version of Python) absolute imports can cause more issues than relative ones. PEP8 also advises not having complex layouts, but there's not much we can do about that when the package itself seeks to solve a complex series of tasks.

My particular view (further reinforced by the difficulty in adjusting the modularity of the previous package as well as porting the project here) is that relative paths that reference the internal source structure are much safer and easier to deal with since they only have to worry about their local place in the structure instead of dealing with the globally imported namespace. This is especially the case for various utility/tool functions that don't have outermost import exposure.

cc @bendichter

CodyCBakerPhD: Also as with most of these types of conventions, it's also worth mimicking the standard set by the rest of our ecosystem - HDMF/PyNWB/DANDI all seem to use the relative import standard like us

h-mayorquin: I don't think that the bug that you point out in the link shows that the situation has changed. In fact, as some answer points out, OP seems confused about the meaning of relative vs absolute imports. See the answer by Brett Cannon.

How does the linked thread shows that "absolute imports can cause more issues than relative ones"?

bendichter: From PEP8:

explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose

bendichter: more to come...

bendichter: There types of style questions are tough, and you are right to ask them @h-mayorquin, since we have no official style guide (maybe we should). When we were coming up with the structure of this library, I wanted to model our styling conventions after similar python projects. sklearn seemed to me to be the closest match for a very established python project that was doing something similar to what we are doing here: creating a standard interface for a variety of classes that all share a few common methods. These classes are meant to be combinable and it should be possible for a user to implement their own custom version and integrate it with the rest of the system. I have always felt that sklearn is very welcoming and well documented, and successfully makes the task of employing and comparing sophisticated machine learning algorithms much less intimidating. We have very similar goals for neuroconv! So, I decided to adopt the [sklearn coding style](https://scikit-learn.org/stable/developers/develop.html#coding-guidelines). Here are the points about imports:

> * Use relative imports for references inside scikit-learn.
> * Unit tests are an exception to the previous rule; they should use absolute imports, exactly as client code would. A corollary is that, if sklearn.foo exports a class or function that is implemented in sklearn.foo.bar.baz, the test should import it from sklearn.foo.
> * Please don’t use import * in any case. It is considered harmful by the [official Python recommendations](https://docs.python.org/3.1/howto/doanddont.html#at-module-level). It makes the code harder to read as the origin of symbols is no longer explicitly referenced, but most important, it prevents using a static analysis tool like [pyflakes](https://divmod.readthedocs.io/en/latest/products/pyflakes.html) to automatically find bugs in scikit-learn.

Now, I will admit, if I had chosen another popular library, like scipy, they don't seem to have a set convention here. I am able to find both absolute and relative imports. But I think what we are building is much closer to a sklearn than it is to a scipy.

I apologize for having these style rules without explicitly stating the motivation.

h-mayorquin: Thanks for taking the time to answer in such a conciliatory and detailed manner. To me, "this is our local convention" is a fine answer. I just wanted to known if the convention had a reason behind it -in this case being consistent with sklearn is the reason it seems- as I read the opposite recommendation from the PEP8 at some point. All of thise are guidelines at the end.

Furthermore, I was thinking along the lines of the second point in the quoted sklearn guidance. As relative imports imply that you used your code "as it is in your repo" and not as you install it. Should we also follow that point and avoid relative imports in unit tests?

h-mayorquin:
OK, I changed the imports to relative and also added the changes in the `CHANGELOG.md`

PR #24, opened 2022-07-21
Add deep lab cut interface

h-mayorquin:
Porting from:
https://github.com/catalystneuro/nwb-conversion-tools/pull/605

### DIFF
--- a/docs/conversion_examples_gallery/behavior/deeplabcut.rst
+++ b/docs/conversion_examples_gallery/behavior/deeplabcut.rst
@@ -1,28 +1,28 @@
-HDF5 data conversion
-^^^^^^^^^^^^^^^^^^^^
+DeepLabCut data conversion
+^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Convert HDF5 imaging data to NWB using :py:class:`~nwb_conversion_tools.datainterfaces.ophys.hdf5.hdf5datainterface.Hdf5ImagingInterface`.
+Convert DeepLabCut imaging data to NWB using :py:class:`~nwb_conversion_tools.datainterfaces.behavior.deeplabcut.deeplabcutdatainterface.DeepLabCutInterface`.
 
 .. code-block:: python
 
     >>> from datetime import datetime
     >>> from dateutil import tz
     >>> from pathlib import Path
-    >>> from neuroconv import Hdf5ImagingInterface
+    >>> from nwb_conversion_tools import DeepLabCutInterface
     >>>
-    >>> file_path = OPHYS_DATA_PATH / "imaging_datasets" / "hdf5" / "demoMovie.hdf5"
-    >>> interface = Hdf5ImagingInterface(file_path=file_path, verbose=False)
+    >>> file_path = BEHAVIOR_DATA_PATH / "DLC" / "m3v1mp4DLC_resnet50_openfieldAug20shuffle1_30000.h5"
+    >>> config_file_path = BEHAVIOR_DATA_PATH / "DLC" / "config.yaml"
+    >>> subject_name = "ind1"
     >>>
+    >>> interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, subject_name=subject_name, verbose=False)
     >>> metadata = interface.get_metadata()
     >>> metadata.update(NWBFile=dict())
     >>> # For data provenance we add the time zone information to the conversion
     >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))
     >>> metadata["NWBFile"].update(session_start_time=session_start_time)
-    >>>
     >>> # Choose a path for saving the nwb file and run the conversion
     >>> nwbfile_path = f"{path_to_save_nwbfile}"
     >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
-    >>>
     >>> # If the conversion was successful this should evaluate to ``True`` as the file was created.
     >>> Path(nwbfile_path).is_file()
     True
--- a/src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
+++ b/src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
@@ -1,0 +1,92 @@
+"""Heberto Mayorquin, Authors: Saksham Sharda, Cody Baker, Ben Dichter."""
+from typing import Optional
+from pathlib import Path
+
+from pynwb.file import NWBFile
+
+from ....basedatainterface import BaseDataInterface
+from neuroconv.utils import dict_deep_update
+from neuroconv.utils import FilePathType, OptionalFilePathType
+from neuroconv.tools.nwb_helpers import make_or_load_nwbfile
+
+try:
+    from dlc2nwb.utils import auxiliaryfunctions, write_subject_to_nwb
+
+    HAVE_DLC2NWB = True
+except ImportError:
+    HAVE_DLC2NWB = False
+
+
+class DeepLabCutInterface(BaseDataInterface):
+    """Data interface for DeepLabCut datasets"""
+
+    def __init__(
+        self,
+        file_path: FilePathType,
+        config_file_path: FilePathType,
+        subject_name: str = "ind1",
+        verbose: bool = True,
+    ):
+        """
+        Interface for writing DLC's h5 files to nwb using dlc2nwb.
+        Parameters
+        ----------
+        file_path: FilePathType
+            path to the h5 file output by dlc.
+        config_file_path: FilePathType
+            path to .yml config file
+        subject_name: str
+            the name of the subject for which the nwbfile is to be created.
+        verbose: bool
+            controls verbosiy. True by default.
+        """
+        file_path = Path(file_path)
+        if "DLC" not in file_path.stem or ".h5" not in file_path.suffixes:
+            raise IOError("The file passed in is not a DeepLabCut h5 data file.")
+        assert HAVE_DLC2NWB, "to use this datainterface: 'pip install dlc2nwb'"
+
+        self._config_file = auxiliaryfunctions.read_config(config_file_path)
+        self.subject_name = subject_name
+        self.verbose = verbose
+        super().__init__(file_path=file_path, config_file_path=config_file_path)
+
+    def get_metadata(self):
+        metadata = dict(
+            NWBFile=dict(session_description=self._config_file["Task"], experimenter=[self._config_file["scorer"]]),
+        )
+        return metadata
+
+    def run_conversion(
+        self,
+        nwbfile_path: OptionalFilePathType = None,
+        nwbfile: Optional[NWBFile] = None,
+        metadata: Optional[dict] = None,
+        overwrite: bool = False,
+    ):
+        """
+        Conversion from DLC output files to nwb. Derived from dlc2nwb library.
+        Parameters
+        ----------
+        nwbfile_path: FilePathType
+            Path for where to write or load (if overwrite=False) the NWBFile.
+            If specified, this context will always write to this location.
+        nwbfile: NWBFile
+            nwb file to which the recording information is to be added
+        metadata: dict
+            metadata info for constructing the nwb file (optional).
+        overwrite: bool, optional
+            Whether or not to overwrite the NWBFile if one exists at the nwbfile_path.
+        """
+
+        base_metadata = self.get_metadata()
+        metadata = dict_deep_update(base_metadata, metadata)
+
+        with make_or_load_nwbfile(
+            nwbfile_path=nwbfile_path, nwbfile=nwbfile, metadata=metadata, overwrite=overwrite, verbose=self.verbose
+        ) as nwbfile_out:
+            write_subject_to_nwb(
+                nwbfile=nwbfile_out,
+                h5file=str(self.source_data["file_path"]),
+                individual_name=self.subject_name,
+                config_file=str(self.source_data["config_file_path"]),
+            )
--- a/tests/test_on_data/test_gin_behavior.py
+++ b/tests/test_on_data/test_gin_behavior.py
@@ -1,10 +1,53 @@
 import unittest
 from datetime import datetime
+from parameterized import parameterized, param
 
 from pynwb import NWBHDF5IO
-from neuroconv import NWBConverter, MovieInterface
+from neuroconv import NWBConverter, MovieInterface, DeepLabCutInterface
 
 from .setup_paths import OUTPUT_PATH, BEHAVIOR_DATA_PATH
+
+
+class TestDeepLabCutInterface(unittest.TestCase):
+    savedir = OUTPUT_PATH
+
+    @parameterized.expand(
+        [
+            param(
+                data_interface=DeepLabCutInterface,
+                interface_kwargs=dict(
+                    file_path=str(BEHAVIOR_DATA_PATH / "DLC" / "m3v1mp4DLC_resnet50_openfieldAug20shuffle1_30000.h5"),
+                    config_file_path=str(BEHAVIOR_DATA_PATH / "DLC" / "config.yaml"),
+                    subject_name="ind1",
+                ),
+            )
+        ]
+    )
+    def test_convert_behaviordata_to_nwb(self, data_interface, interface_kwargs):
+        nwbfile_path = self.savedir / f"{data_interface.__name__}.nwb"
+
+        class TestConverter(NWBConverter):
+            data_interface_classes = dict(TestBehavior=data_interface)
+
+        converter = TestConverter(source_data=dict(TestBehavior=dict(interface_kwargs)))
+        metadata = converter.get_metadata()
+        metadata["NWBFile"].update(session_start_time=datetime.now().astimezone())
+        converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, metadata=metadata)
+
+        with NWBHDF5IO(path=nwbfile_path, mode="r", load_namespaces=True) as io:
+            nwbfile = io.read()
+            assert "behavior" in nwbfile.processing
+            processing_module_interfaces = nwbfile.processing["behavior"].data_interfaces
+            assert "PoseEstimation" in processing_module_interfaces
+
+            pose_estimation_series_in_nwb = processing_module_interfaces["PoseEstimation"].pose_estimation_series
+            expected_pose_estimation_series = ["ind1_leftear", "ind1_rightear", "ind1_snout", "ind1_tailbase"]
+
+            expected_pose_estimation_series_are_in_nwb_file = [
+                pose_estimation in pose_estimation_series_in_nwb for pose_estimation in expected_pose_estimation_series
+            ]
+
+            assert all(expected_pose_estimation_series_are_in_nwb_file)
 
 
 class TestMovieDataNwbConversions(unittest.TestCase):
--- a/docs/api/interfaces.behavior.rst
+++ b/docs/api/interfaces.behavior.rst
@@ -4,3 +4,7 @@ Behavior
 Movie Data
 ----------
 .. automodule:: nwb_conversion_tools.datainterfaces.behavior.movie.moviedatainterface
+
+DeepLabCut
+----------
+.. automodule:: nwb_conversion_tools.datainterfaces.behavior.deeplabcut.deeplabcutdatainterface
--- a/docs/conversion_example_gallery.rst
+++ b/docs/conversion_example_gallery.rst
@@ -42,6 +42,13 @@ Imaging
     Scanbox <conversion_examples_gallery/imaging/scanbox>
     ScanImage <conversion_examples_gallery/imaging/scanimage>
 
+Behavior
+--------
+.. toctree::
+    :maxdepth: 1
+
+    DeepLabCut <conversion_examples_gallery/behavior/deeplabcut>
+
 Segmentation
 ------------
 
--- a/src/neuroconv/datainterfaces/__init__.py
+++ b/src/neuroconv/datainterfaces/__init__.py
@@ -45,6 +45,7 @@ from .ophys.hdf5.hdf5datainterface import Hdf5ImagingInterface
 from .ophys.scanimage.scanimageimaginginterface import ScanImageImagingInterface
 
 from .behavior.movie.moviedatainterface import MovieInterface
+from .behavior.deeplabcut.deeplabcutdatainterface import DeepLabCutInterface
 
 from .icephys.abf.abfdatainterface import AbfInterface
 
@@ -84,6 +85,7 @@ interface_list = [
     TiffImagingInterface,
     Hdf5ImagingInterface,
     MovieInterface,
+    DeepLabCutInterface,
     AbfInterface,
     ScanImageImagingInterface,
 ]


src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -0,0 +1,92 @@
...
+            path to .yml config file
+        subject_name: str
+            the name of the subject for which the nwbfile is to be created.
+        verbose: bool
+            controls verbosiy. True by default.
+        """
+        file_path = Path(file_path)
+        if "DLC" not in file_path.stem or ".h5" not in file_path.suffixes:
+            raise IOError("The file passed in is not a DeepLabCut h5 data file.")
+        assert HAVE_DLC2NWB, "to use this datainterface: 'pip install dlc2nwb'"
```
h-mayorquin, line 46: @bendichter should the error message be for users to pip install 
`git+https://github.com/catalystneuro/DLC2NWB@add_to_existing_nwb` until this is merged to `dlc2nwb`?

CodyCBakerPhD: Would prefer to wait until it's merged upstream

h-mayorquin: Done:
https://github.com/DeepLabCut/DLC2NWB/pull/10#event-7070987720

h-mayorquin: That does not solve the problem here, right? doing pip install dlc2nwb is not going to install a version that contains this as it will revert to the latest release on PiPy.

CodyCBakerPhD: Yeah, I suppose now we're just waiting on it being re-released

docs/conversion_examples_gallery/behavior/deeplabcut.rst
```diff
@@ -0,0 +1,27 @@
...
+    >>> subject_name = "ind1"
+    >>>
+    >>> interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, subject_name=subject_name, verbose=False)
+    >>> metadata = interface.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))
+    >>> metadata.get("NWBFile", dict()).update(session_start_time=session_start_time)
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> nwbfile_path = f"{path_to_save_nwbfile}"
+    >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
```
bendichter, line 24:
```suggestion
    >>> interface.run_conversion(nwbfile_path="nwbfile_path.nwb", metadata=metadata)
```

bendichter: I feel like this is clearer and saves a line of code.

bendichter (changes requested):

docs/conversion_examples_gallery/behavior/deeplabcut.rst
h-mayorquin: I need the variable further down to test that file was actually written though.

docs/conversion_examples_gallery/behavior/deeplabcut.rst
```diff
@@ -0,0 +1,28 @@
...
+    >>> file_path = BEHAVIOR_DATA_PATH / "DLC" / "m3v1mp4DLC_resnet50_openfieldAug20shuffle1_30000.h5"
+    >>> config_file_path = BEHAVIOR_DATA_PATH / "DLC" / "config.yaml"
+    >>> subject_name = "ind1"
+    >>>
+    >>> interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, subject_name=subject_name, verbose=False)
+    >>> metadata = interface.get_metadata()
+    >>> metadata.update(NWBFile=dict())
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))
+    >>> metadata["NWBFile"].update(session_start_time=session_start_time)
```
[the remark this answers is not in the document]
h-mayorquin: I implemented similar updates from your suggestions in the `nwb-conversion-tools` PR. The current code reads as:

```python
    >>> metadata = interface.get_metadata()
    >>> # For data provenance we add the time zone information to the conversion
    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))
    >>> metadata.get("NWBFile", dict()).update(session_start_time=session_start_time)
   ```

**Comment by h-mayorquin** [comment]
@CodyCBakerPhD @bendichter 
I guess the only remaining issue is to see if they do a release within the following week:
https://github.com/DeepLabCut/DLC2NWB/issues/13

**Comment by jeylau** [comment]
@h-mayorquin we will.

File: src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py (hunk 1 above)
Comment (h-mayorquin) [reply src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py:46]: Release is done. I pinned the full requirements to the new version. Should be working now. I will resolve this conversation.

File: docs/conversion_examples_gallery/behavior/deeplabcut.rst (hunk 2 above)
Comment (bendichter) [reply docs/conversion_examples_gallery/behavior/deeplabcut.rst:24]: wbere do you get `path_to_save_nwbfile` from?

Comment (bendichter) [reply docs/conversion_examples_gallery/behavior/deeplabcut.rst:24]: Oh, I see, it's a doctest_namespace fixture. OK, how about this?

File: docs/conversion_examples_gallery/behavior/deeplabcut.rst [hunk 4]
```diff
@@ -0,0 +1,27 @@
...
+    >>> interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, subject_name=subject_name, verbose=False)
+    >>> metadata = interface.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))
+    >>> metadata.get("NWBFile", dict()).update(session_start_time=session_start_time)
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> nwbfile_path = f"{path_to_save_nwbfile}"
+    >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
+    >>> # If the conversion was successful this should evaluate to ``True`` as the file was created.
+    >>> Path(nwbfile_path).is_file()
```
bendichter, line 26:
```suggestion
    >>> interface.run_conversion(nwbfile_path=path_to_save_nwbfile, metadata=metadata)
    >>> # If the conversion was successful this should evaluate to ``True`` as the file was created.
    >>> Path(path_to_save_nwbfile).is_file()
```

h-mayorquin: That works as well, the previous format was trying to make explicit that you can pass the path as a string which I thought would be the simplest way to most people. But from there to "path_to_save_nwbfile` I guess there is not a lot of difference between these two. 

I committed your suggestion.

docs/conversion_examples_gallery/behavior/deeplabcut.rst
```diff
@@ -0,0 +1,26 @@
...
+    >>> from datetime import datetime
+    >>> from dateutil import tz
+    >>> from pathlib import Path
+    >>> from neuroconv import DeepLabCutInterface
+    >>>
+    >>> file_path = BEHAVIOR_DATA_PATH / "DLC" / "m3v1mp4DLC_resnet50_openfieldAug20shuffle1_30000.h5"
+    >>> config_file_path = BEHAVIOR_DATA_PATH / "DLC" / "config.yaml"
+    >>> subject_name = "ind1"
+    >>>
+    >>> interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, subject_name=subject_name, verbose=False)
```
bendichter, line 17:
```suggestion
    >>>
    >>> interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, subject_name="ind1", verbose=False)
```

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -0,0 +1,92 @@
...
+        Parameters
+        ----------
+        file_path: FilePathType
+            path to the h5 file output by dlc.
+        config_file_path: FilePathType
+            path to .yml config file
+        subject_name: str
+            the name of the subject for which the nwbfile is to be created.
+        verbose: bool
+            controls verbosiy. True by default.
```
bendichter, line 41:
```suggestion
        verbose: bool = False,
    ):
        """
        Interface for writing DLC's h5 files to nwb using dlc2nwb.
        Parameters
        ----------
        file_path: FilePathType
            path to the h5 file output by dlc.
        config_file_path: FilePathType
            path to .yml config file
        subject_name: str
            the name of the subject for which the nwbfile is to be created.
        verbose: bool
            controls verbosity. False by default.
```

h-mayorquin: I am of the opinion that all the verbose should be False by default but @CodyCBakerPhD is of a different opinion. In favor of his side of the argument, at the point that this discussion surfaced all the other interfaces had this default.

bendichter: I see. I prefer default False as well, but let's stay consistent for now

tests/test_on_data/test_gin_behavior.py
```diff
@@ -1,12 +1,55 @@
...
+            assert "PoseEstimation" in processing_module_interfaces
+
+            pose_estimation_series_in_nwb = processing_module_interfaces["PoseEstimation"].pose_estimation_series
+            expected_pose_estimation_series = ["ind1_leftear", "ind1_rightear", "ind1_snout", "ind1_tailbase"]
+
+            expected_pose_estimation_series_are_in_nwb_file = [
+                pose_estimation in pose_estimation_series_in_nwb for pose_estimation in expected_pose_estimation_series
+            ]
+
+            assert all(expected_pose_estimation_series_are_in_nwb_file)
```
bendichter, line 50:
```suggestion
            assert all(
                x in processing_module_interfaces["PoseEstimation"].pose_estimation_series for x in ["ind1_leftear", "ind1_rightear", "ind1_snout", "ind1_tailbase"]
            )
```

docs/conversion_examples_gallery/behavior/deeplabcut.rst
```diff
@@ -0,0 +1,25 @@
...
+    >>>
+    >>> interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, subject_name="ind1", verbose=False)
+    >>> metadata = interface.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))
+    >>> metadata.get("NWBFile", dict()).update(session_start_time=session_start_time)
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> interface.run_conversion(nwbfile_path=path_to_save_nwbfile, metadata=metadata)
+    >>> # If the conversion was successful this should evaluate to ``True`` as the file was created.
+    >>> Path(path_to_save_nwbfile).is_file()
```
CodyCBakerPhD, line 24: Where is `path_to_save_nwbfile` defined? The purpose of these galleries is to provide easily copied code to be sent to people that want to paste it, point to their data, and run.

I get that we have to use the global `BEHAVIOR_DATA_PATH` to achieve that testing feature with the pytest fixtures, but other variable references should be standalone and/or more easily modifiable - even by first-time users of Python that otherwise have general understanding of coding

CodyCBakerPhD:
@h-mayorquin Looking good, and with the upstream releases we should be getting close to being able to merge this now - just a reminder to post some example files on slack or but them in a drive to share here

docs/conversion_examples_gallery/behavior/deeplabcut.rst
CodyCBakerPhD: I guess relates to https://github.com/catalystneuro/neuroconv/pull/24#discussion_r933229021

so the user is intended to just know to change the `path_to_save_nwbfile` variable to some place on their system? Feel like that's not super apparent or as user-friendly as it could be

h-mayorquin: It was like this before
```python
    >>> nwbfile_path = f"{path_to_save_nwbfile}"
    >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
```

The idea was to indicate that it should be a string with the path to the nwbfile but then I changed it above after Ben's suggestion and I am less sure it was that clear either. Do you have any suggestion on how to make this more user friendly?

tests/test_on_data/test_gin_behavior.py
h-mayorquin: I really prefer the code that we have now. Two advantages of it:
1)
"assert (that) all(expected_pose_estimations_are_in_nwb_file)"
Reads to me more to me like natural language whereas the one liner has to be parsed to understand what the test is doing.
2) If the test fails and you throw a debugger before the assertion you can readily inspect which ones failed. To do this, you can just inspect the previously created variable `expected_pose_estimation_series_are_in_nwb_file` instead of having the re-do the one-liner again.

h-mayorquin:
> @h-mayorquin Looking good, and with the upstream releases we should be getting close to being able to merge this now - just a reminder to post some example files on slack or but them in a drive to share here

I shared it on slack.

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -0,0 +1,92 @@
+"""Heberto Mayorquin, Authors: Saksham Sharda, Cody Baker, Ben Dichter."""
+from typing import Optional
+from pathlib import Path
+
+from pynwb.file import NWBFile
+
+from ....basedatainterface import BaseDataInterface
+from neuroconv.utils import dict_deep_update
+from neuroconv.utils import FilePathType, OptionalFilePathType
+from neuroconv.tools.nwb_helpers import make_or_load_nwbfile
```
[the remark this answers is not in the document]
h-mayorquin: Yep. That one flew over me. Done in the last commit.

docs/conversion_examples_gallery/behavior/deeplabcut.rst
CodyCBakerPhD: The simplest thing we would want to share via copy/paste is

```python
from neuroconv import SpikeGLXRecordingInterface

spikeglx_file_path = ECEPHYS_DATA_PATH  / "spikeglx_file.ap.bin"
nwbfile_path = "path/to/save/my_nwbfile.nwb"

interface = SpikeGLXRecordingInterface(file_path=spikeglx_file_path)
interface.run_conversion(nwbfile_path=nwbfile_path)
```

from https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/nwbconverter.py#L163-L165 they shouldn't even have to get the metadata for interfaces that support automatic setting of session start time. For those that do something like

```python
from datetime import datetime
from dateutil import tz
from neuroconv import NeuralynxRecordingInterface

session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))  # Y, M, D, H, M, S
neuralynx_folder_path = "path/to/neuralynx_folder"
nwbfile_path = "path/to/save/my_nwbfile.nwb"

interface = NeuralynxRecordingInterface(folder_path=neuralynx_folder_path )
interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
```

but of course more than anything, we should be consistent across the gallery. It looks like most of the currently galleries use the `f"{path_to}"` convention, which I agree with Ben looks a little awkward especially if the user as no idea what an f-string is. I do think that the `ECEPHYS_DATA_PATH` global is fine to use as it is at the top and all in capitals, it should be more apparent to the user to change that for their data. Though we could also leave a comment noting that if you disagree.

I'd also combine all definitions of paths or datetimes needed prior to calling the 2-4 line code sequence to run things so it's more apparent what the 'input' vs. 'output' is. So the order is (i) do imports, (ii) define paths as strings or mixtures of the global data paths needed for fixtures, (iii) 2-4 lines to run the standalone conversion through the data interface.

CodyCBakerPhD: >  metadata.get("NWBFile", dict()).update(session_start_time=session_start_time)

As an aside (just noticing) I would also say that every `get_metadata` call should return an `NWBFile` level to the dictionary structure, if it doesn't then we should add that in where appropriate. The recording interfaces do this automatically: https://github.com/catalystneuro/neuroconv/blob/5f55cdb49d3b613d06fc8cfcde1b2faac0081e41/src/neuroconv/tools/spikeinterface/spikeinterface.py#L82-L85

but the DLC is standalone so maybe it doesn't have this yet?

h-mayorquin: If you find a way to make a notation like "/this/is/a/path" that indicates clearly that the user should change that by a path and work with the testing that would be great. 

Concerning using scripting / scientific programming style for the tests (putting all the definitions at the beginning) I personally don't like it and I find it more difficult to follow but I honestly don't feel strong about it. Let's discuss this issues in the following team meeting as I feel this just about getting to a consensus and sticking with it.

About the metadata comment that can be added here. While doing the conversion gallery I found some that did not have this structure. I don't remember which ones. I agree that we should correct that.

CodyCBakerPhD: > 

We can do it now for the main item in discussion of the `nwbfile_path` ass illustrated above. We don't need the NWBFile to be saved anywhere specifically and for any user it would just be the working directory (or absolute path if they used that on their system).

The data paths are a bit more stringent, I do understand. I guess just have to keep those as-is for now. Down the road we could do something fancy like form a symlink to a made-up folder `/path/to/data` or something but that could also be overkill. My main gripe is just with setting the `nwbfile_path` is the simplest, clearest way which IMO is just to set it equal to a simple string (not fstring or pytest fixture variable)

tests/test_on_data/test_gin_behavior.py
CodyCBakerPhD: Yeah, best to be as readable/modular as possible for test cases like this. For diagnosing it's a lot easier to introspect when they are pre-defined variables in the function prior to assertion fail

CodyCBakerPhD (approved):

docs/conversion_examples_gallery/behavior/deeplabcut.rst
CodyCBakerPhD: Still, no reason to hold up the DataInterface PR for this. Can discuss in team meeting to achieve consistency

CodyCBakerPhD:
Files look good, minus a few things that would take more work upstream. I posted those to slack but also posting here:

1. The data field of each node is uncompressed - can we fix that? It should be easy to do so
2. Logic for using rate instead of timestamps when they are regular
3. For irregular timestamps, enable compression like in (1) above

CodyCBakerPhD (dismissed):

docs/conversion_examples_gallery/behavior/deeplabcut.rst
bendichter: Yes I agree this is a good convo and also that this should not hold up the DLC support. 

I like the idea of `get_metadata` always including an NWBFile. We should adjust the json-schema to make the NWBFile required.

I see that we are using a fixture that creates a temp directory, but I I agree that few of our users are likely to be comfortable with fixtures and it might confuse them. Would it be a problem to just save these files in the local directory with just something like:

```python
    >>> interface.run_conversion(nwbfile_path='output_nwbfile.nwb', metadata=metadata)
```
?

CodyCBakerPhD: > Would it be a problem to just save these files in the local directory with just something like:
> interface.run_conversion(nwbfile_path='output_nwbfile.nwb', metadata=metadata)

That's what I'm proposing, though @h-mayorquin likes to decouple the assignment into two steps

```python
nwbfile_path = "output_nwbfile.nwb"
interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
```

to avoid having a single code line that does two operations (both declaring save path and running conversion).

Personally though I also like having as few a lines as code as possible for these kind of things so I'd also like nothing more than to have 7 lines (above) reduced to 4

```python
interface = NeuralynxRecordingInterface(folder_path="path/to/neuralynx_folder")
metadata["NWBFile"].update(session_start_time=datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific")))
interface.run_conversion(nwbfile_path="path/to/save/my_nwbfile.nwb", metadata=metadata)

```

tests/test_on_data/test_gin_behavior.py
bendichter: My approach is a bit more performant by leveraging [Python's short-circuiting of logical operators](https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not), but I don't expect this to provide any real time improvement, and if you both feel the first way is easier to debug that's a good enough reason to keep it as-is.

CodyCBakerPhD: Yeah if this was actual source code I'd prefer it that way using the lazy generator approach.

Though also worth noting we can obtain even higher degree of test vebosity/introspection here by using this: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual

See https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/tests/test_inspector.py#L285 for example usage - the display is really nice when it fails (as long as maxDiff is set to None in the class)

docs/conversion_examples_gallery/behavior/deeplabcut.rst
```diff
@@ -0,0 +1,25 @@
...
+    >>> from neuroconv import DeepLabCutInterface
+    >>>
+    >>> file_path = BEHAVIOR_DATA_PATH / "DLC" / "m3v1mp4DLC_resnet50_openfieldAug20shuffle1_30000.h5"
+    >>> config_file_path = BEHAVIOR_DATA_PATH / "DLC" / "config.yaml"
+    >>>
+    >>> interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, subject_name="ind1", verbose=False)
+    >>> metadata = interface.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))
+    >>> metadata.get("NWBFile", dict()).update(session_start_time=session_start_time)
```
bendichter, line 20:
```suggestion
    >>> metadata["NWBFile"].update(session_start_time=session_start_time)
```
On second thought I don't think the `.get` call is needed here because we know that in this case `get_metadata` does contain `"NWBFile"`

docs/conversion_examples_gallery/behavior/deeplabcut.rst
h-mayorquin: >That's what I'm proposing, though @h-mayorquin likes to decouple the assignment into two steps
I am fine with Ben proposal of having as a one liner in this case. It is true that I like explicit assignments in code for both auto-documentation and debugging purposes but here just the variable name is clear enough so the correction that Ben did above is already committed. The declaration in this case though was for a different purpose as the variable is needed once again for testing that the conversion run correctly.

I also think that Cody proposal of using a save path that goes something like "this/is/where/stuff/should/be/saved" is OK as well. Although I honestly don't see a lot of difference between that and have a comment there indicating that the user should modify the path with their desired save destination. If you both want to change it I would still think that we should do it for all them at the same time. I was not explicit above about the constrains of testing this way though:
1) Each test file path should be unique or go to a different folder so when you run all the tests together in the CI or in your local machine you don't get input output errors.
2) You might want to create a script that cleans all the files at the end for when you run the testing suit on your local machine (not sure it creates problems in CI, I don't think so).

The fixtures are not intended for the users to modify or even be aware of, the goal of the fixtures is so automatic testing assures that the documentation keeps working. My goal was always that the users should modify all the path themselves. As an example, this is the one in neuralynx:
```python
    >>>  # Choose a path for saving the nwb file and run the conversion
    >>> nwbfile_path = f"{path_to_save_nwbfile}"  # Modify this to suit your needs to something like: "./saved_file.nwb"
    >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
    >>>
    >>> # If the conversion was successful this should evaluate to ``True`` as the file was created.
    >>> Path(nwbfile_path).is_file()
    True
 ```

h-mayorquin: So, to summarize, please feel free to add if I am missing something those are the issues:
1) Make the saving paths unique so they don't need to be modified.
2) Decide between explicit declaration vs implicit (I favor explicit but I think you two favor saving lines so I am fine with reducing it here for all the other interfaces.
3) Script format where everything is declared at the beginning is another option if we chose to keep explicit declaration.

h-mayorquin:
> Files look good, minus a few things that would take more work upstream. I posted those to slack but also posting here:
> 
> 1. The data field of each node is uncompressed - can we fix that? It should be easy to do so
> 2. Logic for using rate instead of timestamps when they are regular
> 3. For irregular timestamps, enable compression like in (1) above

@CodyCBakerPhD 
Can you open an issue on the repo? I will prepare the PR.

bendichter (approved):

PR #27, opened 2022-07-22
Add `NeuroscopeLFPInterface` to gin tests

h-mayorquin:
Porting of:
https://github.com/catalystneuro/nwb-conversion-tools/pull/461

This depends on the master version of Neo. The idea is to merge once NEO does a new release.


h-mayorquin:
For #45 we started using the master branch from neo so I think this is good to go. I am fixing here to the latest available commit as is not good practice to pin to branches because they move.

CodyCBakerPhD:
I think Julia was working on a new release soon so I suppose we can start to move on this now.

CodyCBakerPhD (approved):

CodyCBakerPhD (approved):

PR #28, opened 2022-07-22
First draft of style guide

h-mayorquin:
We had some discussion about style in #19.  I added the basics discussed here as a draft. We can discuss what else we want to add in the next team meeting.


docs/developer_guide/style_guide.rst
```diff
@@ -0,0 +1,28 @@
...
+#. Use the numpy `docstring standard <https://numpydoc.readthedocs.io/en/latest/format.html#numpydoc-docstring-guide>`_ in all your docstring.
+
+Data interfaces conventions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#. Use :code:`file_path` and :code:`folder_path` as arguments for the location of input files and folder respectively.
+#. In opposition to convention number 1 above, we use :code:`nwbfile` to refer to in-memory filesin the nwb format.
+
+Other conventions
+^^^^^^^^^^^^^^^^^
+#. Whenever possible, use dictionary literals :code:`dictionary = dict(foo=bar)`  instead of constructors :code:`dictionary={foo=bar}`.
```
CodyCBakerPhD, line 28:
```suggestion
#. Whenever possible, use dictionary literals :code:`dictionary = dict(foo=bar)`  instead of constructors :code:`dictionary={foo: bar}`.
```

Can you also add a blurb about cases where it's not possible?

I can think of two: dictionary compression from an iterarable, or any time the intended key of the dictionary is not a string, likewise any time you want to use an object/variable as the key (if `foo` is an object in the environment, say an integer equal to `1`, `dict(foo="bar")` will result  in `{'foo': 'bar'}` whereas `{foo: "bar"}` would result in `{1: "bar"}`.

docs/developer_guide/style_guide.rst
```diff
@@ -0,0 +1,28 @@
...
+#. Use relative imports for references inside the source code :code:`src/neuroconv`.
+#. Unit tests are an exception to the previous rule; they should use absolute imports, exactly as client code would.
+#. A corollary is that, if :code:`neuroconv.foo`` exports a class or function that is implemented in :code:`neuroconv.foo.bar.baz`, the test should import it from :code:`neuroconv.foo`.
+#. Please dont use :code:`import *` in any case. It is considered harmful by the official Python recommendations. It makes the code harder to read as the origin of symbols is no longer explicitly referenced, but most important, it prevents using a static analysis tool like pyflakes to automatically find bugs in scikit-learn.
+#. Use the numpy `docstring standard <https://numpydoc.readthedocs.io/en/latest/format.html#numpydoc-docstring-guide>`_ in all your docstring.
+
+Data interfaces conventions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#. Use :code:`file_path` and :code:`folder_path` as arguments for the location of input files and folder respectively.
+#. In opposition to convention number 1 above, we use :code:`nwbfile` to refer to in-memory filesin the nwb format.
```
bendichter, line 24:
```suggestion
#. In opposition to convention number 1 above, we use :code:`nwbfile` to refer to in-memory files in the nwb format.
```

docs/developer_guide/style_guide.rst
bendichter: yes, or if you want to use a key that is a string that starts with a number: `dict(1aa="bar")` does not work but `{'aaa': 'bar'}` does

bendichter:
@h-mayorquin this looks great, thanks for putting it together!

docs/developer_guide/style_guide.rst
CodyCBakerPhD:
```suggestion
#. Use  :code:`nwbfile` as the variable name to refer to an in-memory :py:class:`~pynwb.NWBFile` object.
```

I wouldn't even say it's an exception either - the `_path` references are to places on disk, the `nwbfile` object in memory may not even have been written to disk at all.

bendichter: This is referring to 

> #. Use underscores to separate words in non class names: :code:`n_samples` rather than :code:`nsamples`.

CodyCBakerPhD: That's a bit unclear due to the rendering of the numbers on readthedocs, then

image

link: https://neuroconv--28.org.readthedocs.build/en/28/developer_guide/style_guide.html

CodyCBakerPhD: @h-mayorquin Also there's some trailing ` characters for some of those code references

docs/developer_guide/style_guide.rst
h-mayorquin: So, I was thinking about this and I have two questions: 
1) The python type constructor has three possible type of inputs:
https://docs.python.org/3/library/stdtypes.html#dict:
```python
class dict(**kwargs)
class dict(mapping, **kwargs)
class dict(iterable, **kwargs)
```
The examples that you mention still can be done in this manner by doing it with the third option (using an iterable):

```python
>>> x = 1
>>> dict([(x, 'cody_example')])
{1: 'cody_example'}
>>> dict([('1aa', 'ben_example')])
{'1aa': 'ben_example'}
```
Does not throw any error. 
It is the only the first input parameter to the type constructor dict that is limited by using keyword (therefore converting to string, Cody example) and by requiring them to the valid python identifiers (Ben example). 

**But I guess that the style convention that we want is to use the first option (`class dict(**kwargs)`) when possible and if not then use the constructor with braces (i.e. `{foo:bar}`). Is this correct?**

2) Second but less important, I could not find the dict constructor being referred as a "dictionary literal" in the python documentation. That said, it seems that people do referrer to it in this way here and there, example:
https://stackoverflow.com/questions/6610606/is-there-a-difference-between-using-a-dict-literal-and-a-dict-constructor

Should we refer to `dict()` in the same way that the documentation does and call it the type constructor (I guess type constructor is just the name of the `__init__`):
https://docs.python.org/3/library/stdtypes.html#mapping-types-dict

h-mayorquin:
Btw, I actually don't understand what the second recommendation of the sklearn guide above referrs to:
`#. Avoid multiple statements on one line. Prefer a line return after a control flow statement (if/for)`

Is that referring to the conditional expressiosn?
https://docs.python.org/3.10/reference/expressions.html?highlight=ternary#conditional-expressions
```
conditional_expression ::= or_test ["if" or_test "else" expression]
```
Or is this something else?

bendichter:
@h-mayorquin `for i in range(10): print(i)` is valid python. Also `if True: print('hi')`

docs/developer_guide/style_guide.rst
bendichter: > So, I guess that the style convention that we want is to use the first option (class dict(**kwargs)) when possible and if not then use the constructor with braces (i.e. {foo:bar}). Is this correct?*

That's a good point. I had not thought of the other arg types for `dict`, but I agree if you are writing the keys as strings you might as well use the `{key: val}` notation. The reason I like the `dict` form is that it is the exact same syntax as kwargs, so you can do things like 

```python
my_dict = dict(a=1, b=2)

my_dict.update(a=2, b=3)

# and

run_func(a=1, b=2)
```

all with the same syntax. This is particularly helpful if you want to refactor the kwargs out into a dictionary. If you want to change this

```python
run_func(
    a=1,
    b=2,
    c=3,
    d=4,
)
```
into

```python
kwargs = dict(
    a=1,
    b=2,
    c=3,
    d=4,
)
run_func(**kwargs)
```

or vica versa then you don't need to change as many lines.

h-mayorquin:
> @h-mayorquin `for i in range(10): print(i)` is valid python. Also `if True: print('hi')`

@bendichter 
We have quite a lot of:

```python
x = 'value_when_true' if True else 'value_when_else'
```
In our code, is that the same thing?

docs/developer_guide/style_guide.rst
h-mayorquin: Oh, I see, I never realized that. Good point.

CodyCBakerPhD:
> In our code, is that the same thing?

No, statements are separated by lines (preferred) or semi-colon (should never be used). The modern `if/or` one-liners can be thought of as logical operators like `and` or the binary operators `==/!=`.

docs/developer_guide/style_guide.rst
```diff
@@ -0,0 +1,28 @@
...
+
+#. Use underscores to separate words in non class names: :code:`n_samples` rather than :code:`nsamples`.
+#. Avoid multiple statements on one line. Prefer a line return after a control flow statement (if/for).
+#. Use relative imports for references inside the source code :code:`src/neuroconv`.
+#. Unit tests are an exception to the previous rule; they should use absolute imports, exactly as client code would.
+#. A corollary is that, if :code:`neuroconv.foo` exports a class or function that is implemented in :code:`neuroconv.foo.bar.baz`, the test should import it from :code:`neuroconv.foo`.
+#. Please dont use :code:`import *` in any case. It is considered harmful by the official Python recommendations. It makes the code harder to read as the origin of symbols is no longer explicitly referenced, but most important, it prevents using a static analysis tool like pyflakes to automatically find bugs in scikit-learn.
+#. Use the numpy `docstring standard <https://numpydoc.readthedocs.io/en/latest/format.html#numpydoc-docstring-guide>`_ in all your docstring.
```
CodyCBakerPhD, line 19: I'd recommend multi-section numbering or anything similar to differentiate these rules across the sections: https://neuroconv--28.org.readthedocs.build/en/28/developer_guide/style_guide.html#other-conventions

As it resets to `1.` each time it can get confusing which rule is being referenced in discussion

h-mayorquin: Yes, I have been thinking on how to fix this since the numbering caused a previous confusion in the comments above. For some reason what I tried did not work out. After the updates are available https://github.com/catalystneuro/neuroconv/pull/53 I will try again to re-implement this to have a unique reference to each of the points.

bendichter (dismissed):

docs/developer_guide/style_guide.rst
h-mayorquin: @CodyCBakerPhD I could not find something for this. If you see this in another documentation let me know and I will try to adapt it.

CodyCBakerPhD: For automation it looks like all you're missing is the master headers: https://stackoverflow.com/questions/2362197/restructuredtext-numbered-headers

Otherwise, I mean, you could just manually type out different headings...

1a.
1b.
1c.

2a.
2b.
2c.

etc...

PR #32, opened 2022-07-24
Unittest for `add_electrical_series` I

h-mayorquin:
As discussed, I am adding multi segment support to the `write_recording` function. To pin current behavior and increase unittest coverage I am adding the first batch of tests to `add_electrical_series` in `test_si`. The test here are concerned with default arguments and writing the recording to the correct place (e.g. acquisition vs processing). To test the data I also added the option of  writing the data with no iterator but keep the default behavior identical for backward compatibility (it uses "V2").

### DIFF
--- a/src/neuroconv/tools/spikeinterface/spikeinterface.py
+++ b/src/neuroconv/tools/spikeinterface/spikeinterface.py
@@ -15,6 +15,7 @@ from spikeextractors import RecordingExtractor, SortingExtractor
 from numbers import Real
 from hdmf.data_utils import DataChunkIterator
 from hdmf.backends.hdf5.h5_utils import H5DataIO
+import psutil
 
 from .spikeinterfacerecordingdatachunkiterator import SpikeInterfaceRecordingDataChunkIterator
 from ..nwb_helpers import get_module, make_or_load_nwbfile
@@ -457,6 +458,35 @@ def add_electrodes(
         warnings.warn(f"No information added to the electrodes table")
 
 
+def check_if_recording_traces_fit_into_memory(recording: SpikeInterfaceRecording, segment_index: int = 0) -> None:
+    """Raises an error if the full traces of a recording extractor are larger than psutil.virtual_memory().available
+
+    Parameters
+    ----------
+    recording : SpikeInterfaceRecording
+        A recording extractor object from spikeinterface.
+    segment_index : int, optional
+        The segment index of the recording extractor object, by default 0
+
+    Raises
+    ------
+    MemoryError
+    """
+    element_size_in_bytes = recording.get_dtype().itemsize
+    num_channels = recording.get_num_channels()
+    num_frames = recording.get_num_frames(segment_index=segment_index)
+
+    traces_size_in_bytes = element_size_in_bytes * num_channels * num_frames
+    available_memory_in_bytes = psutil.virtual_memory().available
+
+    if traces_size_in_bytes > available_memory_in_bytes:
+        message = (
+            f"Memory error, full electrical series is {round(traces_size_in_bytes/1e9, 2)} GB) but only"
+            f"({round(available_memory_in_bytes/1e9, 2)} GB are available. Use iterator_type='V2'"
+        )
+        raise MemoryError(message)
+
+
 def add_electrical_series(
     recording: SpikeInterfaceRecording,
     nwbfile=None,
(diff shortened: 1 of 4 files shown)


tests/test_internals/test_si.py
```diff
@@ -599,6 +602,96 @@ def test_group_set_custom_description(self):
...
+
+        processing_module = self.nwbfile.processing
+        assert "ecephys" in processing_module
+
+        ecephys_module = processing_module["ecephys"]
+        assert "LFP" in ecephys_module.data_interfaces
+
+        lfp_container = ecephys_module.data_interfaces["LFP"]
+        assert isinstance(lfp_container, pynwb.ecephys.LFP)
+        assert "ElectricalSeries_lfp" in lfp_container.electrical_series
```
CodyCBakerPhD, line 651: This sequence of assertions (or analogous for acquisition) are repeated multiple times.

I'd recommend defining your own `TestCase` assertion function such as

```python
def assert_processing_container_properties(
    self,
    nwbfile,
    container_name="LFP",  # for example
    object_name="ElectricalSeries_lfp",  # for example
):
        module = self.nwbfile.processing
        assert "ecephys" in  module

        ecephys_submodule = module["ecephys"]
        assert "LFP" in ecephys_submodule.data_interfaces

        lfp_container = ecephys_submodule.data_interfaces["LFP"]
        assert isinstance(lfp_container, pynwb.ecephys.LFP)
        assert "ElectricalSeries_lfp" in lfp_container.electrical_series
```

which would not only reduce the total code by a bit but also make those tests more unit-like by indicating in one line what behavior is being asserted (even if the way that it is asserted involves multiple aspects like this).

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -633,7 +634,9 @@ def add_electrical_series(
         else:
             eseries_kwargs.update(conversion=1e-6)
             eseries_kwargs.update(channel_conversion=channel_conversion)
-    if iterator_type is None or iterator_type == "v2":
+    if iterator_type is None:
+        ephys_data = checked_recording.get_traces(return_scaled=write_scaled, segment_index=segment_index)
```
CodyCBakerPhD, line 638: I would strongly recommend adding perhaps a `psutil` check of available memory (similar to the movie interface) to ensure the user does not load in significantly more data than they really ought to be capable of

CodyCBakerPhD:
More tests are always nice!

tests/test_internals/test_si.py
h-mayorquin: I think that is a good idea but there are really only two cases as the acquisition boilerplate has a simpler structure. Plus, the `isinstance` for the container type is different in each of the two cases so that's an extra parameter.  I feel that introducing indirection in a test to cover only two cases gains in code reduction what it loses in code clarity (even if intent clarity becomes better). What do you think? I am on the edge.

src/neuroconv/tools/spikeinterface/spikeinterface.py
h-mayorquin: Good idea, I will check it out.

tests/test_internals/test_si.py
CodyCBakerPhD: Depends on if that kind of assertion could be used in any other tests through the suite? Or for more general objects (not just ElectricalSeries)

Such an extension of a `TestCase` could be defined in the testing module for usage across the suite.

Up to you, though (can do it in subsequent PR if you decide yes)

CodyCBakerPhD:
@h-mayorquin This PR introduces the feature of non-iterative writing to `write_recording`, can you add something about that to the CHANGELOG?

tests/test_internals/test_si.py
h-mayorquin: Yes, that sounds like a good idea. I will check it out once I add more tests, maybe the pattern will repeat itself more.

src/neuroconv/tools/spikeinterface/spikeinterface.py
h-mayorquin: Done. I added a test for this functionality as well using mocks.

h-mayorquin:
> @h-mayorquin This PR introduces the feature of non-iterative writing to `write_recording`, can you add something about that to the CHANGELOG?

@CodyCBakerPhD 
Done in the last PR.

CodyCBakerPhD (approved):

PR #36, opened 2022-07-26
Propagate default iterator in ecephys to data interfaces

h-mayorquin:
In #32 non-iterative writing capabilities were added to the electrical series writing pipeline. To keep backwards compatibility the defaults of the function were changed to "v2" which is the "in-house" iterator (coincidentally now the docstrings are right, yei!). However, this change of default parameters was not propagated all the way up to the data interfaces so, in the current state, the would be running non-iterative write by default. This PR addresses this and some other minor refactoring of the `add_electrical_series` that I encountered while re-writing this.

### DIFF
--- a/src/neuroconv/tools/spikeinterface/spikeinterface.py
+++ b/src/neuroconv/tools/spikeinterface/spikeinterface.py
@@ -489,7 +489,7 @@ def check_if_recording_traces_fit_into_memory(recording: SpikeInterfaceRecording
 
 def add_electrical_series(
     recording: SpikeInterfaceRecording,
-    nwbfile=None,
+    nwbfile,
     metadata: dict = None,
     segment_index: int = 0,
     starting_time: Optional[float] = None,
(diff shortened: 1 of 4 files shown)


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -664,6 +661,11 @@ def add_electrical_series(
         else:
             eseries_kwargs.update(conversion=1e-6)
             eseries_kwargs.update(channel_conversion=channel_conversion)
+
+    # Iterator
+    if iterator_opts is None:
+        iterator_opts = dict()
```
CodyCBakerPhD, line 667: You can also do this kind of thing as `iterator_opts = iterator_opts or dict()` to save a line. We also usually do that kind of argument parsing as early in the function as possible (right after header).

CodyCBakerPhD:
@h-mayorquin Good catch, thanks

CodyCBakerPhD (approved):

PR #40, opened 2022-07-27
Remove `use_times` argument in `add_electrical_series`

h-mayorquin:
We have been moving away from leaving this as a choice. For roiextractors for example, we automatically detect whether timestamps are uniform or not and then store only the rate if it is or the whole tiemestamps if it is not. This PR does this for ecephys pipeline for recording extractor by removing `use_times` from `add_electrical_function` and propagating up all the way to the interfaces. Also, two unittests to cover the intended behavior are added as well.


src/neuroconv/datainterfaces/ecephys/baselfpextractorinterface.py
```diff
@@ -36,7 +36,7 @@ def run_conversion(
         overwrite: bool = False,
         stub_test: bool = False,
         starting_time: Optional[float] = None,
-        use_times: bool = False,
+        use_times: bool = False,  # To-do to remove, deprecation
```
CodyCBakerPhD, line 39: Not that critical, but using `#TODO` instead will lint it using standard tools like Pylint. Does `# To-do` lint in yours and if so what type are you using?

src/neuroconv/datainterfaces/ecephys/openephys/openephysdatainterface.py
```diff
@@ -42,7 +42,8 @@ def __init__(
                 folder_path=folder_path, experiment_id=experiment_id, recording_id=recording_id, verbose=verbose
             )
             self.recording_extractor = OldToNewRecording(oldapi_recording_extractor=self.recording_extractor)
-
+            # Remove when spikeinterface 0.95 is released, this has an int sampling rate that causes problems
+            self.recording_extractor._sampling_frequency = float(self.recording_extractor.get_sampling_frequency())
```
CodyCBakerPhD, line 46: Shouldn't we use `recording_extractor.set_sampling_frequency()` instead?

CodyCBakerPhD: Just in case there's ever an extractor out there (or a custom extractor as well) that doesn't call it from the private attribute for some reason

h-mayorquin: I could not find such a method in base recording.

CodyCBakerPhD: OK I must have misremembered or assumed it behaved more like an official property, my bad.

CodyCBakerPhD (approved):

PR #23, opened 2022-07-21
Refactor write segmentation IV (add plane segmentation)

weiglszonja:
## Motivation

This is part of a join effort with @h-mayorquin to refactor `write_segmentation` in the `roiextractors` module. 
To create a `PlaneSegmentation` object we need:
- metadata for this plane segmentation (defined by the `plane_segmentation_index`) 
- keyword arguments that define plane segmentation:
    - description
    - output from a `SegmentationExtractor` object added as VectorData columns
    - link to the `ImagingPlane` object from which this data was generated
This is currently all embedded in `write_segmentation`, so the aim of this PR is to organise this into methods that can be tested.

this is a work in progress
## How to test the behavior?
```
Show here how to reproduce the new behavior (can be a bug fix or a new feature)
```

## Checklist

- [ ] Have you checked our [Contributing](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/docs/contribute.rst) document?
- [ ] Have you ensured the PR description clearly describes the problem and solutions?
- [ ] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/nwb-conversion-tools/pulls) for the same change?
- [ ] If this PR fixes an issue, is the first line of the PR description `fix #XXX` where `XXX` is the issue number?

PR #41, opened 2022-07-27
Refactor write segmentation V (add fluorescence)

weiglszonja:
## Motivation

Refactor write segmentation with creating an `add_fluorescence` method to add fluorescence traces to nwbfile

Chained to  #23, it depends on using the `add_plane_segmentation` method from #23

work in progress

## How to test the behavior?

tests are in progress

## Checklist


src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -371,6 +374,96 @@ def image_mask_iterator():
...
+    roi_response_dict = {key: value for key, value in roi_response_dict.items() if np.any(value)}
+
+    # Create fluorescence
+    fluorescence = Fluorescence(name=fluorescence_name)
+
+    rate = (
+        np.nan
+        if segmentation_extractor.get_sampling_frequency() is None
+        else float(segmentation_extractor.get_sampling_frequency())
+    )
```
weiglszonja, line 440: note for testing:
if `segmentation_extractor._sampling_frequency` is `None` then `segmentation_extractor.get_sampling_frequency() `would fail because of float casting the value see:

https://github.com/catalystneuro/roiextractors/blob/1f8a479e90a45595de661ad339fa53fc022294e2/src/roiextractors/segmentationextractor.py#L229-L237

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -371,6 +374,96 @@ def image_mask_iterator():
     return nwbfile
 
 
+def add_fluorescence(segmentation_extractor: SegmentationExtractor, nwbfile: NWBFile, metadata: dict) -> NWBFile:
```
h-mayorquin, line 377: I think this should be called `add_fluoresence_traces`. Adding the Fluoresence MultiContinaer is a only a small part of what this function does. What do you think?

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -371,6 +374,96 @@ def image_mask_iterator():
...
+    # Get plane segmentation from the image segmentation
+    image_segmentation_metadata = metadata_copy["Ophys"]["ImageSegmentation"]
+    image_segmentation_name = image_segmentation_metadata["name"]
+    image_segmentation = ophys.get_data_interface(image_segmentation_name)
+
+    plane_segmentation_name = image_segmentation_metadata["plane_segmentations"][0]["name"]
+    plane_segmentation = image_segmentation.plane_segmentations[plane_segmentation_name]
+
+    # TODO what should be done with plane_no_loop?
+    plane_no_loop = 0
```
h-mayorquin, line 418: While we decided what to do with the multi segmentation extractor case `plane_no_loop` should be passed to the function. The `plane_no_loop` is used to determine the name of the objects and it varies with the outermost loop of `write_segmentation`.  So, to leave the behavior consistent with what we had before we should still name the objects with a variable `plane_no_loop` I think.

In the PR concerning add_summary_images I pass the name that already includes the effects of a variable `plane_no_loop` check that for a different approach as well.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -371,6 +374,96 @@ def image_mask_iterator():
...
+        description=f"region for Imaging plane{plane_no_loop}",
+    )
+
+    # Get traces from the segmentation extractor
+    roi_response_dict = segmentation_extractor.get_traces_dict()
+
+    # Filter empty data
+    roi_response_dict = {key: value for key, value in roi_response_dict.items() if value is not None}
+    # Filter all zero data
+    roi_response_dict = {key: value for key, value in roi_response_dict.items() if np.any(value)}
```
h-mayorquin, line 431: This is a neat trick for `np.any`!

I was wondering on how does this compare against the previous generator solution proposed by @CodyCBakerPhD in terms of speed and I got the following:

```
n [11]: timeit.timeit("np.any(np.ones(int(1e6)))", setup="import numpy as np",
    ...: number=100)
Out[11]: 0.1355249520020152

In [12]: timeit.timeit("any(x!=0 for x in np.ones(int(1e6)))", setup="import num
    ...: py as np", number=100)
Out[12]: 0.0691819530002249
```

So it seems that if you expected the traces to have mostly non-zero data most of the time then @CodyCBakerPhD solution seems better. However, if a lot of the times the tested vectors are going to have mostly zeros we have the opposite result:

```
In [16]: timeit.timeit("np.any(np.zeros(int(1e6)))", setup="import numpy as np",
    ...:  number=100)
Out[16]: 0.11136360900127329

In [17]: timeit.timeit("any(x!=0 for x in np.zeros(int(1e6)))", setup="import nu
    ...: mpy as np", number=100)
Out[17]: 21.058059448998392
```
Where we can see that the generator solution does in fact quite bad. So I don't know what gives, what do you guys think?

tests/test_internals/test_roi_extractors_module.py
```diff
@@ -481,6 +482,140 @@ def test_add_two_plane_segmentation(self):
...
+            nwbfile=self.nwbfile,
+            metadata=self.metadata,
+        )
+
+        fluorescence = self.nwbfile.processing["ophys"].get(self.fluorescence_name)
+
+        self.assertEqual(fluorescence.name, self.fluorescence_name)
+        self.assertEqual(len(fluorescence.roi_response_series), len(self.segmentation_extractor.get_traces_dict()))
+
+    def test_add_fluorescence_one_of_the_traces_is_none(self):
```
h-mayorquin, line 560: The `dummy_segmentation_extractor` has functionality for building examples where the traces are not available so you don't need to set the private attribute to None here:

https://github.com/catalystneuro/roiextractors/blob/1f8a479e90a45595de661ad339fa53fc022294e2/src/roiextractors/testing.py#L54-L66

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -371,6 +374,96 @@ def image_mask_iterator():
...
+    # Set the defaults and required infrastructure
+    metadata_copy = deepcopy(metadata)
+    default_metadata = get_default_ophys_metadata()
+    metadata_copy = dict_deep_update(default_metadata, metadata_copy, append_list=False)
+
+    fluorescence_metadata = metadata_copy["Ophys"]["Fluorescence"]
+    fluorescence_name = fluorescence_metadata["name"]
+
+    ophys = get_module(nwbfile, "ophys")
+    if fluorescence_name in ophys.data_interfaces:
```
h-mayorquin, line 404: What if the fluorescence module already exists but there are still some traces to add?

h-mayorquin (commented):
This is the last one and now the function looks really structured : D

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: It's true, all that logic in numpy is vectorized so it's ideal for algebraic operations on large arrays that are contiguous in memory, but has no early stop condition. So lazy early return generators will behave better whenever the non-zero data density is somewhat high, and arrays will be better for when the non-zero data is very sparse.

Also the generator is lazy in memory so if the `value` of that response series is very large we essentially have iterative read taken care of by default whereas numpy would load it all in at once - how large are these data usually? I guess it may not be too bad (`frames x number_of_rois` I'd guess?). I'd think this is actually the more important consideration and the early return is just a bonus.

So which is better for speed depends on the natural statistics of this application - are the response series we want to ignore predominantly zero with sparse non-zero entries or are they pretty densely populated with non-zero? I guess either way the worst case time for a full array of zero makes numpy the preferred approach dependent on the memory consideration above.

h-mayorquin: I thought about the lazy point as well but right now in roiextractor the response series (traces) are already preloaded numpy arrays so I don't think the point applies yet. I agree that if at some point the return of the series are memmaps or any other lazyly loaded object and the generator reads bit by bit then that would be the most important consideration.

CodyCBakerPhD: Ah, fair enough then. I'd have thought it was possible (for some of our current conversions at least) to treat the segmentation data lazily

h-mayorquin: What you say would work best in the ideal case though. I think that's a good point. Let's see what @weiglszonja thinks.

weiglszonja: Thank you for doing the comparisons @h-mayorquin, I expected the numpy approach to be faster for arrays where the non-zero data is sparse, I did not consider the lazy point though. Thanks @CodyCBakerPhD for explaining it. I guess most of the time we would not expect nearly all-zero traces, then the generator approach is way better. I'd go with that approach here instead of numpy.

tests/test_internals/test_roi_extractors_module.py
weiglszonja: ah thank you!

src/neuroconv/tools/roiextractors/roiextractors.py
weiglszonja: ah I see your point, the early return makes no sense here. I should also add a check to not add those traces again if they already exist in fluorescence.

src/neuroconv/tools/roiextractors/roiextractors.py
h-mayorquin: One more suggestion over here, maybe it would be good to change the key: value variables to something more descriptive? `trace_name` and `trace` is what comes to mind but if you come with something better that's great.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -371,6 +374,108 @@ def image_mask_iterator():
...
+    # Create fluorescence container
+    if fluorescence_name in ophys.data_interfaces:
+        fluorescence = ophys.get_data_interface(fluorescence_name)
+    else:
+        fluorescence = Fluorescence(name=fluorescence_name)
+        # Add fluorescence to the ophys module
+        ophys.add(fluorescence)
+
+    # Get traces from the segmentation extractor
+    roi_response_dict = segmentation_extractor.get_traces_dict()
```
weiglszonja, line 439: open for discussion:
`get_traces_dict()` returns the `dff` signal if exists,
to add `dff` signal to the `DfOverF` object instead of `Fluorescence` container, lines from 452 to 474 should be refactored in a way e.g.:
- create a private method that returns a single roi response series based on its name --> then this can be added to a `DfOverF` object

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -371,6 +389,147 @@ def image_mask_iterator():
...
+            if trace_name not in dff.roi_response_series:
+                response_series_metadata = dff_metadata["roi_response_series"]
+                trace_meta = [trace_metadata for trace_metadata in
+                              response_series_metadata if
+                              trace_name in trace_metadata["name"]]
+                if trace_meta:
+                    roi_response_series_kwargs.update(**trace_meta[0])
+
+                dff.create_roi_response_series(**roi_response_series_kwargs)
+
```
weiglszonja, line 508: @h-mayorquin I'd need some feedback, I understand that this is not ideal this way, and I should still move some code around because the method is called "add_fluorescence_traces" and I'm doing the a lot more here. May I ask for your suggestion how to reduce this duplication?

CodyCBakerPhD: The first if/else block is similar in structure to our nwb_helper function for getting modules, maybe something similar for getting NWB data interfaces could help with that?

CodyCBakerPhD: Otherwise another approach to flattening things like this is to add an early return statement of opposite logic, such as

```
if trace_name in dff.reaponse_series:
    continue  # continue if iteration, return if function 
```

and then all the logic of the positive case can follow in the flattened outer level

h-mayorquin: Hi, sorry for not coming to this before after you pined me. 
My suggestion to eliminate the duplicate code and make the control structure flatter is twofold:

1) Take the logic for extracting the containers out of the loop over the traces.
2) Use the ternary operator (can also be done with default dicts) to assign metadata and containers correctly.

Code:
```python
    # Logic for extracting the containers
    if "Dff" in traces_to_add.keys():
        # Create df/F container
        if dff_name in ophys.data_interfaces:
            dff = ophys.get_data_interface(dff_name)
        else:
            dff = DfOverF(name=dff_name)
            # Add df/F container to the ophys module
            ophys.add(dff)

    remaining_traces = [trace_name for trace_name in traces_to_add.keys() if trace_name != "Dff"]
    if remaining_traces:
        # Create fluorescence container
        if fluorescence_name in ophys.data_interfaces:
            fluorescence = ophys.get_data_interface(fluorescence_name)
        else:
            fluorescence = Fluorescence(name=fluorescence_name)
            # Add fluorescence container to the ophys module
            ophys.add(fluorescence) 

    for trace_name, trace in traces_to_add.items():

        # Extract the response series metadata
        trace_name = "RoiResponseSeries" if trace_name == "raw" else trace_name.capitalize()
        trace_name = trace_name if plane_index == 0 else trace_name + f"_Plane{plane_index}"
        
        metadata = dff_metadata if trace_name == "Dff" else fluorescence_metadata
        response_series_metadata = metadata["roi_response_series"] 
        trace_metadata = next(trace_metadata for trace_metadata in response_series_metadata if trace_name==trace_metadata["name"])    
        
        # Build the roi response series
        roi_response_series_kwargs.update(**trace_metadata)
        roi_response_series = RoiResponseSeries(**roi_response_series_kwargs)
        
        # Add it to the container
        container = dff if trace_name == "Dff" else fluorescence
        container.add_roi_response_series(roi_response_series)
```

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -371,6 +389,147 @@ def image_mask_iterator():
...
+    plane_segmentation_name = image_segmentation_metadata["plane_segmentations"][0]["name"]
+    plane_segmentation = image_segmentation.plane_segmentations[plane_segmentation_name]
+
+    # Create a reference for ROIs from the plane segmentation
+    roi_table_region = plane_segmentation.create_roi_table_region(
+        region=segmentation_extractor.get_roi_ids(),
+        description=f"region for Imaging plane{plane_index}",
+    )
+
```
h-mayorquin, line 463: I think that everything from lines 450 to 463 could be a private method to `_create_table_region` returning the `roi-table_region` that would make the logic clearer.

h-mayorquin (commented):
I have two general comments:
1) Are you testing that the DF/F series is not written anywhere else?
2) Following the logic of the spikeinterface module (https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/spikeinterface/spikeinterface.py) I was adding the functions with the following structure.
- Functions used for `write_imaging`
- `write_imaging` function.
` Functions used for `write_segmentation`
- `write_segmentation` function.

Whereas I see that you are adding these functions before the `write_imaging` method. How are you thinking about? Maybe you are writing the functions in a similar way in which the structures appear on the Ophys schema?
https://nwb-schema.readthedocs.io/en/latest/format.html#optical-physiology

How are you thinking about it?

src/neuroconv/tools/roiextractors/roiextractors.py
weiglszonja: Thank you for the suggestions @CodyCBakerPhD and @h-mayorquin , I wonder how it can be done with `defaultdicts`?

src/neuroconv/tools/roiextractors/roiextractors.py
weiglszonja: Yes, I'll add tests for df/f traces as well -- currently the tests are failing because the logic for adding them was not final, but of course I won't forget to test is as well. I moved around method so it matches the logic in 2. as it should.

src/neuroconv/tools/roiextractors/roiextractors.py
h-mayorquin: Something like this:
```python
from collections.import defaultdict

trace_name_to_container = defaultdict(lambda: fluorescence, "Dff"=dff)
...
...
...
container = trace_name_to_container[trace_name]
container.add_roi_response_series(roi_response_series)
```

CodyCBakerPhD: @h-mayorquin I'm unable to get that example code to run

h-mayorquin: They were not meant to run, just to illustrate the idea. What is it failing?

CodyCBakerPhD: The `defaultdict(lambda: fluorescence, "Dff"=dff)` part. `SyntaxError: expression cannot contain assignment, perhaps you meant "=="?`

h-mayorquin: Ah, what's wrong is the string in "Dff", working example:
```python
In [1]: from collections import defaultdict

In [2]: x = defaultdict(lambda: "default_case", defined_key="defined_value")

In [3]: x["random"]
Out[3]: 'default_case'

In [4]: x["defined_key"]
Out[4]: 'defined_value'

```

CodyCBakerPhD: Gotcha, thanks

weiglszonja: thanks @h-mayorquin

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -586,6 +600,170 @@ def image_mask_iterator():
...
+
+def _create_roi_table_region(
+    segmentation_extractor: SegmentationExtractor,
+    nwbfile: NWBFile,
+    metadata: dict,
+    plane_index: int,
+):
+    """Private method to create ROI table region."""
+    add_plane_segmentation(segmentation_extractor=segmentation_extractor, nwbfile=nwbfile, metadata=metadata)
```
weiglszonja, line 728: This is only making sure that there exists a plane segmentation (if not create one with that call) where we can reference the ROIs. We are not passing the compression/iterator options here though.

weiglszonja:
thank you so much for your comments so far @h-mayorquin and @CodyCBakerPhD, they were really useful! It has become a huge PR with these changes, let me know what you think about tests -- or if there is anything else to change.

CodyCBakerPhD:
> thank you so much for your comments so far @h-mayorquin and @CodyCBakerPhD, they were really useful! It has become a huge PR with these changes, let me know what you think about tests -- or if there is anything else to change.

Majority of changelog is due to new tests, so that's fine. As long as relevant code changes are below +/- 200 I can parse it, thanks for working on this!

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -586,6 +600,170 @@ def image_mask_iterator():
...
+            nwbfile=nwbfile,
+            data_interface_name=fluorescence_name,
+        ),
+        # we expect that the df/F trace name is "Dff"
+        Dff=_get_segmentation_data_interface(
+            nwbfile=nwbfile,
+            data_interface_name=dff_name,
+        ),
+    )
+
```
h-mayorquin, line 692: Right now, the function call which is the first argument of the defaultdict is called every time you use the dict. If you call the `_get_segmentation_data_interface` and then assign the value to the dict you can avoid this repeated call.

tests/test_internals/test_roi_extractors_module.py
```diff
@@ -482,6 +482,289 @@ def test_add_two_plane_segmentation(self):
...
+            DfOverF=dict(
+                name=self.dff_name,
+                roi_response_series=[self.dff_roi_response_series_metadata],
+            )
+        )
+
+        self.metadata["Ophys"].update(fluorescence_metadata)
+        self.metadata["Ophys"].update(dff_metadata)
+
+        self.ophys = get_module(self.nwbfile, "ophys")
```
h-mayorquin, line 555: Having this in the setup means that all your tests are done in an environment where the ophys module is always available, isn't not having this module available some situation that we want to test for?

tests/test_internals/test_roi_extractors_module.py
```diff
@@ -482,6 +482,289 @@ def test_add_two_plane_segmentation(self):
...
+            self.segmentation_extractor.get_sampling_frequency(),
+        )
+
+        traces = self.segmentation_extractor.get_traces_dict()
+
+        assert_array_equal(fluorescence["RoiResponseSeries"].data, traces["raw"].T)
+        assert_array_equal(fluorescence["Deconvolved"].data, traces["deconvolved"].T)
+        assert_array_equal(fluorescence["Neuropil"].data, traces["neuropil"].T)
+
+    def test_add_df_over_f_trace(self):
```
h-mayorquin, line 601: Given that in your previous test you are checking that the dff is not added to the fluorescence container as a roi response, I wonder what do you think of the following: For this test, having a segmentation extractor with only dff and testing that the fluorescence container is non created in this case.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -586,6 +600,170 @@ def image_mask_iterator():
     return nwbfile
 
 
+def add_fluorescence_traces(
+    segmentation_extractor: SegmentationExtractor,
+    nwbfile: NWBFile,
+    metadata: Optional[dict],
+    plane_index: int = 0,
```
h-mayorquin, line 607: This is called `plane_segmentation_index` in the other functions. I also realized that in the top level function `write_segmentation` we have yet another name called `plane_no` (!).

h-mayorquin (commented):
I took a look at the function `write_segmentation` and it looks very nice now with specific calls and looks very short and organized. Don't forget to add the contributions to the CHANGELOG file.

Three comments:
1) There seems to be an unused object `image_segmentation` at the top level that we are no longer using as that is called inside the function.
2) Given that we are doing calling `add_image_segmentation` and `add_imaging_plane` a lot in the functions (As they both check and create) I think it would be a good idea to test that in fact only one of them is created.
3) This is way less important and I think also out of the scope of this PR but what do you think of referring to dff as something more specific such as `dff_normalized_differences` or `flouresence_delta_over_f` or `normalized_fluoresence` or be consistent with pynwb and call it `df_over_f`.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -586,6 +600,170 @@ def image_mask_iterator():
...
+
+    for trace_name, trace in traces_to_add.items():
+        # Extract the response series metadata
+        trace_name = "RoiResponseSeries" if trace_name == "raw" else trace_name.capitalize()
+        trace_name = trace_name if plane_index == 0 else trace_name + f"_Plane{plane_index}"
+
+        if trace_name in container[trace_name].roi_response_series:
+            continue
+
+        metadata = dff_metadata if trace_name == "Dff" else fluorescence_metadata
```
h-mayorquin, line 701: Isn't the trace_name `dff` and not `Dff` which is what you are using for the data interface name?

src/neuroconv/tools/roiextractors/roiextractors.py
weiglszonja: the reason why I tried this is because given this example:

```
if "dff" in traces_to_add:
    dff = _get_segmentation_data_interface("DfOverF")
if remaining_traces:
    fluorescence = _get_segmentation_data_interface("Fluorescence")
x = defaultdict(lambda: fluorescence, Dff=dff) 
```
this will raise an error that local variable `dff` is referenced before assignment in those cases where there are no df over f trace to add. if you have an idea how to overcome this, I'm happy to change it!

tests/test_internals/test_roi_extractors_module.py
weiglszonja: yeah, you're right, I just wanted to reduce code repetition.

tests/test_internals/test_roi_extractors_module.py
weiglszonja: yes, that makes sense.

src/neuroconv/tools/roiextractors/roiextractors.py
weiglszonja: I'll check,

src/neuroconv/tools/roiextractors/roiextractors.py
weiglszonja: Yeah, I can see why that is confusing. I was using the previous intent of naming traces i.e. raw traces are saved as "RoiResponseSeries" and others are strarting capitalized. Like [here](https://pynwb.readthedocs.io/en/stable/tutorials/domain/ophys.html#sphx-glr-tutorials-domain-ophys-py).

src/neuroconv/tools/roiextractors/roiextractors.py
h-mayorquin: Something similar to updating typical dicts works. The only difference is that defaultdicts have this attribut `default_factor` that you can also set post initialization. 

A -probably non working- example of how this could go:
```
trace_to_container = defaultdict()
if remaining_traces:
   trace_to_container.default_factory = lambda: fluorescence # or lambda: call_to_your_function
if "dff" in traces_to_add:
    dff = _get_segmentation_data_interface("DfOverF")
    trace_to_container.update(Dff=dff)
```

That said, if you think that your version is more readable maybe you are right? I was just commenting on the small inefficiency of calling the function every time in the loop.

src/neuroconv/tools/roiextractors/roiextractors.py
h-mayorquin: Leave this as it is, as I think is too big already. I discussed this with @CodyCBakerPhD today and I will take care of this separately after this PR.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -57,12 +60,21 @@ def get_default_ophys_metadata():
...
+        name="RoiResponseSeries", description="array of raw fluorescence traces", unit="n.a."
     )
 
-    default_fluoresence = dict(roi_response_series=[default_roi_response_series])
+    default_fluorescence = dict(
+        name="Fluorescence",
+        roi_response_series=[default_fluorescence_roi_response_series],
+    )
+
+    default_dff_roi_response_series = dict(name="Dff", description="array of df/F traces", unit="n.a.")
```
bendichter, line 72:
```suggestion
    default_dff_roi_response_series = dict(name="RoiResponseSeries", description="array of df/F traces", unit="n.a.")
```

CodyCBakerPhD (approved):

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -690,18 +690,22 @@ def add_fluorescence_traces(
...
+        trace_to_data_interface.update(dff=df_over_f_data_interface)
 
     for trace_name, trace in traces_to_add.items():
+        # Decide which data interface to use based on the trace name
+        data_interface = trace_to_data_interface[trace_name]
         # Extract the response series metadata
-        trace_name = "RoiResponseSeries" if trace_name == "raw" else trace_name.capitalize()
+        # The name of the roi_response_series is "RoiResponseSeries" for raw and df/F traces,
+        # otherwise it is capitalized trace_name.
+        trace_name = "RoiResponseSeries" if trace_name in ["raw", "dff"] else trace_name.capitalize()
```
weiglszonja, line 701: @bendichter 
This is an important change after the last [commit](https://github.com/catalystneuro/neuroconv/pull/41/commits/72104e7810b19fd85826713a9506722ee95b1a1e), that we will write the df/F trace under the name of "RoiResponseSeries". I know this is far from ideal, before it was expected that it is "Dff" and hence the tests started failing. The complication is that we are renaming the trace names here raw->RoiResponseSeries and then we are matching the metadata with this name. But i'm not sure how to rely less on the name of the trace when assigning the metadata to it.

CodyCBakerPhD: Seems highly analogous to the role of `es_key` in the `ecephys` `add_electrical_series` function: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/spikeinterface/spikeinterface.py#L561

For any given `RoiResponseSeries`, I'd think we should therefore follow a similar practice to allow identification of which part of the metadata to use when writing a given response series using a user-definable key.

How does that sound? Would you be able to do that in a separate PR so we can at least get this one in sooner?

CodyCBakerPhD: Similarly, the assigned default name could follow the same practice of appending `RoiResponseSeries` with the type, such as `RoiResponseSeriesDfOverF` or `RoiResponseSeriesRaw`, etc. like in https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/spikeinterface/spikeinterface.py#L630

bendichter: Let's try to follow our [naming conventions](https://nwbinspector.readthedocs.io/en/dev/best_practices/general.html#naming-conventions)

If there's just one RoiResponseSeries, it should be called "RoiResponseSeries"

CodyCBakerPhD: Well sure if there's only one then the extra naming isn't necessary. The additional values would be context in the event that there are more than one

weiglszonja: Yes, I see your point, so instead of saving them as "RoiResponseSeries" and "Deconvolved" if we have raw and deconvolved traces, we'll eventually want to write as RoiResponseRaw and RoiResponseSeriesDeconvolved? but only  "RoiResponseSeries"  if only one trace is there. Yeah, it should be done in a separate PR

CodyCBakerPhD (approved):

PR #51, opened 2022-08-01
Add base metadata to base data interface

h-mayorquin:
This PR enables the assumption that the top level `NWBFile` key will be available in the metadata dict for all the interfaces. This is enabled because all our base interfaces call super in their metadata. As long as we keep this convention all the way to the bottom of the inheritance tree (i.e. the specific interfaces) then calling the `interface.get_metadata()` dict should have the `NWBFile` available.


bendichter (dismissed):

CodyCBakerPhD:
LGTM - update CHANGLOG?

bendichter:
Let's also remove the uuid generation from downstream classes, e.g. here

https://github.com/catalystneuro/neuroconv/blob/73e28849e6bde37c4a57c689297a03527323e44a/src/neuroconv/tools/neo/neo.py#L131

h-mayorquin:
> Let's also remove the uuid generation from downstream classes, e.g. here
> 
> https://github.com/catalystneuro/neuroconv/blob/73e28849e6bde37c4a57c689297a03527323e44a/src/neuroconv/tools/neo/neo.py#L131

I was thinking about removing them but then I realized that they might be necessary if we want to support use of `write_recording`, `write_sorting`, `write_imaging` , etc. without using `run_conversion`. What do you think, is that something that we want to do or is fine to assume that those functions would always be called from inside a `run_conversion` and therefore have the minimal metadata available.

bendichter:
@h-mayorquin good point. OK let's continue as-is with this

h-mayorquin:
> LGTM - update CHANGLOG?

I added now the changes to the CHANGELOG.

bendichter (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * Coerced the recording extractors with `spikeextractors_backend=True` to BaseRecording objects for Axona, Blackrock, Openephys, and SpikeGadgets. [PR #38](https://github.com/catalystneuro/neuroconv/pull/38)
 * Added function to add PlaneSegmentation objects to an nwbfile in `roiextractors` and corresponding unit tests. [PR #23](https://github.com/catalystneuro/neuroconv/pull/23)
 * `use_times` argument to be deprecated on the ecephys pipeline. The function `add_electrical_series` now determines whether the timestamps of the spikeinterface recording extractor are uniform or not and automatically stores the data according to best practices [PR #40](https://github.com/catalystneuro/neuroconv/pull/40)
+* Add `NWBFile` metadata key at the leve of the base data interface so it can always be inherited to be available. [PR #51](https://github.com/catalystneuro/neuroconv/pull/51)
```
bendichter, line 11:
```suggestion
* Add `NWBFile` metadata key at the level of the base data interface so it can always be inherited to be available. [PR #51](https://github.com/catalystneuro/neuroconv/pull/51)
```

h-mayorquin: Thanks.

h-mayorquin:
Thanks for updating the gallery, @bendichter !

bendichter:
ready to merge, @h-mayorquin?

PR #54, opened 2022-08-01
Add ImagingExtractorDataChunkIterator (Restore from catalystneuro/nwb-conversion-tools#575)

weiglszonja:
## Motivation

Restoring PR from catalystneuro/nwb-conversion-tools#575

## How to test the behavior?

Run the unittests in
```
tests/test_internals/test_imagingextractordatachunkiterator.py
```

## Checklist


bendichter:
I don't think it makes sense to add the nwbinspector requirement just for `get_package_version`.

weiglszonja:
Profiling memory usage of run_conversion with  [memory_profiler](https://pypi.org/project/memory-profiler/) to compare the performance of `DataChunkIterator` and the new `ImagingExtractorDataChunkIterator`:
Presentation1
Tried out different sizes of buffer (1GB, 3GB, 10GB) with different chunk sizes (1MB, 2000MB)
- tested on real data: running the conversion of Fee lab imaging data
- the total run time is very similar for both iterators
- used memory profiler to mark the times when the frames are actually accessed with the iterator --> they do not exceed buffer gb

weiglszonja:
cProfiler and memory_profiler files are uploaded [here](https://drive.google.com/drive/folders/1CHZjrkJ0o1hFVLVu9H_AhyIFEbttGc-r)

tests/test_imagingextractordatachunkiterator.py
```diff
@@ -70,7 +72,7 @@ def setUpClass(cls) -> None:
         chunk_shape=(13, 2, 2),
         case_name="chunk_shape_greater_than_buffer_shape",
     )
-    if get_package_version(name="hdmf").base_version >= "3.3.2":
+    if Version(version("hdmf")) >= Version("3.3.2"):
```
weiglszonja, line 75: how about this @bendichter ?

CodyCBakerPhD:
> I don't think it makes sense to add the nwbinspector requirement just for `get_package_version`.

We already have it as a secondary dependency from dandi - if using latest of both we wouldn’t even need to pin here

weiglszonja:
@CodyCBakerPhD, I just noticed the new release of `nwbinspector` so the pin is not needed here anymore indeed.

tests/test_imagingextractordatachunkiterator.py
weiglszonja: reverted for now, see @CodyCBakerPhD's comment

bendichter:
@weiglszonja I mean remove nwbinspector entirely. Don't list as a dep and don't import. 

You can just recreate [the function](https://github.com/NeurodataWithoutBorders/nwbinspector/blob/e47435195e3fa2502d8a0bd7300e1a30299cb57f/nwbinspector/utils.py#L128-L150) in this repo

I don't usually recommend copying code because this leads to code duplication, but in this case I would rather duplicate a few lines of code rather than including a dependency we really don't need.

bendichter:
@weiglszonja oh, we already have nwbinspector as a dep because we have dandi as a dep. OK, I guess this is fine then

bendichter:
ready to merge?

weiglszonja:
@bendichter yes 👍

weiglszonja:
@bendichter it should be good to go now

bendichter (approved):

src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
```diff
@@ -0,0 +1,119 @@
...
+        return (self.imaging_extractor.get_num_frames(),) + self.imaging_extractor.get_image_size()[::-1]
+
+    def _get_data(self, selection: Tuple[slice]) -> np.ndarray:
+        data = self.imaging_extractor.get_video(
+            start_frame=selection[0].start,
+            end_frame=selection[0].stop,
+        )
+        # This condition can be removed when the ImagingExtractors are not
+        # squeezing the data when the selection is a single slice
+        if len(data.shape) < len(selection):
```
h-mayorquin, line 116: Was this necessary even when `get_video`?

weiglszonja: It is probably not necessary anymore. I checked for other ImagingExtractors and now the only extractor left that calls get_frames in get_video is the `MultiImagingExtractor`. The reason why it does that with a `MultiImagingExtractor` we have multiple `ImagingExtractors`, and in a single selection of frames in `get_video` it can happen that some frames are from one extractor and some from another, and this is how I managed to do it in https://github.com/catalystneuro/roiextractors/pull/127. 

I can remove this condition and open a PR to change that behavior of squeezing the frames in `MultiImagingExtractor`.

Issue #60, opened 2022-05-05
Update remaining RecordingInterfaces to use SpikeInterface backend


CodyCBakerPhD:
- [x] IntanRecordingInterface
- [x] BlackrockRecordingExtractorInterface
- [x] BlackrockSortingExtractorInterface 
- [x] CED [Not tested in the old backend]
- [x] Cell Explorer [Not available in spikeinterface]
- [x] Neuralynx 
- [x] NeuroscopeSortingExtractor
- [x] OpenEphys
- [x] PhySortingInterface
- [x] SpikeGadgetsRecordingInterface
- [x] Axona
- [x] Axona LFP

@h-mayorquin Please add the others, thanks! (Done!)

h-mayorquin:
Axona has been added to spikeinterface:
https://github.com/SpikeInterface/spikeinterface/pull/529

PR #61, opened 2022-08-03
Add spikeinterface support to `AxonaRecordingExtractorInterface`

h-mayorquin:
This PR ports the axona interface to use `spikeinterface`.

Related to https://github.com/catalystneuro/neuroconv/issues/60. 

All the previous parsing of the metadata that was done on the interface is now outsourced to the neo side. We access the metadata through `spikeinterface` functionality which is easier and less of a maintenance burden for us.

This is the data in the header that neo provides:
https://gist.github.com/h-mayorquin/90e65fcfd459a5028ddfe9ff610bf74c

The `AxonaPositionDataInterface` still uses these parsing functions so I just moved them together with all the other auxiliary functions for that interface.

As the old spikeextractors function is also using neo behind the scenes no backwards support is offered.


CodyCBakerPhD:
While we’re at it what would you think about moving the auxiliary/helper functions you’re keeping into a separate file in this folder, like we do with certain other interfaces that have a lot of helper functions (neuroscope / movies)

h-mayorquin:
> While we’re at it what would you think about moving the auxiliary/helper functions you’re keeping into a separate file in this folder, like we do with certain other interfaces that have a lot of helper functions (neuroscope / movies)

I think that's a good idea.

docs/conversion_examples_gallery/recording/axona.rst
```diff
@@ -19,9 +19,8 @@ Convert axona data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.axo
     >>> metadata = interface.get_metadata()
     >>> # For data provenance we add the time zone information to the conversion
     >>> tzinfo = tz.gettz("US/Pacific")
-    >>> session_start_time = datetime.fromisoformat(metadata["NWBFile"]["session_start_time"])
-    >>> session_start_time = session_start_time.replace(tzinfo=tzinfo).isoformat()
-    >>> metadata["NWBFile"].update(session_start_time=session_start_time)
+    >>> session_start_time = metadata["NWBFile"]["session_start_time"]
+    >>> metadata["NWBFile"].update(session_start_time=session_start_time.replace(tzinfo=tzinfo).isoformat())
```
bendichter, line 23:
```suggestion
    >>> metadata["NWBFile"].update(session_start_time=session_start_time.replace(tzinfo=tzinfo))
```

bendichter (dismissed):

h-mayorquin:
> While we’re at it what would you think about moving the auxiliary/helper functions you’re keeping into a separate file in this folder, like we do with certain other interfaces that have a lot of helper functions (neuroscope / movies)

@CodyCBakerPhD 
Done.|
I also did the change the Ben requested above to the conversion gallery.

src/neuroconv/datainterfaces/ecephys/axona/axona_utils.py
```diff
@@ -0,0 +1,399 @@
...
+    Creates datetime object (y, m, d, h, m, s) from .set file header
+    and converts it to ISO 8601 format
+    """
+    with open(set_file, "r", encoding="cp1252") as f:
+        for line in f:
+            if line.startswith("trial_date"):
+                date_string = line[len("trial_date") + 1 :].replace("\n", "")
+            if line.startswith("trial_time"):
+                time_string = line[len("trial_time") + 1 :].replace("\n", "")
+    return dateutil.parser.parse(date_string + " " + time_string).isoformat()
```
bendichter, line 166: why not just return a datetime object?

h-mayorquin: I think this was done at a time when the metadata did not accept datetime yet. This is just a function that I moved at the request of @CodyCBakerPhD and is only used in the PositionDataInterface that I did not modify. Do you want me to take the opportunity in this PR to change it?

CodyCBakerPhD: I'd say change in subsequent PR - but the original reasoning is correct, this is from before when we had to cast all datetimes to ISO strings in the metadata structure, which is no longer required

h-mayorquin: OK.

CodyCBakerPhD:
@h-mayorquin Is this good to go?

h-mayorquin:
> @h-mayorquin Is this good to go?

Yes.

CodyCBakerPhD (approved):

PR #62, opened 2022-08-04
add viewcode sphunx extension

bendichter:


CodyCBakerPhD:
Looks good!

bendichter:
I found myself wanting this to see if ScanImage pulls the session start time into metadata. I feel like it would be useful for our users in general

CodyCBakerPhD:
It would but remember it was pointed out to us at the workshop that the ScanImage TIF examples we have right now do not have metadata fields that appear as they do in actual files. I’ll ping Hannah again to see if she can send us some better data for GIN

bendichter:
ok well the ScanImage reader is robust to if this field is missing so I think we are in good shape but yes lets try to get ahold of the latest metadata format

PR #64, opened 2022-08-05
Add unit tests for add electrical series III (compression)

h-mayorquin:
Following on #32 for testing compression options in `add_electrical_series` in our spikeinterface module.

I created a function to wrap the data into H5DataIO as that functionality is likely to present itself multiple times. The function is a simple wrapper that just tests that the parameters make sense so it is just moving the code that was previously present in `add_electrical_series` to its own function so other use cases can benefit from it.


src/neuroconv/tools/nwb_helpers.py
```diff
@@ -115,6 +118,58 @@ def add_device_from_metadata(nwbfile: NWBFile, modality: str = "Ecephys", metada
...
+    H5DataIO
+        The wrapped data in an H5DataIO object.
+    """
+
+    valid_options_for_compression = [
+        "gzip",
+        "lzf",
+        None,
+        True,
+    ]
```
bendichter, line 151: Why are you limiting to these values?

bendichter:
It seems like these checks would be better inside `H5DataIO`

src/neuroconv/tools/nwb_helpers.py
bendichter: See https://github.com/NeurodataWithoutBorders/pynwb/pull/1529

bendichter:
@h-mayorquin this PR would make it impossible to use dynamically loaded filters. I was searching for the tutorial on DLFs to point you to it, and I realized that we don't have one! So I just wrote it. Take a look. https://github.com/NeurodataWithoutBorders/pynwb/pull/1529

CodyCBakerPhD:
> It seems like these checks would be better inside `H5DataIO`

It appears several of them already are: https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/backends/hdf5/h5_utils.py#L528 -> https://github.com/hdmf-dev/hdmf/blob/7be40229d778dea940134753de957f2457a11e1f/src/hdmf/backends/hdf5/h5_utils.py#L572-L573 (for "gzip", for example - also several other options for other filters I didn't know about)

src/neuroconv/tools/nwb_helpers.py
h-mayorquin: I took it from existing code and added None and True options.

h-mayorquin:
> @h-mayorquin this PR would make it impossible to use dynamically loaded filters. I was searching for the tutorial on DLFs to point you to it, and I realized that we don't have one! So I just wrote it. Take a look. [NeurodataWithoutBorders/pynwb#1529](https://github.com/NeurodataWithoutBorders/pynwb/pull/1529)

@bendichter 
Thanks for this. Yes, having this auxiliary function would interfere with assigning compression. 

I just took the code we previously had in the function and then wrongly assumed that if we had those checks here they were not performed upstream (thanks @CodyCBakerPhD for pinpointing them). Given that those checks are performed already by `H5DataIO`, I think that we can get rid of this auxiliary function and just test that the arguments are correctly propagated by the `add_electrical_series` function. What do you guys think?

CodyCBakerPhD:
> Given that those checks are performed already by H5DataIO, I think that we can get rid of this auxiliary function and just test that the arguments are correctly propagated by the add_electrical_series function. What do you guys think?

Yeah, `compression` and `compression_options` are both just intended to be propagated directly into the `H5DataIO`

h-mayorquin:
OK, I eliminated the raw and implemented just propagation tests. This should be ready to go.

CodyCBakerPhD (approved):

PR #65, opened 2022-08-05
Forgotten updates to the CHANGELOG and small reorganization of roiextractors module

h-mayorquin:
This a small administrative PR that covers two points:

1. Two PRs (adding Neuralynx sorting data and enabling source code viewing) did not contain the corresponding modifications to the changelog. 
2. Moves the function `add_plane_segmentation` in the roiextractors module together with all the other `write_segmentation` functions (right now, it is on the write_imaging area) so as to keep the functions for the roiextractors module thematically organized together.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,12 @@
 
 ## Documentation and tutorial enhancements:
 * Unified the documentation of NeuroConv structure in the User Guide readthedocs. [PR #39](https://github.com/catalystneuro/neuroconv/pull/39)
+* Added package for viewing source code in the neuroconv documentation [PR #62](https://github.com/catalystneuro/neuroconv/pull/62)
 
 ### Features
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
@h-mayorquin good catch, thanks. Is there anything I can do at the administrative level to try to add more reminders for the new CHANGELOG update protocol (per-PR)? I thought I had added it to the PR template but I know I'm also guilty of overriding all that sometimes myself

h-mayorquin:
I am not sure and I was thinking about the same. I think this results mainly as a consequence of the CHANGELOG being something that you mostly write by the end of PR. 

Something probably too strong would be to require to pass a small test that detects if changes were made to the CHANGELOG. Maybe something with commit hooks? not sure. I will check out if I see something.

PR #66, opened 2022-08-05
Refactor add electrical series

h-mayorquin:
This is a small refactor of the electrical series name logic in `add_electrical_series` that I need previous to the PR adding multi segment support. As the latter will need a different name for each of the segments and I did not want to have two big changes in the same PR.

In addition, two set of assertions that are already handled by pynwb and are redundant here are removed.


bendichter (approved):

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -584,54 +582,26 @@ def add_electrical_series(
         warn(f"compression_opts ({compression_opts}) were passed, but compression type is 'lzf'! Ignoring options.")
         compression_opts = None
 
+    default_name = f"ElectricalSeries_{write_as}"
```
CodyCBakerPhD, line 585: @h-mayorquin We added a best practice a while back that naming conventions of classes ought to follow CamelCase (we also tend to use it ourselves whenever possible). This mixture of CamcelCase and snake_case (which even goes against ADA 95 suggestions for combining the two) has always looked a bit awkward - how would you feel about adjusting this to `f"ElectricalSeries{write_as.capitalize()}"`?

h-mayorquin: Yes, sure. I think it makes sense. We should uphold our own conventions whenever possible.

PR #67, opened 2022-08-05
Add support for writing multi-segment recording extractors

### LINKED ISSUE #18: Add support for multiple segments recorders in `add_electrical_series`
Right now the `get_traces` method within the `add_electrical_series` function located in the spikeinterface module only support the case when the recording is mono-segment. 

However, some of the data that we have in gin is multi-segment (specifically I know it is the case for Neuralynx).  What we are doing right now is to sub-segment the recorder and only propagate one segment:

https://github.com/catalystneuro/nwb-conversion-tools/blob/ca17dd2146b883d60d641090772f236da1a39869/src/nwb_conversion_tools/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py#L85-L87

We should modify the function `add_electrical_series` to account for this and write multiple segments.

I was discussing with @bendichter today and for this simple cases  we could just iterate over every segment and write each as its own electrical series.

h-mayorquin:
This should be done after #66.
This should close #18 once is done.


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -667,6 +668,7 @@ def add_electrical_series(
     rate = calculate_regular_series_rate(series=timestamps)  # Returns None if it is not regular
 
     if rate:
+        starting_time = starting_time + timestamps[0]
```
h-mayorquin, line 671: It remains to be discussed what to do with the `starting_time` parameter in the case of multiple segment extractors. Right now, this parameter is working as a global shift for the timestamps across all segments.

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -997,24 +997,35 @@ def write_recording(
         metadata = dict_deep_update(recording.nwb_metadata, metadata)
     elif metadata is None:
         metadata = get_nwb_metadata(recording=recording)
+
     with make_or_load_nwbfile(
         nwbfile_path=nwbfile_path, nwbfile=nwbfile, metadata=metadata, overwrite=overwrite, verbose=verbose
     ) as nwbfile_out:
-        add_all_to_nwbfile(
```
h-mayorquin, line 1003: This functions is working as a wrapper for 4 statements. 
As I see it, it is just adding docstring maintenance and lines costs whereas not really providing any useful abstraction and in fact being opaque on its purpose ( not clear what "all" is on "add_all_to_nwbfile"). 

I am writing the whole logic of `write_segmentation` bare at the function as I think is easier to understand and maintain in this way, let me know if I am missing something.

Relevant:
https://matthewrocklin.com/blog/work/2019/06/23/avoid-indirection

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -582,10 +582,11 @@ def add_electrical_series(
         warn(f"compression_opts ({compression_opts}) were passed, but compression type is 'lzf'! Ignoring options.")
         compression_opts = None
 
-    default_name = f"ElectricalSeries_{write_as}"
+    segment_name_signature = "" if checked_recording.get_num_segments() == 1 else segment_index
+    default_name = f"ElectricalSeries{segment_name_signature}_{write_as}"
```
h-mayorquin, line 586:
```suggestion
    default_name = f"ElectricalSeries{segment_name_signature}{write_as.capitalize()}"
```
@CodyCBakerPhD 
Let me know if this makes sense.

src/neuroconv/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Yeah that's fine

src/neuroconv/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: I'd reverse the order - example output: `ElectricalSeriesRaw0`, `ElectricalSeriesProcessed3`. Again from https://nwbinspector.readthedocs.io/en/dev/best_practices/general.html#naming-conventions, though not stated explicitly but numbers likely ought to occur at the end of the name.

h-mayorquin: I am fine with that convention but I would like to do it in another PR as it breaks some tests non related to this feature.

src/neuroconv/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Sooo just to clarify, how are the timestamps from the later segments handled by spikeinterface?

But yeah the intention of the global `starting_time` is to shift all acquisition times by that offset typical of cross-device synchronization.

h-mayorquin: Every segment has either timestamps or t_start assigned to them. If they don't, they produce time by using the sampling frequency and the number of frames.

CodyCBakerPhD: >  If they don't, they produce time by using the sampling frequency and the number of frames.

And applying the shift of subsequent segments under the assumption that all previous ones were contiguous?

h-mayorquin: No. No time shifting is done if neither of those things is provided. So segments with equal frames would produce identical times with `get_times()` in the absence of both times and t_start.

CodyCBakerPhD: But isn't the notion of a 'segment' in neo to represent non-simultaneous recording intervals?

If multiple segments like this share start times, it would result in incorrect data being written by default to the NWBFile (granted, so would assuming perfectly contiguous data), which I would like to avoid.

In practice, do format-specific multi-segment IO's like Neuralynx track and set these start times?

h-mayorquin: What do you mean they share?
Just to clairfy, the 't_start' is different and is not the one that is set with the argument here. 

Concerning your question at the end, what I have seen is that they do.

CodyCBakerPhD: > What do you mean they share?

From what you said before

> So segments with equal frames would produce identical times with get_times()

so they 'share' the the same times even though they represent disjoint non-simultaneous recordings.

> Concerning your question at the end, what I have seen is that they do.

What we should do then is probably at least check and throw a warning if the segments have unset/default/identical start times. But in practice, a supported format ought to set it appropriately every time so we can write the correct starting times to NWB.

h-mayorquin: Ah, yes, the confusion arose because in spikeinterface that case corresponds to recordings that don't have times so there is nothing to share. But yes, I agree it does not make sense for two segments of a multi segment extractor to produce the same result with `get_times()`.

CodyCBakerPhD: Cool, and we can test for that here right?

CodyCBakerPhD: Do you think it should just issue a warning or outright throw an error?

h-mayorquin: I rather tests and issue warnings for this on `spikeinterface`. I will discuss this with Samuel once he is back.

CodyCBakerPhD (approved):

PR #68, opened 2022-08-05
Small re-organization of writing recording and sorting extractors tests

h-mayorquin:
Now that we have move further to have more isolated testing with in the spikeinterface module I think it would be a good idea to separate the previous tests that were added mainly to support writing extractors from `spikeextractors`  (see https://github.com/catalystneuro/nwb-conversion-tools/issues/182) into its own testing module. This PR does that by creating a new testing module called `test_spikeextractors_to_nwb` which contains all the old tests and keeps the new tests on `test_si`.

There are two advantages of this: 
1) It would divide the tests for the function that they are providing (testing writing `spikeinterface` objects vs testing writing `spikeextractors` objects)
2) The test in the `test_si` module are way more isolated and unittest-like so they run faster. It is therefore faster to discover a failure without the need to run the slow tests of the spikeextractors which are some of the slowest tests onr our suite (see https://github.com/catalystneuro/nwb-conversion-tools/issues/449)


CodyCBakerPhD:
Sure, fine by me. Will make cleanup easier when we eventually retire it altogether

CodyCBakerPhD (approved):

PR #70, opened 2022-08-05
Add TDT data interface

### LINKED ISSUE #13: reading TDT
Can we support reading TDT data? It looks like this is not yet supported by SpikeInterface, but is supported by python-neo.

h-mayorquin:
This should close #13 
This needs #67 before it can be merged and has that PR as a base.

There does not seem to be additional metadata that we can extract for this recorder.
I will add a conversion example gallery in a soon to come PR.

I am wondering, the data in the gin repository seem to be LFPs, should this be an LFP data interface?
Maybe someone who knows more about this data format can weight in?
Here are the information of some of the channels form the neo headers.
```
array([("b'LFPs' 1", '1', 595.46496582, 'int16', 'V', 1., 0., '0'),
       ("b'LFPs' 2", '2', 595.46496582, 'int16', 'V', 1., 0., '0'),
       ("b'LFPs' 3", '3', 595.46496582, 'int16', 'V', 1., 0., '0'),
       ("b'LFPs' 4", '4', 595.46496582, 'int16', 'V', 1., 0., '0'),
       ("b'LFPs' 5", '5', 595.46496582, 'int16', 'V', 1., 0., '0'),
       ("b'LFPs' 6", '6', 595.46496582, 'int16', 'V', 1., 0., '0'),
       ("b'LFPs' 7", '7', 595.46496582, 'int16', 'V', 1., 0., '0'),
       ("b'LFPs' 8", '8', 595.46496582, 'int16', 'V', 1., 0., '0'),
```


src/neuroconv/datainterfaces/ecephys/tdt/tdtdatainterface.py
```diff
@@ -0,0 +1,34 @@
...
+        Parameters
+        ----------
+        folder_path: str or Path
+            Path to the edf file
+        verbose: bool, True by default
+            Allows verbose.
+        """
+        super().__init__(
+            folder_path=folder_path,
+            stream_id="0",
```
bendichter, line 27:
```suggestion
    def __init__(self, folder_path: FolderPathType, stream_id: Optional[str] = None, verbose: bool = True):
        """
        Load and prepare data for TDT

        Parameters
        ----------
        folder_path: str or Path
            Path to the TDT file
        stream_id: str, optional
            Select from multiple streams.
        verbose: bool, True by default
            Allows verbose.
        """
        super().__init__(
            folder_path=folder_path,
            stream_id=stream_id,
```

src/neuroconv/datainterfaces/ecephys/tdt/tdtdatainterface.py
```diff
@@ -0,0 +1,34 @@
+"""Author: Heberto Mayorquin"""
+
```
bendichter, line 2:
```suggestion
from typing import Optional
```

bendichter (changes requested):

src/neuroconv/datainterfaces/ecephys/tdt/tdtdatainterface.py
```diff
@@ -0,0 +1,40 @@
...
+        folder_path: str or Path
+            Path to the TDT file
+        stream_id: str, optional
+            Select from multiple streams.
+        verbose: bool, True by default
+            Allows verbose.
+        """
+
+        if stream_id is None:
+            stream_id = "0"  # LFP for gin data. Other streams seem non-electrical.
```
bendichter, line 29: Why aren't you setting this as a default value in the init?

h-mayorquin: That indeed would be better. It slipped my mind.

bendichter:
@h-mayorquin is this ready for review?

CodyCBakerPhD:
@h-mayorquin Probably the easiest way to answer the LFP question would be to share an NWBFile created from this branch using the example data

h-mayorquin:
> @h-mayorquin Probably the easiest way to answer the LFP question would be to share an NWBFile created from this branch using the example data
@CodyCBakerPhD 

Sure, I shared it.
But what would you check to see if it should be LFP? The sampling rate is above there (595.46) is there something else that you would check?

CodyCBakerPhD:
@h-mayorquin Thanks for sharing. A few items of note from that file (though not all of these belong in this PR, but can you fix in subsequent ones?)

1. Device name is still `Device_ecephys`. Can you update that to CamelCase no underscore?
2. Similarly for `ElectricalSeries` name
3. So, I guess our approach to the TDT is much like our approach to SpikeGLX - if 'LFP' is indeed the signal that is acquired by the device, then it is treated as the raw signal (no change needed here by default), and it is sufficient to indicate the channel name as LFP (also already taken care of here). Even though it is LFP data, it is not 'processed' from a stored raw signal so it does not belong in processing. Just taking that from our previous attitude on SpikeGLX LF channels.
4. There are two electrode groups created by default here - one is named 'None` (Hmm, that doesn't seem good...) and the other is `default_group`. I'd say the default name for an electrode group ought to be the classic `ElectrodeGroup`, and that second spurious group should not be made. I assume this can be fixed by correcting the default metadata of this interface, but maybe a follow-up PR adjusting the default names in the core write functions would be merited as well.

h-mayorquin:
> @h-mayorquin Thanks for sharing. A few items of note from that file (though not all of these belong in this PR, but can you fix in subsequent ones?)
> 
> 1. Device name is still `Device_ecephys`. Can you update that to CamelCase no underscore?
> 2. Similarly for `ElectricalSeries` name
> 3. So, I guess our approach to the TDT is much like our approach to SpikeGLX - if 'LFP' is indeed the signal that is acquired by the device, then it is treated as the raw signal (no change needed here by default), and it is sufficient to indicate the channel name as LFP (also already taken care of here). Even though it is LFP data, it is not 'processed' from a stored raw signal so it does not belong in processing. Just taking that from our previous attitude on SpikeGLX LF channels.
> 4. There are two electrode groups created by default here - one is named 'None`(Hmm, that doesn't seem good...) and the other is`default_group`. I'd say the default name for an electrode group ought to be the classic `ElectrodeGroup`, and that second spurious group should not be made. I assume this can be fixed by correcting the default metadata of this interface, but maybe a follow-up PR adjusting the default names in the core write functions would be merited as well.

Yes, I agree with all the changes there. They should be done in a separate PR as you say though.

CodyCBakerPhD:
@h-mayorquin I'd prioritize fixing (4) since that's likely an issue then with any recording interface

CodyCBakerPhD:
@h-mayorquin And maybe fix it first, then check back to see what this interface looks like with the fixes

h-mayorquin:
> @h-mayorquin I'd prioritize fixing (4) since that's likely an issue then with any recording interface

Yes, I think that's the most important and I have been wanting to take a look at the `add_electrodes_groups` function for a while.
Please open issues for the rest if you have time to keep a registry.

I will wait for what @bendichter says about priorities though.

CodyCBakerPhD:
Replaced by #135

PR #71, opened 2022-08-05
refactor imports for better performance

bendichter:
* move all imports to format-specific submodules
* refactor tests and conversion gallery accordingly
* remove tests for DeepLabCut (was causing problems for me)
* make interface_list which is used by the yaml workflow. This can be improved later
* modify jsonschema requirement. This was needed for me. Maybe it's a new thing.

## Motivation

I was having trouble using NeuroConv. Whenever I imported any datainterface, the Python kernel crached. It turns out our DeepLabCutInterface was the problem, but the way we are importing means that if there is a problem with any single interface it could crash the whole system. This refactor moves the imports to their own submodules so this is no longer a problem. It doesn't fix DeepLabCutInterface though. We'll need to look into that to see what is going wrong.

I'd like to make this switch. I know it requires a bit more work to import interfaces, but it makes our library much faster to import and much more robust to problems with individual interfaces. Now is the time to do it while we are in the middle of our renaming and we don't have any released code that currently relies on this version.


requirements-minimal.txt
```diff
@@ -6,7 +6,7 @@ h5py>=2.10.0
 tqdm>=4.60.0
 natsort>=7.1.1
 PyYAML>=5.4
-jsonschema>=3.2.0
+jsonschema[format]>=3.2.0
```
bendichter, line 9:
```suggestion
jsonschema[format]>=4.9.2
```

requirements-full.txt
```diff
@@ -10,7 +10,7 @@ hdf5storage==0.1.18
 scipy>=1.4.1
 h5py>=2.10.0
 PyYAML==5.4
-jsonschema==3.2.0
+jsonschema[format]==4.9.2
```
bendichter, line 13:
```suggestion
jsonschema[format]==4.9.1
```

requirements-minimal.txt
```diff
@@ -6,7 +6,7 @@ h5py>=2.10.0
 tqdm>=4.60.0
 natsort>=7.1.1
 PyYAML>=5.4
-jsonschema>=3.2.0
+jsonschema[format]>=4.9.2
```
bendichter, line 9:
```suggestion
jsonschema[format]>=4.9.1
```

h-mayorquin:
Looks good to me,I like the change. I will take a look at the problem with DeepLabCut.

What was it about the json schema? Cehck this which is important for us as it concerns a licencing issue:
https://github.com/python-jsonschema/jsonschema/pull/619

Looking the format option that you are using for the json schema I am wondering if the `format-nongpl` is not the one we want:

https://github.com/python-jsonschema/jsonschema/blob/658775f6c6228e2aff7a52ac4a20dd85369a304a/pyproject.toml#L47-L66


Also, did you add another interface (Maxwell)?

h-mayorquin (approved):
What should we add to the CHANGELOG for this one?

h-mayorquin:
One possible thing that we could do to preserve the old workflow is to create a new package at the top level called `full` which only keeps the old `__init__` so the following keeps working
`from neuroconv.full import SomeDataInterface`. 

The downside of that is that maybe users will start using that exclusively rendering this change less meaningful. Not sure how to think about it.

tests/test_on_data/test_gin_ophys.py
```diff
@@ -5,17 +5,16 @@
...
+from neuroconv import NWBConverter
+
+from neuroconv.datainterfaces.ophys.scanimage import ScanImageImagingInterface
+from neuroconv.datainterfaces.ophys.tiff import TiffImagingInterface
+from neuroconv.datainterfaces.ophys.hdf5 import Hdf5ImagingInterface
+from neuroconv.datainterfaces.ophys.sbx import SbxImagingInterface
+from neuroconv.datainterfaces.ophys.caiman import CaimanSegmentationInterface
+from neuroconv.datainterfaces.ophys.cnmfe import CnmfeSegmentationInterface
+from neuroconv.datainterfaces.ophys.extract import ExtractSegmentationInterface
+from neuroconv.datainterfaces.ophys.suite2p import Suite2pSegmentationInterface
```
CodyCBakerPhD, line 17: My main concern with this approach is the massive break to back-compatibility that will ensue, which we said we wouldn't do again after the big tool/util modularization of https://github.com/catalystneuro/nwb-conversion-tools/pull/411

It's also a LOT of extra typing for a user (even with tab-completion, which not every IDE supports and not every new user might be aware of) especially if they are loading multiple interfaces at once.

I have a couple other propositions that work more under-the-hood to offer the same benefits, will open those PRs soon. I think we should try to consider different approaches and discuss a weighing of the benefits and drawbacks of each.

CodyCBakerPhD (changes requested):

tests/test_on_data/test_gin_ophys.py
CodyCBakerPhD: Also see https://peps.python.org/pep-0690/ for extension discussion of both high-level and low-level details related this kind of problem and future improvements along the lines we're thinking.

CodyCBakerPhD: See #74 and #75 for some ideas...

Yet another approach to keep this PR nearly as-is but satisfy the backward compatibility breakage would be to override the module `__getattr__` which imports use (submodules are merely attributes of the main module, which is itself just a class). This is the suggestion of the accepted [PEP 562](https://peps.python.org/pep-0562/)

```python
# inside global `__init__.py`
# ...
mapping_from_data_interface_name_to_location_in_module: Dict[str, str]  # would obviously have a better name than this

def __getattr__(name):
    if name in mapping_from_data_interface_name_to_location_in_module:
        return importlib.import_module("." + mapping_from_data_interface_name_to_location_in_module[name], __name__)

# E.g.,
# >>> from neuroconv import DeepLabCutInterface
# 'essentially' calls
# DeepLabCutInterface = getattr(neuroconv, ".behavior.deeplabcut.deeplabcutinterface.DeepLabCutInterface")
```

This way, all imports from `neuroconv` would behave identically to how they did before, but the entire library of interfaces would only be loaded on-demand for each requested interface. Also all external imports in those interfaces would get called right away (PR #74 and #75 have the downside that import errors would only be revealed at initialization).

h-mayorquin: Thanks for putting all those three options together and their implementation in PRs. It was educational for me. If we are going with any of these three options I favor the last one as it confines all the import logic to one file (the global init) and keeps the workflow identical everywhere else.

h-mayorquin:
We importing `spikeinterface`, `roiextractors` and `neo` in the global init?

https://github.com/catalystneuro/neuroconv/blob/4c58489d6cb14d3195f62e9935c9e26c02beb657/src/neuroconv/__init__.py#L3

Should we also not load them until we need them? I was looking at the testing and some of the base data interfaces and they all access those modules through the full path with tools.

CodyCBakerPhD:
> Should we also not load them until we need them? I was looking at the testing and some of the base data interfaces and they all access those modules through the full path with tools.

Yeah that's been on the to-do list for a while, we shouldn't require people using `ophys` modules to install (or, now, import) `spikeinterface`

CodyCBakerPhD:
Main issues here are resolved in #74, good to close this now?

PR #74, opened 2022-08-07
Local-scope imports for external dependencies

CodyCBakerPhD:
Alternative to #71. Demonstration of simply moving the imports of external dependencies into the interface/converter function call workflow (`__init__` -> `get_metadata` -> `run_conversion`).

### Benefits

1. No break to back compatibility. All previous code imports will work in the same manner, and will also be faster due to less global overhead.

2. Acts at the same level as our lazy installation assertions (during interface `__init__`), so any problems with either will be caught after importing the NeuroConv library but before running the conversion. (errors should trigger during initialization of interface)

### Drawbacks

a. Goes against a spirit of [PEP8 import styling](https://pep8.org/#imports)

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

which is in place to detect any import errors prior to attempting to run any code. Though another common exception to this PEP standard is to do what is suggested here as a measure to avoid circular dependencies (not our issue, but tangentially related).

b. If the same package is needed multiple times over multiple functions, we would have to re-import them within each local function namespace since the reference would not be available to the global namespace. Note however that there is no downside to execution speed even if we need to import the same module multiple times across multiple functions in this manner - the first time it is ever imported it will be cached into the `sys.modules`; the downside comes only from restricted access to the reference.


src/neuroconv/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
```diff
@@ -22,8 +18,6 @@
 class BlackrockRecordingExtractorInterface(BaseRecordingExtractorInterface):
     """Primary data interface class for converting a BlackrockRecordingExtractor."""
 
-    RX = BlackrockRecordingExtractor
```
bendichter, line 25: oh wait, I don't know about this. Are you making this an instance attribute instead of a class attribute?

CodyCBakerPhD: It's always defined as a class attribute, but with lazy imports the default of `None` can only ever be set to the corresponding appropriate class after the external dependency has loaded, which is now occurring only when the object is initialized (that's the 'lazy' part - otherwise, setting it at the usual level of the class definition occurs simultaneous to the import of that class).

So we can either set it alongside `__init__` or at the `__new__` for the class - both have the same effect of triggering the import of the external dependency only when initializing the interface, and not otherwise.

bendichter: https://medium.com/python-features/class-vs-instance-variables-8d452e9abcbd

bendichter: hmm, I like the idea of being able to call `BlackrockRecordingExtractorInterface.RX` on the *class* not an instance of the class

CodyCBakerPhD: Another alternative could be setting the *name* (as as string) of the extractors class at that DataInterface class attribute level, which would have the same effect as we've always had for producing a corresponding associated name. But setting the *actual* extractor class as an *actual* class attribute at the same time you define the class itself is a non-lazy operation that occurs whenever you import the class.

CodyCBakerPhD: I also tried finding some ways that could allow such class attributes to be retrieved lazily (only imported/set when actually requested via `SomeRecordingInterface.RX` - not dissimilar from the previous `__getattr__` override stuff, but this time for class attributes) but this seems rather nuanced: https://stackoverflow.com/a/13095408

CodyCBakerPhD: OK you can now retrieve all extractor classes as you used to, but they are set only at the time of access

CodyCBakerPhD: See https://github.com/catalystneuro/neuroconv/pull/98 comments for some specific details and descriptions

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -11,13 +10,27 @@
...
+class BaseRecordingExtractorInterface(BaseDataInterface, metaclass=_LazyRecordingExtractor):
+    """Parent class for all RecordingExtractorInterfaces."""
+
+    RXModule: Optional[str] = None  # Defaults to "spikeinterface.extractors". Manually override in subclass if needed.
+    RXName: Optional[str] = None  # Defaults to __name__.replace("Interface", "Extractor"). Manually override if needed.
+    RX = None  # Loads dynamically on first access attempt
+
+    def __new__(cls, *args, **kwargs):
+        cls.RX = getattr(cls, "RX")
+        return object.__new__(cls)
```
CodyCBakerPhD, line 33: This code structure repeats with slight differences over each extractor interface. One way to do some code reduction would be to define a `BaseExtractorInterface` with a general attribute like `ExtractorClass` instead of specific `RX/SegX/IX/SX`

CodyCBakerPhD:
Summary of changelog:

* +41 new `BaseExtractorInterface` class
* +61 for common `get_package` helper function for all lazy installation checks
* +94 for explicit tests of import structure for the package

Then a series of +/- 10 to 50 lines for each interface in the package to nest the external package imports inside the corresponding functions.

CodyCBakerPhD:
@bendichter OK this is ready for review now (assuming that pesky `sonpy` test is passing now)

> Summary of changelog:

> +41 new BaseExtractorInterface class
> +61 for common get_package helper function for all lazy installation checks
> +94 for explicit tests of import structure for the package
> Then a series of +/- 10 to 50 lines for each interface in the package to nest the external package imports inside the corresponding functions.

src/neuroconv/datainterfaces/ecephys/intan/intandatainterface.py
```diff
@@ -47,7 +37,7 @@ def extract_electrode_metadata_with_pyintan(file_path):
     return electrodes_metadata
 
 
-def extract_electrode_metadata(recording_extractor: BaseRecording):
+def extract_electrode_metadata(recording_extractor):
```
bendichter, line 40: why remove the typehint here?

src/neuroconv/datainterfaces/ecephys/kilosort/kilosortdatainterface.py
```diff
@@ -1,16 +1,14 @@
...
-
 from ..basesortingextractorinterface import BaseSortingExtractorInterface
 from ....utils import FolderPathType
 
 
 class KilosortSortingInterface(BaseSortingExtractorInterface):
     """Primary data interface class for converting a KiloSortingExtractor from spikeinterface."""
 
-    SX = KiloSortSortingExtractor
+    ExtractorName = "KiloSortSortingExtractor"
```
bendichter, line 11: Do we need to define this because there is a difference in camel case with capital vs. lower case S? Perhaps it would be better to change the name of the class to be consistent with spikeinterface.

src/neuroconv/datainterfaces/ecephys/phy/phydatainterface.py
```diff
@@ -21,5 +16,7 @@ def __init__(
         spikeextractors_backend: bool = False,
     ):
         if spikeextractors_backend:
-            self.SX = se.PhySortingExtractor
+            from spikeextractors import PhySortingExtractor
+
+            self.SX = PhySortingExtractor
```
bendichter, line 21: shouldn't this be self.Extractor?

bendichter (approved):

bendichter:
phew, this was a lot of work!

It looks like one additional downside of this approach is that we can't use all of the typehints we want to, but I think that's ok.

src/neuroconv/datainterfaces/ecephys/kilosort/kilosortdatainterface.py
CodyCBakerPhD: Yes, that's why it's needed - we can definitely consider renaming to match SpikeInterface. Maybe in a follow-up PR though since this is enough as-is.

CodyCBakerPhD:
> It looks like one additional downside of this approach is that we can't use all of the typehints we want to, but I think that's ok.

There is a workaround to that, for example we can just define the `BaseRecording = typing.TypeVar` (similar to how we handle [special path types](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/utils/types.py#L8)) as a type hint that does not require importing SpikeInterface but would read exactly the same. Since we use type hints only for human-readability at this point nothing would change or be lost.

I can issue a follow-up PR for that to see what that looks like in practice.

PR #75, opened 2022-08-08
Dynamic module loading as interface attributes

CodyCBakerPhD:
Another alternative to #71 that is a bit more sophisticated than #74.

A couple ideas going on here

i. Define a general utility method for safely importing a module if it is installed and giving a more informative error message to perform the installation if not. This method could be used to reduce code around the repo any time we use that type of lazy import logic.

ii. Define two new attributes internal to any `DataInterface` - the user would not even need to know they are there. The first is the `modules_to_import`, which is just a list of all the names of modules that one would need to load for the interface. The second is `loaded_modules` (should maybe be renamed to 'imported_modules'?) which is a dictionary mapping from all those strings in `modules_to_import` to the actual modules themselves. The loading of the modules is performed as a generic step in the `__init__` common to all data interfaces, though this part may need small adjustments for appropriate behavior depending on how customized the most downstream initialization is for any particular interface. This workflow adjustment could also be done a bit more effectively by making the `__init__` of any interface focus on module loading and setting up other class attributes, with a `__post_init__` steep for the recording/sorting/imaging/segmentation interfaces where we actually initialize the extractor after successfully loading the required modules.

### Benefits

1. No break to back compatibility. All previous code imports will work in the same manner, and will also be faster due to less global overhead.

2. Acts at the same level as our lazy installation assertions (during interface `__init__`), so any problems with either will be caught after importing the NeuroConv library but before running the conversion. (errors should trigger during initialization of interface)

3. References to imported modules are readily available to a downstream user. Even using the current `main` branch, when I import a `DataInterface` I do not automatically import any associated modules, just the classes/objects of that module that have been attached to the interface. To use any of those modules I would have to re-import them in the script I'm using the data interface in to establish a reference. But this approach allows immediate use without re-establishing the reference since the reference is attached to the interface itself. Granted, there aren't too many situations you'd actually want to use this in.

### Drawbacks

a. Goes against a spirit of [PEP8 import styling](https://pep8.org/#imports)

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

which is in place to detect any import errors prior to attempting to run any code. Though another common exception to this PEP standard is to do what is suggested here as a measure to avoid circular dependencies (not our issue, but tangentially related).

b. Unsure as of yet, but could be a bit 'touchy' as a generic property of every data interface. Also a more sophisticated thing to have to teach in the tutorials of creating custom data interfaces.

### DIFF
--- a/src/neuroconv/utils/imports.py
+++ b/src/neuroconv/utils/imports.py
@@ -1,0 +1,15 @@
+"""Helper functions for performing more complicated import operations."""
+import importlib
+from typing import Optional
+
+
+def safe_import(module_name: str, pypi_package_name: Optional[str] = None):
+    """Import a module if it is installed or raise a more informative installation error if not."""
+    try:
+        return importlib.import_module(name=module_name)
+    except ModuleNotFoundError:
+        pypi_package_display_name = pypi_package_name or module_name
+        raise ModuleNotFoundError(
+            f"\nThe attempt to import '{module_name}' failed!\n"
+            f"To install this package, please run\n\n\tpip install {pypi_package_display_name}\n"
+        )
--- a/docs/conversion_examples_gallery/behavior/deeplabcut.rst
+++ b/docs/conversion_examples_gallery/behavior/deeplabcut.rst
@@ -9,12 +9,12 @@ Convert DeepLabCut imaging data to NWB using :py:class:`~neuroconv.datainterface
     >>> from dateutil import tz
     >>> from pathlib import Path
     >>> from neuroconv import DeepLabCutInterface
-    ...
     >>>
     >>> file_path = BEHAVIOR_DATA_PATH / "DLC" / "m3v1mp4DLC_resnet50_openfieldAug20shuffle1_30000.h5"
     >>> config_file_path = BEHAVIOR_DATA_PATH / "DLC" / "config.yaml"
     >>>
     >>> interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, subject_name="ind1", verbose=False)
+    ...
     >>> metadata = interface.get_metadata()
     >>> # For data provenance we add the time zone information to the conversion
     >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))
--- a/docs/conversion_examples_gallery/behavior/deeplabcut.rst
+++ b/docs/conversion_examples_gallery/behavior/deeplabcut.rst
@@ -9,6 +9,7 @@ Convert DeepLabCut imaging data to NWB using :py:class:`~neuroconv.datainterface
     >>> from dateutil import tz
     >>> from pathlib import Path
     >>> from neuroconv import DeepLabCutInterface
+    ...
     >>>
     >>> file_path = BEHAVIOR_DATA_PATH / "DLC" / "m3v1mp4DLC_resnet50_openfieldAug20shuffle1_30000.h5"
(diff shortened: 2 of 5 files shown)


CodyCBakerPhD:
Only test failing right now for linux is a strange `stdout` that is unexpected by the `doctest` gallery. Actual behavior of the interface is as expected.

@h-mayorquin is there no way to tell the `doctest` for the gallery to ignore verbose output from a certain line? Or otherwise capture it or test that is satisfies a certain regex?

h-mayorquin:
> Only test failing right now for linux is a strange `stdout` that is unexpected by the `doctest` gallery. Actual behavior of the interface is as expected.
> 
> @h-mayorquin is there no way to tell the `doctest` for the gallery to ignore verbose output from a certain line? Or otherwise capture it or test that is satisfies a certain regex?

@CodyCBakerPhD 
It seems to be coming from deep lab cut, right? One of the drawbacks of verbose by default options. 

You can try the ellipsis notation:
https://docs.python.org/3/library/doctest.html#doctest.ELLIPSIS

src/neuroconv/utils/imports.py
```diff
@@ -0,0 +1,15 @@
+"""Helper functions for performing more complicated import operations."""
+import importlib
+from typing import Optional
+
+
+def safe_import(module_name: str, pypi_package_name: Optional[str] = None):
+    """Import a module if it is installed or raise a more informative installation error if not."""
+    try:
```
h-mayorquin, line 8: What do you think of this pattern here vs the try-except block?
https://docs.python.org/3/library/importlib.html#checking-if-a-module-can-be-imported

CodyCBakerPhD: Certainly feels like a more modern approach.

CodyCBakerPhD:
> It seems to be coming from deep lab cut, right? One of the drawbacks of verbose by default options.

Yeah it's coming from the initial import of DLC, which has some stuff about version number, looks like this:

```
    Loading DLC 2.2.1.1...
    DLC loaded in light mode; you cannot use any GUI (labeling, relabeling and standalone GUI)
```

which could be useful if we were actually using DLC as a tool. I wouldn't actually call it a drawback of the package but rather a drawback of the sensitivity to output that doctest has - in this case it even depends on which platform of the CI it runs on: https://github.com/catalystneuro/neuroconv/runs/7725840103?check_suite_focus=true

h-mayorquin:
> > It seems to be coming from deep lab cut, right? One of the drawbacks of verbose by default options.
> 
> Yeah it's coming from the initial import of DLC, which has some stuff about version number, looks like this:
> 
> ```
>     Loading DLC 2.2.1.1...
>     DLC loaded in light mode; you cannot use any GUI (labeling, relabeling and standalone GUI)
> ```
> 
> which could be useful if we were actually using DLC as a tool. I wouldn't actually call it a drawback of the package but rather a drawback of the sensitivity to output that doctest has - in this case it even depends on which platform of the CI it runs on: https://github.com/catalystneuro/neuroconv/runs/7725840103?check_suite_focus=true

Is not about usefulness of the information provided by verbose is about an unintended side effect of sending information to the standard output by default in downstream use.

src/neuroconv/utils/imports.py
CodyCBakerPhD: This could still be useful to include eventually though, we repeat this block so many times as is

CodyCBakerPhD:
Closing in favor of approach https://github.com/catalystneuro/neuroconv/pull/74 that preserves all possible auto-parsing effects across IDEs and interactive consoles

PR #76, opened 2022-08-08
Demonstrate __getattr__ overload for lazy imports

CodyCBakerPhD:
Demonstration of comment: https://github.com/catalystneuro/neuroconv/pull/71#discussion_r939743746

If implemented for the other interfaces, would resolve the break in back-compatability while still granting the performance and safety improvement offered by lazier external imports.

Works for me locally but we'll see what the CI says in the morning.

### DIFF
--- a/src/neuroconv/__init__.py
+++ b/src/neuroconv/__init__.py
@@ -1,4 +1,11 @@
 from .nwbconverter import NWBConverter
-from .datainterfaces import *
 from .tools import spikeinterface, roiextractors, neo
 from .tools.yaml_conversion_specification import run_conversion_from_yaml
+
+
+def __getattr__(name: str):
+    from importlib import import_module
+
+    interface_location = dict(DeepLabCutInterface="datainterfaces.behavior.deeplabcut")
+    if name == "DeepLabCutInterface":
+        return getattr(import_module("." + interface_location[name], __name__), name)
--- a/docs/conversion_examples_gallery/behavior/deeplabcut.rst
+++ b/docs/conversion_examples_gallery/behavior/deeplabcut.rst
@@ -9,8 +9,7 @@ Convert DeepLabCut imaging data to NWB using :py:class:`~neuroconv.datainterface
     >>> from dateutil import tz
     >>> from pathlib import Path
     >>> from neuroconv import DeepLabCutInterface
-    Loading DLC 2.2.1.1...
-    DLC loaded in light mode; you cannot use any GUI (labeling, relabeling and standalone GUI)
+    ...
     >>>
     >>> file_path = BEHAVIOR_DATA_PATH / "DLC" / "m3v1mp4DLC_resnet50_openfieldAug20shuffle1_30000.h5"
     >>> config_file_path = BEHAVIOR_DATA_PATH / "DLC" / "config.yaml"
--- a/docs/conversion_examples_gallery/behavior/deeplabcut.rst
+++ b/docs/conversion_examples_gallery/behavior/deeplabcut.rst
@@ -3,24 +3,24 @@ DeepLabCut data conversion
 
 Convert DeepLabCut imaging data to NWB using :py:class:`~neuroconv.datainterfaces.behavior.deeplabcut.deeplabcutdatainterface.DeepLabCutInterface`.
 
-#.. code-block:: python
+.. code-block:: python
 
-#    >>> from datetime import datetime
-#    >>> from dateutil import tz
-#    >>> from pathlib import Path
-#    >>> from neuroconv import DeepLabCutInterface
-#    >>>
-#    >>> file_path = BEHAVIOR_DATA_PATH / "DLC" / "m3v1mp4DLC_resnet50_openfieldAug20shuffle1_30000.h5"
-#    >>> config_file_path = BEHAVIOR_DATA_PATH / "DLC" / "config.yaml"
-#    >>>
-#    >>> interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, subject_name="ind1",
-#verbose=False)
-#    >>> metadata = interface.get_metadata()
-##    >>> # For data provenance we add the time zone information to the conversion
-#    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))
-#    >>> metadata["NWBFile"].update(session_start_time=session_start_time)
-#    >>> # Choose a path for saving the nwb file and run the conversion
-#    >>> interface.run_conversion(nwbfile_path=path_to_save_nwbfile, metadata=metadata)
-#    >>> # If the conversion was successful this should evaluate to ``True`` as the file was created.
-#    >>> Path(path_to_save_nwbfile).is_file()
-#    True
+    >>> from datetime import datetime
+    >>> from dateutil import tz
+    >>> from pathlib import Path
+    >>> from neuroconv import DeepLabCutInterface
+    >>>
+    >>> file_path = BEHAVIOR_DATA_PATH / "DLC" / "m3v1mp4DLC_resnet50_openfieldAug20shuffle1_30000.h5"
+    >>> config_file_path = BEHAVIOR_DATA_PATH / "DLC" / "config.yaml"
+    >>>
+    >>> interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, subject_name="ind1",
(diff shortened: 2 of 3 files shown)


bendichter:
I see how this works. It's pretty clever, and it seems like something like this would be needed for the yaml approach regardless of if we do it for the standard python library usage. I think the biggest positive here is that the user has each of the interfaces at their fingertips instead of having to do a deeper import. I think I still prefer the classic approach for standard imports. Here are some thoughts:

1. I'm not too worried about backwards compatibility at this point. We are still in alpha versions and we just recently renamed the library so there aren't any codebases we would break right now. If we are going to make a breaking change, now is the time to do it.

2. Deep imports are the norm in larger packages. Consider scikit learn:
```python
from sklearn.linear_model import (
    LinearRegression,
    TheilSenRegressor,
    RANSACRegressor,
    HuberRegressor,
)
from sklearn.metrics import mean_squared_error
from sklearn.preprocessing import PolynomialFeatures
from sklearn.pipeline import make_pipeline
```
This informs the user about the structure of the project and how imports are layered, so there is some benefit to it. For me, having the nested import structure reflect the optional dependencies makes sense. We'll have the conversion galleries which will illustrate how to use each `DataInterface` with proper importing, so I don't expect that will pose a big problem for users. We can still have an `interface_location` dictionary you have started here for users to query against. This will be useful for the yaml approach.

3. I don't think tab-complete or IDE auto parsing would work for this approach.

4. The biggest reason I am reluctant to use this approach is I am worried it will cause problems down the road. How will this work cross platform? How will it work with threading, memory profiling, test coverage, etc.? Even if we test that those use-cases do not pose a problem, it would be impossible to know if other issues might (even if it's someone else's fault, not ours). When we use non-standard approaches like this we open the door for months of work when something comes up we didn't anticipate.

@CodyCBakerPhD what are you thoughts on these points?

CodyCBakerPhD:
> Deep imports are the norm in larger packages. Consider scikit learn:

The `scikit` examples are only one-level deep, which I would be fine with (that's what we use for the `spikeinterface`/`roiextractors` tools after all). `from neuroconv.datainterfaces import SomeRandomInterface` would be fine, but I don't think we need to force them to think about modality and format nesting on top of that. It would also be compatible with my other complaint that importing multiple interfaces across modalities requires a lot more typing, you could still do that on one line as `from neuroconv.datainterfaces import SomeEcephysInterface, SomeSortingInterface, SomeBehaviorInterface` which is super convenient.

Going more then 2 levels deep at most just generates headaches for trying to remember how things are nested in any given project.


> I don't think tab-complete or IDE auto parsing would work for this approach.

Just tested it out - the only thing required to make this work is to use the second part of the [PEP 562](https://peps.python.org/pep-0562/) (updating the `__dir__` structure), which is how all auto-parsers should be based (the way that structure is set and retrieved is via `dir(module_name)`)


> The biggest reason I am reluctant to use this approach is I am worried it will cause problems down the road. When we use non-standard approaches like this we open the door for months of work when something comes up we didn't anticipate.

As we say when taking more control over programmatic behavior ([example from chunking tutorial [in Notes]](https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/h5dataio.html?highlight=great#chunking)), "with great power comes great responsibility".

The great thing however is that this has been an accepted [PEP 562](https://peps.python.org/pep-0562/) standard for the past 5 years - the reason more people don't use it is likely because they don't have to and because there are so many other alternative approaches (like local scope definitions) that achieve the same benefit.

Examples like `sklearn` only seem to be based off of a few external packages like `numpy`/`scipy` so they don't have the same problems we do with needing to interface with so many outside dependencies. They also use a separate approach for something similar, which is also something we could look into: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/feature_selection/__init__.py#L33-L53 with reference to https://peps.python.org/pep-0008/#global-variable-names 

Even `neo` appears to work the less-efficient way (global `*` init from https://github.com/NeuralEnsemble/python-neo/blob/50abfbce832edb1c0f8364ac34c8a6442e1a3edf/neo/__init__.py#L10-L11 -> interface listing and simultaneous imports https://github.com/NeuralEnsemble/python-neo/blob/50abfbce832edb1c0f8364ac34c8a6442e1a3edf/neo/io/__init__.py#L278-L379 -> 'lazy' w.r.t. installation but if installed, imports https://github.com/NeuralEnsemble/python-neo/blob/50abfbce832edb1c0f8364ac34c8a6442e1a3edf/neo/rawio/cedrawio.py#L29-L33). Though it's not a problem for them or us until DLC because the initialization of the DLC module (likely due to tensorflow) is very heavy but no other package we deal with is (at least by comparison).


> How will this work cross platform? How will it work with threading, memory profiling, test coverage, etc.? Even if we test that those use-cases do not pose a problem, it would be impossible to know if other issues might (even if it's someone else's fault, not ours).

As the PEP indicates, we are in fact following the official and appropriate process of importing modules and assigning references within those modules, so there are no differences between what we're doing with manual redirection and how it would otherwise naturally work through other adjustments to the package structure (either propagating the customization to downstream `__init__.py` or outright placing the classes at the outer level - which of course would be terrible for organization purposes). So any usage case would behave as it normally would.

Basically, putting `from .deeplabcutinterface import DeepLabCutInterface` in the `datainterfaces/behavior/deeplabcut/__init__.py` is identical to adding the `.datainterfaces.behavior.deeplabcut.DeepLabCutInterface` attribute to the `neuroconv` module class. This is doing the same thing but with less typing that allows us to benefit from an easier (flatter) import structure while maintaining a more readable (deeper) module organization.

h-mayorquin:
> 1. I'm not too worried about backwards compatibility at this point. We are still in alpha versions and we just recently renamed the library so there aren't any codebases we would break right now. If we are going to make a breaking change, now is the time to do it.

+1 over this.

CodyCBakerPhD:
Also for future maintainability, if [PEP 690](https://peps.python.org/pep-0690/) or something similar gets approached at the standard library level, then future versions of Python would have it built in. Some of them are proposing keyword semantics like `lazy import some_package` which can act at the same level as `import some_other_package` but have the  desired behavior, but that also seems to be most of the 'how to best implement this in standard' discussion that populates the thread (as well as the social movement and education of Python developers on how to best use such a feature).

But even if it doesn't get built-in to Python there are many other packages starting to go in this direction:

see [`scikit-image` block](https://discuss.scientific-python.org/t/lazy-loading-has-landed/114) and their [more generalized approach](https://github.com/scikit-image/scikit-image/blob/main/skimage/_shared/lazy.py#L7-L56) for what some other packages are doing about this on their own.

bendichter:
@CodyCBakerPhD thanks for pointing me to PEP 562. I'll take a look.

bendichter:
I guess, since all of the tests pass, this works on all of our supported versions of Python

src/neuroconv/__init__.py
```diff
@@ -1,4 +1,11 @@
 from .nwbconverter import NWBConverter
-from .datainterfaces import *
 from .tools import spikeinterface, roiextractors, neo
 from .tools.yaml_conversion_specification import run_conversion_from_yaml
+
+
+def __getattr__(name: str):
+    from importlib import import_module
```
bendichter, line 7: why have this import here?

src/neuroconv/__init__.py
```diff
@@ -1,4 +1,11 @@
...
+
+    interface_location = dict(DeepLabCutInterface="datainterfaces.behavior.deeplabcut")
```
bendichter, line 9: I'd prefer if this were outside the scope of the function as well

src/neuroconv/__init__.py
CodyCBakerPhD: If we import in the global space it would be included in the `__dir__` so you could do something weird like `from neuroconv import importlib`. I guess we could overwrite that however. IDK feels safer to do it locally to where it's needed

src/neuroconv/__init__.py
CodyCBakerPhD: That would be a good idea for portability, yes.

h-mayorquin:
Relevant to this approach and in favor of @CodyCBakerPhD argument's of wider applicability is the facs that the scientific python community has endorsed and implemented an specification with a library following this approach:
https://scientific-python.org/specs/spec-0001/

CodyCBakerPhD:
Closing in favor of approach #74 that preserves all possible auto-parsing effects across IDEs and interactive consoles

PR #78, opened 2022-08-08
Remove top level imports of big modules

h-mayorquin:
From the discussion at #71:
https://github.com/catalystneuro/neuroconv/pull/71#issuecomment-1208061343

Maybe this is something we can do to improve import performance as well.


src/neuroconv/__init__.py
```diff
@@ -1,4 +1,3 @@
 from .nwbconverter import NWBConverter
 from .datainterfaces import *
-from .tools import spikeinterface, roiextractors, neo
```
CodyCBakerPhD, line 3: Similar to #76, I would not remove access to this alias from the top level. Breaks backcompatability and the decision to have this kind of structure stems from https://github.com/catalystneuro/nwb-conversion-tools/pull/411 and extensive in-person discussions around it. But we now know of ways to have access to that import structure without actually loading each respective module.

tests/test_minimal/test_spikeextractor_to_nwb.py
```diff
@@ -17,7 +17,6 @@
     get_default_nwbfile_metadata,
 )
 
-from neuroconv import spikeinterface  # testing aliased import
```
CodyCBakerPhD, line 20: Just tested locally, `from neuroconv.tools import spikeinterface` should still and be lazy and is good enough for me. Can you try re-adding those aliased tests with that alteration to the import?

tests/test_minimal/test_import_structure.py
```diff
@@ -32,10 +33,25 @@ def test_outer_import_structure(self):
...
+            "neo",
             "spikeinterface",
             "roiextractors",
-            "neo",
-            "run_conversion_from_yaml",
+            # Helper functions
+            "get_package",
+            "importing",
+            # Other stuff that I don't know why it is here (for PR)
+            "nwb_helpers",
```
h-mayorquin, line 51: What about this stuff here? Why are they imported with `from neuroconv import tools`?

tests/test_minimal/test_spikeextractor_to_nwb.py
h-mayorquin: Sure.

tests/test_minimal/test_import_structure.py
CodyCBakerPhD: Half of these are package structure available from all import navigation (those are sub-modules). Can't hide those without `__dir__` override magic.

But this test is good, definitely keep it and keep it up to date with what we intend the package structure to appear as.

h-mayorquin: My question was more like, when are they imported? the init of tools does not have them and neither does the top level init.

CodyCBakerPhD: If you call `import neuroconv`, only the names of the submodules are loaded into the `__dir__`, not their contents. So actually trigger import code in the submodule you'd have to call `neuroconv.tools.submodule.some_function` and it will load that entire submodule. That's the classic way of achieving 'lazy' imports that Bens original solution utilized (albeit at deeper levels of nesting).

CodyCBakerPhD: Alternative way to trigger imports there is `from neuroconv.tools import submodule`

h-mayorquin:
So, as things stand the import of spikeinterface is trigger when the tools module is used and for the minimal installation this is not available.  Any way of using the newly developed functionality to avoid that?

src/neuroconv/tools/__init__.py
```diff
@@ -1,8 +1,7 @@
 # To be treated as sub-modules or plugins
-# TODO: temporarily disabled until decision to remove in other PR
-# from .spikeinterface import spikeinterface
-# from .roiextractors import roiextractors
-# from .neo import neo
+from .spikeinterface import spikeinterface
+from .roiextractors import roiextractors
+from .neo import neo
```
CodyCBakerPhD, line 4: @h-mayorquin This is why that import is still happening - when I said I tested locally, I meant on main (with the comments). So these should just be deleted, as in https://github.com/catalystneuro/neuroconv/pull/124)

May have to update the new import test to reflect this, however.

tests/test_minimal/test_import_structure.py
h-mayorquin: Ok so my understanding is the following.  When I call `from neuroconv import tools`, "nwb_helpers" is added to tools.__dict__  because of two conditions:
1) nwb_helpers is a submodule of tools 
AND
2) Some functions form nwb_helpers are called in the import hierarchy. What I mean is that the import `NWBConverter` in the top __init__ here triggers this lines here:
https://github.com/catalystneuro/neuroconv/blob/ae08b4ae627f9e5ced0e0d59d15c1bd329abffb8/src/neuroconv/nwbconverter.py#L10

CodyCBakerPhD: Yes, that is true as well

CodyCBakerPhD (approved):

PR #81, opened 2022-08-08
[Hotfix]: disable auto-import of DLC interface

CodyCBakerPhD:
Simpler fix for issue in PR #71 pending ecosystem-wide agreement on lazy import approach.

@bendichter can you test this resolves your issue for the time being?


h-mayorquin:
This has been merged now:
https://github.com/DeepLabCut/DLC2NWB/pull/15

bendichter:
@h-mayorquin https://github.com/DeepLabCut/DLC2NWB/pull/15 isn't going to help me because you are still importing deeplabcut

h-mayorquin:
> @h-mayorquin [DeepLabCut/DLC2NWB#15](https://github.com/DeepLabCut/DLC2NWB/pull/15) isn't going to help me because you are still importing deeplabcut. Well, I suppose it could if I just uninstall deeplabcut so it can't import, but it would be better to not import deeplabcut at all.

Yes, you will need to uninstall deeplabcut for it to work. That's true.

bendichter:
Ok that's fine then

h-mayorquin:
> Ok that's fine then

@bendichter 
Yeah, I left that `try except` block as I wanted to have that merged as soon as possible. The developer there already agree that that's probably not the place for it so maybe it can be removed in a subsequent PR. Maybe you can comment [there](https://github.com/DeepLabCut/DLC2NWB/pull/15#discussion_r942172220) to strengthen the case?

bendichter (approved):

CodyCBakerPhD:
This is actually better implemented in #74 along with ELLIPSIS functionality for the doctests

PR #82, opened 2022-08-08
Try auto parsing from __all__

CodyCBakerPhD:
I think all I had to do (no pun intended) was specify the `__all__` attribute synchronized with the changes to `__dir__`

@bendichter @weiglszonja Can you two try to checkout this branch and before the following?

```python
import neuroconv

neuroconv.[type tab and see what it shows]
```

and

```python
from neuroconv import [type tab and see what it shows]
```

Also @h-mayorquin not sure how you prefer to run in-line auto-completion in VSCode, but if you can try and let me know.

I tested it on my system with IPython 7.34.0 and got the following

image

image

Then on latest jupyter version

image

image

### DIFF
--- a/src/neuroconv/__init__.py
+++ b/src/neuroconv/__init__.py
@@ -2,10 +2,28 @@ from .nwbconverter import NWBConverter
 from .tools import spikeinterface, roiextractors, neo
 from .tools.yaml_conversion_specification import run_conversion_from_yaml
 
+interface_location = dict(
+    DeepLabCutInterface="datainterfaces.behavior.deeplabcut",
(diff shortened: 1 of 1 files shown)


weiglszonja:
Jupyter Notebook:
1/
Screenshot 2022-08-09 at 17 30 12
2/
Screenshot 2022-08-09 at 17 30 47

In PyCharm 2022.1.3 for me only 1/ works, but I think it is PyCharm issue if this was working in VC.
Screenshot 2022-08-09 at 17 21 31

bendichter:
This is a step in the right direction, but we still wouldn't get within-function tools like automatic typehint parsing and signature hinting

PR #37, opened 2022-07-26
Add unit tests for add  electrical series II (scaling to volts)

h-mayorquin:
I added tests for the `add_electrical_series` function in the spikeinterface module. Also, some change in the code of the module itself. Relevant to the discussion here is the discussion in https://github.com/catalystneuro/neuroconv/issues/34.

PR #84, opened 2022-08-09
Add unit tests for add electrical series IV (chunking) 

h-mayorquin:
Following on #64 and #37 this PR adds unit tests for the chunking capabilities of `add_electrical_series` and refactors the function  with a separation of the functionality for wrapping the traces in an iterator in a separate function.

### DIFF
--- a/src/neuroconv/tools/spikeinterface/spikeinterface.py
+++ b/src/neuroconv/tools/spikeinterface/spikeinterface.py
@@ -13,7 +13,7 @@ from spikeinterface import BaseRecording, BaseSorting
 from spikeinterface.core.old_api_utils import OldToNewRecording, OldToNewSorting
 from spikeextractors import RecordingExtractor, SortingExtractor
 from numbers import Real
-from hdmf.data_utils import DataChunkIterator
+from hdmf.data_utils import DataChunkIterator, AbstractDataChunkIterator
 from hdmf.backends.hdf5.h5_utils import H5DataIO
 import psutil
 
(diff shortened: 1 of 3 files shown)


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -488,6 +488,68 @@ def check_if_recording_traces_fit_into_memory(recording: SpikeInterfaceRecording
         raise MemoryError(message)
 
 
+def iterize_recording_traces(
```
CodyCBakerPhD, line 491: I'd go with `iterate_recording_traces` as being a bit clearer

h-mayorquin: I don't feel comfortable with "iterize" to be honest. 
On the other hand, "iterate" is used in Python for iterating over something so I don't think that overloading the semantics of this with "making something an iterator" is a good choice either.

What do you think of `build_recording_traces_iterator`, `wrap_recording_traces_on_iterator`?
Maybe I can think on something better.

CodyCBakerPhD: `setup_trace_iteration`?

h-mayorquin: I decided to be more specific as I realized iterator is overly broad:

`_recording_traces_to_hdmf_iterator` 

I would prefer three words instead but as it is a private method I think is fine for it to focus on readability instead of type-ability.

h-mayorquin:
Is there still a use case for using the iterator `v2` that extracts the traces and then passes them to a `DataChunkIterator`?

src/neuroconv/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Sure thing

CodyCBakerPhD:
> Is there still a use case for using the iterator v2 that extracts the traces and then passes them to a DataChunkIterator?

?

I don't understand what you mean

CodyCBakerPhD (approved):

PR #87, opened 2022-08-12
Update ExtractSegmentationInterface

weiglszonja:
Work in progress

## Checklist

### DIFF
--- a/tests/test_on_data/test_gin_ophys.py
+++ b/tests/test_on_data/test_gin_ophys.py
@@ -132,7 +132,8 @@ class TestOphysNwbConversions(unittest.TestCase):
         name_func=custom_name_func,
     )
     def test_convert_segmentation_extractor_to_nwb(self, data_interface, interface_kwargs):
-        nwbfile_path = str(self.savedir / f"{data_interface.__name__}.nwb")
+        extractor_name = data_interface.SegX.extractor_name
+        nwbfile_path = str(self.savedir / f"{data_interface.__name__}_{extractor_name}.nwb")
 
         class TestConverter(NWBConverter):
             data_interface_classes = dict(TestSegmentation=data_interface)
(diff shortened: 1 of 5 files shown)


tests/test_on_data/test_gin_ophys.py
```diff
@@ -124,7 +132,8 @@ class TestConverter(NWBConverter):
         name_func=custom_name_func,
     )
     def test_convert_segmentation_extractor_to_nwb(self, data_interface, interface_kwargs):
-        nwbfile_path = str(self.savedir / f"{data_interface.__name__}.nwb")
+        extractor_name = data_interface.SegX.extractor_name
+        nwbfile_path = str(self.savedir / f"{data_interface.__name__}_{extractor_name}.nwb")
```
h-mayorquin, line 136: Is this something that changes with the case (the name of the extractor) or why are you adding it?

CodyCBakerPhD: The same interface can now map onto multiple different underlying classes, yes.

weiglszonja: ExtractSegmentationInterface is being tested for the old version of the extract output file, and I added the new one in. This would mean nwbfile_path with the same name would be used when testing the two files, and I got an error for that the same file cannot be opened. Instead of this maybe we can add a line to remove the file after the test passed?

CodyCBakerPhD: > Instead of this maybe we can add a line to remove the file after the test passed?

Nah, way more informative to just write an extra file with more specific name, especially when running tests with `save_output=True`

h-mayorquin: @CodyCBakerPhD 
> The same interface can now map onto multiple different underlying classes, yes.
You mean extractors by classes?

@weiglszonja 
I get you. Thanks for answering.
In the `test_gin_ecephys` you can see another soultion for this problem where I have assigned a case name to characterize with a different names tests that concern the same interface.

CodyCBakerPhD (approved):

PR #90, opened 2022-08-15
Integrate ImagingDataChunkIterator with write tools

CodyCBakerPhD:
## Motivation

PR has two linked parts needed for my current conversion.

1a) Adds the new `ImagingDataChunkIterator` to the `add_two_photon` helper function in `tools.roiextractors`.
1b) TODO - need to add to `write_imaging` as well.
2) 'technically' adds support to the `ImagingDataChunkIterator` for volumetric data (i.e., an extra axis on the end for depth). Works on top of existing roiextractors API with no actual volumetric support needed (just depends on image size retrieved through the extractor)

- [x] Also TODO is remove the print statements.

## How to test the behavior?
@weiglszonja What tests would you want for this? Could you add them if you have the time?

## Checklist

- [x] Have you thoroughly read our [Developer Guide](https://neuroconv.readthedocs.io/en/main/developer_guide.html)?
- [x] Have you ensured the PR description clearly describes the problem and solutions?
- [x] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/neuroconv/pulls) for the same change?
- [x] If this PR fixes an issue, is the first line of the PR description `fix #XXX` where `XXX` is the issue number?
- [x] Did you update the [CHANGLOG.md](https://github.com/catalystneuro/neuroconv/blob/main/CHANGELOG.md) file on your branch to describe your changes?

PR #91, opened 2022-08-16
Create aux method for choosing iterator type (for #90)

weiglszonja:
To be merged into #90

### DIFF
--- a/src/neuroconv/tools/roiextractors/roiextractors.py
+++ b/src/neuroconv/tools/roiextractors/roiextractors.py
@@ -295,7 +295,6 @@ def add_two_photon_series(
             "Specify as a key in the new 'iterator_options' dictionary instead."
         )
     iterator_type = iterator_type or "v2"
-    assert iterator_type in ["v1", "v2"], "'iterator_type' must be either 'v1' or 'v2' (recommended)."
     iterator_options = iterator_options or dict()
 
     metadata_copy = deepcopy(metadata)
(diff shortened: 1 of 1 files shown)


h-mayorquin (approved):
Looks good to me. I am curios about the default of `buffer_size=10` for the old data chunk iterator, isn't that useful for the new one as a default?

Also, not related to this but I am of the opinion that there is no reason to maintain the old functionality given that we have explicitly designed 'v2' to be better.

CodyCBakerPhD (approved):

CodyCBakerPhD:
> Looks good to me. I am curios about the default of buffer_size=10 for the old data chunk iterator, isn't that useful for the new one as a default?

New defaults are better, keeping this one to maintain consistent behavior.


> Also, not related to this but I am of the opinion that there is no reason to maintain the old functionality given that we have explicitly designed 'v2' to be better.

The 'better' isn't quite as clear cut in this case - their behavior matches more closely than it did in the ecephys case and performance wasn't different by orders of magnitude. They do, however, work in markedly different ways and produce different behavior (`v1` keeps dimensions 'open' for example, which can be undesired in some circumstances but desired in others), and again they are powered by completely different code bases so if a bug appeared in `v2` there's a very good chance you could rollback and use `v1` safely.

CodyCBakerPhD:
Thanks a bunch @weiglszonja

h-mayorquin:
> > Looks good to me. I am curios about the default of buffer_size=10 for the old data chunk iterator, isn't that useful for the new one as a default?
> 
> New defaults are better, keeping this one to maintain consistent behavior.
> 
> > Also, not related to this but I am of the opinion that there is no reason to maintain the old functionality given that we have explicitly designed 'v2' to be better.
> 
> The 'better' isn't quite as clear cut in this case - their behavior matches more closely than it did in the ecephys case and performance wasn't different by orders of magnitude. They do, however, work in markedly different ways and produce different behavior (`v1` keeps dimensions 'open' for example, which can be undesired in some circumstances but desired in others), and again they are powered by completely different code bases so if a bug appeared in `v2` there's a very good chance you could rollback and use `v1` safely.

Thanks for explaining this one.

PR #92, opened 2022-08-16
Add more tests for `add_two_photon_series` ( for #90)

weiglszonja:
To be merged into https://github.com/catalystneuro/neuroconv/pull/90

### DIFF
--- a/tests/test_internals/test_roi_extractors_module.py
+++ b/tests/test_internals/test_roi_extractors_module.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from datetime import datetime
 
 import numpy as np
+from hdmf.data_utils import DataChunkIterator
 from hdmf.testing import TestCase
 from numpy.testing import assert_array_equal, assert_raises
 from parameterized import parameterized, param
@@ -847,6 +848,7 @@ class TestAddTwoPhotonSeries(TestCase):
         )
 
     def test_default_iterator(self):
+        """Test adding two photon series with default iterator."""
         add_two_photon_series(imaging=self.imaging_extractor, nwbfile=self.nwbfile, metadata=self.metadata)
 
         # Check data
@@ -872,6 +874,7 @@ class TestAddTwoPhotonSeries(TestCase):
         assert len(imaging_planes_in_file) == 1
 
     def test_invalid_iterator_type_raises_error(self):
+        """Test error is raised when adding two photon series with invalid iterator type."""
         with self.assertRaisesWith(
                 AssertionError,
                 "'iterator_type' must be either 'v1' or 'v2' (recommended).",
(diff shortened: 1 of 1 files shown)


tests/test_internals/test_roi_extractors_module.py
```diff
@@ -862,6 +862,8 @@ def test_default_iterator(self):
         # NWB stores images as num_columns x num_rows
         expected_two_photon_series_shape = (self.num_frames, self.num_columns, self.num_rows)
         assert two_photon_series_extracted.shape == expected_two_photon_series_shape
+        expected_two_photon_series_data = self.imaging_extractor.get_video().transpose((0, 2, 1))
+        assert_array_almost_equal(two_photon_series_extracted, expected_two_photon_series_data)
```
weiglszonja, line 864: following the discussion [here](https://github.com/catalystneuro/roiextractors/issues/193#issuecomment-1213338165) I wasn't sure whether to use strong equality test, so I chose to be consistent with [ecephys test](https://github.com/catalystneuro/neuroconv/blob/27357f8145746f2279e580b8a5565eaa26a98f96/tests/test_internals/test_si.py#L427)

tests/test_internals/test_roi_extractors_module.py
```diff
@@ -882,7 +945,7 @@ def test_add_two_photon_series_roundtrip(self):
 
             # Check data
             acquisition_modules = read_nwbfile.acquisition
-            self.two_photon_series_name in acquisition_modules
+            assert self.two_photon_series_name in acquisition_modules
```
h-mayorquin, line 948: Thanks : )

tests/test_internals/test_roi_extractors_module.py
```diff
@@ -828,33 +837,31 @@ def setUp(self):
...
         self.imaging_extractor = generate_dummy_imaging_extractor(
             self.num_frames, num_rows=self.num_rows, num_columns=self.num_columns
         )
 
-    def test_add_two_photon_series(self):
-
-        metadata = self.metadata
-
-        add_two_photon_series(imaging=self.imaging_extractor, nwbfile=self.nwbfile, metadata=metadata)
+    def test_default_iterator(self):
```
h-mayorquin, line 848: I think that this first test could be better called `test_default_values` as is not only the iterator that is tested. What do you think?

weiglszonja: yes, thank you.

tests/test_internals/test_roi_extractors_module.py
h-mayorquin: I think the argument of @CodyCBakerPhD  is stronger here. There is no reason why the float representation of this object should be different so maybe you want to know if they do change.

The argument over there in the linked thread is because `check_segmentation_equal` is used to compare objects that come from "farther away" (such as the use we do in our gin tests to compare the extractors with their versions from nwb or to compare the extractor with is marshaled version). Over there, in `check_segmentations_equal` I think we might want to have the two tests and allow the stricter version to be skipped in some cases.

h-mayorquin (approved):
Looks good to me.

CodyCBakerPhD (approved):

CodyCBakerPhD:
Thanks you two @weiglszonja @h-mayorquin

PR #97, opened 2022-08-18
Add missing annotation types in tools.ROIExtractors

CodyCBakerPhD:
@h-mayorquin As requested


h-mayorquin (approved):
Thanks for taking care of that!

PR #98, opened 2022-08-18
Lazy IX class attribute via metaclasses

CodyCBakerPhD:
From discussion https://github.com/catalystneuro/neuroconv/pull/74#discussion_r949517850

Allows class attributes like `SomeRecordingExtractor.RX` to be set lazily from external dependencies only at the time of call for that attribute.

```python
>>> from neuroconv.datainterfaces import CEDRecordingInterface

>>> # spikeinterface will not have been imported yet

>>> CEDRecordingInterface.RX
spikeinterface.extractors.neoextractors.ced.CedRecordingExtractor
```


src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -14,7 +14,7 @@
 from ...utils import get_schema_from_hdmf_class, get_base_schema, OptionalFilePathType
 
 
-class BaseRecordingExtractorInterface(BaseDataInterface, ABC):
+class BaseRecordingExtractorInterface(BaseDataInterface):
```
CodyCBakerPhD, line 17: Requires removal of this mix-in to avoid metaclass conflicts.

Also, why is this mix-in defined anyhow on `main` since the parent is an `ABC` anyway?

src/neuroconv/basedatainterface.py
```diff
@@ -8,7 +8,7 @@
 from .utils import get_base_schema, get_schema_from_method_signature
 
 
-class BaseDataInterface(ABC):
+class BaseDataInterface(object):
```
CodyCBakerPhD, line 11: Only thing about this is the `ABC` MRO can cause some metaclass conflicts with this approach

CodyCBakerPhD: Nevermind - figured out the resolution to the the metaclass conflict: https://stackoverflow.com/a/28727066

src/neuroconv/datainterfaces/ecephys/ced/ceddatainterface.py
```diff
@@ -9,12 +9,20 @@ def _test_sonpy_installation() -> None:
...
 
-class CEDRecordingInterface(BaseRecordingExtractorInterface):
-    """Primary data interface class for converting data from CED (Cambridge Electronic Design)."""
+class LazyExtractorClass(type(BaseRecordingExtractorInterface), type):
+    def __getattribute__(self, name):
+        if name == "RX":
+            _test_sonpy_installation()
+            spikeinterface = get_package(package_name="spikeinterface")
+            RX = spikeinterface.extractors.CedRecordingExtractor
+            self.RX = RX
```
CodyCBakerPhD, line 18: Couple touchy things to note - the attribute must also be 'set' at this call in order for it to persist later. Otherwise an instance of the class will not have it set (calling `self.RX` on an object does not itself seem to trigger `getattr(cls, "RX")`.

src/neuroconv/datainterfaces/ecephys/ced/ceddatainterface.py
```diff
@@ -9,12 +9,20 @@ def _test_sonpy_installation() -> None:
...
+
+            return RX
+        return super().__getattribute__(name)
```
CodyCBakerPhD, line 21: Also a bit touchy - this must be defined this way in order for decorators like `classmethod` to become bound to the corresponding functions like `get_source_schema`.

src/neuroconv/datainterfaces/ecephys/ced/ceddatainterface.py
```diff
@@ -28,11 +36,17 @@ def get_all_channels_info(cls, file_path: FilePathType):
         """Retrieve and inspect necessary channel information prior to initialization."""
         return cls.RX.get_all_channels_info(file_path=file_path)
 
+    def __new__(cls, *args, **kwargs):
+        getattr(cls, "RX")
```
CodyCBakerPhD, line 40: While issuing a manual call like the one in the PR description works fine with the metaclass override, in order for a fresh instance of the class to automatically import and set the class, the call must be triggered at this level.

src/neuroconv/datainterfaces/ecephys/axona/axonadatainterface.py
```diff
@@ -11,11 +9,9 @@
 from ....utils import get_schema_from_method_signature, FilePathType
 
 
-class AxonaRecordingExtractorInterface(BaseRecordingExtractorInterface):
+class AxonaRecordingInterface(BaseRecordingExtractorInterface):
```
CodyCBakerPhD, line 12: Just leaving this here as a demonstration - some of the current interfaces have the word 'Extractor' in the name. While we're breaking back-compatibility with moving things into `neuroconv.datainterfaces`, what would you all think about standardizing the interface names a bit? @h-mayorquin @weiglszonja @bendichter

src/neuroconv/datainterfaces/ecephys/blackrock/blackrockdatainterface.py
```diff
@@ -19,6 +19,8 @@
 class BlackrockRecordingExtractorInterface(BaseRecordingExtractorInterface):
     """Primary data interface class for converting a BlackrockRecordingExtractor."""
 
+    RXName = "BlackrockRecordingExtractor"
```
CodyCBakerPhD, line 22: This is an alternative demonstration that maintains the current name of the extractor and merely corrects it to match SpikeInterface. Though something like this could also just be added to the default auto-naming rule, like `__name__.strip("Interface") if "Extractor" in __name__ else __name__.replace("Interface", "Extractor")` or something...

src/neuroconv/datainterfaces/ecephys/ced/ceddatainterface.py
```diff
@@ -9,12 +9,12 @@ def _test_sonpy_installation() -> None:
...
-class CEDRecordingInterface(BaseRecordingExtractorInterface):
+class CEDRecordingInterface(
+    BaseRecordingExtractorInterface,
+):
     """Primary data interface class for converting data from CED (Cambridge Electronic Design)."""
 
-    _test_sonpy_installation()
-    spikeinterface = get_package(package_name="spikeinterface")
-    RX = spikeinterface.extractors.CedRecordingExtractor
+    RXName = "CedRecordingExtractor"
```
CodyCBakerPhD, line 17: Most of the interfaces could have extractor names auto-inferred, but this and Neuroscope (SI: 'NeuroScope') are some of the conditions where a manual name override is necessary and not difficult to do.

src/neuroconv/datainterfaces/ecephys/axona/axonadatainterface.py
h-mayorquin: I think that is a good idea to eliminate the redundancy for those which have Extractor in their name

weiglszonja: I agree with @h-mayorquin.

h-mayorquin: Note that the base class is named `BaseRecordingExtractorInterface`

weiglszonja: I thought that is intentional, or do you think it is redundant in the base class too @h-mayorquin ?

h-mayorquin: To some degree, we only have RecordingExtractors we don't have Recording by themselves. I am fine with either RecordingInterface or RecordingExtractorInterface but I think we should have consistent names.

CodyCBakerPhD: Ben and I discussed this and we agreed to remove the "Extractor" from all names of the type "<format name>ExtractorInterface" but keep it in all the Base<extractor type>ExtractorInterfaces to indicate that they must specify an Extractor class of some type from either SpikeInterface or ROIExtractors.

I'll also be making a single general BaseExtractorClass to contain the lazy import logic to reduce duplicated code.

PR #99, opened 2022-08-19
Fix Icephys gallery

CodyCBakerPhD:
@luiztauffer This prevents the file from getting left around when tests are done.

Also, we standardized the `save_path` argument to `nwbfile_path` among the various tools a while back but I suppose we missed the `neo` one. Added a deprecation warnings for that here as well.


h-mayorquin (approved):
Looks good, I think this interface is missing using the context manager that you developed. I have been meaning to add it for a bit.

CodyCBakerPhD:
> Looks good, I think this interface is missing using the context manager that you developed. I have been meaning to add it for a bit.

@h-mayorquin Ah yes, feel free to do that whenever

Issue #94, opened 2022-08-17
modular interface-based installation configurations


bendichter:
NeuroConv has many dependencies that are specific to particular data formats. We have been working on how to make imports modular. This works right now, but we might end up with a problem where different libraries have conflicting dependencies, so it would not be possible to install all of them. Therefore, I think it is also important to make installation modular. Currently, we have a few different installation options. You can run:

`pip install neuroconv`

for the minimal dependencies and 

`pip install neuroconv[full]`

for the full dependencies. I propose a solution where you can specify modalities during install and pip will know to install the dependencies specifically for those modalities, e.g.:

`pip install neuroconv[openephys]`

would install the minimum plus `pyopenephys`. A user would also be able to specify multiple dependencies, e.g.:

`pip install neuroconv[spikeglx,kilosort,phy,deeplabcut]`

and the environment would be set up with the necessary libraries for those data sources.

In order to accomplish this, we will need to formally define the requirements of each type of source data. One way we can do this is to have `requirements.txt` files in each data type directory. For instance: 
`neuroconv/src/neuroconv/datainterfaces/behavior/deeplabcut/requirements.txt` would be:

```
dlc2nwb>=0.2
```

The `setup.py` file would use `glob` to iterate through all of the datainterface modules and create a `extra_requires` `dict` for each source data type.

CodyCBakerPhD:
Another approach could be to still only have one file at the top level (say, `extra_requires.json` or `extra_requires.pkl`) and what is saved within is a dictionary of format keywords and list of specific dependencies as values.

Main benefit would be readability as it would make it really easy to compare at a glance the dependencies across each formats - with the above approach I'd have to open separate tabs for each file on the web interface (on top of having to navigate down to each file), or similarly open each file separately (again, after navigating) on my local clone.

bendichter:
I would rather not use pickle files. Not transparent to users and no version control. json is much better than pickle for this.

neo uses a dict in the setup.py file: https://github.com/NeuralEnsemble/python-neo/blob/50abfbce832edb1c0f8364ac34c8a6442e1a3edf/setup.py#L10-L18

I think I would still prefer the individual requirements.txt files. I know this is more work, but I'd like to go for more of a plugin style architecture where DataInterfaces can be contributed without altering any of the core files.

bendichter:
Having separate `requirements.txt` files also allows for the option to install these options via `pip install -r path/to/requirements.txt

h-mayorquin:
[...]

PR #100, opened 2022-08-19
Extra setup keys and sub-requirement files

CodyCBakerPhD:
A first step towards #94 specific to demonstrating the usage for the `ophys` interfaces.

The `/datainterfaces/ophys` level requirements specifies the `roiextractors` version needed, and each `/ophys/<format name>` specifies the specific requirements for each particular format.

The setup now includes the extra installation keys for each format name as they appear from the `/ophys/` directory, as well as a general `ophys` key that installs everything you would need to use any Ophys interface.


.github/workflows/testing.yml
```diff
@@ -32,10 +32,17 @@ jobs:
...
+      - name: Install with minimal requirements
         run: pip install .[test]
       - name: Run minimal tests
-        run: pytest tests/test_internals -n auto --dist loadscope
+        run: pytest tests/test_minimal -n auto --dist loadscope
+        
+      - name: Install with ophys requirements
+        run: pip install .[ophys]
+      - name: Run ophys tests
+        run: pytest tests/test_ophys -n auto --dist loadscope
```
CodyCBakerPhD, line 44: This demonstrates one possible way we could deploy tests for the specific keys. Personally I think I'd be fine with just one of these steps per each modality.

Another option would be to utilize a matrix, which would use a fresh environment for each fresh hyper-specific installation. This would be the strongest form of testing for this as it would truly represent the intended user workflow. However, it would result in an awful lot of action deployments, which could get spammy and could be considered a fair amount of 'wasted' computational resources if set to trigger on every push. I'd say if that's the route we want to go it would be better to have it set as a daily schedule (or longer).

setup.py
```diff
@@ -1,23 +1,34 @@
...
     testing_suite_dependencies = f.readlines()
 
-extras_require = dict(full=full_dependencies, test=testing_suite_dependencies, docs=documentation_dependencies)
+extras_require = defaultdict(list)
+subpaths = (root / "src" / "neuroconv" / "datainterfaces").rglob("*/requirements.txt")
+for subpath in subpaths:
+    with open(subpath) as f:
+        subpath_requirements = f.readlines()
+    extras_require[subpath.parent.name].extend(subpath_requirements)
+    full_dependencies.update(subpath_requirements)
```
CodyCBakerPhD, line 25: With this approach, we eventually we wouldn't even need to maintain a separate `requirements-full.txt` file with duplicated info - just rely on this machinery to amalgamate all the sub requirements automatically.

setup.py
```diff
@@ -1,23 +1,36 @@
...
+
+extras_require = defaultdict(list)
+subpaths = (root / "src" / "neuroconv" / "datainterfaces").rglob("*/requirements.txt")
+for subpath in subpaths:
+    with open(subpath) as f:
+        subpath_requirements = f.readlines()
+    extras_require[subpath.parent.name].extend(subpath_requirements)
+    full_dependencies.update(subpath_requirements)
+
+    if "ophys" in subpath.parts:
```
bendichter, line 29: I think this is a little fragile, as it assumes "ophys" is not anywhere else in the path. I think a better approach might be to rglob `root / "src" / "neuroconv" / "datainterfaces"/ "ophys"`

CodyCBakerPhD: Sure

bendichter:
@CodyCBakerPhD so you would prefer to have empty requirements.txt files to indicate that this data modality has no additional requirements? I take it you prefer this to just not having a requirements file for that data type

CodyCBakerPhD:
> so you would prefer to have empty requirements.txt files to indicate that this data modality has no additional requirements

That does mimic the approach of leaving empty `__init__.py` files required of package setup tools... which I've maybe gotten used to...

The advantage is to avoid user confusion from being able to install with something like `pip install neuroconv[pyopenephys]` but not be able to make a call like `pip install neuroconv[spikeglx]`...

Of course to avoid that you could have simply defined the `spikeglx` key in the defaultdict in setup, but to do that for every interface that does not have extra requirements removes the benefit of this all being taken care of 'automatically' a the setup level...

So be defining an empty `requirements.txt` file there, the glob takes care of everything at the expense of having a bunch of empty files cluttering up each submodule...

Then again, at least then each submodule will have a very standard appearance (`__init__.py` + `requirements.txt` + `<interface name>_utils.py` + `<interface name>.py`)...

IDK what the best solution is, that's just my thought process right now

bendichter:
> Of course to avoid that you could have simply defined the spikeglx key in the defaultdict in setup, but to do that for every interface that does not have extra requirements removes the benefit of this all being taken care of 'automatically' a the setup level...

I feel like you could find a way to do this automatically. Something like:

```python

for modality in ['ecephys', 'ophys', 'icephys', 'behavior']:
    for data_type in os.listdir(datainterfaces_path / modality):
    modality_requirements = []
        requirements_path = datainterfaces_path / modality / data_type / 'requirements.txt'
        if os.isfile(requirements_path):
            requirements = read_requirements(requirements_path)
            extras_require.update({data_type: requirements})
            modality_requirements.extend(requirements)
        else:
            extras_require.update(data_type=[])
    extras_require.update({modality: list(set(modality_requirements))})
extras_require.update
    
```
That's an option, if you want it. It might be better than having a bunch of empty files.

CodyCBakerPhD:
> That's an option, if you want it. It might be better than having a bunch of empty files.

Sure, but then we're not making full use of recursive globbing - instead, we structurally define the type of recursion to perform on the package.

I guess I don't really care either way - any thoughts from the rest of the team @h-mayorquin @weiglszonja @alejoe91 @luiztauffer ?

luiztauffer:
@CodyCBakerPhD I like this organization of dependencies a lot!
Maybe we should discourage the use of `full`, or even not have it at all? It would decrease the chances of incompatible dependencies, and make it easier to spot these incompatibilities once they occur within specific combinations of datainterfaces

> Personally I think I'd be fine with just one of these steps per each modality.

I agree one step per modality sounds good enough for the CI tests

> Sure, but then we're not making full use of recursive globbing - instead, we structurally define the type of recursion to perform on the package.

I don't see any clear objective advantage in either way, it looks like it will boil down to aesthetics feelings: 
`having a bunch of functionally useless empty files` vs `explicit organization consistency of data interfaces plugins`
I slightly favor the latter.

.github/workflows/testing.yml
h-mayorquin: I think we should deploy the tests somewhere where the basic version has not be installed before as this is the way that I think the functionality will be used. That is, it has more ecological validity. What do you think?

h-mayorquin:
I have some comments in #94 where I advocate for a more gradual approach to this changes that take into account recent developments in python packaging.

.github/workflows/testing.yml
CodyCBakerPhD: Both ways need to work - I'm finding it very convenient personally to start from a minimal base (probably the first thing people will intuitively think to do) then if I need to add something (or more than one thing) to use an interface I can just do `install neuroconv[format_name]` and it will quickly skip the installation of everything except for the items needed for what I requested

CodyCBakerPhD: Although, don't you think `setuptools` are trustworthy enough to know they will also install minimal requirements any time an extra key is specified? I don't think that's something we would actually have to explicitly test.

I can add a separate gallery for each individual format key (starting from a minimal base) though, to ensure they did in fact wind up in the `extra_requires` with the correct dependencies to run

CodyCBakerPhD:
OK the minimized and modularized installation strategy is ready for review here...

`requirements-minimal.txt` is now scrapped down to the bare minimum of PyNWB/DANDI and higher version pins of their dependencies that we require (like scipy/h5py).

All other external dependencies are added to plug-in style individual `requirements.txt` files at the corresponding levels across the repo, but not including empty files if a given interface doesn't require anything extra.

Current strategy for how this affects testing is to...

Install bare minimum dependencies -> test bare minimum

Go through each modality and install according to modality key -> test everything that modality covers (not including actual data tests; mostly just SpikeInterface or ROIExtractors tools, or OpenCV stuff for behavior).

After those, everything in `full` should be installed, so run all data tests like usual.

bendichter (approved):

.github/workflows/testing.yml
h-mayorquin: > I'm finding it very convenient personally to start from a minimal base (probably the first thing people will intuitively think to do) then if I need to add something (or more than one thing) to use an interface I can just do install neuroconv[format_name] and it will quickly skip the installation of everything except for the items needed for what I requested

So I guess this is the key point. My impression is that it would be more common for users to just install pip install neuroconv[modality] but it could be that they first install the minimum and they go for the modalities as you describe. I don't know what is more common. 

As far as I remember, pip install package[modality] runs both the minimal+extra_requirements at the same time. My feeling is that dependency resolution will work differently in these two cases (sequential with minimal then modality vs installing single modality) so different dependencies might end up being installed on each case.

h-mayorquin:
> Go through each modality and install according to modality key -> test everything that modality covers (not including actual data tests; mostly just SpikeInterface or ROIExtractors tools, or OpenCV stuff for behavior).

Aren't the data tests the most important ones as people install the modality to do conversions over there? Isn't that what we should guarantee? that the conversions in that modality work with the installation of that modality.

.github/workflows/testing.yml
CodyCBakerPhD: > As far as I remember, pip install package[modality] runs both the minimal+extra_requirements at the same time.

Yep!

> My feeling is that dependency resolution will work differently in these two cases (sequential with minimal then modality vs installing single modality) so different dependencies might end up being installed on each case.

Hence the setup for https://github.com/catalystneuro/neuroconv/pull/115 to test that at least to first-order. Should be 'good enough'?

CodyCBakerPhD:
> Aren't the data tests the most important ones as people install the modality to do conversions over there? Isn't that what we should guarantee? that the conversions in that modality work with the installation of that modality.

That's why I asked for a one-liner to use the doctests, which will run on actual data.

There's no way, currently, to use the full tests on data to only run a single format, but that could be something to think about. We've always liked the use of `parametrized` to reduce code there, however, but IDK ultimately we just have to do what we think is best for testing.

Going to transition the discussion of how to do this into https://github.com/catalystneuro/neuroconv/pull/115, though, since that's where it's best.

PR #102, opened 2022-08-20
Download GIN data from S3

CodyCBakerPhD:
Companion to #101 that updates the testing workflows to download the data directly from the S3 bucket instead of messing with datalad and git-annex


CodyCBakerPhD:
Annoyingly, all of the GitHub actions used for S3 transfers are based off of some container that is only compatible with linux - as such, I adjusted the workflow to do the direct installation and usage of the AWS CLI (which is what all the actions ever seemed to need to do anyway).

Now working well on all platforms.

Also, worth noting that I notice the transfer speeds are several orders of magnitude faster than they were with datalad, so that's a plus as well.

bendichter:
This is great! Would it make sense to update the README as well to get data from s3 rather than gin?

h-mayorquin:
Seems good. We have copies of all the gin data in an S3 bucket that we download to all the platforms using AWS client. Increases the complexity of our CI a little bit to avoid the problem of using git-annex / datalad outside of linux.

.github/workflows/testing.yml
```diff
@@ -65,51 +67,30 @@ jobs:
...
-      - if: steps.cache-ophys-datasets.outputs.cache-hit == false && (matrix.os == 'ubuntu-latest' || matrix.os == 'windows-latest')
-        name: "Force GIN: ophys download"
-        run: datalad install -rg https://gin.g-node.org/CatalystNeuro/ophys_testing_data
-      - if: steps.cache-behavior-datasets.outputs.cache-hit == false && (matrix.os == 'ubuntu-latest' || matrix.os == 'windows-latest')
-        name: "Force GIN: behavior download"
-        run: datalad install -rg https://gin.g-node.org/CatalystNeuro/behavior_testing_data
+          pip install awscli==1.25.27
+          aws configure set aws_access_key_id ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws configure set aws_secret_access_key ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
```
h-mayorquin, line 80: Shouldn't the region be configured as well?

.github/workflows/testing.yml
```diff
@@ -65,51 +67,30 @@ jobs:
...
+      - if: steps.cache-ephys-datasets.outputs.cache-hit == false
+        name: Download ephys dataset from S3
+        run: aws s3 cp --recursive ${{ secrets.S3_GIN_BUCKET }}/ephy_testing_data ./ephy_testing_data
+      - if: steps.cache-ophys-datasets.outputs.cache-hit == false
+        name: Download ophys dataset from S3
+        run: aws s3 cp --recursive ${{ secrets.S3_GIN_BUCKET }}/ophys_testing_data ./ophys_testing_data
```
h-mayorquin, line 86: `secrets.S3_GIN_BUCKET`. Is this just the S3 URI? I am curios on why this need to be a secret as even if the URI is public you will not be able to access, right?

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -113,17 +111,20 @@ class TestConverter(NWBConverter):
...
-            interface_kwargs=dict(file_path=str(DATA_PATH / "spike2" / "m365_1sec.smrx")),
-            case_name="smrx",
-        ),
         param(
             data_interface=EDFRecordingInterface,
             interface_kwargs=dict(file_path=str(DATA_PATH / "edf" / "edf+C.edf")),
             case_name="artificial_data",
         ),
     ]
+    if platform != "darwin" or version.parse(python_version()) >= version.parse("3.8"):
```
h-mayorquin, line 120: Just FYI you could do all of this with the platform module:
```
if platform.system != "darwin" or  platform.python_version() > '3.8'
```
People consider `platform.system` more reliable than `sys.platform` but I don't think it matters in this case, so up to you. I think it is nice to use only library so it avoids some overloading of platform here.

tests/test_internals/test_interfaces.py
```diff
@@ -24,16 +24,17 @@
 from neuroconv.datainterfaces.ecephys.basesortingextractorinterface import BaseSortingExtractorInterface
 
 
+@pytest.mark.skipif(
+    platform != "darwin" or version.parse(python_version()) >= version.parse("3.8"),
```
h-mayorquin, line 28: I somehow find the negative  of this more readable `platform == 'drawin' and version < 3.8`. Just a comment.

h-mayorquin (commented):
Everything looks good to me.

Any reason for using the cli utility here `awscli` and not the `keithweaver/aws-s3-github-action@v1.0.0` as you do with the cron for caches?

.github/workflows/testing.yml
CodyCBakerPhD: No need apparently

.github/workflows/testing.yml
CodyCBakerPhD: Don't want to confuse people mainly - the public-facing avenues are the GIN repo's, and that's where individuals should try to get it from (either through GIN CLI or datalad), it's just that S3 is much easier for automated workflows in the cloud (which are essentially just containers).

tests/test_on_data/test_gin_ecephys.py
CodyCBakerPhD: Going with https://stackoverflow.com/a/11674977 here for platform detection

tests/test_internals/test_interfaces.py
CodyCBakerPhD: I had it that way initially but found it less readable to have a big `not` in front of everything casted in parentheses. With this logic you can think of the conditions one at a time.

Not on a `darwin` platform? Skip.
Using Python version 3.8 or above? Skip.

No need to think about joint conditions this way

CodyCBakerPhD:
> Any reason for using the cli utility here awscli and not the keithweaver/aws-s3-github-action@v1.0.0 as you do with the cron for caches?

Yep, reason given in https://github.com/catalystneuro/neuroconv/pull/102#issuecomment-1221411627

tests/test_internals/test_interfaces.py
h-mayorquin: I see, I was confused. I agree about it being more readable without distributing negation in your mind (the big negation in front of everything)

My confusion was that I read the reason
`reason="Not testing on MacOSX with Python<3.8!"`
And I thought that you wanted to skip the test if:
you are in the Darwin platform and the version of python was strictly before 3.8

That's why I wrote:
skip if:
`platform == 'drawin' and version < 3.8`

Am I mistaken that you are doing the opposite? Is the reason written wrong?

.github/workflows/testing.yml
h-mayorquin: > > Any reason for using the cli utility here awscli and not the keithweaver/aws-s3-github-action@v1.0.0 as you do with the cron for caches?
> 
> Yep, reason given in [#102 (comment)](https://github.com/catalystneuro/neuroconv/pull/102#issuecomment-1221411627)

So you can't use the action in other platforms that are not linux. Why not use the cli everywhere then? (to make our action code more similar)

.github/workflows/testing.yml
h-mayorquin: Great. How does it do it? Maybe that's the default one? Maybe it is because of where the machines that execute are? Maybe it gets the default location from the account associated with the secrets.

tests/test_on_data/test_gin_ecephys.py
h-mayorquin: Is that thread and the linked comment that made me think that `platform.system` is more reliable? Did you get the opposite message from it?

tests/test_internals/test_interfaces.py
CodyCBakerPhD: The test, as seen below, only triggers when the user attempts to run the CED interface on a Mac device with Python Version 3.7.

Thus, running this test on anything other than that would fail because it wouldn't be triggered.

Thus, we skip this test if we are not testing on a Mac, or if we are using any Python version strictly above 3.7

Updated the `reason` to reflect this

.github/workflows/testing.yml
CodyCBakerPhD: Theoretically, haven't put that to the test

tests/test_internals/test_interfaces.py
h-mayorquin: Thanks. It was the reason that threw me off.

h-mayorquin (approved):
This is great, all the rest were just suggestions/questions. Thanks for answering them!

.github/workflows/testing.yml
CodyCBakerPhD: Dunno. If I had to guess I'd say it checks the credentials for all existing S3 buckets and only uses the region if there are multiple buckets of the same name?

.github/workflows/testing.yml
h-mayorquin: I tried it this morning. You can't even look at it from the URI.

.github/workflows/testing.yml
h-mayorquin: I see, I was somehow thinking that you that transaction costs or speeds would be different if you specify the wrong region.

CodyCBakerPhD:
> Why not then use the cli everywhere to have similar code across the action files?

B/c one came before the other, and the upload step does run on linux (since that's the only place datalad works without a hitch). Feel free to change it yourself if it bugs you

h-mayorquin:
> > Why not then use the cli everywhere to have similar code across the action files?
> 
> B/c one came before the other, and the upload step does run on linux (since that's the only place datalad works without a hitch). Feel free to change it yourself if it bugs you

Fair enough.

tests/test_on_data/test_gin_ecephys.py
h-mayorquin: It is "the most reliable" part that I am not getting. My reading of the thread that you linked and the linked thread therein seems to suggest that `platform.system` is the most reliable:
https://stackoverflow.com/questions/1854/how-to-identify-on-which-os-python-is-running-on/58071295#58071295

I don't think it matters here though, and from reading above it does seem that `platform.system` does more than `sys.platform` in terms of computation. I just wanted to see if I was misreading something on the threads.

.github/workflows/testing.yml
CodyCBakerPhD: That's definitely something to always be aware of when deploying EC2 or related instances through AWS - using a different region from where your S3 bucket is will generate spurious costs (and this bucket does exist on us-east-2, same as DANDI).

tests/test_on_data/test_gin_ecephys.py
CodyCBakerPhD: Here is the source code for `platform.system`

```python
def uname():

    """ Fairly portable uname interface. Returns a tuple
        of strings (system, node, release, version, machine, processor)
        identifying the underlying platform.

        Note that unlike the os.uname function this also returns
        possible processor information as an additional tuple entry.

        Entries which cannot be determined are set to ''.

    """
    global _uname_cache

    if _uname_cache is not None:
        return _uname_cache

    # Get some infos from the builtin os.uname API...
    try:
        system, node, release, version, machine = infos = os.uname()
    except AttributeError:
        system = sys.platform
        node = _node()
        release = version = machine = ''
        infos = ()

    if not any(infos):
        # uname is not available

        # Try win32_ver() on win32 platforms
        if system == 'win32':
            release, version, csd, ptype = win32_ver()
            machine = machine or _get_machine_win32()

        # Try the 'ver' system command available on some
        # platforms
        if not (release and version):
            system, release, version = _syscmd_ver(system)
            # Normalize system to what win32_ver() normally returns
            # (_syscmd_ver() tends to return the vendor name as well)
            if system == 'Microsoft Windows':
                system = 'Windows'
            elif system == 'Microsoft' and release == 'Windows':
                # Under Windows Vista and Windows Server 2008,
                # Microsoft changed the output of the ver command. The
                # release is no longer printed.  This causes the
                # system and release to be misidentified.
                system = 'Windows'
                if '6.0' == version[:3]:
                    release = 'Vista'
                else:
                    release = ''

        # In case we still don't know anything useful, we'll try to
        # help ourselves
        if system in ('win32', 'win16'):
            if not version:
                if system == 'win32':
                    version = '32bit'
                else:
                    version = '16bit'
            system = 'Windows'

        elif system[:4] == 'java':
            release, vendor, vminfo, osinfo = java_ver()
            system = 'Java'
            version = ', '.join(vminfo)
            if not version:
                version = vendor

    # System specific extensions
    if system == 'OpenVMS':
        # OpenVMS seems to have release and version mixed up
        if not release or release == '0':
            release = version
            version = ''

    #  normalize name
    if system == 'Microsoft' and release == 'Windows':
        system = 'Windows'
        release = 'Vista'

    vals = system, node, release, version, machine
    # Replace 'unknown' values with the more portable ''
    _uname_cache = uname_result(*map(_unknown_as_blank, vals))
    return _uname_cache

### Direct interfaces to some of the uname() return values

def system():

    """ Returns the system/OS name, e.g. 'Linux', 'Windows' or 'Java'.

        An empty string is returned if the value cannot be determined.

    """
    return uname().system
```

As you can see, the initial value starts off as `sys.platform` and then gets mapped through a whole bunch of conditions, 99% of which (and this seems to hold for all secondary functions in the `platform` package) are for classifying configurations of Windows and Linux. The most they touch MacOS is to capitlize the `sys.platform` output of `"darwin"` to `"Darwin"`.

So again, for simply determining if I am specifically on (or not on, but don't care specifically what I am on) a Mac platform, all I need is `sys.platform`. If I wanted to know specifically the kind of device otherwise, yes I would use `platform.system`.

h-mayorquin: Thank you for looking into this.
> As you can see, the initial value starts off as sys.platform

This is only true if ` os.uname()` does not work (which I read only works on Unix, so mostly read). 

It seems that the confusion from the linked thread is that they wrote this as python 3.7 but this function changed after, check it out:
https://github.com/python/cpython/blob/3.7/Lib/platform.py#L949

PR #107, opened 2022-08-24
Updated some interface names to match SpikeInterface

CodyCBakerPhD:
Names Kilosort -> "KiloSort" and Neuroscope -> "NeuroScope"


h-mayorquin (commented):
Are we changing how we refer to them in the gallery and doc strings as well? If so, there are references (uses of the name) at the top of the conversion gallery files that use the old way.

Also, I was concern about the links the API as they use a reference to the API of the following form:
```
[:py:class:`~neuroconv.datainterfaces.ecephys.neuroscope.neuroscopedatainterface.NeuroscopeRecordingInterface`](https://github.com/catalystneuro/neuroconv/blob/main/docs/conversion_examples_gallery/recording/neuroscope.rst#id1).
```

They seem to be non broken but lead to a class with the wrong name.

CodyCBakerPhD:
@h-mayorquin Yep, good catch. Think I fixed them all now but you can double check me on that

h-mayorquin:
> @h-mayorquin Yep, good catch. Think I fixed them all now but you can double check me on that

They look all right to me, now. 

What about the names at this level:
https://neuroconv--107.org.readthedocs.build/en/107/conversion_example_gallery.html#recording

CodyCBakerPhD:
> What about the names at this level:
https://neuroconv--107.org.readthedocs.build/en/107/conversion_example_gallery.html#recording

Fixed that now too

h-mayorquin (approved):

PR #109, opened 2022-08-25
Use ELLIPSIS more in gallery

CodyCBakerPhD:
PR #74 introduced ELLIPSIS into the doctests, which enables some options that weren't available before in terms of style.

This showcases the simplified appearance on the SpikeGLX gallery.

### DIFF
--- a/docs/conversion_examples_gallery/recording/spikeglx.rst
+++ b/docs/conversion_examples_gallery/recording/spikeglx.rst
@@ -13,7 +13,7 @@ Convert spikeglx data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.
     >>> # For this interface we need to pass the location of the ``.bin`` file
     >>> file_path = f"{ECEPHY_DATA_PATH}/spikeglx/Noise4Sam_g0/Noise4Sam_g0_imec0/Noise4Sam_g0_t0.imec0.ap.bin"
     >>> # Change the file_path to the location in your system
-    >>> interface = SpikeGLXRecordingInterface(file_path=file_path, verbose=False)
+    >>> interface = SpikeGLXRecordingInterface(file_path=file_path)
     >>>
     >>> # Extract what metadata we can from the source files
     >>> metadata = interface.get_metadata()
@@ -26,7 +26,8 @@ Convert spikeglx data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.
     >>> # Choose a path for saving the nwb file and run the conversion
     >>> nwbfile_path = f"{path_to_save_nwbfile}"
     >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
-    >>>
-    >>> # If the conversion was successful this should evaluate to ``True`` as the file was created.
-    >>> Path(nwbfile_path).is_file()
-    True
+    Source data is valid!
+    Metadata is valid!
+    Conversion options are valid!
+    
+    NWBFile saved at ...


docs/conversion_examples_gallery/recording/spikeglx.rst
```diff
@@ -26,7 +26,4 @@ Convert spikeglx data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.
     >>> # Choose a path for saving the nwb file and run the conversion
     >>> nwbfile_path = f"{path_to_save_nwbfile}"
     >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
-    >>>
-    >>> # If the conversion was successful this should evaluate to ``True`` as the file was created.
-    >>> Path(nwbfile_path).is_file()
-    True
```
CodyCBakerPhD, line 32: I always thought this looked a little awkward to introduce what is basically an assertion statement into code that is meant to be easily shared and interpretable by novices to coding in general.

With this approach, we can save 3 lines per gallery and the output is just as (if not more) informative when the user themselves tries to run it on their data and file paths.

docs/conversion_examples_gallery/recording/spikeglx.rst
```diff
@@ -13,7 +13,7 @@ Convert spikeglx data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.
     >>> # For this interface we need to pass the location of the ``.bin`` file
     >>> file_path = f"{ECEPHY_DATA_PATH}/spikeglx/Noise4Sam_g0/Noise4Sam_g0_imec0/Noise4Sam_g0_t0.imec0.ap.bin"
     >>> # Change the file_path to the location in your system
-    >>> interface = SpikeGLXRecordingInterface(file_path=file_path, verbose=False)
+    >>> interface = SpikeGLXRecordingInterface(file_path=file_path)
```
CodyCBakerPhD, line 16: Always thought it was a bit awkward as well to ask a first-time user to disable verbosity, when verbosity exists to allow human-readable communication with the user (not that it does much of this besides confirmation of validations at the NWBConverter level and success return at end of all `run_conversion` commands, but principle is it could always be enhanced).

docs/conversion_examples_gallery/recording/spikeglx.rst
h-mayorquin: That line was not meant to be informative to the user. It was just as a means to test that the conversion run OK for us (to alert ourselves that our documentation was not working with our code base). To be honest, I am not sure that what I did does a great job at that goal but is appropriate to ask about this new approach: does the doctest now fail if the conversion fails? What do you think?

I sort of agree with the virtue of your approach by the way. I think that hiding details from first time users is great (that's why I think making them modify metadata is noise). That said, you seem to justify it in terms of saving lines which I think is always the wrong metric to focus on (reducing number of lines). Is there something I am missing there?

h-mayorquin (commented):
I still think we should just not have verbose by default. This is not the standard in most tools and introduce side effects to anyone that uses our code with default parameters (printing to standard output). 

If I remember a while ago you were very concerned that we would need to deactivate `tqdm` if we did this but I think we can just separate that and leave on by default if that is your major concern.

CodyCBakerPhD:
> That line was not meant to be informative to the user. It was just as a means to test that the conversion run OK for us (to alert ourselves that our documentation was not working with our code base anymore and therefore we should make it up to date). To be honest, I am not sure that what I did does a great job at that goal but is appropriate to ask the same about this new approach: does the doctest now fail if the conversion fails? What do you think?

Yes, if there is an error at the `run_conversion` step, the doctest will fail. Also, if the output does not match "NWBFile saved at ...", the doctest will also fail. So it's functionally equivalent but reads better as documentation/example code

CodyCBakerPhD:
>I sort of agree with the virtue of your approach by the way. I think that hiding details from first time users is great (that's why I think making them modify metadata is noise). That said, you seem to justify it in terms of saving lines which I think is always the wrong metric to focus on (reducing number of lines). Is there something I am missing there?

If a user is trying to figure out at a glance if our tools are worth their effort to learn to use, showing demo code that is as concise, readable, and succinct as possible is a huge plus. The more unnecessary complications that get added in, the more they have to read, the more they have to think about, and thus the lower the likelihood they want to use it.

One of the sales lines often included in presentations of NeuroConv at workshops/conferences is the 'ability to convert data from proprietary formats to NWB in only a few lines of code' compared to using PyNWB directly, which would be hundreds of lines.

CodyCBakerPhD:
> I still think we should just not have verbose by default.

For people using `neuroconv.tools` as an API I can still see that argument. For packages that exist more as core tools or basic API's (like neo/SI), that is definitely true.

But our package is supposed to be user-friendly. For people using the `neuroconv.datainterfaces`, `NWBConverter`, or CLI interactivity is essential to a good user experience - people need feedback and/or progress display that the software is behaving as expected throughout the process.

Also we are a DANDI tool now, and DANDI functions with verbosity by default as well (could you imagine if it didn't?)

h-mayorquin:
> If a user is trying to figure out at a glance if our tools are worth their effort to learn to use, showing demo code that is as concise, readable, and succinct as possible is a huge plus. The more unnecessary complications that get added in, the more they have to read, the more they have to think about, and thus the lower the likelihood they want to use it.
> 
> One of the sales lines often included in presentations of NeuroConv at workshops/conferences is the 'ability to convert data from proprietary formats to NWB in only a few lines of code' compared to using PyNWB directly, which would be hundreds of lines.

My whole point is that complications is different from lines. Clarity is what matters, brevity is a means to and end. Easy is the presentation.

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -172,6 +172,6 @@ def make_or_load_nwbfile(
                 io.write(nwbfile)
 
                 if verbose:
-                    print(f"NWB file saved at {nwbfile_path}!")
+                    print(f"NWBFile saved at {nwbfile_path}!")
```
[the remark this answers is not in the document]
bendichter: IMO NWBFile (nwbfile) refer to the `NWBFile` object. When you are talking about a file that is in NWB format, I use a space.

docs/conversion_examples_gallery/recording/spikeglx.rst
bendichter: honestly, I'd prefer verbose to be False by default and to not set it to True here. I think verbose should really only be used for debugging purposes if the code does not work as expected in non-verbose mode. This is similar to how sklearn uses it. See e.g. https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html

h-mayorquin:
> > I still think we should just not have verbose by default.
> 
> For people using `neuroconv.tools` as an API I can still see that argument. For packages that exist more as core tools (like HDMF/PyNWB) or basic API's (like neo/SI), that is definitely true.
> 
> But our package is supposed to be user-friendly. For people using the `neuroconv.datainterfaces`, `NWBConverter`, or CLI interactivity is essential to a good user experience - people need feedback and/or progress display that the software is behaving as expected throughout the process.

I don't think this is an obvious as you make it to be. Many packages that are used by our target community (scikit-learn, scipy, pandas) don't have anything of the sort. You don't get "pandas data frame transformed successfully" in `pandas`, "fit process ran succesffuly" in `scikit-learn` or "root found as expected" in scipy. At a personal level, I don't go with the expectation to get confirmation form a library that the core operation is correct when I run it. I assume it is working fine unless it fails and in that case I should be informed by assertions or warnings  while verbose is commonly reserved for debugging.

> Also we are a DANDI tool now, and DANDI functions with verbosity by default as well (could you imagine if it didn't?)

Rclone has the -P flag to display transfer statistics and is not on by default, is that what you mean? I have not used the DANDI CLI as much as you have, what is the part that you feel is essential to display?

PR #110, opened 2022-08-25
Showcase tzinfo idea for gallery and general usage

CodyCBakerPhD:
Another idea I had for simplifying much of the code in the galleries.

Basically a convenience functionality for setting timezone information in metadata. Not necessarily specific to `session_start_time`, but could have it alter any metadata info that is ultimately specified as ISO/datetime objects.

Saves ~6 lines per gallery (~2 unnecessary imports, ~4 lines that parse the existing ISO string as datetime just to cast it back in the update)

### DIFF
--- a/docs/conversion_examples_gallery/recording/spikeglx.rst
+++ b/docs/conversion_examples_gallery/recording/spikeglx.rst
@@ -5,8 +5,6 @@ Convert spikeglx data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.
 
 .. code-block:: python
 
-    >>> from datetime import datetime
-    >>> from dateutil import tz
     >>> from pathlib import Path
     >>> from neuroconv.datainterfaces import SpikeGLXRecordingInterface
     >>>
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
I also wouldn't be opposed to setting `tzinfo` at some global level, like at `__init__` of interfaces or convert objects.

Or even set as a global variable via some convenience function `neuroconv.utils.set_timezone(tzinfo="US/Pacific")` that could be called after imports but before anything else?

bendichter:
Cool idea! Maybe there is a way we can make this work across all data interfaces...


```python

class BaseDataInterface:
    
    def get_metadata(tzinfo):
        metadata = self._get_metadata()
        if metadata["NWBFile"]["session_start_time"] and tzinfo:
            tzinfo_object = dateutil.tz.gettz(tzinfo) if isinstance(tzinfo, str) else tzinfo
            metadata["NWBFile"]["session_start_time"].replace(tzinfo=tzinfo_object)
        return metadata

    @abstractmethod
    def _get_metadata():
        pass

class SpikeGLXDataInterface(RecordingExtractorInterface):
    def _get_metadata():
        # we'd need to rename all of the individual class methods to "_get_metadata"

**Comment by bendichter** [comment]
Another option besides global variables is to have a `.neuroconv` config files that lives in your home directory.

**Comment by CodyCBakerPhD** [comment]
> Another option besides global variables is to have a .neuroconv config files that lives in your home directory.

A config file would be fantastic, actually. More than `tzinfo`, you could also set a `verbose` level to use in that way, which might be the best solution to the discussion in #109.

We'd also want to be `.toml` compatible in any package that uses neuroconv for converting data, *e.g.* in `ahrens-lab-to-nwb/pyproject.toml`, have

```python
[neuroconv]
tzinfo = "US/Eastern"
verbose = True # or an integer, default to be discussed
# anything else we can think of
```

bendichter:
> you could also set a verbose level to use in that way

I'd rather only set variables that we don't expect to change. You might want to change `verbose` run-to-run. This could also be the place where we put the data paths for testing.

CodyCBakerPhD:
Could also set data transfer credentials like a DANDI API token or Rclone credentials at the NeuroConv config level instead of setting them as environment variables (which NeuroConv could take care of), which would be a bit of a convenience.

h-mayorquin:
I think that the current form of modifying the current time zone can be improved. It is important to state that while in this case you can can simplify a lot because spikeglx already provides a session start time.  This is no the case for all the interfaces so you will only be saving the lines related to the time zone. Now, the most important part of the `session_start_time` field is the datetime part as we use it for synchronization purposes and it is good to establish its importance  in the conversion gallery for that reason alone. So what to do with the time zone part? How should we simplify that? The conversions run fine without them so maybe this information should not even be in the conversion gallery. This is to me the first important to discuss:

** Do we want to describe in the conversion gallery how users can add timezone info to their conversions**

To me, once we decide this we can discuss how to do it and how to add time-zone in the conversion workflow (either as a general argument in the metadata, in the __init__ or in a general config file as it is discussed above).

PR #112, opened 2022-08-25
Sync with latest ROIExtractors

CodyCBakerPhD:
Recent improvements to ROIExtractors altered the behavior of a component of the `write_imaging` method, specifically `add_epochs`.

For the same justification that it was removed in ROIExtractors, this method has always gone untested and unused across the supported formats, and unused in actual conversion projects as well. Epoch information, similar to trials information, is instead typically added via secondary data interfaces that read it from non-imaging files.


h-mayorquin (approved):
Great!

Issue #14, opened 2021-10-25
Centralize definitions for metadata schemas


CodyCBakerPhD:
Adding this to the global to-do list, low priority. Would be another good first issue to tackle for someone learning the metadata details. 

**Describe the solution you'd like**
Some of the base interfaces, such as the [RecordingInterface](https://github.com/catalystneuro/nwb-conversion-tools/blob/master/nwb_conversion_tools/datainterfaces/ecephys/baserecordingextractorinterface.py#L50-L63), explicitly set definitions for their metadata schema structures.

These could and probably should be centralized into a single util location, and can also probably be generalized due to similarities across interfaces.

## Checklist

- [X] Have you ensured the feature or change was not already [reported](https://github.com/catalystneuro/nwb-conversion-tools/issues)?
- [X] Have you included a brief and descriptive title?
- [X] Have you included a clear description of the problem you are trying to solve?
- [X] Have you checked our [Contributing](https://nwb-conversion-tools.readthedocs.io/en/master/developer_guide.html) document?

h-mayorquin:
Linking here from #113 so this discussion is not lost. I agree with @CodyCBakerPhD that this is desirable as our metadata and the one in the API that is generated on the fly do not match. I would rather prefer to have ours hard-coded so we can consult it in an easier way (plus we have less code to maintain).  There must be another way of assuring ourselves that we are up to date but I am not that concerned about this as the schema does not -should not?- change this fast.

I add the link here to Cody's specific comment so it does not get lost once we merge the other issue:
https://github.com/catalystneuro/neuroconv/pull/113#issuecomment-1227724347

How are you thinking about this, @bendichter ?

h-mayorquin:
Another advantage of having the schema right now is that we can point it to users to it in the documentation (ideally, we should be able to point them to the real schema but as we mentioned, it is not identical).

CodyCBakerPhD:
Was closed accidentally by the first of several PRs necessary to finish this across the metadata structure

PR #113, opened 2022-08-25
[Testing]: Dev branch gallery workflows

CodyCBakerPhD:
Something I've wanted to do for a while is have some workflows that test against dev branches of key packages we depend on, like PyNWB.

I actually just ran across something while debugging a conversion and wanted to see if the workflows can replicate what I'm seeing before attempting a fix.

Now, a question remains how we want to implement such an approach. My current approach is to simply keep this on a separate branch from `main`, perhaps with a daily or weekly scheduled run, and either keep the draft PR open so it directly displays when the workflows are passing or failing, or keep it closed (I would still get the emails whenever it fails, but others wouldn't have to worry about it).

DANDI has a part of this in their testing matrix: https://github.com/dandi/dandi-cli/blob/master/.github/workflows/test.yml#L35-L43 but our suite is considerably heavier than theirs is so I don't think that would scale very well.


CodyCBakerPhD:
Yeaaahhh, this is what I see locally as well: https://github.com/catalystneuro/neuroconv/runs/8022980924?check_suite_focus=true

Basically I think something must have changed with docval upstream, and now our `get_schema_from_hdmf_class` method is dropping various fields.

If anyone knows what that might be they're welcome to try to fix it on a splinter from this branch, otherwise I'm going to place a higher priority on the long-standing issue #14 to just have interfaces load the less-obfuscated [metadata schema definition](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/schemas/metadata_schema.json).

bendichter:
> If anyone knows why that might be they're welcome to try to fix it on a splinter from this branch, otherwise I'm going to place a higher priority on the long-standing issue https://github.com/catalystneuro/neuroconv/issues/14 to just have interfaces load the less-obfuscated [metadata schema definition](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/schemas/metadata_schema.json).

I'd rather have the json schema be derived directly from the API rather than managing those fields separately.

It looks like the fields that are breaking are floats. In pynwb, several floats were changed in docval from 'float' to `float`. [here](https://github.com/NeurodataWithoutBorders/pynwb/commit/d75cc825fe4aaeae18feeac3cbcb3f70651a4c09)

CodyCBakerPhD:
> I'd rather have the json schema be derived directly from the API rather than managing those fields separately.

But we do not always define or use the structure in the same way - one example is the `rate` of various time series objects. PyNWB would see that as 'metadata' but we do not treat it as a field of our own metadata structure (we add it from underlying extractor objects).

We also need to keep our own metadata definitions like this up to date anyway for the YAML CLI specification: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/schemas/yaml_conversion_specification_schema.json so it's work that we're doing anyway.

The main complaint about having it autogenerated from the upstream classes is it becomes very difficult to debug since you can only diagnose a potential issue after you generate for an interface initialized on some data. @h-mayorquin and I actually found some bugs this way a couple weeks ago in the ophys interfaces (BTW any status on those fixes?) and it took like 20+ minutes to figure out why the structure wasn't validating correctly (and he was the only one with immediate access to the data, the only way I could load the `metadata_schema` he generated was for him to dump it to a JSON file and share it with me). Whereas if we had it in a central place (which we already do, we just need to update it a bit and utilize it in the interfaces) we could tell such things at a glance  from anywhere.

bendichter:
The current error:

```pytb
   def _error_on_new_warn_on_construct(self, error_msg: str):
        """
        Raise an error when a check is violated on instance creation.
        To ensure backwards compatibility, this method throws a warning
        instead of raising an error when reading from a file, ensuring that
        files with invalid data can be read. If error_msg is set to None
        the function will simply return without further action.
        """
        if error_msg is None:
            return
>       if not self._in_construct_mode:
E       AttributeError: 'ImageSeries' object has no attribute '_in_construct_mode'
```

looks like we aren't keeping up with the latest hdmf. See this PR: https://github.com/hdmf-dev/hdmf/commit/727636d8caf9e60042f989b3fc8567db71e171e5

CodyCBakerPhD:
> Should we have this PR install the latest from hdmf as well?

One would have hoped that installing and using the main dev branch of PyNWB would have also installed any further upstream dev branch dependencies, or at least have an extra requirements option for it, but I suppose not. So yeah, we'll always have to pair HDMF and PyNWB dev together

CodyCBakerPhD:
Hmm IDK, working with `dev` of HDMF didn't resolve it either...

h-mayorquin:
This is good, I was thinking on doing the same at the level of roiextractors with `neuroconv` for a while.

CodyCBakerPhD:
@h-mayorquin Yeah OK, IDK. Something happened and everything seems to magically pass CI now.

Requested/needed to fully test #170

h-mayorquin (commented):
I am confused. As it is, this would change our `testing.yml` to `dev-testing.yml` and since we added the deploy testing mechanism (to not run the suit when docstrings are only changed) this references `testing.yml` absolutely from main:

https://github.com/catalystneuro/neuroconv/blob/ebf3939c7036b9d64e431ec70b2b965789a81810/.github/workflows/deploy-tests.yml#L13

So this will fail once we merge, right?

Moreover, I feel that testing against dev should be a cron job like the ones we have for the formatwise installation tests and have a matrix for each of the dev branches (maybe we should also roiextractors?). 

Am I missing something?

CodyCBakerPhD:
> I am confused. As it is, this would change our testing.yml to dev-testing.yml and since we added the deploy testing mechanism (to not run the suit when docstrings are only changed) this references testing.yml absolutely from main:

Err yeah not sure what happened there - I think originally I just meant for this to be a thing that existed on this branch and the PR kept open, but it would be even better to have it run on all PRs so then I adjusted that. Should be fixed now - there should be two separate workflows, `testing` which tests whatever branch triggered on latest PyPI releases of everything, and does minimal tests before full tests, and now `dev-testing` which only runs full against all the dev branches

 
> So this will fail once we merge, right?

I'll adjust the target after merge, but I did just check and it was indeed not running this branches intended workflow, so fixing that now


> cron job like the ones we have for the formatwise installation tests and have a matrix for each of the dev branches

I think I set a cron on it so we get notifications when it fails, but I do think it should trigger on PRs so things like #170 can benefit from it, too.

Right now all dev branches are considered together at once - this is especially recommended for HDMF+PyNWB since they often have dev branches that work in tandem. The trouble with a matrix over those is then we do end up deploying an awful lot of testing, and I feel like any bugs that would should up on a single part of that matrix would also show up in the one that uses all dev branches simultaneously. Doing it this way also alerts us to inter-incompatible changes, *e.g.*, changes made to dev of HDMF somehow breaks something with changes to SI or ROIExtractors

> (maybe we should also roiextractors?).

Yeah, sure

CodyCBakerPhD:
@h-mayorquin With the `dev-testing.yml` as a separate workflow, do you think we should still test it over a matrix of Python versions and platforms? Or just pin it to 3.9 ubuntu?

h-mayorquin:
> @h-mayorquin With the `dev-testing.yml` as a separate workflow, do you think we should still test it over a matrix of Python versions and platforms? Or just pin it to 3.9 ubuntu?

The latter. Just latest version on Ubuntu.

h-mayorquin:
> 
> > cron job like the ones we have for the formatwise installation tests and have a matrix for each of the dev branches
> 
> I think I set a cron on it so we get notifications when it fails, but I do think it should trigger on PRs so things like #170 can benefit from it, too.
> 

My only concern is that if the dev is failing for reasons other than our own it might block some of our PRs. What do you think?

CodyCBakerPhD:
> My only concern is that if the dev is failing for reasons other than our own it might block some of our PRs. What do you think?

Right, that is why I originally wanted it to be something visible but separate... We could always just have it be non-required for merge, but still shows up to inform us when something is wrong just in case it is our fault and not theirs

Really, ideally, those sorts of upstream things shouldn't happen too often (or at all, if the upstream projects used a similar testing workflow to alert them of dev-level breaking changes)

CodyCBakerPhD:
Looks like there might still be problems to fix on our side, putting this back in draft for now

CodyCBakerPhD:
Leaving To-Do's: 

- [ ] Add test for new TypeError for `starting_frames` in `MovieInterface`
- [ ] Add test for new ValueError for `starting_frames` in `MovieInterface`

.github/workflows/dev-testing.yml
```diff
@@ -0,0 +1,110 @@
...
+        with:
+          python-version: 3.9
+
+      - name: Global Setup
+        run: |
+          python -m pip install -U pip  # Official recommended way
+          pip install pytest-xdist
+          git config --global user.email "CI@example.com"
+          git config --global user.name "CI Almighty"
+          pip install wheel # Needed for scan image
```
h-mayorquin, line 40: Should we move this line to the corresponding scan image requirements btw?

h-mayorquin (changes requested):
Looks good to me, two small things:
1) Can you add something about the tests that you added to the CHANGELOG?
2) Can you make the `starting_frames` on the new tests that you added integers?

.github/workflows/dev-testing.yml
CodyCBakerPhD: A good question, but unfortunately no - I tried that back when this was added but it needs `wheel` installed _prior_ to installing dependencies from requirements/setup including the `ScanImageReader` package. Very silly, I know.

CodyCBakerPhD:
> Can you add something about the tests that you added to the CHANGELOG?

Done

> Can you make the starting_frames on the new tests that you added integers?

Derp, fixed. Also improved annotation typing to make structure clearer.

.github/workflows/dev-testing.yml
h-mayorquin: What a shame indeed.

h-mayorquin (approved):
Great.

PR #114, opened 2022-08-26
Add workflow for checking external links in docs

CodyCBakerPhD:
This workflow has been around for a while, but it hasn't made its way here yet.

Now it has, though.


CodyCBakerPhD:
Looks like a few aren't in fact working: https://github.com/catalystneuro/neuroconv/runs/8040199100?check_suite_focus=true

@h-mayorquin or @weiglszonja could one of you take a look and make updates when you get some time?

h-mayorquin:
The two links that the functionality indicate as broken are false positives. They do work. I am not sure why this is failing. Some problem with links with anchors (#) but looking the [source code ](https://www.sphinx-doc.org/en/master/_modules/sphinx/builders/linkcheck.html)this should be handled. No idea.

.github/workflows/test_external_links.yml
```diff
@@ -0,0 +1,21 @@
...
+      - uses: actions/checkout@v2
+      - run: git fetch --prune --unshallow --tags
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - name: Install sphinx Dependencies
+        run: pip install -r requirements-rtd.txt
+      - name: Install package for API docs
+        run: pip install -e .
```
h-mayorquin, line 19:
```suggestion
        run: pip install -e .[docs]
```
This should remove the need of installing `requirements.rtd`

CodyCBakerPhD:
@h-mayorquin any luck figuring out this one?

h-mayorquin:
> @h-mayorquin any luck figuring out this one?

No luck so far. I had the hope that other would help.

h-mayorquin:
@CodyCBakerPhD last thing of today, this should get us going. Apparently checking the anchors in some sites is not trivial. I still need to read more about it but this should make it work at least at the link level.

CodyCBakerPhD:
Thanks - relaxing that can be a longer-term thing, I did just want the main workflow here for our own sanity.

h-mayorquin (approved):

PR #116, opened 2022-08-28
Add SegmentationInterface frame stubbing

CodyCBakerPhD:
Relies on https://github.com/catalystneuro/roiextractors/pull/201

Necessary for efficient debugging of my current conversion (entire segmentation data is over a GB, but a single frame is only ~1 MB; for comparison the typical 100 frame stub of a SpikeGLX recording is about 1/10 of that).


src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -888,7 +888,7 @@ def add_summary_images(
 
 
 def write_segmentation(
-    segext_obj: SegmentationExtractor,
+    segmentation_extractor: SegmentationExtractor,
```
h-mayorquin, line 891: Great! This has bothered me for a while.

h-mayorquin (commented):
Looks good to me

Looking at the future what do you think about just having making a `stub_test` a dict with stubggin options instead of a boolean? Advantages of this approach:
* No (or minimal) break of backwards compatibility.
* Different modalities might require different stubbing options (as a far fetched example we might want to stub images not only by frames but by position, or to stub units in sorter) and therefore their `run_conversion` will require different parameters. Having a single dict will allow us to have this flexibility while at the same time keep a consistency interface of the `run_conversion` signature across the different modalities.
* Avoids the [code smell of too many parameters](https://en.wikipedia.org/wiki/Code_smell#:~:text=Too%20many%20parameters%3A%20a%20long,a%20more%20clean%2Dcut%20way.) and keeps things that are related together in a single object (in this case a dict) which allows a more general testing procedure.

CodyCBakerPhD:
> Looking at the future what do you think about just having making a stub_test a dict with stubbing options instead of a boolean?

Yeah I've been thinking about that for a while, but if/when we do it we should do it for everything with a deprecation warning like with `save_path`. So we'll have 3 `*_options: dict` where applicable for `iterator`, `compression`, and `stub`. Fields of `stub` would be `enable`, `frames` to start with, and possible add `channels` or `subimage_shape` in the future.

CodyCBakerPhD:
But best to do in separate PR so we can get them all at once.

h-mayorquin:
> > Looking at the future what do you think about just having making a stub_test a dict with stubbing options instead of a boolean?
> 
> Yeah I've been thinking about that for a while, but if/when we do it we should do it for everything with a deprecation warning like with `save_path`. So we'll have 3 `*_options: dict` where applicable for `iterator`, `compression`, and `stub`. Fields of `stub` would be `enable`, `frames` to start with, and possible add `channels` or `subimage_shape` in the future.

Great. I agree that is totally out of the scope of this PR.

h-mayorquin (approved):
Tests seem to be passing now. Ready to go I guess.

CodyCBakerPhD:
@h-mayorquin Requires https://github.com/catalystneuro/roiextractors/pull/201, can you give that a review too?

PR #117, opened 2022-08-28
[DO NOT MERGE - splintering]: Segmentation fixes for volumetric data

CodyCBakerPhD:
Various improvements to `write_segmentation` methods and control over which fields get added.

- boolean flag for controlling whether or not ROI centroids are written to the plane segmentation table. This was proving to be a very expensive step for my current conversion (122K+ ROIs per file) so I am disabling it when running stubs
- The current method always writes image masks by default, but there are other types of masks available in PyNWB. In particular, voxel masks are essentially the way my data is stored in the source format, so I'd prefer to write it that way as well. So I added control over what type of mask gets written to the NWBFile. However, see comments below for some difficulties getting this to work with compression.
- I noticed that none of the Fluorescence traces for my SegmentationExtractor were getting chunked or compressed. Added both of those in with some basic iterative write as well.


h-mayorquin:
I would suggest breaking this PR into one for each of your points above. At least 2 and 3 should be separated and tests for them added as well.

h-mayorquin (changes requested):
I would suggest breaking this PR into one for each of your points above. At least 2 and 3 should be separated to keep good PR practices. I also think they deserve some minimal tests and that will make review easy.

PR #118, opened 2022-08-29
Add ROI centroid flag

CodyCBakerPhD:
First modularization of #117

### DIFF
--- a/src/neuroconv/tools/roiextractors/roiextractors.py
+++ b/src/neuroconv/tools/roiextractors/roiextractors.py
@@ -15,6 +15,7 @@ from pynwb.device import Device
 from pynwb.ophys import (
     ImageSegmentation,
     ImagingPlane,
+    PlaneSegmentation,
     Fluorescence,
     OpticalChannel,
     TwoPhotonSeries,
@@ -568,11 +569,13 @@ def add_plane_segmentation(
     nwbfile: NWBFile,
     metadata: Optional[dict],
     plane_segmentation_index: int = 0,
+    include_roi_centroids: bool = True,
     iterator_options: Optional[dict] = None,
     compression_options: Optional[dict] = None,
 ) -> NWBFile:
     """
     Adds the plane segmentation specified by the metadata to the image segmentation.
+
     If the plane segmentation already exists in the image segmentation, it is not added again.
 
     Parameters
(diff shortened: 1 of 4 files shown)


src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -671,9 +674,14 @@ def image_mask_iterator():
             ],
             id=roi_ids,
         )
-
-        image_segmentation.create_plane_segmentation(**plane_segmentation_kwargs)
-
+        plane_segmentation = PlaneSegmentation(**plane_segmentation_kwargs)
+        if include_roi_centroids:
+            tranpose_roi_location_axes = (1, 0) if len(segmentation_extractor.get_image_size()) == 2 else (1, 0, 2)
```
h-mayorquin, line 679: Can you write a brief comment on which case each of this if covers. Or maybe an appropriate name for the boolean variable instead? `dealing_with_non_volumetric_data = len(segmentation_extractor.get_image_size())==2` or something like this?

I just think that it took me a minute or two what this was for and I think I will forget in the future.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -671,9 +674,14 @@ def image_mask_iterator():
...
+            roi_locations = segmentation_extractor.get_roi_locations()[tranpose_roi_location_axes, :].T
```
h-mayorquin, line 680: I was kind of confused about the double transpose (as another transpose is repeated on line 680) at first glance. Maybe some comment about that? 

I get that you require to write the images in the expected format (columns vs heights) [hence transpose in lines 677] and then you transpose all the roi_locations to get num_rois in the first dimension. Not related tot his PR but should not we return the roi locations in shape `(num_rois, shape_of_the_images)`?

h-mayorquin (changes requested):
Looks good to me. Thanks for breaking this apart. Some minimal request for the sake of future readability of the code.

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: Added comment

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: Moved transpose to emphasize connection to writing as column to table

h-mayorquin (commented):
There are some on-going conflicts with this one.

h-mayorquin (approved):

Issue #96, opened 2022-08-18
Use typing.Literals


CodyCBakerPhD:
### What would you like to see added to NeuroConv?

There's a bunch of places throughout NeuroConv (and other packages in the ecosystem, too...) where [typing.Literal](https://docs.python.org/3.8/library/typing.html#typing.Literal) would make more sense than what the current annotation types are.

However, this annotation type is not supported in Python 3.7 (only 3.8+), so this will have to wait until we drop support for Python 3.7 (likely when security updates stop going out for it in summer 2023).

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

bendichter:
Good idea!

CodyCBakerPhD:
I've at least gotten in the habit of putting this in currently in a commented state. I.e.,

```python
def SomeFunc(
    some_arg: str,  # Literal["option_1", "option_2"]
...
```

I can go through and add these everywhere needed for now in order to call this issue 'done' to remove it from backlog, then in the future replace actual annotation with the comment.

PR #119, opened 2022-08-29
Add mask type control

CodyCBakerPhD:
Continuation of #117 - adds `mask_type` specification to `add_plane_segmentation`.

This one in particular has some difficulties to discuss as shown below


src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -672,15 +680,56 @@ def image_mask_iterator():
                     description="1 if ROI was rejected or 0 if accepted as a cell during segmentation operation",
                 ),
             ],
-            id=roi_ids,
         )
+        if mask_type != "pixel" and mask_type != "voxel":  # roi_ids will not be written row-wise
+            plane_segmentation_kwargs.update(id=roi_ids)
```
CodyCBakerPhD, line 685: This is only necessary due to the mixed column-wise/row-wise writing below

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -672,15 +680,56 @@ def image_mask_iterator():
...
+
         if include_roi_centroids:
             tranpose_roi_location_axes = (1, 0) if len(segmentation_extractor.get_image_size()) == 2 else (1, 0, 2)
             roi_locations = segmentation_extractor.get_roi_locations()[tranpose_roi_location_axes, :].T
             plane_segmentation.add_column(
                 name="ROICentroids", description="The x, y, (z) centroids of each ROI.", data=roi_locations
             )
+
+        if mask_type is not "None" and mask_type == "image":
+            plane_segmentation.add_column(
```
CodyCBakerPhD, line 697: First difficulty - the JSON schema validator complained if I set the upstream annotation to `Optional[str]` and set the value to `None` (not a 'recognized' JSON type).

Best fix would probably be to allow the schema parser to understand the Python `None` type.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -672,15 +680,56 @@ def image_mask_iterator():
...
+
+            mask_type_kwarg = f"{mask_type}_mask"
+            for roi_id, pixel_mask in zip(roi_ids, pixel_masks):
+                plane_segmentation.add_roi(**{"id": roi_id, mask_type_kwarg: [tuple(x) for x in pixel_mask]})
+            # plane_segmentation.add_column(
+            #     name=mask_type_kwarg,
+            #     description=f"{mask_type}.capitalize() masks for each ROI.",
+            #     data=[[tuple(x) for x in pixel_mask] for pixel_mask in pixel_masks],
+            #     index=2,
+            # )
```
CodyCBakerPhD, line 731: This part was the biggest difficulty - I would prefer to write it column-wise to allow compression, but the indexed (indeed, doubly jagged) nature of it did not play well with the standard ways of adding it as a column.

This may be an unintended consequence of the NWB schema definition of these types, which is not in terms of a 'dataset'  with 'shapes' but as individual x/y/z properties? https://nwb-schema.readthedocs.io/en/latest/format.html#sec-planesegmentation-src

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -672,15 +680,56 @@ def image_mask_iterator():
...
+            if mask_type == "voxel" and num_pixel_dims == 3:
+                warn(
+                    "Specified mask_type='voxel', but ROIExtractors returned 3-dimensional masks. "
+                    "Using mask_type='pixel' instead."
+                )
+                mask_type = "pixel"
+
+            mask_type_kwarg = f"{mask_type}_mask"
+            for roi_id, pixel_mask in zip(roi_ids, pixel_masks):
+                plane_segmentation.add_roi(**{"id": roi_id, mask_type_kwarg: [tuple(x) for x in pixel_mask]})
```
CodyCBakerPhD, line 725: See other comment for how the column-wise was not working, but this is the only way I was able to successfully do this.

Note that the returned structure from ROIExtractors (the `pixel_masks` object) is a `np.ndarray` of shape `(num_rois, num_dimensions)` - attempting to add this without the list-compression tuple wrapping results in shape errors. The list-compression tuple wrapping was taken from the working example of the PyNWB documentation: https://pynwb.readthedocs.io/en/stable/tutorials/domain/ophys.html#pixel-masks

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -672,15 +680,56 @@ def image_mask_iterator():
...
             roi_locations = segmentation_extractor.get_roi_locations()[tranpose_roi_location_axes, :].T
             plane_segmentation.add_column(
                 name="ROICentroids", description="The x, y, (z) centroids of each ROI.", data=roi_locations
             )
+
+        if mask_type is not "None" and mask_type == "image":
+            plane_segmentation.add_column(
+                name="image_mask",
+                description="Image masks for each ROI.",
+                data=H5DataIO(DataChunkIterator(image_mask_iterator(), **iterator_options), **compression_options),
```
CodyCBakerPhD, line 700: I noticed here we may wish to do away with the iterator - the data as returned by ROIExtractors is an in-memory array, so there's no need for any iteration (though a `SliceableIterator` would still allow better default chunk shaping).

src/neuroconv/tools/roiextractors/roiextractors.py
h-mayorquin: Yes, I see, alt ought their type is still `VectorData` so maybe there is a way.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -650,16 +658,57 @@ def image_mask_iterator():
...
+
         if include_roi_centroids:
             # ROIExtractors uses height x width x (depth), but NWB uses width x height x depth
             tranpose_image_convention = (1, 0) if len(segmentation_extractor.get_image_size()) == 2 else (1, 0, 2)
             roi_locations = segmentation_extractor.get_roi_locations()[tranpose_image_convention, :]
             plane_segmentation.add_column(
                 name="ROICentroids", description="The x, y, (z) centroids of each ROI.", data=roi_locations.T
             )
+
+        if mask_type is not "None" and mask_type == "image":
```
h-mayorquin, line 675: No need for the testing against None if `mask_type== "image"` , right?

src/neuroconv/tools/roiextractors/roiextractors.py
h-mayorquin: You can add this together in the other if in line 675 I think.

h-mayorquin (commented):
So, do you have a segmentation_extractor that returns volumetric pixel_masks?  Can we add some tests for this? Also, the CHANGELOG.

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: Mmmmm so hacky, I really don't like that lol. But I suppose it would work

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: This is from previously when it was a proper `None` type. Can't compare `None == "some_string"

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: Well, they have to be VectorData above because you're passing it directly as the `columns` of the parent `DynamicTable` of the `PlaneSegmentation`, which is separate from calling `add_column` on it.

Critically, this needs `index=2`, but VectorData does not have that: https://hdmf.readthedocs.io/en/stable/hdmf.common.table.html#hdmf.common.table.VectorData (since the meaning of `index` is actually to pair a flat VectorData with a VectorIndex)

h-mayorquin:
> > So, do you have a segmentation_extractor that returns volumetric pixel_masks? Can we add some tests for this? Also, the CHANGELOG.
> 
> This PR isn't quite ready to go, there's a lot of artifacts not working that I wanted your take on like the column-wise approach

Yep, that's what I imagined -it is a draft after all- Sorry for not being able to help you there about that specific issue other than to share your perplexity.

src/neuroconv/tools/roiextractors/roiextractors.py
h-mayorquin: Yep, not sure what's better, we are choosing between two ugly options here I think. I agree that the best global fix is what you describe. Fixing the parser.

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: So that bit is important actually - if you know that the type of mask being written is one of the two types that currently has to be added by-row (since the column approach of #157 has yet to work), then you had to perform that row writing after the plane segmentation object is in memory but before adding other column data.

But also, when adding the data by row you have to set the ID there instead of as a separate column.

But if you're not writing any data row-wise, then you can pass the column of IDs directly into the initialization of the table as we have always done.

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: Trying to fix now then

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: This'll be a lesser priority to fix - outside of this PR

CodyCBakerPhD:
I'm simplifying this PR just to get the basic functionality in. Splintering column-wise stuff into #157.

Last thing this should need is the debug to the validation to allow `None` types.

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: Just gonna splinter this into #157

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: Fixed. `Optional[...]` wrappers now work in any signature. As long as it's also not containing a `Union` of other types, which could indicate conflicts.

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: Actually, I guess I did have to fix it here since the iterator wasn't operating with this approach (the original approach of directly column attachment bypasses a lot of the PyNWB API, for better or worse).

Wound up just removing it entirely because the underlying `_image_mask` attribute is already in memory for current formats so it's not even doing anything here. That said, this is an area of ROIExtractors that could really be improved

CodyCBakerPhD:
> So, do you have a segmentation_extractor that returns volumetric pixel_masks?

Yep. In NWB they're actually called `voxel_mask`'s

> Can we add some tests for this?

Done

> Also, the CHANGELOG.

Done

h-mayorquin (changes requested):
Looks good to me. Two requests in order of importance. 

1) Adding the other modifications done in this PR to the changelog. mask_type=None and the schema modifications.
2) Adding a test for the case when mask_type = None

h-mayorquin:
@CodyCBakerPhD 
I am wondering something now though. Right now `SegmentationExtractor` returns either a voxel_mask or a pixel_mask through the method `get_roi_pixel_masks`. Given that, would not make more sense to just chose between these two and write either a `voxel_mask` or a `pixel_mask` depending on what the extractor actually returns. 

Is it possible that a segmentation extractor should return both pixel and voxel? It does not seem to make sense  I don't see why create an option to control something (pixel vs voxel) that they in fact do not control as the segmentation extractor controls this. The choice should be between "store_dense_mask" or "store_sparse_mask" (in space)  representation and the code should take care of choosing between pixel and voxel automatically. Maybe there is something I am missing out?

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -549,6 +549,7 @@ def add_plane_segmentation(
     metadata: Optional[dict],
     plane_segmentation_index: int = 0,
     include_roi_centroids: bool = True,
+    mask_type: Optional[str] = "image",  # Optional[Literal["image", "pixel"]]
```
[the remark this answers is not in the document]
CodyCBakerPhD: Comment for #96, does nothing yet

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: Might want to raise an issue in PyNWB for that lol

src/neuroconv/tools/roiextractors/roiextractors.py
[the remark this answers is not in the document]
h-mayorquin: Thanks for explaining this.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -958,7 +1007,7 @@ def write_segmentation(
         nwbfile_path=nwbfile_path, nwbfile=nwbfile, metadata=metadata_base_common, overwrite=overwrite, verbose=verbose
     ) as nwbfile_out:
 
-        ophys = get_module(nwbfile=nwbfile_out, name="ophys", description="contains optical physiology processed data")
+        _ = get_module(nwbfile=nwbfile_out, name="ophys", description="contains optical physiology processed data")
```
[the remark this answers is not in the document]
CodyCBakerPhD: Just linter cleanup, they're not doing anything. Technically the call to `get_module` might be initializing the processing module, but the `ophys` return is unused and the stuff below is unused

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -624,39 +636,64 @@ def image_mask_iterator():
...
-            ],
-            id=roi_ids,
-        )
+        plane_segmentation_kwargs = dict(**plane_segmentation_metadata, imaging_plane=imaging_plane)
+        if not (mask_type == "pixel" or mask_type == "voxel"):  # If roi_ids will not be written row-wise
+            plane_segmentation_kwargs.update(id=roi_ids)
+
         plane_segmentation = PlaneSegmentation(**plane_segmentation_kwargs)
+
+        if mask_type is not None and mask_type == "image":
```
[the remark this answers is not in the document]
CodyCBakerPhD: Logic simplified to avoid additional cyclomatic complexity and mimic the switch statement that we'll eventually be able to use for this type of logic

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -728,18 +765,19 @@ def add_fluorescence_traces(
...
-    timestamps = segmentation_extractor.frame_to_time(np.arange(segmentation_extractor.get_num_frames()))
-    rate = calculate_regular_series_rate(series=timestamps)
-    if rate:
-        roi_response_series_kwargs.update(starting_time=timestamps[0], rate=rate)
+    if getattr(segmentation_extractor, "_times") is None:
+        rate = segmentation_extractor.get_sampling_frequency()
+        roi_response_series_kwargs.update(starting_time=0.0, rate=rate)
     else:
-        roi_response_series_kwargs.update(timestamps=H5DataIO(data=timestamps, **compression_options))
+        timestamps = segmentation_extractor.frame_to_time(np.arange(segmentation_extractor.get_num_frames()))
```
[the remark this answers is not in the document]
CodyCBakerPhD: > I think these modifications should be in another PR.

Reverted


> I still think that frame_to_time should return all the times when called without arguments. Can you remind me why you were against it?

Really, it should be `get_times()` and mimic w/e the default behavior of SI is. The rest of the code here was a minor speedup by not forcing it to generate artificial timestamps for a series that's already known to be regular.

src/neuroconv/utils/json_schema.py
```diff
@@ -78,11 +78,17 @@ def get_schema_from_method_signature(class_method: classmethod, exclude: list =
...
+                    num_params = len(set(param_types))
+                    conflict_message = (
+                        "Conflicting json parameter types were detected from the annotation! "
+                        f"{param.annotation.__args__} found."
+                    )
+                    # Normally cannot support Union[...] of multiple annotation types
+                    if num_params > 2:
+                        raise ValueError(conflict_message)
+                    # Special condition for Optional[...]
+                    if num_params == 2 and not args[1] is type(None):  # noqa: E721
```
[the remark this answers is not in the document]
CodyCBakerPhD: Done

CHANGELOG.md
```diff
@@ -42,7 +42,7 @@
 * Propagate `output_struct_name` argument to `ExtractSegmentationInterface` to match its extractor arguments. [PR #128](https://github.com/catalystneuro/neuroconv/pull/128)
 * Added compression and iteration (with options control) to all Fluorescence traces in `write_segmentation`. [PR #120](https://github.com/catalystneuro/neuroconv/pull/120)
 * For irregular recordings, timestamps can now be saved along with all traces in `write_segmentation`. [PR #130](https://github.com/catalystneuro/neuroconv/pull/130)
-
+* Added `mask_type` argument to `tools.roiextractors.add_plane_segmentation` function and all upstream calls. This allows users to request writing not just the image_masks (still the default) but also pixels or voxels. [PR #119](https://github.com/catalystneuro/neuroconv/pull/119)
```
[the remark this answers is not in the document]
CodyCBakerPhD: Done

CodyCBakerPhD:
> Given that, would not make more sense to just chose between these two and write either a voxel_mask or a pixel_mask depending on the shape of what the extractor returns?

Hence the warnings that decide that based on the shape and issue a simple warning if the person didn't specify the right one.


> Also, it could be that plane segmentation is different and a second pass does not make sense here. Plane segmentation refers to an imaging plane and I don't see why further rois segmented by another extractor could not be added to the same imaging plane. We can always add new plane segmentation and then link them to the same plane which could make things simpler for us but if that the case, maybe we could do the same for the units table and simplify the code there. What do you think?

Sounds like a use case to explore in the future

> Critically, the limitation over there is that adding new units or electrodes requires writing them by rows. I realize that there is probably no way of adding compression when adding information row wise, is that correct?

Hmm that is interesting... You cannot use a `H5DataIO` for compression writing row-wise, but I'm unsure how it would work if you wrote the column that way first, then in append mode added extra rows. One would hope the dataset would still be compressed, but would be worth checking out what problems it causes.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -582,14 +594,14 @@ def add_plane_segmentation(
     NWBFile
         The nwbfile passed as an input with the plane segmentation added.
     """
+    assert mask_type in ["image", "pixel", "voxel", "None"], (
```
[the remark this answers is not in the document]
CodyCBakerPhD: Yep, fixed

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -624,39 +636,64 @@ def image_mask_iterator():
...
-                VectorData(
-                    data=rejected_ids,
-                    name="Rejected",
-                    description="1 if ROI was rejected or 0 if accepted as a cell during segmentation operation",
-                ),
-            ],
-            id=roi_ids,
-        )
+        plane_segmentation_kwargs = dict(**plane_segmentation_metadata, imaging_plane=imaging_plane)
+        if not (mask_type == "pixel" or mask_type == "voxel"):  # If roi_ids will not be written row-wise
```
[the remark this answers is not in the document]
CodyCBakerPhD: Agreed, fixed

CodyCBakerPhD:
> Adding the other modifications done in this PR to the changelog. mask_type=None and the schema modifications.

Done

> Adding a test for the case when mask_type = None

Done

> Adding a test for the case of mask_type =="pixel"

Done

h-mayorquin:
> > Given that, would not make more sense to just chose between these two and write either a voxel_mask or a pixel_mask depending on the shape of what the extractor returns?
> 
> Hence the warnings that decide that based on the shape and issue a simple warning if the person didn't specify the right one.
> 

Yeah, but why give them an non-working option in the first place though? You are kind of semi patching a mistake of our making with the warning. How are you thinking about it? What is the purpose of having that option?

CodyCBakerPhD:
> Yeah, but why give them an non-working option in the first place though? You are kind of semi patching a mistake of our making with the warning. How are you thinking about it? What is the purpose of having that option?

Because those are how they are written and accessed in the NWB file itself. Going with your suggested option wouldn't be clear to them how the final mask field would show up as a field of the neurodata type. Also not entirely true to call one dense and other sparse since the storage on disk in HDF5 will be sparse either way, it's just the structure returned in memory (which you can convert one to the other in-memory with a single line anyway).

I'd call it an 'educational' warning to help teach people in that situation the difference between the two masks. It's subtle enough that we *can* easily correct it in the event they knew they wanted a non-image mask but weren't sure about the nomenclature for their data. TBH though, this will probably be a very rare use case since volumetric data is not the norm. Also most people might not even think about this option at all unless it affects performance, which is only in the cases like I'm seeing for whole-brain imaging.

Added tests for the warnings and this functionality.

h-mayorquin (approved):

PR #120, opened 2022-08-29
Add compression to traces in write_segmentation

CodyCBakerPhD:
Continuation of #117 to allow compression of segmentation extractor traces


src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -797,7 +804,7 @@ def add_fluorescence_traces(
 
         # Build the roi response series
         roi_response_series_kwargs.update(
-            data=np.array(trace).T,
+            data=H5DataIO(SliceableDataChunkIterator(trace, **iterator_options), **compression_options),
```
CodyCBakerPhD: This is temporary - an actual `SegmentationExtractorDataChunkIterator(GenericDataChunkIterator)` would be preferred, but best in follow-up PR.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -691,6 +694,10 @@ def add_fluorescence_traces(
     NWBFile
         The nwbfile passed as an input with the fluorescence traces added.
     """
+    if iterator_options is None:
```
h-mayorquin, line 697: What about using the or notation in these cases here that I have seen you using in other places?
`iterator_options = iterator_options or dict()`

src/neuroconv/datainterfaces/ophys/basesegmentationextractorinterface.py
```diff
@@ -67,6 +67,8 @@ def run_conversion(
         stub_test: bool = False,
         stub_frames: int = 100,
         include_roi_centroids: bool = True,
+        iterator_options: Optional[dict] = None,
```
h-mayorquin, line 70: What do you think on making this more specific by naming this `hdmf_iterator_options` and `hdmf_compression_options`?

h-mayorquin (changes requested):
I think we should keep the option for non-iterative write. I think this has the advantage to decouple the testing in the following way: we can test everything related to the writing in the right place (naming and of series and location of processing module), data shape and pynwb parameters with the non-iterative (non wrapped in `H5DataIO`) and then just test propagation of options to the hdmf wrappers as we have done in `test_tools_spikeinterface `(so we also have a consistent testing style). This would keep the testing of everything above separate from wrapping and simplify the tests a little bit as you avoid the data extraction nesting that the hdmf wrapper imply (i.e. `processing_modue["SeriesName'].data.data.data`). What do you think? Of course, if you agree with this view,  I think should be done in another PR (there are some things that need to be changed in the two photon series testing as well).

Some minor comments, but add the changes to the CHANGELOG and this should be good to go.

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: Sure

src/neuroconv/datainterfaces/ophys/basesegmentationextractorinterface.py
CodyCBakerPhD: Too much typing, and no one needs to know that it has anything to do with HDMF underneath

CodyCBakerPhD: Also - just imagining - what if we one day have compression/iterators that aren't reliant on HDMF? I'd rather not have to add or modify the keyword argument name to correct that or worse, add a separate argument. Better to just have all that be a part of the dictionary

CodyCBakerPhD:
> I think we should keep the option for non-iterative write.

Sure, that's fine. More control is always better and there are cases where writing contiguously can be preferred for performance.

I'd say let's do it in a follow-up where we standardize all the `iterator_options` and remove `iterator_type` across the other tools, what do you think? Can add more robust testing then as well.

> I think this has the advantage to decouple the testing in the following way: we can test everything related to the writing in the right place (naming and of series and location of processing module), data shape and pynwb parameters with the non-iterative (non wrapped in H5DataIO) and then just test propagation of options to the hdmf wrappers as we have done in test_tools_spikeinterface (so we also have a consistent testing style). This would keep the testing of everything above separate from wrapping and simplify the tests a little bit as you avoid the data extraction nesting that the hdmf wrapper imply (i.e. processing_modue["SeriesName'].data.data.data).

I don't mind testing all of it - while it's not our responsibility per se to test that the `H5DataIO` and such actually work, it's not difficult to test that they apply and produce expected results. More tests are always better

h-mayorquin:
> > I think we should keep the option for non-iterative write.
> 
> Sure, that's fine. More control is always better and there are cases where writing contiguously can be preferred for performance.

> I'd say let's do it in a follow-up where we standardize all the `iterator_options` and remove `iterator_type` across the other tools, what do you think? Can add more robust testing then as well.

Yep, I think this makes sense.
 
> > I think this has the advantage to decouple the testing in the following way: we can test everything related to the writing in the right place (naming and of series and location of processing module), data shape and pynwb parameters with the non-iterative (non wrapped in H5DataIO) and then just test propagation of options to the hdmf wrappers as we have done in test_tools_spikeinterface (so we also have a consistent testing style). This would keep the testing of everything above separate from wrapping and simplify the tests a little bit as you avoid the data extraction nesting that the hdmf wrapper imply (i.e. processing_modue["SeriesName'].data.data.data).
> 
> I don't mind testing all of it - while it's not our responsibility per se to test that the `H5DataIO` and such actually work, it's not difficult to test that they apply and produce expected results. More tests are always better

I think that testing  more is fine but to me the important thing is to separate this two scenarios.

src/neuroconv/datainterfaces/ophys/basesegmentationextractorinterface.py
h-mayorquin: Yes, I am on the fence on that one. I agree that that's probably leaking too much information to users which can be better described in the docstring. The concern comes from the fact that the word iterator has a wider meaning in Python that does not match with the specific and narrow meaning* of the word in HDMF and I wanted to indicate or disambiguate that. It does not come from a desire to let the users know that the we use THAT iterator.

What I mean is that the iterator here is a specific type of iterator that is focused mainly in solving the problem of writing data efficiently without loading everything into memory.  Maybe `iterative_write_options` would be a better description? That's what they use in their tutorial to describe the problem / context that the tool is solving:
https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/iterative_write.html#sphx-glr-tutorials-advanced-io-iterative-write-py

h-mayorquin (approved):

PR #115, opened 2022-08-26
[DO NOT MERGE][Testing]: Format-wise gallery

CodyCBakerPhD:
Demonstration of how we could have a matrix deployment over each formatwise key to ensure the data interface corresponding each one works as expected.

@h-mayorquin The doctests would be a great way to harness that, can you provide a command to run doctest on only a single file in the gallery here?

PR #121, opened 2022-08-29
[Possible to merge][Testing]: Format-wise installation

CodyCBakerPhD:
Replaces to #115, could be merged without affecting primary workflows (would not show up on PRs, but can be triggered manually and will run on a schedule)

### DIFF
--- a/.github/workflows/formatwise_installation_testing.yml
+++ b/.github/workflows/formatwise_installation_testing.yml
@@ -1,22 +1,19 @@
-name: Minimal and Full Tests
+name: Formwatwise gallery tests
 on:
   schedule:
     - cron: "0 16 * * 1"  # Every Monday at noon EST
   workflow_dispatch:
 
-env:
-  DANDI_API_KEY: ${{ secrets.DANDI_API_KEY }}
-
 jobs:
   run:
-    name: Minimal and full tests on ${{ matrix.os }} with Python ${{ matrix.python-version }}
+    name: Formatwise gallery tests for ${{ format.type }}:${{ format.name }} on ${{ matrix.os }} with Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


h-mayorquin (approved):
I have one comment and one question:
### Comment
As far as I can see this still tests the sequential path (installing minimal dependencies which is done in line 35, and the install the module specific dependencies on line 72). As stated in a previous [discussion](https://github.com/catalystneuro/neuroconv/pull/100#discussion_r950413371) I have the feeling that the most common use of this installation method would not be sequential but stand-alone (which could be tested by removing line 35 and changing line 72 to `pip install .[${{ matrix.format.name }}, test]` instead). 

I think that is hard to know what will be the most common use case and you might be right that it will be sequential instead of stand-alone.  Given that you put all of this work (thanks!) I think it is fair that we go with your intuition in a situation of uncertainty. My only comment is that we should remember,when we write the user facing documentation of this installation modality, to alert users that we support and test the sequential installation and not the stand-alone one. 

### Question
I have this memory that we were getting cache non-hits when for different PRs (caches being PR exclusively or something like this) That is, one PR would not get the caches of another. How does this work here for the cron job? Which previous caches will this be accessing?

CodyCBakerPhD:
> # Comment

It doesn't have to be intuition - boot up a fresh environment and try it out for yourself. Minimal requirements are always attached in all cases.

> # Question

Haven't seen that issue for a while now - also only happens if we keep the branches separate. If it goes on main, it sees what main sees. Main uses the caches established from the primary S3 workflow now. I'm not forcing download in non-hit event here since that would destroy our data transfer bill, but if something happens we can troubleshoot when it does (would just need to sync the key hit name with the other workflow). So shouldn't have to worry

PR #122, opened 2022-08-29
[Bug fix]: Context output during raised error

CodyCBakerPhD:
Standalone fix peeled from #117 - an issue was discovered when an error was raised *during* the context (not in the enter/exit) for making or loading NWBFiles where the success statement would print prior to the traceback.

This fixes the issue and added an explicit test for it.

### DIFF
--- a/src/neuroconv/tools/nwb_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers.py
@@ -153,6 +153,7 @@ def make_or_load_nwbfile(
     )
 
     load_kwargs = dict()
+    success = True
     if nwbfile_path:
         load_kwargs.update(path=nwbfile_path)
         if nwbfile_path_in.is_file() and not overwrite:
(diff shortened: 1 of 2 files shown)


bendichter (approved):

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -166,12 +167,15 @@ def make_or_load_nwbfile(
...
+    except Exception as e:
+        success = False
+        raise e
     finally:
         if nwbfile_path:
             try:
                 io.write(nwbfile)
 
-                if verbose:
+                if success and verbose:
```
h-mayorquin, line 178: Should the file still be written if (line 176) if an exception was risen?

CodyCBakerPhD: Probably not 😆

h-mayorquin: Line 172 now in this preview somehow.

h-mayorquin (approved):

PR #123, opened 2022-08-30
Remove upcoming deprecations

CodyCBakerPhD:
We were a bit delayed by the SpikeInterface release, but since we're changing the import structure anyway we might as well clean up other stuff ahead of the upcoming release.

The `icephys` module still has the `save_path` deprecation warning it since we caught that one much later than the others - should we remove it here as well?


h-mayorquin:
@CodyCBakerPhD I am curios how much the coverage will increase with this one.

CodyCBakerPhD:
> @CodyCBakerPhD I am curios how much the coverage will increase with this one.

Lol. We probably should have used `# pragma` on that originally.

CodyCBakerPhD:
But it's meaningless coverage either way so it's just artificial inflation/deflation

h-mayorquin:
> But it's meaningless coverage either way so it's just artificial inflation/deflation

Don't be a party popper, @CodyCBakerPhD  : P

CodyCBakerPhD:
This branch got tangled with the ROI centroid stuff somehow. Re-doing

PR #124, opened 2022-08-30
Remove tools alias

CodyCBakerPhD:
testing splinter of #78

### DIFF
--- a/src/neuroconv/tools/__init__.py
+++ b/src/neuroconv/tools/__init__.py
@@ -1,7 +1,1 @@
-# To be treated as sub-modules or plugins
-from .spikeinterface import spikeinterface
-from .roiextractors import roiextractors
-from .neo import neo
-
-# Basically utils but rely on external dependencies
 from .importing import get_package
--- a/tests/test_minimal/test_import_structure.py
+++ b/tests/test_minimal/test_import_structure.py
@@ -41,17 +41,13 @@ class TestImportStructure(TestCase):
         current_structure = _strip_magic_module_attributes(dictionary=tools.__dict__)
         expected_structure = [
             # Sub-modules
-            "neo",
-            "spikeinterface",
-            "roiextractors",
-            # Helper functions
-            "get_package",
             "importing",
-            # Other stuff that I don't know why it is here (for PR)
             "nwb_helpers",
             "yaml_conversion_specification",
             "hdmf",
             "data_transfers",
+            # Helper functions
+            "get_package",
         ]
         self.assertCountEqual(first=current_structure, second=expected_structure)
 
--- a/tests/test_minimal/test_import_structure.py
+++ b/tests/test_minimal/test_import_structure.py
@@ -5,7 +5,7 @@ from neuroconv import datainterfaces
 from neuroconv import tools
 
 
-def _strip_magic_module_attributes(dictionary: dict) -> dict:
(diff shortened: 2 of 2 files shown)


CodyCBakerPhD:
Well that is interesting. The import structure passes minimal but something about full alters it? Will have to investigate further...

CodyCBakerPhD:
@h-mayorquin OK I believe I figured it out. A couple things here

a) swapped the tests to use direct `dir()` calls, return structure is better to work with and behavior is identical (mismatches would only result if you were doing fancy injections like we were trying before)

b) submodules unlisted in the `__init__` at a certain level are not included in the default `dir()`, but this updates dynamically over time if you ever import those submodules later in the process - important here, because for minimal tests we only import a couple of things but for full tests we import everything. So the `dir()` structure of tools changes depending on how you call pytest. I've adjusted the test to only assert the minimal structure we expect on stand-alone load from a fresh kernel

h-mayorquin:
I am wondering, would not doing the imports inside the test functions for this case would be more appropriate?

I am confused about this pytest behavior. I somehow thought that at least each testing module would be run independently. Are you saying that the imports done in one test (say, the file `test_tools_spikeextractors`) are shared with the imports in other test (say, the file `test_tools_spikeinterface)?

I somehow thought that this could be related to the difference between `loadscope` and `loadfile` here:
https://github.com/catalystneuro/neuroconv/blob/ea8e5c50d6eaf15f23b47832c01c1bdc7c4a5672/.github/workflows/testing.yml#L122

See:
https://pytest-xdist.readthedocs.io/en/latest/distribution.html?highlight=loadscope#running-tests-across-multiple-cpus

CodyCBakerPhD:
> I am wondering, would not doing the imports inside the test functions for this case would be more appropriate?

Makes no difference, upon first call of the import anywhere it goes to the sys.modules

> I am confused about this pytest behavior. I somehow thought that at least each testing module would be run independently. Are you saying that the imports done in one test (say, the file test_tools_spikeextractors) are shared with the imports in other test (say, the file `test_tools_spikeinterface)?

It appears so - I assume all global imports in all tests are performed at the beginning. Actually, I always knew that in a way, because if there's an import error in any part of the test suite, it gets raised first thing when you call `pytest`

> I somehow thought that this could be related to the difference between loadscope and loadfile here:

`--dist` is only for specifying the grouped deployment of linked `test_` named functions, important for us because otherwise those grouped in `TestCase` classes get called without performing proper setups.

Point is still, the actual running of any/all `test_` named functions in the suite occurs after successful global imports from test files.

h-mayorquin:
I see. Why not just skip this test on the full testing and run it independently by itself? What we really want to test is the import system in a fresh start. Testing the state of `sys.modules` in the middle of a complicated run seems like far away from what we care and probably pointless. Am I am missing something?

CodyCBakerPhD:
@h-mayorquin You're welcome to figure out a way to do that

h-mayorquin:
> @h-mayorquin You're welcome to figure out a way to do that

Yes, if you agree that that makes sense then I will take a stab at that.

CodyCBakerPhD:
> Yes, if you agree that that makes sense then I will take a stab at that.

It does, please do

CodyCBakerPhD:
Though I wouldn't spend more than an hour or so trying to figure it out - otherwise just testing minimal inclusion structure as done in the current state of the PR should be sufficient.

PR #125, opened 2022-08-30
Remove upcoming deprecations

CodyCBakerPhD:
We were a bit delayed by the SpikeInterface release, but since we're changing the import structure anyway we might as well clean up other stuff ahead of the upcoming release.

The `icephys` module still has the `save_path` deprecation warning it since we caught that one much later than the others - should we remove it here as well?


h-mayorquin (dismissed):
Yei, deprecations!

PR #128, opened 2022-09-01
Propagate `output_struct_name` argument to `ExtractSegmentationInterface`

weiglszonja:
## Motivation

The segmentation output from the Fee Lab has a custom name ("exOut") for `output_struct_name`, which is not possible to be modified from the interface level. Here I propose to propagate  `output_struct_name` to the interface level, so that it can be changed during conversion. 

Alternatively, I can do this in a custom interface, where the `__init__` would also be modified with adding timestamps for the segmentation data. 

```python
class ScherrerOphysSegmentationInterface(ExtractSegmentationInterface):
    """Data interface for ExtractSegmentationExtractor."""
    Extractor = ExtractSegmentationExtractor

    def __init__(
            self,
            file_path: FilePathType,
            timestamps_file_path: FilePathType,
            output_struct_name: str = "exOut",
            verbose: bool = True,
    ):

        super().__init__(
            file_path=file_path,
            sampling_frequency=30.0,
            output_struct_name=output_struct_name,
        )

        timestamps = get_timestamps_from_csv(file_path=timestamps_file_path)
        if not calculate_regular_series_rate(timestamps):
            # only use timestamps if they are not regular
            self.segmentation_extractor.set_times(times=timestamps)
        self.verbose = verbose

```
## How to test the behavior?
```
```

## Checklist


CodyCBakerPhD:
Yup, that's good. Always need to keep interfaces up to date with their corresponding extractor arguments.

weiglszonja:
@CodyCBakerPhD should I modify any of the current tests in `test_gin_ophys.py`?

CodyCBakerPhD:
> @CodyCBakerPhD should I modify any of the current tests in test_gin_ophys.py?

I was wondering that but since the example data has the default name, specifying the keyword there wouldn't effectively test anything.

I suppose what you could do to ensure it's being passed is add a test where the passed group name is incorrect for the example files and use a `self.assertRaisesWith` to grab whatever error gets raised from ROIExtractors, just to ensure that the argument is getting passed

tests/test_on_data/test_gin_ophys.py
```diff
@@ -148,6 +149,17 @@ class TestConverter(NWBConverter):
         nwb_segmentation = NwbSegmentationExtractor(file_path=nwbfile_path)
         check_segmentations_equal(segmentation, nwb_segmentation)
 
+    def test_extract_segmentation_interface_non_default_output_struct_name(self):
+        """Test that the value for 'output_struct_name' is propagated to the extractor level
+        where an error is raised."""
+        file_path = OPHYS_DATA_PATH / "segmentation_datasets" / "extract" / "extract_public_output.mat"
+        with self.assertRaisesWith(KeyError, '''"Unable to open object (object 'filters' doesn't exist)"'''):
```
weiglszonja, line 156: I feel like this is not the correct error that should be thrown by the extractor:
https://github.com/catalystneuro/roiextractors/blob/263b57598b91fc146632cdcb463485d8b33b0970/src/roiextractors/extractors/schnitzerextractor/extractsegmentationextractor.py#L82-L83
The current behavior is to redirect to the Legacy extractor when the struct name is not found in the file. what do you think @CodyCBakerPhD ?

CodyCBakerPhD: Ahhhh very good point.

What I would recommend is making https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/extractors/schnitzerextractor/extractsegmentationextractor.py#L37

instead

```python
output_struct_name: Optional[str] = None,
```

and then split the version checking into two parts with different corresponding errors.

a) `if output_struct_name is not None` (which means it was specified by the user), then check if it is in the file. If not in the file, raise a specific error that the specified name is not in the file.

b) if it is `None`, then check if the name "output" is in the file, if it is try to check for the version info. If the name "output" is not in the file, inform user to specify the name of the struct in the file that has the info.

c) if the version field itself is not in the `self.output_struct_name` (with that name specified or not, but present either way) then we can assume at that point it is LEGACY

weiglszonja: Connecting PR https://github.com/catalystneuro/roiextractors/pull/210

CodyCBakerPhD:
@weiglszonja Is this good to go?

weiglszonja:
@CodyCBakerPhD it seems like there are still some tests that are failing, looking into it.

CodyCBakerPhD:
Will approve this once https://github.com/catalystneuro/roiextractors/pull/210 is merged

weiglszonja:
Great, thank you. Should I go ahead an merge https://github.com/catalystneuro/roiextractors/pull/210?

CodyCBakerPhD:
> Great, thank you. Should I go ahead and merge https://github.com/catalystneuro/roiextractors/pull/210?

Yep, any time you're done with a PR you're assigned to and it has approval, feel free to do the merge and branch deletion on your own 😃

CodyCBakerPhD (approved):

PR #130, opened 2022-09-05
Add timestamps or rate to fluorescence traces

weiglszonja:
## Motivation

Similarly to writing imaging data, `timestamps` can be added to the `RoiResponseSeries` when timestamps are irregular. For regular timings, `rate` and `starting_time` is used instead.

This will be used in https://github.com/catalystneuro/fee-lab-to-nwb/pull/28

## How to test the behavior?
Unittests are added to check `RoiResponseSeries` is created with timestamps when timestamps are irregular, but `rate` and `starting_time` is used when the series is regular.

## Checklist


CodyCBakerPhD (approved):

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -730,21 +730,19 @@ def add_fluorescence_traces(
...
         unit="n.a.",
     )
 
+    # Add timestamps or rate
+    timestamps = segmentation_extractor.frame_to_time(np.arange(segmentation_extractor.get_num_frames()))
+    rate = calculate_regular_series_rate(series=timestamps)
+    if rate:
+        roi_response_series_kwargs.update(starting_time=timestamps[0], rate=rate)
+    else:
+        roi_response_series_kwargs.update(timestamps=H5DataIO(data=timestamps, **compression_options))
```
weiglszonja, line 744: @CodyCBakerPhD looked into the tests once more and doing

        self.assertAlmostEqual(
            df_over_f[trace_name].rate,
            segmentation_extractor.get_sampling_frequency(),
            places=6
        )

doesn't fix the tests, because  `df_over_f[trace_name].rate` is `None`.

CodyCBakerPhD: Are the timestamps for the dummy extractor in that test irregular? I didn't think they were, but if the `rate` isn't being written to the object in the tests that's the only reason I can think as to why... does the object in the failing tests have `timestamps` instead?

weiglszonja: yeah they have `timestamps` instead in this case, but still I thought it is a rounding issue because if we use `tolerance_decimals = 5` when trying to infer rate from timestamps in `calculate_regular_series_rate`, then it's regular.

CodyCBakerPhD: OK thanks that helps pin down the issue.

I think the actual problem is with the `SegmentationExtractor.frame_to_time()` method here: https://github.com/catalystneuro/roiextractors/blob/master/src/roiextractors/segmentationextractor.py#L305

which if you compare to SpikeInterface (which doesn't have this problem) does not round the timestamps: https://github.com/catalystneuro/spikeinterface/blob/master/spikeinterface/core/baserecording.py#L448-L449

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -609,9 +609,10 @@ def test_add_fluorescence_traces(self):
             self.neuropil_roi_response_series_metadata["unit"],
         )
 
-        self.assertEqual(
+        self.assertAlmostEqual(
             fluorescence["Neuropil"].rate,
             self.segmentation_extractor.get_sampling_frequency(),
+            places=3,
         )
```
CodyCBakerPhD, line 616: You can actually try returning this to `assertEqual` and pinning to https://github.com/catalystneuro/roiextractors/pull/212, which should fix it

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -675,7 +676,11 @@ def test_add_df_over_f_trace(self):
 
         self.assertEqual(df_over_f[trace_name].unit, "n.a.")
 
-        self.assertEqual(df_over_f[trace_name].rate, segmentation_extractor.get_sampling_frequency())
+        self.assertAlmostEqual(
+            df_over_f[trace_name].rate,
+            segmentation_extractor.get_sampling_frequency(),
+            places=3,
+        )
```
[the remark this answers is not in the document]
weiglszonja: @CodyCBakerPhD thanks for working on that PR!

Not exactly, the segmentation extractor would return `30.0`, while there is `0.0003` difference:
```
Expected :30.000300003000028
Actual   :30.0
```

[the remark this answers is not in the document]
weiglszonja: If we were to change the tolerance decimals from 6 to 9 here:
https://github.com/catalystneuro/neuroconv/blob/948be7cfb4866e121edfff20479468d08106f1a4/src/neuroconv/utils/checks.py#L7
then we would have `30.000000300000004`, then we could be closer, but not sure this is the right direction.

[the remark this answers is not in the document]
CodyCBakerPhD: Gotcha, I see. That's fine then

weiglszonja:
Tried merging this PR with #151 in #152, CI tests should pass after merging #151 into main.

CodyCBakerPhD (approved):

CodyCBakerPhD (approved):

PR #132, opened 2022-09-06
Bump `spikeinterface` version to `0.95.1` or higher

### LINKED ISSUE #127: [Bug]: Latest ProbeInterface broke SpikeGLXInterface
### What happened?

https://github.com/catalystneuro/neuroconv/runs/8136438434?check_suite_focus=true

For now, pinning to previous version of ProbeInterface as upper bound.

@h-mayorquin You know the most about this, do you think you can fix it?

### Steps to Reproduce

```python
pytest tests/test_on_data/test_gin_ecephys.py
```

### Traceback

_No response_

### Operating System

Linux

### Python Executable

Python

### Python Version

3.9

### Package Versions

Same as full requirements, latest of all. Also included in the log above.

### Code of Conduct

weiglszonja:
## Motivation

Bump `spikeinterface` version to `0.95.1` to fix issue with probe annotations `KeyError` for "imDatPrb_type". This ought to resolve #127, an issue also described at https://github.com/SpikeInterface/spikeinterface/issues/923 which is fixed in `0.95.1.`

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/requirements.txt
+++ b/src/neuroconv/datainterfaces/ecephys/requirements.txt
@@ -1,1 +1,2 @@
 spikeinterface>=0.95.1
+spikeextractors>=0.9.10
--- a/src/neuroconv/datainterfaces/ecephys/requirements.txt
+++ b/src/neuroconv/datainterfaces/ecephys/requirements.txt
@@ -1,4 +1,4 @@
 neo>=0.11
 spikeinterface>=0.95.1
 spikeextractors>=0.9.10
-probeinterface<=0.2.9 # https://github.com/catalystneuro/neuroconv/issues/127
+probeinterface>=0.2.10
--- a/src/neuroconv/datainterfaces/ecephys/requirements.txt
+++ b/src/neuroconv/datainterfaces/ecephys/requirements.txt
@@ -1,4 +1,4 @@
 neo @ git+https://github.com/NeuralEnsemble/python-neo@50abfbce832edb1c0f8364ac34c8a6442e1a3edf
-spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git@4346eff5706408435e8aa8345be22fdd452afa27
+spikeinterface>=0.95.1
 spikeextractors>=0.9.10
 probeinterface<=0.2.9 # https://github.com/catalystneuro/neuroconv/issues/127
--- a/src/neuroconv/datainterfaces/ecephys/requirements.txt
+++ b/src/neuroconv/datainterfaces/ecephys/requirements.txt
@@ -1,4 +1,4 @@
-neo @ git+https://github.com/NeuralEnsemble/python-neo@50abfbce832edb1c0f8364ac34c8a6442e1a3edf
+neo>=0.11
 spikeinterface>=0.95.1
 spikeextractors>=0.9.10
 probeinterface<=0.2.9 # https://github.com/catalystneuro/neuroconv/issues/127
--- a/src/neuroconv/datainterfaces/ecephys/requirements.txt
+++ b/src/neuroconv/datainterfaces/ecephys/requirements.txt
@@ -1,4 +1,2 @@
-neo>=0.11
 spikeinterface>=0.95.1
 spikeextractors>=0.9.10
-probeinterface>=0.2.10
--- a/src/neuroconv/datainterfaces/ecephys/requirements.txt
+++ b/src/neuroconv/datainterfaces/ecephys/requirements.txt
@@ -1,2 +1,1 @@
 spikeinterface>=0.95.1
-spikeextractors>=0.9.10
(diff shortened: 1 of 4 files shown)


src/neuroconv/datainterfaces/ecephys/requirements.txt
```diff
@@ -1,4 +1,4 @@
-neo @ git+https://github.com/NeuralEnsemble/python-neo@50abfbce832edb1c0f8364ac34c8a6442e1a3edf
+neo>=0.11
```
weiglszonja, line 1: This was due to spikeinterface 0.95.1 depends on neo>=0.11

src/neuroconv/datainterfaces/ecephys/requirements.txt
```diff
@@ -1,4 +1,4 @@
-neo @ git+https://github.com/NeuralEnsemble/python-neo@50abfbce832edb1c0f8364ac34c8a6442e1a3edf
-spikeinterface @ git+https://github.com/SpikeInterface/spikeinterface.git@4346eff5706408435e8aa8345be22fdd452afa27
+neo>=0.11
+spikeinterface>=0.95.1
 spikeextractors>=0.9.10
 probeinterface<=0.2.9 # https://github.com/catalystneuro/neuroconv/issues/127
```
CodyCBakerPhD, line 4: Can you also try to relax the upper bound on probeinterface? Does latest of that play well with latest of SI now?

CodyCBakerPhD: Actually looks like SI pins to probeinterface >= 0.2.10 anyway, so you can probably remove that altogether here

weiglszonja: oh yeah, sorry I went ahead and changed it to 0.2.10 or higher. I just finally would like to see whether the tests are passing or not 😄 from the local install that I was testing SI is installed with `neo-0.11.0` and `probeinterface-0.2.10`

src/neuroconv/datainterfaces/ecephys/requirements.txt
weiglszonja: ah yes, so we don't need neo and probeinterface here at all right?

weiglszonja: the same goes actually for spikeextractors>=0.9.10, SI 0.95.1. has the same pin

CodyCBakerPhD: Are you sure? I don't see spikeextractors listed in SI requirements - would be weird for them to do that, too, since it's in LEGACY

CI seems to agree, need to restore our deps for spikeextractors here

h-mayorquin:
I added some minor modifications on #133 . 

Concerning the use of strict vs lower bound versioning I was thinking that now that we don't have a `requirements.full` but we gather from here the tests should have specific dependencies. This will make the tests more reproducible (at least at the dependency level). Yes, maybe more conflicts will arise but then we can handle them and provide assurance that at least with these specific dependencies our pipe-line works. What do you think?

CodyCBakerPhD:
> Concerning the use of strict vs lower bound versioning I was thinking that now that we don't have a requirements.full but we gather from here the tests should have specific dependencies. This will make the tests more reproducible (at least at the dependency level). Yes, maybe more conflicts will arise but then we can handle them and provide assurance that at least with these specific dependencies our pipe-line works. What do you think?

If you wanted to add a separate `requirements-full-testing.txt` that exactly version pins things for testing purposes (mostly used in the suite) instead of the new default `neuroconv[full]` behavior, which loosely installs latest for safety, that would be fine. There can then be two dependency gallery tests - those that runs with the latest releases of that day, and the other that runs on dev branches for each dependency. The main non-gallery PR workflow would then use the exact pinning when running the last step for coverage.

h-mayorquin:
> > Concerning the use of strict vs lower bound versioning I was thinking that now that we don't have a requirements.full but we gather from here the tests should have specific dependencies. This will make the tests more reproducible (at least at the dependency level). Yes, maybe more conflicts will arise but then we can handle them and provide assurance that at least with these specific dependencies our pipe-line works. What do you think?
> 
> If you wanted to add a separate `requirements-full-testing.txt` that exactly version pins things for testing purposes (mostly used in the suite) instead of the new default `neuroconv[full]` behavior, which loosely installs latest for safety, that would be fine. There can then be two dependency gallery tests - those that runs with the latest releases of that day, and the other that runs on dev branches for each dependency. The main non-gallery PR workflow would then use the exact pinning when running the last step for coverage.

Does the lose dependency tries to install as new as possible?

CodyCBakerPhD:
> Does specifying lose/lower bound dependencies means to install as new as possible?

Lower bound dependencies always install latest release, yep.

h-mayorquin:
> > Does specifying lose/lower bound dependencies means to install as new as possible?
> 
> Lower bound dependencies always install latest release, yep.

Ah, OK, then that's very predictable as well and will also add the benefit of detecting if something new breaks the pipeline.

CodyCBakerPhD:
Which is why they're much safer for resolving conflicts; if you have two separate dependencies that each pin a separate lower bound, the highest will be used to automatically resolve. The problems with conflicts occur if an upper bound is used that produces an interval of non-intersection with other lower or upper bounds. And equality conflicts occur likewise if another dependency has a higher lower bound set for that package. Don't get me wrong, freezing things for testing is a lot more reliable, but we also need to keep our eyes ahead to stay up to date with the extensive secondary and tertiary dependency structure, such as the recent issue with the DANDI keyrings.

CodyCBakerPhD (approved):

PR #133, opened 2022-09-06
Bump spikeinterface and neo versions

h-mayorquin:
This PR objective is to synch with the latest release and catch any bugs on the CI if they appear


CodyCBakerPhD:
Already in progress on https://github.com/catalystneuro/neuroconv/pull/132

Also, let's try using lower bound versioning whenever possible with the new structure in order to minimize chance of conflicts (which have potential to arise any time we use exact pinning or upper bounds).

Individual conversion projects should still generally use a frozen set of exact version pins however, to ensure that the code functions well into the future.

h-mayorquin:
> Already in progress on #132

I needed to do some changes as they modify the format of the probe as I mentioned on the other issue. Now it seem to be working and I will set the base of this PR to #132.

@weiglszonja, feel free to just copy and paste the changes or merge from here to there.

CodyCBakerPhD (approved):

PR #135, opened 2022-09-06
Add TDT data interface (II)

### LINKED ISSUE #13: reading TDT

h-mayorquin:
This substitutes #70 
This should close #13 

We are pushing this for the release soon to come.

I solved issue 4 that @CodyCBakerPhD  pointed out here https://github.com/catalystneuro/neuroconv/pull/70#issuecomment-1213239985 which was deemed the most critical.

I will add the example to the conversion gallery and solve the other issues in another PR.


src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -56,11 +56,13 @@ def get_metadata_schema(self):
...
     def get_metadata(self):
         metadata = super().get_metadata()
+        channel_groups = np.unique(self.recording_extractor.get_channel_groups()) or ["ElectrodeGroup"]
+
         metadata["Ecephys"] = dict(
             Device=[dict(name="Device_ecephys", description="no description")],
             ElectrodeGroup=[
                 dict(name=str(group_id), description="no description", location="unknown", device="Device_ecephys")
-                for group_id in np.unique(self.recording_extractor.get_channel_groups())
+                for group_id in channel_groups
```
h-mayorquin, line 65: This was generating the None group that @CodyCBakerPhD saw on https://github.com/catalystneuro/neuroconv/pull/70#issuecomment-1213239985

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -348,10 +348,10 @@ def add_electrodes(
         if "group" in data_to_add:
             group_name_array = data_to_add["group"]["data"].astype("str", copy=False)
         else:
-            default_group_name = "default_group"
-            group_name_array = np.full_like(channel_name_array, fill_value=default_group_name)
+            default_group_name = "ElectrodeGroup"
```
h-mayorquin, line 351: Changing the default as discussed in https://github.com/catalystneuro/neuroconv/pull/70#issuecomment-1213239985 . 

Moreover, changing to `np.full` from `np.full_like` because using numpy array with strings is always tricky. Numpy implicitly defines a character size as the type and then cuts the size of your strings silently if they are longer than than the defined type (in this case I was confused for a while why a group called "Electro" was being created (turned out it was np.full_like cutting the strings).

CodyCBakerPhD:
@h-mayorquin Can you re-share the new NWB file for this format for comparison?

h-mayorquin:
> @h-mayorquin Can you re-share the new NWB file for this format for comparison?

The last commit re-introduce the None group mistake. Trying to fix that up now.

CodyCBakerPhD (approved):

CodyCBakerPhD (dismissed):

CodyCBakerPhD (approved):

PR #136, opened 2022-09-07
remove metadata.update(NWBFile=dict()) from conversion gallery

bendichter:
## Motivation

These lines were effectively deleting metadata that is populated by get_metadata.


h-mayorquin (approved):
Yep, my oversight, those should have been removed when we added NWBFile as a default in #51.

CodyCBakerPhD (approved):

PR #138, opened 2022-09-07
modality headings for conversion gallery

bendichter:


CodyCBakerPhD (approved):

CodyCBakerPhD:
LGTM

h-mayorquin:
Amazing, I have been thinking on doing this for a while.

docs/conversion_example_gallery.rst
```diff
@@ -61,6 +73,7 @@ Segmentation
 
 Behavior
 --------
+
```
h-mayorquin, line 76:
```suggestion
Pose estimation
~~~~~~~~~~~~
```

?

CodyCBakerPhD: Would you also want to put together a gallery + named section for the general purpose MovieInterface (since there are examples files available through the fixtures anyway)?

h-mayorquin: Yes, I think that's something to add.

bendichter: we can do that later once there is more than one interface here

PR #139, opened 2022-09-07
Add TDT data interface (II) - conversion gallery example-

h-mayorquin:
Adding a conversion gallery example for #135. This has it as its base.


bendichter:
@h-mayorquin this looks good to me. Ready for review?

PR #140, opened 2022-09-07
Try to skip CI for non-/src/ code changes

CodyCBakerPhD:
Saw this on SpikeInterface a while back and made a note to look into implementing similar logic here.

Also noticed all the PRs from @bendichter today making small improvements of the docs, but they all trigger full platform tests even though no actual code is changed (so all we really care about is the readthedocs build display).

On the topic of potentially applying this to a more fine-grain scale (*i.e.*, only run tests for a certain interface if code related to that interface is changed) I'd potentially be open to it but it would have to be very carefully configured to ensure the system doesn't unintentionally skip tests in situations they might be meant to run in - I'm normally of the opinion that it's best to run all tests all the time just to be sure some given change doesn't unintentionally break or alter something 'small' that is tangentially related.


h-mayorquin:
It seems to me that this in its current state will not trigger the doctests even if they are changed. Should not we also enable the doctests if changes occur in the `/docts/conversion_example_gallery`.  Maybe we can do this by allowing the doctests to run in their own action.

Another edge case if someone messing with `requirements-minimal.txt,  tests or setup files. We probably want to run tests after that.

CodyCBakerPhD:
Windows doesn't like the bash syntax so exploring other possiblities. But yes I'll add

- [ ] requirements changes
- [ ] extra step for explicitly running any of the galleries are changed

CodyCBakerPhD:
GitHub Actions are pretty cool.

I've rigged this now to have a pre-step that calculates file changes on push to a PR, then decides whether or not to trigger the main testing workflow. Can be refined in the future to whatever level we desire.

Will prevent _a lot_ of unnecessary testing when only submitting PRs for documentation changes.

In the future, could be rigged to reduce the total runtime of tests that aren't necessary to that PR, which should save quite a lot of CI time (not that we're being billed for it or anything, just saving on human wait times).

Still need to implement what @h-mayorquin suggested.

Also, specifying secrets as environment variables is a bit more nuanced, still need to debug that (hence the -0.75% coverage diff): https://github.com/orgs/community/discussions/25238#discussioncomment-3247035

CodyCBakerPhD:
Allright this should be done now.

Demonstration of usage:

(doctests)

[c563489](https://github.com/catalystneuro/neuroconv/pull/140/commits/c563489687893e07cad74911513745acd7a24731) introduced a change to a gallery, which properly triggered deployment of doctests only here: https://github.com/catalystneuro/neuroconv/actions/runs/3083818263/jobs/4985217734

The change also technically counted as a doctest failure which was later debugged and tests as of [d7fc07bd](d7fc07bd9ef937884c43f94b3e74887bb7c50d55) were passing

The change was then reverted in [bb0dcd1](https://github.com/catalystneuro/neuroconv/pull/140/commits/bb0dcd1118b44fe6019e56939bdab11897b4aa77) which then allows the deployment to properly not trigger the respective tests, saving quite a bit of CI time.


(source)

Similarly, [c57199d3](https://github.com/catalystneuro/neuroconv/pull/140/commits/c57199d3a6afab35ee5a584dd31ab55aa9339a07) introduced a dummy change to a source file, which properly triggered deployment of the full tests only thereafter: 

The change was then reverted in [8ae0ed63c](https://github.com/catalystneuro/neuroconv/pull/140/commits/8ae0ed63ce939da41f6d97878e58a98c89b417aa) which then allows the deployment to properly not trigger the respective tests, saving a massive amount of CI time.

CodyCBakerPhD:
@h-mayorquin @bendichter See https://github.com/catalystneuro/neuroconv/pull/140#issuecomment-1251271310 for summary

This will be very useful for the upcoming docathon, which is why I was pushing it a bit today

.github/workflows/deploy-tests.yml
```diff
@@ -0,0 +1,20 @@
...
+  run-tests:
+    needs: assess-file-changes
+    if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
+    uses: catalystneuro/neuroconv/.github/workflows/testing.yml@disable_tests_if_no_changes
+    secrets:
+      DANDI_API_KEY: ${{ secrets.DANDI_API_KEY }}
+
+  run-doctest:
+    needs: assess-file-changes
+    if: ${{ needs.assess-file-changes.outputs.CONVERSION_GALLERY_CHANGED == 'true' && needs.assess-file-changes.outputs.SOURCE_CHANGED != 'true' }}
```
h-mayorquin, line 19:
```suggestion
    if: ${{ needs.assess-file-changes.outputs.CONVERSION_GALLERY_CHANGED == 'true' || needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
```

I think that the doctest should also be run when the source changes. One of their goals is to ensure that our documentation is up to date with our code base so we want to couple them.

.github/workflows/deploy-tests.yml
```diff
@@ -0,0 +1,20 @@
+name: Deploy tests
+
+on:
+  pull_request:
+
+jobs:
+  assess-file-changes:
+    uses: catalystneuro/neuroconv/.github/workflows/assess-file-changes.yml@disable_tests_if_no_changes
```
h-mayorquin, line 8: What is the @disable_tests_if_no_changes doing in this section? Is it referencing this branch? Why do you need to reference the files in such an absolute way? Interesting

h-mayorquin (changes requested):
This looks good to me. As mentioned on the comments I think the doctests should also run when the source code, requirements of setup is changed. Is there a reason not to do this?

h-mayorquin:
Another question, right now we are requiring the full tests on python 3.9 as a condition for merge. Right now here is showing as required for me and is marked as "waiting for status to be reported". Should this condition be removed when neither of the tests were ran such as in this last actions iteration?

.github/workflows/deploy-tests.yml
CodyCBakerPhD: B/c the workflow only exists on this branch at the moment 😆 would have to update that to main after merge

.github/workflows/deploy-tests.yml
CodyCBakerPhD: I did that because doctest IS run as a part of the general testing workflow, no need to run it twice. The standalone workflow is just for explicitly running doctests and nothing else

CodyCBakerPhD:
> As mentioned on the comments I think the doctests should also run when the source code, requirements of setup is changed.

They are, they are always a part of the main `testing.yml` workflow, so the creation of this separate `doctets.yml` workflow only very specifically runs the doctests and nothing else if those are the only thing that ever changed.

CodyCBakerPhD:
> Another question, right now we are requiring the full tests on python 3.9 as a condition for merge. Right now here is showing as required for me and is marked as "waiting for status to be reported". Should this condition be removed when neither of the tests were ran such as in this last actions iteration?

Yes, good catch. I'd switch it to just requiring the deployment step since that's the only thing guaranteed to be run. Unfortunately I don't think that PR condition can be enhanced to that level of detail in the settings, it's purely binary pass/fail on latest commit, which doesn't play well with dynamic testing workflows like this. Hopefully they improve that in the future.

.github/workflows/deploy-tests.yml
h-mayorquin: Yes, I guess it is because the action machinery sits outside the repo and then you can't just call relative scripts.

.github/workflows/deploy-tests.yml
h-mayorquin: My bad, that makes perfect sense. 
One thing we could do to avoid this confusion for someone else is to change the line:
https://github.com/catalystneuro/neuroconv/blob/a5881e98d6eb2f80e92f92ab640c32982e8ce07d/.github/workflows/testing.yml#L127-L128

To
```
        run: pytest tests/ -n auto --dist loadscope --cov=./ --cov-report xml:./codecov.xml
```

So the testing workflow files (.yml) are segregated by function (run source code tests vs run doctests). What do you think?

h-mayorquin (approved):
Out of curiosity, can you say what did you need to add the report skipped tests -rst function? I read above that you had problems on the side of the code coverage and I guess that's related but I would like to know.

.github/workflows/deploy-tests.yml
CodyCBakerPhD: I'd probably still want it to include doctests in that single run because that's where code coverage comes from, so if there's a single line of coverage that somehow comes from the doctests but not from any other tests (kind of unlikely I suppose) we would miss it. Also they don't take very long so it doesn't take much to include them for simplicity

CodyCBakerPhD:
> Out of curiosity, can you say what did you need to add the report skipped tests -rst function? I read above that you had problems on the side of the code coverage and I guess that's related but I would like to know.

Oh I just noted that recently from the Inspect (which now skips certain tests if there's no network connectivity or other conditions) so I use the `-rx` flag to actually print out in the log the text-assigned 'reason' for skipping that test.

CodyCBakerPhD:
@h-mayorquin Thanks for taking a look - we'll see how this goes

PR #141, opened 2022-09-07
update user guide

bendichter:
## Motivation

Update user guide and make more modular.


CodyCBakerPhD:
Kind of an odd appearance for the schema sections - they show as separate tabs on the left but are contained on the same actual page (index just routes to the other section heading). Guessing this is intentional but it looked/felt a bit strange for them to not be on separate pages since all the other tabs are.

Not a huge deal, just noting.

CodyCBakerPhD:
There are some merge conflicts, though. CLI-level ones that can't be resolved on the web interface.

CodyCBakerPhD (approved):

PR #145, opened 2022-09-08
Conversion gallery grid / two columns

h-mayorquin:
This should work as an illustration of a grid / doube column design for the conversion gallery using [sphinx-design](https://sphinx-design.readthedocs.io/en/rtd-theme/index.html). (check out their other examples!)

I mainly illustrate two things here:
1) We can add some grid design to organize the different sub-modalities  (recording vs sorting).
2) We can add elements to the boxes in the grids such as images. 

@bendichter I am not sure if this is what you imagined when you mentioned the two columns. The other approach that I tried requiring modifying the css but I could not make the things align or have the headers show in a nice way so I doped after some trying.  This is how it looks:

Layout:
image
Another layout:
image

The images that I have chosen for illustrating the image displaying capabilities of sphinx-design are probably poor in taste but I just wanted to illustrate that possibility.  Note also that I illustrate two types of headers for the grid elements in sorting. I prefer the bold one that is shown in recording and the rest of the boxes. 

A drawback of this approach is that I could not organize the general navigation to showcase the contents separated sub-modality as it was before:
Before:
image
Now:
image

If we like this approach and the capabilities probably we can do some trick to make this work but I could not find it.

@weiglszonja it would be great to get some feedback from you if you have some time.


CodyCBakerPhD:
Another option for images, if desired, could be to have one per modality and just re-use the ones from PyNWB: https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/docs/source/figures/gallery_thumbnails_ophys.png

Otherwise I think layout 1 looks good but it is unfortunate that the nav bar does not reflect the sectioning...

bendichter:
Thanks for taking a look at this, @h-mayorquin This is a lookup/index/glossary page and I think the images are distracting here. I also think the panels are distracting and I prefer the current layout.

h-mayorquin:
> Thanks for taking a look at this, @h-mayorquin This is a lookup/index/glossary page and I think the images are distracting here. I also think the panels are distracting and I prefer the current layout.

All right, let's close this then.

h-mayorquin:
> > All right, let's close this then.
> 
> No way to get the columns without the graphical panels around them?

I tried briefly and it seems to me that there is an option for this:
https://sphinx-design.readthedocs.io/en/rtd-theme/css_classes.html#borders

However, I spent 2 hours and half already and I felt that I was spending too much time into that. What do you think, @bendichter ? Something worth looking into?

PR #149, opened 2022-09-08
Re enable tools on the API documentation

### LINKED ISSUE #148: Tools API not working in the documentation
So, when we look at the tools we only get this:

image

Check it out here:
https://neuroconv.readthedocs.io/en/main/api/tools.html
The API is in fact missing for all save the yaml functionality. Maybe related to the init import.

h-mayorquin:
This should close #148 .

I also added NEO and the iterator for roi that were not in the API before.

This works locally for me but somehow is failing here. Trying different things.


CodyCBakerPhD:
Would be nice if there was some sort of workflow that could check for empty sections of text in the docs. Do you think that could be possible?

h-mayorquin:
> Would be nice if there was some sort of workflow that could check for empty sections of text in the docs. Do you think that could be possible?

Yes, I will think about it.

h-mayorquin:
This seems to be working now. 

@CodyCBakerPhD, is there a way to not trigger the full tests when the `requirement.rst` is modified?

Any pattern we can block here?
https://github.com/catalystneuro/neuroconv/blob/main/.github/workflows/assess-file-changes.yml#L36-L39

requirements-rtd.txt
```diff
@@ -1,7 +1,9 @@
-myst_parser==0.18.0
 Jinja2<3.1
 Sphinx==5.1.1
 sphinx_rtd_theme==1.0.0
 readthedocs-sphinx-search==0.1.2
 sphinx-toggleprompt==0.2.0
 sphinx-copybutton==0.5.0
+spikeinterface>=0.95.1  # Necessary to display tools.spikeinterface
+spikeextractors>=0.9.10 # Necessary to display tools.spikeinterface
+roiextractors>=0.5.0  # Necessary to displays tools.roiextractors
```
CodyCBakerPhD, line 9: Can't we just add `-full` to the extras of the `.readthedocs.yaml`?

h-mayorquin: I think that can be done, do you think is preferable?

CodyCBakerPhD: Yeh, keeps the requirements files from non overlapping (esp. w.r.t. the new recursive setup logic)

CodyCBakerPhD:
> @CodyCBakerPhD, is there a way to not trigger the full tests when the requirement.rdt is modified?
> 
> Any pattern we can block here?
https://github.com/catalystneuro/neuroconv/blob/main/.github/workflows/assess-file-changes.yml#L36-L39

Increased the specificity. Should be good now.

CodyCBakerPhD:
LGTM, thanks for fixing.

Minor note: I don't know why, but https://neuroconv--149.org.readthedocs.build/en/149/api/tools.roiextractors.html#neuroconv.tools.roiextractors.roiextractors._imaging_frames_to_hdmf_iterator has text overflowing for parameter descriptions whereas the other ones on that page do not...?

requirements-rtd.txt
h-mayorquin: I actually thought that you would not like to have as many unnecessary dependencies on the documentation environment. I think it makes it more brittle. As in, if any of all of the dependencies in full breaks or fails to install then the documentation will fail to build. This would not be a problem if we had a dev branch separated from a stable branch.

What do you think?

CodyCBakerPhD (approved):

h-mayorquin:
> LGTM, thanks for fixing.
> 
> Minor note: I don't know why, but https://neuroconv--149.org.readthedocs.build/en/149/api/tools.roiextractors.html#neuroconv.tools.roiextractors.roiextractors._imaging_frames_to_hdmf_iterator has text overflowing for parameter descriptions whereas the other ones on that page do not...?

Yeah, I don't know, something about the way we write our docstrings I think. But I guess this will have to wait to the next dockaton x D

PR #151, opened 2022-09-08
Increase precision of inferred regularity rates

CodyCBakerPhD:
@weiglszonja How about this?

### DIFF
--- a/tests/test_behavior/test_movie_interface.py
+++ b/tests/test_behavior/test_movie_interface.py
@@ -251,8 +251,8 @@ class TestMovieInterface(TestCase):
 
         fps = 25.0
         expected_warn_msg = (
-            "The fps=25.0 from movie data is unequal to the difference in "
-            "regular timestamps. Using fps=5.0 from timestamps instead."
+            "The fps=25 from movie data is unequal to the difference in "
+            "regular timestamps. Using fps=5 from timestamps instead."
         )
 
         with self.assertWarnsWith(warn_type=UserWarning, exc_msg=expected_warn_msg):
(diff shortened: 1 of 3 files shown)


weiglszonja:
Yes, that will also be needed, also https://github.com/catalystneuro/roiextractors/pull/212 is more consistent to `spikeinterface`. I'm rerunning the tests now, but so far it seems good.

weiglszonja (approved):

tests/test_behavior/test_movie_interface.py
```diff
@@ -248,11 +249,10 @@ def test_movie_regular_timestamps(self):
             Movie=dict(starting_times=self.starting_times, timestamps=timestamps, external_mode=True)
         )
 
-        rate = 1.0 / np.round(timestamps[1] - timestamps[0], 9)
         fps = 25.0
         expected_warn_msg = (
-            f"The fps={fps} from movie data is unequal to the difference in "
-            f"regular timestamps. Using fps={rate} from timestamps instead."
+            "The fps=25.0 from movie data is unequal to the difference in "
+            "regular timestamps. Using fps=5.0 from timestamps instead."
```
weiglszonja, line 255: @CodyCBakerPhD 
Looking at the CI run, this should either be fps=5.000000000000007, or should use rate = 1.0 / (timestamps[1] - timestamps[0]), but I can see why you removed that before, but hopefully how rate is calculated won't be changed soon.

CodyCBakerPhD: Right, that's just a display issue though(with the print statement). Fixing now

weiglszonja (approved):
Thank you @CodyCBakerPhD for helping with this.

PR #156, opened 2022-09-13
Remove unnecessary call to `get_roi_locations` in segmentation pipeline

h-mayorquin:
This PR removes an unecessary call to `get_roi_locations` on the segmentation pipeline. This is called early and in fact unecessarily when `include_roi_centroids` is True as it is called again and used only in that case.


CodyCBakerPhD:
Wonder how this got through #118 - thanks for catching this! Suspect it may have been an incorrect merge conflict resolution...

CodyCBakerPhD (approved):

PR #158, opened 2022-09-14
Bump/test ROIExtractors latest release

CodyCBakerPhD:
In advance of the upcoming latest NeuroConv release, testing against latest ROIExtractors release.

@h-mayorquin had mentioned something about seeing some errors but it worked fine when I tried, see what the CI says.


h-mayorquin:
It is good that it worked. The error was related to some floating point number when calculating the rate from timestamps.

PR #159, opened 2022-09-14
Fix segmentation rate/timestamps with metadata

CodyCBakerPhD:
@weiglszonja I found a bug when writing through a segmentation interface with irregular timestamps

I've added some example tests that here that are failing, I'll fix it if I have the time

The problem is mainly that the call to `get_nwb_segmentation_metadata` sets the `rate` value which acts in addition to the new rate/timestamps stuff from #130, but when writing an `RoiResponseSeries` you cannot use *both* `rate` (always in the metadata) *and* `timestamps` (in the event they are irregular).


h-mayorquin:
Maybe we should remove the rate from the default metadata now that we are doing are always testing for timestamps.

weiglszonja:
@h-mayorquin I'd support that.

CodyCBakerPhD:
> Maybe we should remove the rate from the default metadata now that we are doing are always testing for timestamps.

Yep, that's the fix needed. Rate info should be attached via the ROIExtractors objects, not the overhead metadata dictionary

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -1054,6 +1055,53 @@ def test_add_fluorescence_traces_regular_timestamps(self):
...
+        add_fluorescence_traces(
+            segmentation_extractor=segmentation_extractor,
+            nwbfile=self.nwbfile,
+            metadata=self.metadata,
+        )
+
+        ophys = get_module(self.nwbfile, "ophys")
+        roi_response_series = ophys.get(self.fluorescence_name).roi_response_series
+        for series_name in roi_response_series.keys():
+            self.assertEqual(roi_response_series[series_name].rate, None)
```
h-mayorquin, line 1077: I don't understand this test. The dummy segmentation extractor has a constant sampling rate (30 by default). What do you mean by default timestamps in the string?

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -1054,6 +1055,53 @@ def test_add_fluorescence_traces_regular_timestamps(self):
...
+    def test_add_fluorescence_traces_irregular_default_timestamps_with_metadata(self):
+        """Test adding traces with default timestamps and metadata rates."""
+        segmentation_extractor = generate_dummy_segmentation_extractor(
+            num_rois=2,
+            num_frames=5,
+            num_rows=self.num_rows,
+            num_columns=self.num_columns,
+        )
+        metadata = deepcopy(self.metadata)
+        metadata["Ophys"]["Fluorescence"]["roi_response_series"][0].update(rate=1.23)
```
h-mayorquin, line 1090: Is this second one to test whether we can update / overwrite a specific rate by means of the metadata?

CodyCBakerPhD: This is just showcasing the current issue (which has more to do with how the default behavior adds rate info to the metadata dictionary regardless of if timestamps are set in the segmentation extractor). This test could probably be removed if that default behavior is changed

CodyCBakerPhD: To clarify, default behavior when using any SegmentationInterface (which does stuff to the metadata dictionary where our unit tests of the `add_fluorescence_traces` function does not)

tests/test_ophys/test_tools_roiextractors.py
CodyCBakerPhD: I think some code might have been lost in between pushes, updated it to what I think is the expected behavior (but again, I don't thing any `rate` info should be in overhead metadata, should just be attached to the extractor object

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -1054,6 +1055,61 @@ def test_add_fluorescence_traces_regular_timestamps(self):
...
+        )
+
+        ophys = get_module(self.nwbfile, "ophys")
+        roi_response_series = ophys.get(self.fluorescence_name).roi_response_series
+        for series_name in roi_response_series.keys():
+            self.assertEqual(roi_response_series[series_name].rate, 1.23)
+            self.assertEqual(roi_response_series[series_name].starting_time, 0)
+            self.assertEqual(roi_response_series[series_name].timestamps, None)
+
+    def test_add_fluorescence_traces_irregular_timestamps_with_metadata(self):
```
h-mayorquin, line 1086: OK, so if I got you right, the behavior is that if the user does pass the metadata (in this case rate) then this should overwrite any other behavior. That is, even if the timestamps are indeed irregular the file will be written as a regular one. 

I modified the code so it does this.

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -1054,6 +1055,61 @@ def test_add_fluorescence_traces_regular_timestamps(self):
...
+        segmentation_extractor = generate_dummy_segmentation_extractor(
+            num_rois=2,
+            num_frames=5,
+            num_rows=self.num_rows,
+            num_columns=self.num_columns,
+        )
+        segmentation_extractor.set_times(times)
+
+        metadata = deepcopy(self.metadata)
+        metadata["Ophys"]["Fluorescence"]["roi_response_series"][0].update(rate=1.23)
```
h-mayorquin, line 1070: I have hammered this a couple of times but I still feel that that representing stuff as lists as the metadata (instead of dicts) confuses me. Like, what is the index 0 doing there?

I added the modification for the other container which I guess from the loop below you intended. I also passed the right metadata and no the one in self that you had before. Please let me know if those were as you intended.

h-mayorquin (commented):
I removed things from default metadata. The tests are working on my machine but there is some things to discuss. We should be able to merge this soon.

tests/test_ophys/test_tools_roiextractors.py
CodyCBakerPhD: > I have hammered this a couple of times but I still feel that that representing stuff as lists as the metadata (instead of dicts) confuses me. Like, what is the index 0 doing there?

Yeah I believe this is on the to-do list or we are at least aware that that zero-index is a current placeholder (here and in other places, like the actual roiextractors submodule) and will eventually be replaced with the more informative link

tests/test_ophys/test_tools_roiextractors.py
CodyCBakerPhD: Thanks, I'll check it out in practice

h-mayorquin:
I think this is ready after the last commit, what do you think, @CodyCBakerPhD ?

CodyCBakerPhD:
Well, I'm past the rate/timestamp error but I'm getting a new one, will play with it a bit to see what the source cause is

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -126,13 +126,9 @@ def get_nwb_imaging_metadata(imgextractor: ImagingExtractor):
                     description="An optical channel of the microscope.",
                 )
             )
-    # set imaging plane rate:
-    rate = np.nan if imgextractor.get_sampling_frequency() is None else float(imgextractor.get_sampling_frequency())
 
-    # adding imaging_rate:
-    metadata["Ophys"]["ImagingPlane"][0].update(imaging_rate=rate)
```
weiglszonja, line 133: Should this be also done in `get_nwb_segmentation_metadata`?

CodyCBakerPhD: Yep, thanks for catching that

h-mayorquin: Indeed. Thanks.

h-mayorquin (commented):
The failing this is generated because there is mismatch between the rate that the segmentation extractor has and the one that ends in the nwbfile. They are not very big:

```
segmentation_extractor2.get_sampling_frequency()
1.855481130340034
segmentation_extractor1.get_sampling_frequency()
1.8554811303400338
```

Maybe they could be compared with almost equal instead of with equal as they are being done.  This I think would be a solution to this problem.

However, this also highlights another problem we have.  We store in the file the rate determined empirically with the function `calculate_regular_series_rate` which has an empirical estimation that sometimes can end up being a little bit different from the actual rate as well because  of floating point and rounding issues. Compounded by the fact that this segmentation extractor that is failing has a strange sampling rate I think.

Related to:
https://github.com/catalystneuro/roiextractors/issues/153
https://github.com/catalystneuro/roiextractors/issues/193

What do you guys think? @CodyCBakerPhD @weiglszonja

CodyCBakerPhD:
> Maybe they could be compared with almost equal instead of with equal as they are being done. This I think would be a solution to this problem.

Yeah that's the solution I've been working on (by having `check_segmentations_equal` have an optional `almost_equal` argument that we would set only for this one test)


But it is a good question... Since `sampling_frequency` is required of all extractor objects, should perhaps use that set value instead of the one inferred by processing of `frame_to_time` (which may use `timestamps`, regular or no, or interpolated values from the sampling frequency, which can lead to a small rounding here we see here).

I'll get something resolved for this early next week

weiglszonja:
I was thinking that maybe we could refactor this part, change `calculate_regular_series_rate()` to only return whether the series is regular or not, and if so, use the rate from `get_sampling_frequency()`?
Do you think this could help? @CodyCBakerPhD @h-mayorquin 

```python
    timestamps = segmentation_extractor.frame_to_time(np.arange(segmentation_extractor.get_num_frames()))
    is_regular = check_if_series_is_regular(series=timestamps)

    if is_regular:
        roi_response_series_kwargs.update(starting_time=timestamps[0], rate=segmentation_extractor.get_sampling_frequency())
    else:
        roi_response_series_kwargs.update(timestamps=H5DataIO(data=timestamps, **compression_options))
```

https://github.com/catalystneuro/neuroconv/blob/d4bf05c1523c39a297be0ebb4376c8378be1b4ce/src/neuroconv/tools/roiextractors/roiextractors.py#L321-L327

CodyCBakerPhD:
> I was thinking that maybe we could refactor this part, change calculate_regular_series_rate() to only return whether the series is regular or not, and if so, use the rate from get_sampling_frequency()?
Do you think this could help? @CodyCBakerPhD @h-mayorquin

That's the better way to go if the extractor does not have timestamps explicitly set to it, yes. Otherwise let's say someone did set the timestamps explicitly, but they are regular nonetheless at a sampling frequency that does not match (significantly or otherwise) the scalar set to that extractor as the `sampling_frequency` attribute. In that case I'd say it makes more sense to infer it via the calculation function. Really we just need a better way to determine what was manually vs. automatically assigned to the extractor, which is not something we have right now since we rely only on the output generated by `frame_to_time` (which uses either pathway, depending on available attributes)...

weiglszonja:
> Really we just need a better way to determine what was manually vs. automatically assigned to the extractor,

Would this help, if it has a time vector then we can assume it was manually defined?
https://spikeinterface.readthedocs.io/en/latest/api.html#spikeinterface.core.BaseSorting.has_time_vector

CodyCBakerPhD:
> Would this help, if it has a time vector then we can assume it was manually defined?
https://spikeinterface.readthedocs.io/en/latest/api.html#spikeinterface.core.BaseSorting.has_time_vector

Indeed! That would be a great solution, good find!

CodyCBakerPhD:
@weiglszonja @h-mayorquin OK this should be fixed now

weiglszonja (approved):
Looks good to me now, thank you for working on this!

PR #160, opened 2022-09-15
Add SLEAP data interface

### LINKED ISSUE #48: [Feature]: Add SLEAP
### What would you like to see added to nwb-conversion-tools?

The SLEAP team recently added NWB support.

Would be nice to have a DataInterface here for that.

Similar to the way DLC initially did it, none of the logic writing to an in-memory file was peeled out so we would have to do/request that: https://github.com/talmolab/sleap/blob/develop/sleap/io/format/ndx_pose.py#L262-L339

This issue is primarily just for the pose estimation series output for each session, they are also doing a lot of work representing the training data through a new neurodata type as well: https://github.com/rly/ndx-pose/pull/9

They have testing data available from their repo: https://github.com/talmolab/sleap/tree/develop/tests/data for example usage of it refer to https://github.com/talmolab/sleap/blob/e2ec3b3b9eec243d943f46ced6e61e3c4f69b27a/tests/io/test_formats.py#L365-L399

_No response_

### What solution would you like?

Goals is to make a SLEAPInterface for interfacing with their data.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/nwb-conversion-tools/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/nwb-conversion-tools/issues)?

h-mayorquin:
Adding datainterface for conversions with sleap data. This should close #48.


tests/test_on_data/test_gin_behavior.py
```diff
@@ -1,14 +1,139 @@
...
+        interface.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, metadata=metadata)
+
+        import sleap_io
+
+        slp_predictions_path = interface_kwargs["file_path"]
+        labels = sleap_io.load_slp(slp_predictions_path)
+
+        with NWBHDF5IO(path=nwbfile_path, mode="r", load_namespaces=True) as io:
+            nwbfile = io.read()
+            # nwbfile.processing["SLEAP_VIDEO_000_20190128_113421"]["track=track_0"].pose_estimation_series
```
CodyCBakerPhD, line 44: Is this needed?

src/neuroconv/datainterfaces/behavior/sleap/sleapdatainterface.py
```diff
@@ -0,0 +1,70 @@
...
+
+        base_metadata = self.get_metadata()
+        metadata = dict_deep_update(base_metadata, metadata)
+
+        with make_or_load_nwbfile(
+            nwbfile_path=nwbfile_path, nwbfile=nwbfile, metadata=metadata, overwrite=overwrite, verbose=self.verbose
+        ) as nwbfile_out:
+
+            labels = self.sleap_io.load_slp(self.file_path)
+            nwbfile_out = self.sleap_io.append_labels_data_to_nwb(labels, nwbfile_out)
```
CodyCBakerPhD, line 68: Nice and simple! Well done

src/neuroconv/datainterfaces/behavior/sleap/sleapdatainterface.py
```diff
@@ -0,0 +1,70 @@
...
+
+from pynwb.file import NWBFile
+
+from ....basedatainterface import BaseDataInterface
+from ....tools.nwb_helpers import make_or_load_nwbfile
+from ....tools import get_package
+from ....utils import dict_deep_update, FilePathType, OptionalFilePathType
+
+
+class SleapInterface(BaseDataInterface):
```
CodyCBakerPhD, line 13: I know it's kind of minor, but references on their site are as SLEAP since it's an abbreviation. Should this then reflect that as the `SLEAPInterface`, like we do for other things like EDF, CED, SpikeGLX, etc?

tests/test_on_data/test_gin_behavior.py
h-mayorquin: Nop, thanks.

tests/test_on_data/test_gin_behavior.py
```diff
@@ -1,14 +1,139 @@
...
+    )
+    def test_sleap_to_nwb_interface(self, data_interface, interface_kwargs):
+        nwbfile_path = str(self.savedir / f"{data_interface.__name__}.nwb")
+
+        interface = SleapInterface(**interface_kwargs)
+        metadata = interface.get_metadata()
+        metadata["NWBFile"].update(session_start_time=datetime.now().astimezone())
+        interface.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, metadata=metadata)
+
+        import sleap_io
```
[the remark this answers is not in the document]
h-mayorquin: Nop. I was testing some errors earlier.

src/neuroconv/datainterfaces/behavior/sleap/sleapdatainterface.py
h-mayorquin: That makes sense to me. Let's see what @bendichter decides.

bendichter: I vote for SLEAP

h-mayorquin: All right .I changed it here and on the gallery.

CodyCBakerPhD:
A couple minor things I noticed from the file shared on Slack - not necessary to fix in this PR but just to be aware of...

a) None of the labels data fields are compressed. Not a huge deal for now but for a longer session with lots of labels it could add up. Recommend enabling compression for all such datasets.

b) `labeled_videos` and `original_videos` are written as absolute paths (probably depends on how you specified the paths given to SLEAP, I'm guessing) - would be great if this could be moved, copied to be adjacent to the `nwbfile_path` (similar to how `dandi organize` offers options for this), and/or the path itself altered to be a relative path to the NWB file to make just sharing easier. As it stands, the paths there only be meaningful on the system that wrote them (yours). Even if including those fields is just for provenance, then there's still a secondary concern about privacy (what if the names of upstream folders to the video path contain sensitive or privatized information?) so if it's just for provenance maybe just write the file stem?

c) The `rate` linked to some `starting_time`s (*e.g.* `track_0/eyeR`) is `1.0` - is that just for simplicity in the example file? I would have thought it would be the `fps` of the associated movie.

d) Some of the labels on `track_0` have rate/starting time and others have timestamps. Is this intentional and/or typical? I had assumed all time information would be consistent and drawn from the movie itself.

CodyCBakerPhD (approved):

h-mayorquin:
Thanks for taking the time to go over this, @CodyCBakerPhD 

> A couple minor things I noticed from the file shared on Slack - not necessary to fix in this PR but just to be aware of...
> 
> a) None of the labels data fields are compressed. Not a huge deal for now but for a longer session with lots of labels it could add up. Recommend enabling compression for all such datasets.

Yep, this is on the works.

> b) `labeled_videos` and `original_videos` are written as absolute paths (probably depends on how you specified the paths given to SLEAP, I'm guessing) - would be great if this could be moved, copied to be adjacent to the `nwbfile_path` (similar to how `dandi organize` offers options for this), and/or the path itself altered to be a relative path to the NWB file to make just sharing easier. As it stands, the paths there only be meaningful on the system that wrote them (yours). Even if including those fields is just for provenance, then there's still a secondary concern about privacy (what if the names of upstream folders to the video path contain sensitive or privatized information?) so if it's just for provenance maybe just write the file stem?

Yep, to be fixed. 

> c) The `rate` linked to some `starting_time`s (_e.g._ `track_0/eyeR`) is `1.0` - is that just for simplicity in the example file? I would have thought it would be the `fps` of the associated movie.

This is because sleap files do not have timestamps or the frame rates of the movies by default.  As the timstamps is a required argument they make up fake timestamps with one frame per second. I am trying to work with them to add that.

> d) Some of the labels on `track_0` have rate/starting time and others have timestamps. Is this intentional and/or typical? I had assumed all time information would be consistent and drawn from the movie itself.

This is intentional, not all the nodes of the pose estimation series are present in every frame (sometimes the prediction fails and therefore some frames do not have a prediction for let's say the left leg) so the timestamps are choppy. When they are not choppy and the timestamps are regular then a rate is loaded instead.

CodyCBakerPhD:
> This is intentional, not all the nodes of the pose estimation series are present in every frame (sometimes the prediction fails and therefore some frames do not have a prediction for let's say the left leg) so the timestamps are choppy. When they are not choppy and the timestamps are regular then a rate is loaded instead.

Ohhh I see.

An alternative way of representing that would be `observation_intervals` like in the Units table: https://nwb-schema.readthedocs.io/en/latest/format.html#units. See https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/tests/unit_tests/test_ecephys.py#L192-L205 for some practical examples of usage.

h-mayorquin:
But that's for a dynamic table, each of the nodes is a PoseEstimationSeries which is just a TimeSeries with extra properties? Can those have observational units as well?

CodyCBakerPhD:
> But that's for a dynamic table, each of the nodes is a PoseEstimationSeries which is just a TimeSeries with extra properties? Can those have observational units as well?

Ah, that's true, you're right. You *could* technically adapt the schema for the `PostEstimationSeries` to take an extra field that functions identically to the observation intervals idea but that would be quite a bit more work. But then you would also have the potential for Best Practice violations similar to what I linked from the Inspector tests where you could have manually specified timestamps occurring outside of those intervals, so maybe best to just stick to timestamps only.

bendichter:
thanks for the hard work on this, guys!

Issue #42, opened 2022-05-09
Discussion: Unify `run_conversion` interfaces between base classes and `nwbconverter`


h-mayorquin:
As we have been using `run_conversion` directly from the interfaces lately I have been discussing this with @CodyCBakerPhD . I am opening this issue to have a template to discuss tomorrow.

I think that we should have a common set of arguments (i.e. unify the keyword arguments that control input-output). Here I mean things like passing an nwbfile or a save_path. Another one is over-write functionality.

Other important considerations listed in no particular order are:
* The base data interfaces do not perform any validation.
* Division of responsibilities between `spikeinterface` and `nwb-conversion-tools`. 

For reference, the signatures of all the `run_conversions` methods at the moment of writing this issue are:
```python
# NWB general
def run_conversion(
    self,
    metadata: Optional[dict] = None,
    save_to_file: Optional[bool] = True,
    nwbfile_path: Optional[str] = None,
    overwrite: Optional[bool] = False,
    nwbfile: Optional[NWBFile] = None,
    conversion_options: Optional[dict] = None,
)

# Sorting extractor
def run_conversion(
    self,
    nwbfile: NWBFile = None,
    metadata: dict = None,
    stub_test: bool = False,
    save_path: OptionalFilePathType = None,
    overwrite: bool = False,
    write_ecephys_metadata: bool = False,
):

# Recording extractor
def run_conversion(
    self,
    nwbfile: NWBFile = None,
    metadata: dict = None,
    stub_test: bool = False,
    starting_time: Optional[float] = None,
    use_times: bool = False,
    save_path: OptionalFilePathType = None,
    overwrite: bool = False,
    write_as: str = "raw",
    write_electrical_series: bool = True,
    es_key: str = None,
    compression: Optional[str] = "gzip",
    compression_opts: Optional[int] = None,
    iterator_type: Optional[str] = None,
    iterator_opts: Optional[dict] = None,
)

# Movie data interface
def run_conversion(
    self,
    nwbfile: NWBFile,
    metadata: dict,
    stub_test: bool = False,
    external_mode: bool = True,
    starting_times: Optional[list] = None,
    chunk_data: bool = True,
    module_name: Optional[str] = None,
    module_description: Optional[str] = None,
    compression: Optional[str] = "gzip",
    compression_options: Optional[int] = None,
):

# Imaging
def run_conversion(
    self,
    nwbfile: NWBFile = None,
    metadata: dict = None,
    overwrite: bool = False,
[...]

PR #164, opened 2022-09-19
[Proposal] Combine simple examples in gallery with NWBConverter

h-mayorquin:
Right now we have the conversion gallery which as simple as possible illustrates how to do conversion with a simple interface:
https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/conversion_example_gallery.html#extracellular-electrophysiology

In order to combine more than one interface we use the NWBConverter object but the tutorial, in my view, is not as simple as the one for the simple gallery:
https://neuroconv.readthedocs.io/en/main/user_guide/nwbconverter.html

To be specific, the complexities that I see to the end user are the following:
1) They need to define a class themselves and inherit from the converter
2)  To define all the properties they need to use a structure of nested dictionaries for the source data, conversion options and metadata. 

I feel that this creates an unnecessary gap in complexity between single interface conversion and multiple interface conversion and I want to propose some solutions.  This PR is an unpolished prototype of what I believe is the simplest idea:

If we add the possibility to the `NWBConverter` object to take as an input already initialized interfaces (as defined in the conversion gallery) then we retain all the functionality of the NWBConverter object that we had before. In addition, users can use all the information of the single interface conversion examples in the gallery to build their interfaces and then just combine them using a simple input pass to the `NWBConverter` object. 

To illustrate this idea I add to this PR the necessary changes to enable this and a conversion gallery example which shows how to combine recording and sorting interfaces in a single conversion.

### DIFF
--- a/docs/conversion_examples_gallery/combinations/recording_and_sorting.rst
+++ b/docs/conversion_examples_gallery/combinations/recording_and_sorting.rst
@@ -1,0 +1,43 @@
+Conversion combining electrophysiological recordings and spike sorting data
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A common workflow is to record electrophysiological data and then extract spiking units from the stream. The following is an
+example of how to combine electrophysiological recordings and spike sorting data in the same conversion.
+
+Note that for this specific example were using SpikeGLXRecordingInterface (insert link) and PhySortingInterface (insert link)but any
+of the other examples in our conversion gallery (insert link to recorders and sorters) can be combined in this way.
+
+.. code-block:: python
+
+    >>> from datetime import datetime
+    >>> from dateutil import tz
+    >>> from pathlib import Path
+    >>> from neuroconv.datainterfaces import SpikeGLXRecordingInterface
+    >>> from neuroconv.datainterfaces import PhySortingInterface
+    >>>
+    >>> from neuroconv import NWBConverter
+    >>> ECEPHY_DATA_PATH = Path("/home/heberto/ephy_testing_data/")
+    >>>
+    >>> # For this interface we need to pass the location of the ``.bin`` file
+    >>> file_path = f"{ECEPHY_DATA_PATH}/spikeglx/Noise4Sam_g0/Noise4Sam_g0_imec0/Noise4Sam_g0_t0.imec0.ap.bin"
+    >>> # Change the file_path to the location in your system
+    >>> interface_spikeglx = SpikeGLXRecordingInterface(file_path=file_path, verbose=False)
+    >>>
+    >>> folder_path = f"{ECEPHY_DATA_PATH}/phy/phy_example_0"
+    >>> # Change the folder_path to the location of the data in your system
+    >>> interface_phy = PhySortingInterface(folder_path=folder_path, verbose=False)
+    >>>
+    >>>  # Now that we have defined the two interfaces we pass them to the NWBConverter which will coordinate the
+    >>>  # concurrent conversion of the data
+    >>> converter = NWBConverter(data_interfaces=[interface_spikeglx, interface_phy])
+    >>>
+    >>> # Extract what metadata we can from the source files
+    >>> metadata = converter.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> tzinfo = tz.gettz("US/Pacific")
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific")).isoformat()
+    >>> metadata["NWBFile"].update(session_start_time=session_start_time)
+    >>>
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> nwbfile_path = f"test.nwb"
+    >>> converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, metadata=metadata)
--- a/docs/conversion_examples_gallery/conversion_example_gallery.rst
+++ b/docs/conversion_examples_gallery/conversion_example_gallery.rst
@@ -80,3 +80,12 @@ Behavior
 
     DeepLabCut <behavior/deeplabcut>
     SLEAP <behavior/sleap>
+
+Examples of common interface combinations
+-----------------------------------------
+
+.. toctree::
+    :maxdepth: 1
+
+    Combine electrophysiological recordings and spike sorting <combinations/recording_and_sorting>
+    SLEAP <behavior/sleap>
--- a/src/neuroconv/nwbconverter.py
+++ b/src/neuroconv/nwbconverter.py
@@ -1,7 +1,7 @@
 """Authors: Cody Baker and Ben Dichter."""
 import json
 from jsonschema import validate
-from typing import Optional, Dict
+from typing import Optional, Dict, List
 from pathlib import Path
 
 from pynwb import NWBFile
@@ -60,15 +60,27 @@ class NWBConverter:
         """Validate source_data against Converter source_schema."""
         cls._validate_source_data(source_data=source_data, verbose=verbose)
 
-    def __init__(self, source_data: Dict[str, dict], verbose: bool = True):
+    def __init__(self, source_data: Dict[str, dict] = None, data_interfaces: List = None, verbose: bool = True):
         """Validate source_data against source_schema and initialize all data interfaces."""
         self.verbose = verbose
-        self._validate_source_data(source_data=source_data, verbose=self.verbose)
-        self.data_interface_objects = {
-            name: data_interface(**source_data[name])
-            for name, data_interface in self.data_interface_classes.items()
-            if name in source_data
-        }
+
+        if source_data is not None:
+            self._validate_source_data(source_data=source_data, verbose=self.verbose)
+
+            self.data_interface_objects = {
+                name: data_interface(**source_data[name])
+                for name, data_interface in self.data_interface_classes.items()
+                if name in source_data
+            }
+
+        elif data_interfaces is not None:
+
+            self.data_interface_objects = {
+                f"{interface.__class__.__name__}{index}": interface for index, interface in enumerate(data_interfaces)
+            }
+
+        else:
+            raise IOError("Need to pass either source_data or a list with initialized data interfaces")
 
     def get_metadata_schema(self):
         """Compile metadata schemas from each of the data interface objects."""
@@ -166,9 +178,12 @@ class NWBConverter:
 
         if conversion_options is None:
             conversion_options = dict()
+
         default_conversion_options = self.get_conversion_options()
         conversion_options_to_run = dict_deep_update(default_conversion_options, conversion_options)
-        self.validate_conversion_options(conversion_options=conversion_options_to_run)
+
+        if self.data_interface_classes is not None:
+            self.validate_conversion_options(conversion_options=conversion_options_to_run)
 
         with make_or_load_nwbfile(
             nwbfile_path=nwbfile_path,


src/neuroconv/nwbconverter.py
```diff
@@ -60,15 +60,27 @@ def validate_source(cls, source_data: Dict[str, dict], verbose: bool = True):
...
+                name: data_interface(**source_data[name])
+                for name, data_interface in self.data_interface_classes.items()
+                if name in source_data
+            }
+
+        elif data_interfaces is not None:
+
+            self.data_interface_objects = {
+                f"{interface.__class__.__name__}{index}": interface for index, interface in enumerate(data_interfaces)
+            }
```
bendichter, line 80: I see you are using `index` here to ensure all of the DataInterfaces have unique names. I think it might be better to start without a number and then add numbers if you have more than one of the same type. I think that would be a bit too complicated for a dictionary comprehension though

bendichter: I also think it would be nice to allow the user to input data_interfaces as a dictionary

src/neuroconv/nwbconverter.py
```diff
@@ -60,15 +60,27 @@ def validate_source(cls, source_data: Dict[str, dict], verbose: bool = True):
...
+
+        else:
+            raise IOError("Need to pass either source_data or a list with initialized data interfaces")
```
bendichter, line 83: This would be a `ValueError`

bendichter: https://docs.python.org/3/library/exceptions.html#ValueError

docs/conversion_examples_gallery/combinations/recording_and_sorting.rst
```diff
@@ -0,0 +1,41 @@
+Conversion combining electrophysiological recordings and spike sorting data
```
bendichter, line 1:
```suggestion
spikeGLX + Phy
```

docs/conversion_examples_gallery/combinations/recording_and_sorting.rst
```diff
@@ -0,0 +1,41 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A common workflow is to record electrophysiological data and then extract spiking units from the stream. The following is an
+example of how to combine electrophysiological recordings and spike sorting data in the same conversion.
+
+Note that for this specific example were using SpikeGLXRecordingInterface (insert link) and PhySortingInterface (insert link)but any
```
bendichter, line 7:
```suggestion
Note that for this specific example were using :py:class:`~neuroconv.datainterfaces.ecephys.spikeglxrecordingingerface.SpikeGLXRecordingInterface` (insert link) and PhySortingInterface (insert link)but any
```

docs/conversion_examples_gallery/conversion_example_gallery.rst
```diff
@@ -80,3 +80,12 @@ Behavior
...
     DeepLabCut <behavior/deeplabcut>
     SLEAP <behavior/sleap>
+
+Examples of common interface combinations
+-----------------------------------------
+
+.. toctree::
+    :maxdepth: 1
+
+    Combine electrophysiological recordings and spike sorting <combinations/recording_and_sorting>
```
bendichter, line 90:
```suggestion
    spikeGLX & Phy <combinations/recording_and_sorting>
```

src/neuroconv/nwbconverter.py
```diff
@@ -60,15 +60,27 @@ def validate_source(cls, source_data: Dict[str, dict], verbose: bool = True):
...
         """Validate source_data against source_schema and initialize all data interfaces."""
         self.verbose = verbose
-        self._validate_source_data(source_data=source_data, verbose=self.verbose)
-        self.data_interface_objects = {
-            name: data_interface(**source_data[name])
-            for name, data_interface in self.data_interface_classes.items()
-            if name in source_data
-        }
+
+        if source_data is not None:
```
bendichter, line 67: we should make sure a user does not supply both `source_data` and `data_interfaces`

CodyCBakerPhD:
Also depends on where the distinction between 'gallery' and 'tutorial' is. I generally consider NWBConverter stuff to be tutorial level, so you could also port over these https://github.com/NeurodataWithoutBorders/nwb_hackathons/tree/main/HCK13_2022_Janelia/projects/neuroconv_tutorial

src/neuroconv/nwbconverter.py
```diff
@@ -166,9 +178,12 @@ def run_conversion(
 
         if conversion_options is None:
             conversion_options = dict()
+
         default_conversion_options = self.get_conversion_options()
         conversion_options_to_run = dict_deep_update(default_conversion_options, conversion_options)
-        self.validate_conversion_options(conversion_options=conversion_options_to_run)
+
+        if self.data_interface_classes is not None:
```
bendichter, line 185: Maybe it would be better to add the `self.data_interface_classes` in the init when data_interfaces are provided.

src/neuroconv/nwbconverter.py
h-mayorquin: I agree with this. My initial instinct was to have the interface variable name as the name. That is, if someone defined:

```
interface_phy = PhySortingInterface(folder_path=folder_path, verbose=False)
interface_spikeglx = SpikeGLXRecordingInterface(file_path=file_path, verbose=False)
```
And passed those to the NWBConverter as a list:
```
converter = NWBConverter(data_interfaces=[interface_spikeglx, interface_phy], verbose=False)
```

Then their names in the `self.data_interface_objects` should be `interface_phy` and `interface_spikeglx`. This will care of uniqueness of name for free and is also a natural reference that the user already knows because they defined themselves.  The problem with this is that it can only be done very easily after python=3.8, python 3.7 requires a hacky solution.

bendichter:
@h-mayorquin I like this idea! I think it would probably be easier for most users who don't need to use the advanced features of NWBConverter to do things like navigate conflicting metadata. @CodyCBakerPhD, what do you think?

src/neuroconv/nwbconverter.py
h-mayorquin: > I also think it would be nice to allow the user to input data_interfaces as a dictionary

Yep, this make sense. We can have a general variable called `data_interfaces` and users can pass either a dict or a list. If there is a list then we ourselves determine our names with the logic mentioned above.

src/neuroconv/nwbconverter.py
h-mayorquin: Thanks, this is no the first time that I get this wrong!

src/neuroconv/nwbconverter.py
h-mayorquin: Yep, probably we could use some of the new import functionality to build that on the fly so we have the workflow as similar as it was before. 

One important point to add though should do those checks by themselves though. This has been on the to-do list for a while but we have not done it. Check:
https://github.com/catalystneuro/neuroconv/issues/42

src/neuroconv/nwbconverter.py
h-mayorquin: Agree.

src/neuroconv/nwbconverter.py
bendichter: I don't think we'll need to import. If we have objects, we can just do `obj.__class__`

h-mayorquin: True!

docs/conversion_examples_gallery/combinations/recording_and_sorting.rst
```diff
@@ -0,0 +1,41 @@
...
+    >>> # Extract what metadata we can from the source files
+    >>> metadata = converter.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> tzinfo = tz.gettz("US/Pacific")
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific")).isoformat()
+    >>> metadata["NWBFile"].update(session_start_time=session_start_time)
+    >>>
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> nwbfile_path = f"test.nwb"
+    >>> nwfile = converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, metadata=metadata)
```
CodyCBakerPhD, line 41:
```suggestion
    >>> nwbfile = converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, metadata=metadata)
```

CodyCBakerPhD:
> @h-mayorquin I like this idea! I think it would probably be easier for most users who don't need to use the advanced features of NWBConverter to do things like navigate conflicting metadata. @CodyCBakerPhD, what do you think?

Conflicting metadata will still happen with this approach, it's just bypassing the immediate `super()` call for the `NWBConverter` that takes `source_data` and does the initialization on-the-fly.

The real benefit I see to this approach towards the `NWBConverter` attributes as dynamic properties is (and I think this ought to be included in this PR) to have a `setter` functionality for it which is usable even after the first initialization of that `NWBConverter` object. That is

```python
my_interface = IDKSomeInterface(...)
my_converter = MyNWBConverter(...)  # maybe on source_data even
my_converter.attach_interface(interface=my_interface)  # but only if interface is initialized in-memory
```

so long as this occurs prior to the main deployments `get_metadata`+`run_conversion` it's no problem. That way the `NWBConverter` class can be used anonymously instead of having to define a child class (which I grant is a bit more technical from a first-time programmer perspective, and somewhat unnecessary to make full use of NeuroConv), but also has separate pathways for instantiating certain interfaces using `source_data` (on disk usage) or post-init attachment (in-memory usage) depending on which is more convenient for which interfaces.


<details>
<summary>
To summarize, you've introduced an alternative to the classic</summary>

```python
from neuroconv import NWBConverter

MyConverterClass(NWBConverter):
    data_interface_classes = dict(...)

my_converter = MyConverterClass(source_data=dict(my_interface=dict(some_path="...")))
```
</details>

<details>
<summary>which is</summary>

```python
from neuroconv import NWBConverter
from neuroconv.datainterfaces import SomeInterface

my_interface = SomeInterface(some_path="...")
my_converter = NWBConverter(data_interfaces=[my_interface])
```
</details>

<details>
<summary>or if implementing my suggestion could even start with a blank slate</summary>

```python
from neuroconv import NWBConverter
from neuroconv.datainterfaces import SomeInterface

my_interface = SomeInterface(some_path="...")
my_converter = NWBConverter(data_interfaces=[])
my_converter.attach_interface(interface=my_interface)
```
</details>


<details>
<summary>The latter could also allow conditional looping like</summary>

```python
from neuroconv import NWBConverter

my_converter = NWBConverter(data_interfaces=[])

for SomeInterface, some_path in some_items_loop:
    if some_path.exists():
        this_interface = SomeInterface(some_path)
        my_converter.attach_interface(interface=my_interface)
```

which would be pretty convenient.
</details>

CodyCBakerPhD:
Still though, for the purposes of calling it a gallery or a tutorial, a gallery would imply that we would have more than one showcase of this, in the upper limit the outer product of all compatible modalities (possible combinations of recording+sorting or imaging+segmentation). That sounds like overkill, but tutorials for the different ways of using an NWBConverter are necessary and will likely get added at the upcoming docathon.

src/neuroconv/nwbconverter.py
h-mayorquin: Done in the last commit.

src/neuroconv/nwbconverter.py
h-mayorquin: I implemented both of these things in the last commit.

h-mayorquin:
> Still though, for the purposes of calling it a gallery or a tutorial, a gallery would imply that we would have more than one showcase of this, in the upper limit the outer product of all compatible modalities (possible combinations of recording+sorting or imaging+segmentation). That sounds like overkill, but tutorials for the different ways of using an NWBConverter are necessary and will likely get added at the upcoming docathon.

@CodyCBakerPhD 
My idea is to add examples of common workflows. Right now I have in mind the following conversion examples:
1) Recording + Sorting
2) Imaging + Segmentation
3) Recording + Sorting + PoseEstimation (for an example with three interfaces)

And emphasize that their respective pieces are interchangeably and can be chosen piece-wise from around the conversion gallery.

h-mayorquin:
> > @h-mayorquin I like this idea! I think it would probably be easier for most users who don't need to use the advanced features of NWBConverter to do things like navigate conflicting metadata. @CodyCBakerPhD, what do you think?
> 
> Conflicting metadata will still happen with this approach, it's just bypassing the immediate `super()` call for the `NWBConverter` that takes `source_data` and does the initialization on-the-fly.

I understood that Ben meant that the common NWBConverter workflow (not this) avoids some conflicting metadata problem. That is, this approach is less robust.

CodyCBakerPhD:
> I understood that Ben meant that the common NWBConverter workflow (not this) avoids some conflicting metadata problem. That is, this approach is less robust.

How? By the time you hit the metadata steps the internals of the converter should appear exactly the same

h-mayorquin:
> ```python
> my_interface = IDKSomeInterface(...)
> my_converter = MyNWBConverter(...)  # maybe on source_data even
> my_converter.attach_interface(interface=my_interface)  # but only if interface is initialized in-memory
> ```
> 
> so long as this occurs prior to the main deployments `get_metadata`+`run_conversion` it's no problem. That way the `NWBConverter` class can be used anonymously instead of having to define a child class (which I grant is a bit more technical from a first-time programmer perspective, and somewhat unnecessary to make full use of NeuroConv), but also has separate pathways for instantiating certain interfaces using `source_data` (on disk usage) or post-init attachment (in-memory usage) depending on which is more convenient for which interfaces.

@CodyCBakerPhD 

I think that those two uses cases that your point in the end are indeed not accomplished by what I have here. But the goal I have here is simpler: I want to make it as easy as possible for first time users to combine multiple interfaces and I feel that passing arguments to a function is what most scientists / PhDs / Post-docs  are familiar and find natural when they build scripts (that is scripting programming like the one in Matlab is the main paradigm out there I feel). Instantiating an object and then modifying it dynamically according to context seems to me more like a power user functionality.  Am I missing something?

That said, I don't see why we can't have both and I volunteer to implement the dynamic setter in a separate PR after this if @bendichter agrees with that. 

In brief, I would like to separate a simple initialization feature for non-power users focused mainly on the conversion gallery (as proposed in this PR) from a more ambitious feature of a dynamic setter PR-wise. I want to do this for testing and design purposes as they have different scope and use cases I feel.

bendichter:
I don't see the point of `my_converter.attach_interface`. Something like this:

```python
for SomeInterface, some_path in some_items_loop:
    if some_path.exists():
        this_interface = SomeInterface(some_path)
        my_converter.attach_interface(interface=my_interface)
```

could instead be implemented as:

```python
interfaces = []
for SomeInterface, some_path in some_items_loop:
    if some_path.exists():
        interfaces.append(SomeInterface(some_path))
NWBConverter(data_interfaces=interfaces)
```

docs/conversion_examples_gallery/combinations/recording_and_sorting.rst
```diff
@@ -0,0 +1,41 @@
...
+Note that for this specific example were using :py:class:`~neuroconv.datainterfaces.ecephys.spikeglxrecordingingerface.SpikeGLXRecordingInterface` and PhySortingInterface (insert link)but any
+of the other examples in our conversion gallery (insert link to recorders and sorters) can be combined in this way.
+
+.. code-block:: python
+
+    >>> from datetime import datetime
+    >>> from dateutil import tz
+    >>> from pathlib import Path
+    >>> from neuroconv.datainterfaces import SpikeGLXRecordingInterface
+    >>> from neuroconv.datainterfaces import PhySortingInterface
```
bendichter, line 16:
```suggestion
    >>> from neuroconv.datainterfaces import SpikeGLXRecordingInterface, PhySortingInterface
```

docs/conversion_examples_gallery/combinations/recording_and_sorting.rst
```diff
@@ -0,0 +1,41 @@
...
+    >>>  # Now that we have defined the two interfaces we pass them to the NWBConverter which will coordinate the
+    >>>  # concurrent conversion of the data
+    >>> converter = NWBConverter(data_interfaces=[interface_spikeglx, interface_phy], verbose=False)
+    >>>
+    >>> # Extract what metadata we can from the source files
+    >>> metadata = converter.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> tzinfo = tz.gettz("US/Pacific")
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific")).isoformat()
+    >>> metadata["NWBFile"].update(session_start_time=session_start_time)
```
bendichter, line 37: I don't think you need this, since the session start time should be fetched from spikeGLX

docs/conversion_examples_gallery/conversion_example_gallery.rst
```diff
@@ -80,3 +80,12 @@ Behavior
...
     SLEAP <behavior/sleap>
+
+Examples of common interface combinations
+-----------------------------------------
+
+.. toctree::
+    :maxdepth: 1
+
+    spikeGLX & Phy <combinations/recording_and_sorting>
+    SLEAP <behavior/sleap>
```
bendichter, line 91:
```suggestion
```

src/neuroconv/nwbconverter.py
```diff
@@ -39,8 +40,7 @@ def get_source_schema(cls):
             source_schema["properties"].update({interface_name: unroot_schema(data_interface.get_source_schema())})
         return source_schema
 
-    @classmethod
-    def get_conversion_options_schema(cls):
+    def get_conversion_options_schema(self):
```
bendichter, line 43: Is changing this from a classmethod really necessary? I'd rather not do that

src/neuroconv/nwbconverter.py
```diff
@@ -60,15 +60,44 @@ def validate_source(cls, source_data: Dict[str, dict], verbose: bool = True):
...
+                interface_names = dict()
+                for interface in data_interfaces:
+                    counter[interface] += 1
+                    unique_signature = f"{counter[interface]:03}" if total_counts[interface] > 1 else ""
+                    interface_name = f"{interface.__class__.__name__}{unique_signature}"
+                    interface_names[interface] = interface_name
+
+                self.data_interface_objects = {name: interface for interface, name in interface_names.items()}
+            else:
+                self.data_interface_objects = data_interfaces
```
bendichter, line 95:
```suggestion
            elif isinstance(data_interfaces, dict):
                self.data_interface_objects = data_interfaces
            else:
                ValueError("data_interfaces musts be a list or a dict")
```

src/neuroconv/nwbconverter.py
```diff
@@ -60,15 +60,44 @@ def validate_source(cls, source_data: Dict[str, dict], verbose: bool = True):
...
+
+            self.data_interface_classes = {name: interface.__class__ for interface, name in interface_names.items()}
```
bendichter, line 97: The keys in this dict need to match the keys of the `data_interface_objects` dict

src/neuroconv/nwbconverter.py
```diff
@@ -60,15 +60,44 @@ def validate_source(cls, source_data: Dict[str, dict], verbose: bool = True):
...
+                counter = {interface: 0 for interface in data_interfaces}
+                total_counts = Counter(data_interfaces)
+                interface_names = dict()
+                for interface in data_interfaces:
+                    counter[interface] += 1
+                    unique_signature = f"{counter[interface]:03}" if total_counts[interface] > 1 else ""
+                    interface_name = f"{interface.__class__.__name__}{unique_signature}"
+                    interface_names[interface] = interface_name
+
+                self.data_interface_objects = {name: interface for interface, name in interface_names.items()}
```
bendichter, line 93:
```suggestion
                    self.data_interface_objects[interface_name] = interface
```

CodyCBakerPhD:
@h-mayorquin We should have a bunch of core units tests for these new usages of the NWBConverter as well instead of relying purely on doctests

docs/conversion_examples_gallery/conversion_example_gallery.rst
h-mayorquin: Done in the last commit.

docs/conversion_examples_gallery/combinations/recording_and_sorting.rst
h-mayorquin: Done in the last commit.

src/neuroconv/nwbconverter.py
h-mayorquin: The problem is making line 52 reference the instance attribute instead of the class one. Maybe there is another way? I will think about it.

Another option is to set the class attribute data_interface_classes from inside the init . Is there a standard way to do that?

src/neuroconv/nwbconverter.py
h-mayorquin: Done in the last commit.

src/neuroconv/nwbconverter.py
h-mayorquin: Done.

docs/conversion_examples_gallery/conversion_example_gallery.rst
h-mayorquin: Done.

h-mayorquin:
> @h-mayorquin We should have a bunch of core units tests for these new usages of the NWBConverter as well instead of relying purely on doctests

@CodyCBakerPhD I don't see tests for the old usage of it either. What kind of tests do you have on mind?

CodyCBakerPhD:
> I don't see tests for the old usage of it either. What kind of tests do you have on mind?

Then both would be good to ensure equivalent behavior - ensure that the attribute structures match between both methods after initializations using simple mock interfaces as input

src/neuroconv/nwbconverter.py
bendichter: why can't you just do `cls.data_interface_classes = ...`

CodyCBakerPhD: We discussed this in our meeting, this is easily fixable via a small `__new__` override to allow communication between these attributes for the instance of the class, very similar to how the `BaseExtractorInterface` works.

I also pointed out that an alternative approach here would be to have a separate but related `NWBConverter` class (possibly `DynamicNWBConverter`) that takes the instantiated list of data interfaces as input and adjusts any of the methods as required by the adjusted workflow, not dissimilar to this old PR of a utility-like function for generating such a class from a list of data interface classes: https://github.com/catalystneuro/nwb-conversion-tools/pull/110/files

h-mayorquin:
@CodyCBakerPhD I added some tests now. This requires some modifications on #167 I feel but now, at this point in time, the tests are not failing. I think there is something fishy going on.

src/neuroconv/nwbconverter.py
h-mayorquin: @bendichter It can be done. Check the last commit as I have reverted this. 

My concern with updating class attributes is that if you initialize another `NWBConverter` instance the updates of one will propagate to the other entangling them in hard to predict (and debug) ways.

```
interface1 = NWBConverter()  # This modifies the class attributes
interface2 = NWBConverter()  # This modifies the class attributes as well

interface1.run_conversion() #  <- here the interface1 will have the class attributes as set by interface2 
```
If this is an exotic concern that you think is unlikely to affect us then we can keep the code as it is (check the last commit). If on the other hand this is a concern after thinking more about it I think that I favor the alternative proposed by @CodyCBakerPhD of another class that works as a NWBConverter child factory.  We can try to emulate the [similar process in scikit-learn](https://scikit-learn.org/stable/modules/compose.html) as much as possible to leverage their data model as well 

What do you think?

h-mayorquin: I really want to avoid using meta programming as I think is harder to do introspection and therefore debug. A simple way to:
1) Keep the code `NWBConverter` without breaking backwards compatibility.
2) Provide the functionality disused in this PR
3) Not using metaprogramming .

Is to introduce use something like the following class instead:

```python
class InterfacePipeline(NWBConverter):

    def get_conversion_options_schema(self):
        """Compile conversion option schemas from each of the data interface classes."""
        conversion_options_schema = get_base_schema(
            root=True,
            id_="conversion_options.schema.json",
            title="Conversion options schema",
            description="Schema for the conversion options",
            version="0.1.0",
        )
        for interface_name, data_interface in self.data_interface_classes.items():
            conversion_options_schema["properties"].update(
                {interface_name: unroot_schema(data_interface.get_conversion_options_schema())}
            )


    def __init__(self, data_interfaces, verbose=True):
        self.verbose = verbose
        if isinstance(data_interfaces, list):
            # Create unique names for each interface
            counter = {interface.__class__.__name__: 0 for interface in data_interfaces}
            total_counts = Counter([interface.__class__.__name__ for interface in data_interfaces])
            self.data_interface_objects = dict()
            for interface in data_interfaces:
                class_name = interface.__class__.__name__
                counter[class_name] += 1
                unique_signature = f"{counter[class_name]:03}" if total_counts[class_name] > 1 else ""
                interface_name = f"{class_name}{unique_signature}"
                self.data_interface_objects[interface_name] = interface
        elif isinstance(data_interfaces, dict):
            self.data_interface_objects = data_interfaces
            
            
        self.data_interface_classes = {
                name: interface.__class__ for name, interface in self.data_interface_objects.items()
            }
```

I tested this locally and it works. Critically, we need to make the class method `get_conversion_options_schema` and instance method here as the class method relies on the class attribute. The other private method  to validate the source schema is not required since the instances of the interfaces are already initialized. 

You can find a complete script with this here:
https://gist.github.com/h-mayorquin/72b83aaa9c19375862819651e64000c5

src/neuroconv/nwbconverter.py
```diff
@@ -60,15 +61,48 @@ def validate_source(cls, source_data: Dict[str, dict], verbose: bool = True):
...
+            elif isinstance(data_interfaces, dict):
+                self.data_interface_objects = data_interfaces
+
+            else:
+                ValueError("data_interfaces musts be a list or a dict")
+
+            # Build the data interface classess on the move
+            type(self).data_interface_classes = {
+                name: interface.__class__ for name, interface in self.data_interface_objects.items()
+            }
```
bendichter, line 103: I don't want to overwrite this class attribute. Why can't we just do 

```python
self.data_interface_classes = {
                name: interface.__class__ for name, interface in self.data_interface_objects.items()
            }
```

CodyCBakerPhD: I thought he was opting for a different approach now: https://github.com/catalystneuro/neuroconv/pull/164#discussion_r975901112?

h-mayorquin: @bendichter 
If you assign `data_interface_classes` as an instance attribute instead of as a class one then line 52 in the following call fails:
https://github.com/catalystneuro/neuroconv/blob/45ade589959092f3e2a9942890b1a7b1a3c70172/src/neuroconv/nwbconverter.py#L42-L56

The reason is that class methods don't have access to instance attributes and `cls.data_interface_classes.items()` is `None`.

src/neuroconv/nwbconverter.py
h-mayorquin: Ok, I implemented this in #169.

src/neuroconv/nwbconverter.py
h-mayorquin: The other version is implemented in #169

PR #169, opened 2022-09-28
Combine simple examples in gallery with NWBConverter 

### LINKED ISSUE #165: Simplify conversions with multiple interfaces
Right now we have the conversion gallery which as simple as possible illustrates how to do conversion with a single interface:
https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/conversion_example_gallery.html#extracellular-electrophysiology

On the other hand, our workwflow for handling conversions with multiple interfaces relies on the `NWBConverter` object for which we also have another specific tutorial:
https://neuroconv.readthedocs.io/en/main/user_guide/nwbconverter.html

In my view, the use of the `NWBConverter` is not as simple as it could be. To be specific, two complexities that I see to the end user are the following:
1) They need to define a class themselves and inherit from the converter
2)  To define all the properties they need to use a structure of nested dictionaries for the source data, conversion options and metadata. 

I feel that this creates an unnecessary gap in complexity between A single interface conversion and a multiple interface conversion. Moreover, the workflow for multiple interfaces does not build on the single interface one which seems like a lost purpose. Concretely, the single interface conversions rely on their specific interfaces for conversion whereas the multiple ones rely on the NWBConverter object. **It would be great if we could simplify the step from single to multiple interface and that the multiple interface conversions built on what we have in the conversion gallery already**. I want to propose some solutions that go in this direction:

1. The first solution is to adapt the `NWBConverter` object to take as an input previously initialized interfaces. This allows us to mostly rely on the machinery that we have already in place while at the same time it allows users to pass their already initialized interfaces (that they copy-paste from the conversion gallery) to the NWBConveter object to build more complex pipelines. #164 shows a prototype for this.
2. @bendichter has mentioned a couple of times that he takes inspiration from `scikit-learn` and I think we can follow the same course here. We could use something like their [pipeline objects ](https://scikit-learn.org/stable/modules/compose.html) for users to concatenate conversions. Of course, the pipelines would rely on `NWBConverter` behind the scenes and this is just a matter of simplifying the interaction with the object. The advantage of this approach is that it relies on another widely used data model that is out there in a popular library so we can leverage this knowledge from the community.
3. Since we introduced contexts to handle writing most of our conversion can and some do return an nwb-file with the data already attached to them. We could leverage this capability and instruct the users that they can chain their single interface conversions to attach data from other interfaces to the same file.

Example of pipeline:
```
build nwb_file with pynwb or by running the first conversion with `nwbfile_path` instead.  # Step 1
nwbfile = interface1.run_conversion(nwbfile)  # Step 2 
nwbfile = interface2.run_conversion(nwbfile)  # Step 3 
save the file to disk   # Step 4
```

The drawback of this last approach is that we lost all the work that we did with the NWBConverter object. Moreover, we have less control of how the data is combined and therefore is harder to test.

h-mayorquin:
I am re-doing #164 for an easier to read diff after the modifications outlined in the following comments: 

https://github.com/catalystneuro/neuroconv/pull/164#discussion_r975901112

This should close #165 and also close #164

### DIFF
--- a/docs/conversion_examples_gallery/combinations/ecephys_pose_estimation.rst
+++ b/docs/conversion_examples_gallery/combinations/ecephys_pose_estimation.rst
@@ -41,4 +41,7 @@ For this specific example were are combining a OpenEphys recording with KiloSort
     >>>
     >>> # Choose a path for saving the nwb file and run the conversion
     >>> nwbfile_path = f"{path_to_save_nwbfile}"
-    >>> nwbfile_out = converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, metadata=metadata)
+    >>> converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, metadata=metadata)
+    root pynwb.file.NWBFile ...
+    >>> Path(nwbfile_path).is_file()
+    True
--- a/docs/conversion_examples_gallery/combinations/ecephys_pose_estimation.rst
+++ b/docs/conversion_examples_gallery/combinations/ecephys_pose_estimation.rst
@@ -13,7 +13,7 @@ For this specific example were are combining a OpenEphys recording with KiloSort
     >>> from datetime import datetime
     >>> from dateutil import tz
     >>> from pathlib import Path
-    >>> from neuroconv import NWBConverterPipe
+    >>> from neuroconv import ConverterPipe
     >>> from neuroconv.datainterfaces import BlackrockRecordingInterface, KiloSortSortingInterface, SLEAPInterface
     >>>
     >>>
@@ -29,9 +29,9 @@ For this specific example were are combining a OpenEphys recording with KiloSort
     >>> file_path = BEHAVIOR_DATA_PATH / "sleap" / "predictions_1.2.7_provenance_and_tracking.slp"
     >>> interface_sleap = SLEAPInterface(file_path=file_path, verbose=False)
     >>>
-    >>>  # Now that we have defined the two interfaces we pass them to the NWBConverterPipe which will coordinate the
+    >>>  # Now that we have defined the two interfaces we pass them to the ConverterPipe which will coordinate the
     >>>  # concurrent conversion of the data
-    >>> converter = NWBConverterPipe(data_interfaces=[interface_blackrock, interface_kilosort, interface_sleap], verbose=False)
+    >>> converter = ConverterPipe(data_interfaces=[interface_blackrock, interface_kilosort, interface_sleap], verbose=False)
     >>>
     >>> # Extract what metadata we can from the source files
     >>> metadata = converter.get_metadata()
--- a/docs/conversion_examples_gallery/combinations/ecephys_pose_estimation.rst
+++ b/docs/conversion_examples_gallery/combinations/ecephys_pose_estimation.rst
@@ -29,14 +29,13 @@ For this specific example were are combining a OpenEphys recording with KiloSort
     >>> file_path = BEHAVIOR_DATA_PATH / "sleap" / "predictions_1.2.7_provenance_and_tracking.slp"
     >>> interface_sleap = SLEAPInterface(file_path=file_path, verbose=False)
     >>>
-    >>>  # Now that we have defined the two interfaces we pass them to the ConverterPipe which will coordinate the
-    >>>  # concurrent conversion of the data
+    >>> # Now that we have defined the two interfaces we pass them to the ConverterPipe which will coordinate the
+    >>> # concurrent conversion of the data
     >>> converter = ConverterPipe(data_interfaces=[interface_blackrock, interface_kilosort, interface_sleap], verbose=False)
     >>>
     >>> # Extract what metadata we can from the source files
     >>> metadata = converter.get_metadata()
     >>> # For data provenance we add the time zone information to the conversion
-    >>> tzinfo = tz.gettz("US/Pacific")
     >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific")).isoformat()
     >>> metadata["NWBFile"].update(session_start_time=session_start_time)
     >>>
--- a/docs/conversion_examples_gallery/combinations/ecephys_pose_estimation.rst
+++ b/docs/conversion_examples_gallery/combinations/ecephys_pose_estimation.rst
@@ -1,0 +1,44 @@
+Electrophysiology and Behavior
+------------------------------
+
+This example showcases a conversion that combines two modalities of data electrophysiology and behavior in the form of pose estimation.
+For this specific example were are combining a OpenEphys recording with KiloSort sorting results and PoseEstimation from sleap using the
+:py:class:`~neuroconv.datainterfaces.ecephys.blackrock.blackrockdatainterface.BlackrockRecordingInterface`,
+:py:class:`~neuroconv.datainterfaces.ecephys.kilosort.kilosortdatainterface.KiloSortSortingInterface`, and
+:py:class:`~neuroconv.datainterfaces.behavior.sleap.sleapdatainterface.SLEAPInterface`. classes.
+
+.. code-block:: python
+
+
+    >>> from datetime import datetime
+    >>> from dateutil import tz
+    >>> from pathlib import Path
+    >>> from neuroconv.datainterfaces import BlackrockRecordingInterface, KiloSortSortingInterface, SLEAPInterface
+    >>>
+    >>> from neuroconv import NWBConverter
+    >>>
+    >>> file_path = f"{ECEPHY_DATA_PATH}/blackrock/FileSpec2.3001.ns5"
+    >>> # Change the file_path to the location in your system
+    >>> interface_blackrock = BlackrockRecordingInterface(file_path=file_path, verbose=False)
+    >>>
+    >>> folder_path = f"{ECEPHY_DATA_PATH}/phy/phy_example_0"
+    >>> # Change the folder_path to the location of the data in your system
+    >>> interface_kilosort = KiloSortSortingInterface(folder_path=folder_path, verbose=False)
+    >>>
+    >>> # Change the file_path so it points to the slp file in your system
+    >>> file_path = BEHAVIOR_DATA_PATH / "sleap" / "predictions_1.2.7_provenance_and_tracking.slp"
+    >>> interface_sleap = SLEAPInterface(file_path=file_path, verbose=False)
+    >>>
+    >>> converter = NWBConverter(data_interfaces=[interface_blackrock, interface_kilosort, interface_sleap], verbose=False)
+    >>>
+    >>> # Extract what metadata we can from the source files
+    >>> metadata = converter.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> tzinfo = tz.gettz("US/Pacific")
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific")).isoformat()
+    >>> metadata["NWBFile"].update(session_start_time=session_start_time)
+    >>>
+    >>> # Choose a path for saving the nwb file and run the conversion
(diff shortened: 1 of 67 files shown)


src/neuroconv/nwbconverter.py
```diff
@@ -184,3 +183,42 @@ def run_conversion(
                 )
 
         return nwbfile_out
+
+
+class NWBConverterPipe(NWBConverter):
+    """Takes pre-initialized interfaces as arguments to build an NWBConverter class"""
+
+    def get_conversion_options_schema(self):
```
CodyCBakerPhD, line 191: Should maybe also do

```python
def get_source_schema():
     raise NotImplementedError("Informative error about why this doesn't make sense here...")
```

to avoid having a confusing method available for the class?

h-mayorquin: You can get the source schema if you want.

src/neuroconv/nwbconverter.py
```diff
@@ -184,3 +183,42 @@ def run_conversion(
...
+            description="Schema for the conversion options",
+            version="0.1.0",
+        )
+        for interface_name, data_interface in self.data_interface_classes.items():
+            conversion_options_schema["properties"].update(
+                {interface_name: unroot_schema(data_interface.get_conversion_options_schema())}
+            )
+        return conversion_options_schema
+
+    def __init__(self, data_interfaces, verbose=True):
```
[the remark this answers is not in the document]
h-mayorquin: Thanks, I added annotation for both dict and list.

src/neuroconv/nwbconverter.py
CodyCBakerPhD: But how? That's a class method as well and won't mean anything since the `__init__` here doesn't take `source_data`, so what is the meaning?

src/neuroconv/nwbconverter.py
```diff
@@ -184,3 +183,42 @@ def run_conversion(
...
+            description="Schema for the conversion options",
+            version="0.1.0",
+        )
+        for interface_name, data_interface in self.data_interface_classes.items():
+            conversion_options_schema["properties"].update(
+                {interface_name: unroot_schema(data_interface.get_conversion_options_schema())}
+            )
+        return conversion_options_schema
+
+    def __init__(self, data_interfaces: Union[List, Dict], verbose=True):
```
CodyCBakerPhD, line 206:
```suggestion
    def __init__(self, data_interfaces: Union[List[BaseDataInterface], Dict[str, BaseDataInterface]], verbose=True):
```

To make content of iterables clearer

src/neuroconv/nwbconverter.py
h-mayorquin: Oh, what I mean is that we could have the equivalent method that returns the source schema as that is extracted by the data classes which are available here as well.  The problem will be `_validate_source_data` that requires `source_data`.

src/neuroconv/nwbconverter.py
```diff
@@ -184,3 +183,42 @@ def run_conversion(
                 )
 
         return nwbfile_out
+
+
+class NWBConverterPipe(NWBConverter):
+    """Takes pre-initialized interfaces as arguments to build an NWBConverter class"""
```
CodyCBakerPhD, line 189: Structure of the class definitely looks better, question now is what name to choose... other options might be `DynamicNWBConverter` (from https://github.com/catalystneuro/nwb-conversion-tools/pull/110/files#diff-91a2f17a5200d62cf54bb3dc0051c2df6c96650f247faf2bc514327211840919R138), `NWBConverterContainer` (reflecting treatment of acting as a bucket/wrapper of already created objects), or any others we can think of to consider...

Any thoughts @bendichter?

src/neuroconv/nwbconverter.py
CodyCBakerPhD: Yeah I mean you *could* have it defined as an instance method that tells you what the input keyword argument *were* for the already passed interfaces, it just seems redundant at that point though since the person obviously knows how to instantiate the interfaces to begin with

src/neuroconv/nwbconverter.py
h-mayorquin: My argument for something along the lines of Pipe is that a pipeline is a common metaphor in the scientific python community for a series of processing operations. See [scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html) and [pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.pipe.html).

I think @CodyCBakerPhD mention that there were some overlap with some objects in Datalad.

I don't think the word dynamic is specific enough. As I see it, it makes reference to the python data model (as classes are created dynamically, meta-programming, etc.) but I don't see what is the utility of communicating that to the end users. 

This is meant to be a function that end users access frequently and I think it should be named by their function. Maybe there is something better than pipe?

src/neuroconv/nwbconverter.py
CodyCBakerPhD: I'd just recommend removing it through, w.r.t. this new subclass any method having to do with source data/schema just doesn't make sense

src/neuroconv/nwbconverter.py
CodyCBakerPhD: 100% agree that a NWBConverter, as a class structure, is analogous to a data pipeline in the sense that you specify which data interfaces to use (which themselves are pipelines of which types of conversion protocols to use) but my point is that is true even of the base NWBConverter class as well - this new subclass does not change or introduce that base behavior, so my philosophy for choosing a name for a child class like this would be to find the one-two simplest word(s) that minimally differentiates it from the parent and summarizes that difference

h-mayorquin: I am not sure I agree with that naming philosophy. I want to have a name that is useful and easy to remember for users. I don't think the goal of the name here should be to describe how this object is different from the NWBConverter (parent) as I don't think that that information is useful to the user.

I still think that `InterfacePipeline` is a good name as it refers to both the atomic pieces of the conversion (the interfaces) and it states what it does with them. I changed it here to `NWBConverterPipe` because of the ambiguity concerns that you mentioned, that is, the mention of `NWBConverter` was accidental as its goal was to disambiguate.

Other possibilities are just calling it  `Pipe` like scikit-learn or `ConversionPipe` if we want to be more specific.

src/neuroconv/nwbconverter.py
h-mayorquin: The only thing that I was saying is that `get_source_schema` makes sense here (even it is not useful). The one that does not make sense is `_validate_source_data` as that requires `source_data`. 

I agree it would be better to remove it. Any suggestions for how to go about? To me this indicates that there should be a third class from were both `NWBConveter` and the class proposed here both inherit from. Maybe there is a simpler way.

src/neuroconv/nwbconverter.py
h-mayorquin: Changed it to `ConverterPipe` to avoid a possible confusion with the `NWBConverter`. Other possible name is `InterfacePipe`.

h-mayorquin:
Ok, I just discussed this with @bendichter, it should be ready to go as soon as we pick a name that we feel satisfied with @CodyCBakerPhD .

docs/conversion_examples_gallery/combinations/ecephys_pose_estimation.rst
```diff
@@ -0,0 +1,45 @@
...
+    >>> interface_sleap = SLEAPInterface(file_path=file_path, verbose=False)
+    >>>
+    >>>  # Now that we have defined the two interfaces we pass them to the ConverterPipe which will coordinate the
+    >>>  # concurrent conversion of the data
+    >>> converter = ConverterPipe(data_interfaces=[interface_blackrock, interface_kilosort, interface_sleap], verbose=False)
+    >>>
+    >>> # Extract what metadata we can from the source files
+    >>> metadata = converter.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> tzinfo = tz.gettz("US/Pacific")
```
bendichter, line 39:
```suggestion
```

docs/conversion_examples_gallery/combinations/ecephys_pose_estimation.rst
```diff
@@ -0,0 +1,45 @@
...
+    >>> folder_path = f"{ECEPHY_DATA_PATH}/phy/phy_example_0"
+    >>> # Change the folder_path to the location of the data in your system
+    >>> interface_kilosort = KiloSortSortingInterface(folder_path=folder_path, verbose=False)
+    >>>
+    >>> # Change the file_path so it points to the slp file in your system
+    >>> file_path = BEHAVIOR_DATA_PATH / "sleap" / "predictions_1.2.7_provenance_and_tracking.slp"
+    >>> interface_sleap = SLEAPInterface(file_path=file_path, verbose=False)
+    >>>
+    >>>  # Now that we have defined the two interfaces we pass them to the ConverterPipe which will coordinate the
+    >>>  # concurrent conversion of the data
```
bendichter, line 33:
```suggestion
    >>> # Now that we have defined the two interfaces we pass them to the ConverterPipe which will coordinate the
    >>> # concurrent conversion of the data
```

docs/conversion_examples_gallery/combinations/tiff_and_suite2p.rst
```diff
@@ -0,0 +1,37 @@
...
+    >>> interface_suit2p = Suite2pSegmentationInterface(folder_path=folder_path, verbose=False)
+    >>>
+    >>> # Now that we have defined the two interfaces we pass them to the ConverterPipe which will coordinate the
+    >>> # concurrent conversion of the data
+    >>> converter = ConverterPipe(data_interfaces=[interface_tiff, interface_suit2p], verbose=False)
+    >>>
+    >>> # Extract what metadata we can from the source files
+    >>> metadata = converter.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> tzinfo = tz.gettz("US/Pacific")
```
bendichter, line 31:
```suggestion
```

src/neuroconv/nwbconverter.py
CodyCBakerPhD: `ConverterPipe` is fine

h-mayorquin:
I think this is ready now that that @CodyCBakerPhD and I have reached common understanding concerning naming. @bendichter, anything else needed?

docs/conversion_examples_gallery/combinations/tiff_and_suite2p.rst
```diff
@@ -0,0 +1,36 @@
...
+    >>>
+    >>> # Extract what metadata we can from the source files
+    >>> metadata = converter.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))
+    >>> metadata["NWBFile"].update(session_start_time=session_start_time)
+    >>>
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> nwbfile_path = f"{path_to_save_nwbfile}"
+    >>> nwbfile_out = converter.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, metadata=metadata)
```
CodyCBakerPhD, line 36: The name `nwbfile_out` here differs from `nwbfile` in the other gallery, might want to standardize for consistency. 

Also, are we sure we want to demonstrate the return value of `converter.run_conversion` here, since the workflow seems to instead focus on writing the file to disk? Might confuse the user in this case to show the file as being 'returned' in-memory, such as confusion about why making changes to the `nwbfile` object doesn't make direct changes to the file (of course we can explain all of this in other tutorials, just saying for the sake of simplicity we w

h-mayorquin: I agree with you. I will change that.The only thing on the way is that ellipsis need to be used as the NWBInterface, in opposition to the interfaces, does return a in memory file and therefore `run_conversion` does throw something out to the standard output.

CodyCBakerPhD (approved):

Issue #43, opened 2022-05-05
Automatic metadata extraction for remaining RecordingInterfaces


CodyCBakerPhD:
@h-mayorquin Please add check boxes for what interfaces don't have much metadata extraction as of yet, thanks!

h-mayorquin:
I will wait till I get to partake on an OpenEphys conversion to work into this.

CodyCBakerPhD:
@h-mayorquin That could be a while - @weiglszonja did a lot of work on that in the last year, maybe consult with here and then we can call this done?

h-mayorquin:
Let' close this. I was thinking that I would had a OpenEphys project this August so I could finish the ecephys part but that will have to wait. I think we can open some more specific issues in the future.

PR #170, opened 2022-10-03
[neuralynx] exploit metadata available from neo.neuralynxrawio 

JuliaSprenger:
Exploit metadata available from neo NeuralynxIO file headers and annotations. This PR is an update of https://github.com/catalystneuro/nwb-conversion-tools/pull/596
Contributes to https://github.com/catalystneuro/neuroconv/issues/43

## Motivation

Recent versions of Neo provide access to header metadata from recording files. There is no need to re-read and interpret the headers twice.

## Checklist


src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -105,25 +39,73 @@ def initialize_in_spikeextractors(self, folder_path, verbose):
...
+        neo_metadata = {k: v for k, v in neo_metadata.items() if not k.lower().startswith("dsp")}
+
+        # map Neuralynx metadata to NWB
+        nwb_metadata = {"NWBFile": {}, "Ecephys": {"Device": []}}
+        if "session_id" in neo_metadata:
+            session_id = neo_metadata.pop("session_id", None)
+            if session_id is not None:
+                nwb_metadata["NWBFile"]["session_id"] = neo_metadata.pop("session_id")
+        if "SessionUUID" in neo_metadata:
+            nwb_metadata["NWBFile"]["identifier"] = neo_metadata.pop("SessionUUID")
```
bendichter, line 96: SessionUUID should map to session_id. "identifer" is generated from uuid4() by default and we can use that.

bendichter:
@JuliaSprenger this looks great! Can you add tests please?

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
JuliaSprenger: I reverted back to storing the Neuralynx UUID as session_id

JuliaSprenger:
This PR will be simplified based on #174

JuliaSprenger:
The coverage seems to decrease as code lines intended for the next neo version are not covered and test files are not downloaded as they are cached.

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
```diff
@@ -13,56 +10,88 @@
...
+    assert file_metadata["session_start_time"] == datetime.datetime(2017, 2, 16, 17, 56, 4)
+    # the session id can only be extracted from the file headers exposed for neo > 0.11.0
+    if version.LooseVersion(neo.__version__) > version.LooseVersion("0.11.0"):
+        assert file_metadata["session_id"] == "d8ba8eef-8d11-4cdc-86dc-05f50d4ba13d"
+
+    # Metadata extracted directly from file header (neo >= 0.11)
+    assert '"FileType": "NCS"' in file_metadata["notes"]
+    assert '"recording_closed": "2017-02-16 18:01:18"' in file_metadata["notes"]
+    assert '"ADMaxValue": "32767"' in file_metadata["notes"]
+    # the sampling rate and device is part of the file headers exposed for neo > 0.11.0
```
h-mayorquin, line 29: Is this > or >= ? 
I am asking because it seems that the latest dependency of spikeinterface is `neo>=0.11` . If this only works after neo > 0.11 then we should add this is a requirement. 

I also thing that the if statement is probably not necessary in the testing files as they have a a explicit environment in which they are meant to be run, maybe I am missing something, what do you think @CodyCBakerPhD ?

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -118,3 +83,109 @@ def __init__(self, folder_path: FolderPathType, sampling_frequency: float = None
...
+    # use metadata provided as array_annotations for each channel (neo version <=0.11.0)
+    else:
+        headers = []
+        neo_annotations = neo_reader.raw_annotations
+        for stream_annotations in neo_annotations["blocks"][0]["segments"][0]["signals"]:
+            for chan_idx in range(len(stream_annotations["__array_annotations__"]["channel_names"])):
+                headers.append({k: v[chan_idx] for k, v in stream_annotations["__array_annotations__"].items()})
+
+    # extract common attributes across channels
+    common_header = _dict_intersection(headers)
```
h-mayorquin, line 157: Can you mention what this function is doing?

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
JuliaSprenger: The latest version of neo is 0.11.0. However, in the current dev version more metadata will be exposed, and this is what I am testing already here. I wasn't sure if you run the tests with multiple neo versions (or maybe some users run different versions), so I wanted to make the test as robust as possible right now. In the future if you ever upgrade your package dependency to neo >=0.12 we can simplify the code here again.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
JuliaSprenger: I extended the comment. Also the function itself has a docstring.

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
h-mayorquin, line 29:
```suggestion
    # the sampling rate and device is only exposed on neo dev as of now. TODO: Remove if after neo >=0.12
```

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
```diff
@@ -13,56 +10,88 @@
...
+
     folder_path = NLX_PATH / "Cheetah_v5.7.4" / "original_data"
-    assert get_metadata(str(folder_path)) == dict(
-        session_start_time=datetime.datetime(2017, 2, 16, 17, 56, 4), session_id="d8ba8eef-8d11-4cdc-86dc-05f50d4ba13d"
-    )
+    metadata = NeuralynxRecordingInterface(folder_path).get_metadata()
+    file_metadata = metadata["NWBFile"]
+
+    assert file_metadata["session_start_time"] == datetime.datetime(2017, 2, 16, 17, 56, 4)
+    # the session id can only be extracted from the file headers exposed for neo > 0.11.0
```
h-mayorquin, line 21:
```suggestion
    # the session id is only exposed on neo dev as of now. TODO: Remove if after neo >=0.12
```

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
```diff
@@ -13,56 +10,88 @@
...
+    file_metadata = metadata["NWBFile"]
+
+    assert file_metadata["session_start_time"] == datetime.datetime(2016, 11, 28, 21, 50, 33, 322000)
+
+    # Metadata extracted directly from file header (neo >= 0.11)
+    assert '"FileType": "CSC"' in file_metadata["notes"]
+    assert '"recording_closed": "2016-11-28 22:44:41.145000"' in file_metadata["notes"]
+    assert '"ADMaxValue": "32767"' in file_metadata["notes"]
+
+    # the sampling rate and device is part of the file headers exposed for neo > 0.11.0
```
h-mayorquin, line 52:
```suggestion
    # the sampling rate and device is only exposed on neo dev as of now. TODO: Remove if after neo >=0.12
```

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
```diff
@@ -13,56 +10,88 @@
...
-    )
+    metadata = NeuralynxRecordingInterface(folder_path).get_metadata()
+    file_metadata = metadata["NWBFile"]
+
+    assert file_metadata["session_start_time"] == datetime.datetime(2001, 1, 1, 0, 0)
+
+    assert '"recording_closed": "2001-01-01 00:00:00"' in file_metadata["notes"]
+    assert '"ADMaxValue": "32767"' in file_metadata["notes"]
+
+    # the sampling rate and device is part of the file headers exposed for neo > 0.11.0
```
h-mayorquin, line 73:
```suggestion
   # the sampling rate and device is only exposed on neo dev as of now. TODO: Remove if after neo >=0.12
```

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
CodyCBakerPhD: > I wasn't sure if you run the tests with multiple neo versions (or maybe some users run different versions), so I wanted to make the test as robust as possible right now.

We do not, but we can (and should) add a dev gallery of tests and can (and should) include `neo` in that matrix. I'll see if I can get that setup today.


> In the future if you ever upgrade your package dependency to neo >=0.12 we can simplify the code here again.

Currently we do not pin `neo` on the `neuroconv` level, instead we rely on `spikeinterface` (atalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ecephys/requirements.txt), which in turn relies on latest patch version of `0.11` (https://github.com/SpikeInterface/spikeinterface/blob/master/requirements.txt#L2). Though if `spikeinterface does not update that after the next major `neo` release, we probably would here just to make reliance on this metadata stuff more reliable for our purposes.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -118,3 +83,109 @@ def __init__(self, folder_path: FolderPathType, sampling_frequency: float = None
...
+            for key, value in channel_filter_dict.items():
+                if not isinstance(value, str):
+                    value = " ".join(value)
+                channel_filter_dict[key] = value
+            filter_info.append(json.dumps(channel_filter_dict, ensure_ascii=True))
+
+    return filter_info
+
+
+def extract_metadata_single_reader(neo_reader) -> dict:
```
h-mayorquin, line 126:
```suggestion
def extract_neo_header_metadata(neo_reader) -> dict:
```

More specific, I guess that the single reader was for the previous case when multi readers were considered for the old backend.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -118,3 +83,109 @@ def __init__(self, folder_path: FolderPathType, sampling_frequency: float = None
         """
 
         super().__init__(folder_path=folder_path, sampling_frequency=sampling_frequency, verbose=verbose)
+
+
+def get_filtering_multi_channel(neo_reader) -> List[str]:
```
h-mayorquin, line 88:
```suggestion
def extract_filtering_metadata(neo_reader) -> List[str]:
```
I am confused about the multi channel here, are there some single channel  readers out there?

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
h-mayorquin: I added more specific comments and TODO for this.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
h-mayorquin: I was wondering more about _dict_intersection, what are the cases when:

```
    if len(dictionaries) == 0:
        return {}
    if len(dictionaries) == 1:
        return dictionaries[0]
  
```

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -1,106 +1,71 @@
...
         self.recording_extractor = self.recording_extractor.select_segments(segment_indices=0)
-
+        # preserve neo_reader attribute for SelectSegmentRecording
+        self.recording_extractor.neo_reader = neo_reader
         self.add_recording_extractor_properties()
 
     def add_recording_extractor_properties(self):
+        filtering = get_filtering_multi_channel(self.recording_extractor.neo_reader)
 
         try:
```
h-mayorquin, line 33: Can you remove the try structure? it is no longer needed here.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -1,106 +1,71 @@
...
-
-
 class NeuralynxRecordingInterface(BaseRecordingExtractorInterface):
     """Primary data interface for converting Neuralynx data. Uses
     :py:class:`~spikeinterface.extractors.NeuralynxRecordingExtractor`."""
 
     def __init__(self, folder_path: FolderPathType, verbose: bool = True):
 
-        self.nsc_files = natsorted([str(x) for x in Path(folder_path).iterdir() if ".ncs" in x.suffixes])
+        self.ncs_files = natsorted([str(x) for x in Path(folder_path).iterdir() if ".ncs" in x.suffixes])
```
h-mayorquin, line 21: I don't think this or natsorted is needed here anymore.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -1,106 +1,71 @@
-"""Authors: Heberto Mayorquin, Cody Baker and Ben Dichter."""
+"""Authors: Heberto Mayorquin, Cody Baker, Ben Dichter and Julia Sprenger."""
 import json
 import warnings
 from pathlib import Path
+from typing import List
 
 from natsort import natsorted
```
h-mayorquin, line 7: I don't think this is needed anymore.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
h-mayorquin: All right, I was finding this function confusing, Here is my suggestion of how to make it more readable that also includes type hints:

```
def _dict_intersection(dict_list: List) -> Dict:
    """
    Intersect dict_list and return only common keys and values
    Parameters
    ----------
    dict_list: list of dicitionaries each representing a header
    Returns
    -------
    dict:
        Dictionary containing key-value pairs common to all input dicitionary_list
    """
    if len(dict_list) == 0:
        return {}
    if len(dict_list) == 1:
        return dict_list[0]

    # Collect keys appearing in all dictionaries
    common_keys = set.intersection(*[set(dict.keys()) for dict in dict_list])

    # Add values for common keys if the value is identical across all headers (dict_list)
    first_dict = dict_list[0]
    all_dicts_have_same_value_for = lambda key: all([first_dict[key] == dict[key] for dict in dict_list])
    common_header = {key: first_dict[key] for key in common_keys if all_dicts_have_same_value_for(key)}
    return common_header
```

I am still not sure the cases are necessary. I don't think the first can happen as it will fail before and the second should work anyway with the functions as it is.

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
```diff
@@ -13,56 +10,88 @@
...
-        '"DspDelayCompensation": "Enabled", '
-        '"DspFilterDelay_µs": "984"}'
-    )
-    assert filtering == expected_filtering
+    file_path = NLX_PATH / "Cheetah_v5.7.4" / "original_data"
+    interface = NeuralynxRecordingInterface(file_path)
+    filtering = interface.recording_extractor.get_property("filtering")
+
+    assert interface.recording_extractor.get_num_channels() == len(filtering)
+    assert '"DSPLowCutFilterEnabled": "True"' in filtering[0]
```
h-mayorquin, line 87: This should not be a json but we should have a channel property for each of those values. That is, instead of having a single property whose values is a json dictionary with the x, y and z values, we should have a properties x, y and z with values for each channels. This can be done automatically by spikeinterface I think.

h-mayorquin (changes requested):
I think this looks good save the filtering part which is adding the channel properties as a json dict instead of as channel properties (which in turn implies that they are not cleanly written in the electrode table). 

Spikeinterface already has the capability of propagating some of the neo metadata that we are extracting here and provide them as channel properties automatically. For some reason, this is not enabled by default but we can make them available by changing the init of the interface here in the following way:

```
  def __init__(self, folder_path: FolderPathType, verbose: bool = True):

      super().__init__(folder_path=folder_path, verbose=verbose, all_annotations=True)
      self.add_recording_extractor_properties()
``` 

I tested this and I saw that the properties extracted in this PR by the `get_filtering` function are already loaded as channel properties this way. The code that performs this function which I think we are duplicating here is:
https://github.com/SpikeInterface/spikeinterface/blob/9b9ae370fdaa80544022a05c7f4848cb3b8371cb/spikeinterface/extractors/neoextractors/neobaseextractor.py#L115-L129

@JuliaSprenger

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
```diff
@@ -118,3 +83,109 @@ def __init__(self, folder_path: FolderPathType, sampling_frequency: float = None
...
+        for downstream things that expect ASCII.
+    """
+
+    # check if neuralynx file header objects are present and use these metadata extraction
+    if hasattr(neo_reader, "file_headers"):
+        # use only ncs files as only continuous signals are extracted
+        headers = [header for filename, header in neo_reader.file_headers.items() if filename.lower().endswith(".ncs")]
+
+    # use metadata provided as array_annotations for each channel (neo version <=0.11.0)
+    else:
```
h-mayorquin, line 149:
```suggestion
    else: TODO: Remove else after neo >=0.12
```

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
JuliaSprenger: Done.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
JuliaSprenger: Done.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
JuliaSprenger: I removed it.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
JuliaSprenger: I added the comment.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
JuliaSprenger: Yes, in the old `spikeextractors` version there was also an `get_filtering_single_channel`. I renamed the method.

h-mayorquin:
Thanks for implementing the suggestions above @JuliaSprenger .

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
JuliaSprenger: Yes, I see, so spikeinterface properties correspond to neo array_annotations. I will see how I can make use of this.

JuliaSprenger: @h-mayorquin Thanks for the `all_annotations` hint. I am using this now to automatically convert all neo array_annotations to properties. To still keep track of which of those properties are related to filtering I added an annotation that lists all filter related property keys. What do you think of this solution?

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
h-mayorquin: I don't understand, this is an aux function that should only be used inside this interface (it is marked as private). If we are using it in this interface then it means that the recordings should have continous data or the interface should have failed before. Asked in another way, what would lead to a situation where there would be no headers in the context of this interface? Can you flesh out the case a little bit more so I can think on what's the best way of building a safety test for this as well.

h-mayorquin:
@JuliaSprenger 
The tests that is failing is because some of the filtered columns is not in the following types: a list, np.ndarray, string or Real object (from the numbers library). Can you show here the column types of the filtered related column, I wonder what is the column that is failing the type check.

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
JuliaSprenger: The case I have in mind is that at the end you want to run the conversion on many recording sessions, most of which are complete in the sense that among other data they contain continuous neuralynx data. However, knowing the difficulty of having 100% complete datasets there might be a few sessions where the continuous neuralynx data might not be present. In that case,  do you prefer the conversion process to break and raise an error here or do you want to convert as much of the data as you can extract and just leave the continuous data empty? I agree this is kind of a corner case, but for later usage of the package potentially important.

JuliaSprenger:
All properties were numpy arrays, but it seems an object (e.g. datetime.datetime objects) or bool dtype is not accepted by nwb. So I added some lines to convert these to a string representation.

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
JuliaSprenger: So where would you add the additional metadata then instead to the NeuralynxInterface? Is there' a way how to add this directly to the NWB metadata without nwb complaining about invalid keys being used?

In general if there would be a standardized way on how to specify filter settings (e.g. specifying multiple filters for each channel with pre-defined keys for each channel already in spikeinterface that would be very useful in my opinion. Possible keys could be filter type ('low'/'high'), specification (e.g. 'butterworth',...), order (int), mode ('analog'/'digital'), ....

src/neuroconv/datainterfaces/ecephys/neuralynx/neuralynxdatainterface.py
h-mayorquin: OK, I got this. I will discuss this in the next meeting on how to have a common way of dealing with this problem.

h-mayorquin (approved):
Oh, yes, this is OK, thanks for changing the types, we should accept bools and I will fix that into another PR. One last thing, can you remove the annotation out of the initialization, as I mentioned in our discussion we don't really propagate those.

h-mayorquin (approved):
Thanks for all the hard work.

tests/test_on_data/test_metadata/test_neuralynx_metadata.py
```diff
@@ -13,56 +10,122 @@
...
+    expected_single_channel_props = {
+        "DSPLowCutFilterEnabled": "True",
+        "DspLowCutFrequency": "10",
+        "DspLowCutNumTaps": "0",
+        "DspLowCutFilterType": "DCO",
+        "DSPHighCutFilterEnabled": "True",
+        "DspHighCutFrequency": "9000",
+        "DspHighCutNumTaps": "64",
+        "DspHighCutFilterType": "FIR",
+        "DspDelayCompensation": "Enabled",
```
bendichter, line 119: It looks like there are int and bool values here there encoded as strings. Is there a reason for that?

JuliaSprenger: For export of these properties we need to convert bool values to strings, as we otherwise get an error in nwb (see https://github.com/catalystneuro/neuroconv/issues/185 first checkpoint).

bendichter: the electrodes table should be able to handle ints and bools with no problem. What's the error in NWB?

CodyCBakerPhD: It's not an error with NWB, it's an error with channel properties that are bool which are then used in our `add_electrodes` function (see https://github.com/catalystneuro/neuroconv/issues/185) which doesn't have a default type set for that class, which is something needed when writing multiple recording objects (not all of which have that same boolean channel property)

CodyCBakerPhD: The `float` and `int` values here shouldn't have to be strings though

PR #177, opened 2022-10-17
Propose simplification to gallery codes

CodyCBakerPhD:
@h-mayorquin What would you think about removing the last bit of the conversion galleries that involves printing out the `is_file()` result on the path?

I've always thought it looks a bit awkward, and it occurred to me now that it's also not a great indicator of success. In fact, there are many points that can occur during any given `run_conversion` call that will result in a file being created, but the contents are corrupted and the NWB file is unreadable. The `is_file()` check would still be true, but not a good indicator that nothing went wrong.

Likewise, if anything does go wrong, it will show up as a traceback (however complicated and hdmf or h5py-nested it might be) from the `interface.run_converssion(...)` call, so it's more like we're asserting that there is no unexpected `stderr` from that line above


h-mayorquin:
Yes, I don't think checking for file existence is doing a great check at verifying that the conversion ran successfully. Ultimately, we need a better way (any suggestions?). If you feel that the advantages of this weak test are not as large as the cost of the extra line I am fine with removing them, I don't have a strong feeling for this one

Can you describe some situations when the test would throw a false success?

CodyCBakerPhD:
> Ultimately, we need a better way (any suggestions?)

The test would fail if such an error occurred during the `run_conversion` call wouldn't it? So the doctest being treated as an actual test unto itself would fail correspondingly and even have the traceback stack for it. I'd be fine with just relying on that as the assurance of pass/fail.

If you really do still want to keep a more unit-test like vibe to it, though, where there's some kind of structural assertion at the end, the most reliable thing would be a roundtrip access, _i.e_., 

```python
with NWBHDF5IO(path=nwbfile_path, mode="r") as io:
    nwbfile = io.read()
```

which might be nice as well so then the user can look at the file contents/structure right away (if they're using a notebook, for instance, in the next cell they could just immediately print that variable to see a rough outline of content).




> Can you describe some situations when the test would throw a false success?

Basically anything that would cause an error after a write-io is established. Basically, all `run_conversion` operations using `nwbfile_path` can be generalized to a structure of

```python
with NWBHDF5IO(...) as io:
    # do a bunch of stuff to some in-memory nwbfile object
    io.write(nwbfile)
```

So if there's an error that occurs after the context creates the file, you'll still be left with a file and anything that had been partially written to it prior to the error.

Some of the worst types of this though are errors that can occur during the `io.write()` step, such as when you setup an iterative write for a dataset that doesn't raise any problem when you first make the neurodata type, but then encounters an unexpected read-access or shaping error after the iteration starts.

h-mayorquin:
Thanks for explaining. Yes, this is fine to me. For the moment we can rely on the doctest just throwing any assertion as a failure. In the future, we can add a function that does a round-trip which we can use for code reduction in the gin-tests and doctests and users can use to check their conversions.

bendichter:
I like that these galleries are tested, but I don't want to sacrifice readability to improve tests here. If we want to do thorough tests, we should do that in the testing suite

bendichter (dismissed):

h-mayorquin:
Yes, this makes sense to me. I committed the remaining changes.

bendichter (approved):

PR #178, opened 2022-10-17
Remove unneeded GIN files on S3 copy

CodyCBakerPhD:
These files aren't needed for our testing purposes with the S3 copy of the data, and since the AWS CLI recursive inclusion/exclusion rules are so tricky, it's easier to just remove them prior to sync to reduce the number of billed requests (and storage space, but that's a way smaller amount).

### DIFF
--- a/.github/workflows/update-caches.yml
+++ b/.github/workflows/update-caches.yml
@@ -113,7 +113,7 @@ jobs:
         run: datalad install -rg https://gin.g-node.org/CatalystNeuro/behavior_testing_data
       - name: Cleanup unneeded files to minimize AWS billing
         if: steps.cache-behavior-datasets.outputs.cache-hit == false
-        run: |  
+        run: |
           rm -rf ./ephy_testing_data/.datalad/
           rm -rf ./ephy_testing_data/.git/
           rm -rf ./ephy_testing_data/.gitattributes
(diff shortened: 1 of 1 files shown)


h-mayorquin (commented):
Can you add a comment explaining why we are doing this on the code or, alternatively,  do the removal under a specific action with a corresponding name?

Out of curiosity, does this makes a large difference? I have no idea.

CodyCBakerPhD:
> Can you add a comment explaining why we are doing this on the code or, alternatively, do the removal under a specific action with a corresponding name?

Good idea, done


> Out of curiosity, does this makes a large difference? I have no idea.

The `ephy_testing_data` especially has like tens of thousands of items stored in the private files of the 'git' repo, whereas all we need really are the data files themselves. The thing is, when you push these many files to the S3 bucket (or pull them via download, or scan them with a search) it bills you per API call to each item, so each time we need to update the cache we're getting billed for tens of thousands of files we simply don't need.

Now, a lot of that fits under the free tier, and the billing scaling is very forgiving, we're talking about saving maybe $0.10 per month during a busy month of updating the datasets.

I know, not the biggest issue in the world, but it does bother me and was quick to fix.

h-mayorquin (approved):
Thanks for explaining to me how it works.

PR #179, opened 2022-10-18
Auto-cancel previous workflow runs

CodyCBakerPhD:
Noticed this one on PyNWB and thought it might be useful given how heavy our tests are. When there's a new push to a PR where the tests of a previous push were running, this will automatically cancel them. I kind of do this already myself when I need to free up resources (or avoid feeling bad about the cost to the environment) but it's a bit annoying to do it manually.


h-mayorquin (approved):
This is good!

PR #181, opened 2022-10-18
stub Movie interface timestamps

h-mayorquin:
While working in the current conversion with Murthy I am facing with some rather large video files and stubbing is very useful for fast iteration. However,  the timestamps themselves are not stubbed and OpenCV goes frame by frame through the whole video rendering the stubbing process useless (even in `external_mode=True`). This PR implements stubbing in the timestamp extraction as well limiting the frame reading in OpenCV.

I added a test and tested this on the Murthy conversion.


CodyCBakerPhD (approved):

CodyCBakerPhD:
I do wish there was a lazier way of getting the timestamps to begin with, but I guess that takes us out of the OpenCV territory so a bit bigger undertaking

PR #183, opened 2022-10-19
Add `VideoInterface` to conversion gallery

h-mayorquin:
As stated in the title. This has #181 as a base. Note that the MovieInterface was not using the context manager which we were using in all the other interfaces (`make_or_load_nwbfile`). This was added her e to keep consistent style access across the conversion gallery.

PR #190, opened 2022-10-21
Replace `MovieInterface` with `VideoInterface`

h-mayorquin:
As discussed in #183, we are changing Video for Movie as the term is more accurate. I introduced a deprecation warning for the the `MovieInterface` indicating that it will be deprecated in six months. 

This PR introduces the changes in naming and the interface functions but keeps most of the testing (gin, data and utils) for the `MovieInterface` the same. After I verify that everything is running properly I will introduce the changes in the testing names as well. This is just out of caution.


h-mayorquin:
> @h-mayorquin this is great, thanks for putting this together! The only thing I can find that you missed is the filenames within the test directory. test_movie_interface.py -> test_video_interface.py and test_movie_utils.py -> test_video_utils.py

Yep, I will change that now.

h-mayorquin:
@bendichter done.
I changed this on all the other places as well.

PR #193, opened 2022-10-27
Add flag for disabling accepted+rejected ROIs

CodyCBakerPhD:
Branched from #159 

For my conversion with 100's of thousands of ROIs, I was wondering why my conversion was taking so long (even in stub mode) and some investigation led to these lines, kind of similar to the ROI centroid calculation (but not as bad as that one was) so adding optional control over whether this gets called.


CodyCBakerPhD:
Two future PR's

- Remove redundancy of both Accepted and Rejected columns (since if you have one, you necessarily have the other)
- More efficient assessment at lines https://github.com/catalystneuro/neuroconv/pull/193/files#diff-5fc5d8ed0a5b946f2a3aa5f55f3c91020f9700071563487cbdda6da8e0314110R622-R623 for detecting acceptance for each ROI ID

h-mayorquin (approved):

PR #194, opened 2022-10-28
Added capture for memory leak in ImagingExtractorDataChunkIterator

CodyCBakerPhD:
Branched from #193 

WIP (need to add test and figure out how to fix it)

The `ImagingExtractorDataChunkIterator` resulted in a memory leak during my non-stub attempt to write a file for the Ahrens project.

I've identified the source (thanks to warnings from numpy) but to make things safe (otherwise the excess RAM usage eventually kills my entire device) I've added a capture that elevates the problem to an error and makes it more informative.

Error message appears like

image


CodyCBakerPhD:
Found the source of the issue, summarized in sister PR on HDMF: https://github.com/hdmf-dev/hdmf/pull/780

Added fix here, too. Will add tests tomorrow or next week

CodyCBakerPhD:
Replaced by #195

PR #195, opened 2022-11-01
Prevent buffer overflow

CodyCBakerPhD:
The major fix for this should be taken care of in https://github.com/hdmf-dev/hdmf/pull/780

Replaced #194


h-mayorquin (approved):
I guess the cycle of HDMF is too slow to wait for the stable version?

CodyCBakerPhD:
> I guess the cycle of HDMF is too slow to wait for the stable version?

It's something to consider.. If this is merged here we will lock ourselves out of another NeuroConv release until it is, but it's also just a bug fix so maybe we could ask them to do a small release for that

h-mayorquin:
I would like to have a release for neuroconv after the Murthy project. But that probably would be in December if Ben Cowley manages to get the sleap data by then. I have some small fix to add to the sleap interface here.

bendichter (approved):

CodyCBakerPhD:
This and #238 are failing due to a new release of `packaging` yesterday (after our daily scheduled run no less...) introducing a breaking change to version parsing down at the `neo` level: https://github.com/pypa/packaging/issues/631

I'll try to get a fix here quickly

CodyCBakerPhD:
OK that fix is going to take a little bit, handed it off to SI team: https://github.com/SpikeInterface/probeinterface/issues/147

and trying to skip the tests here

bendichter:
Can we just specify an older version for packaging for now and fix it in a separate PR?

PR #199, opened 2022-11-10
Include session_id in automatic dandi upload test

CodyCBakerPhD:
## Motivation

A part of our helper functions for running a fully automated transfer+conversion+upload workflow is a one-liner for an automatic dandi upload (given the api token is set as an environment variable). This goes to https://gui-staging.dandiarchive.org/dandiset/200560/draft/files?location=sub-foo, which I just noticed has a `None` as the session ID. So this sets it to be a bit clearer.

### DIFF
--- a/tests/test_minimal/test_tools.py
+++ b/tests/test_minimal/test_tools.py
@@ -186,7 +186,7 @@ class TestAutomaticDANDIUpload(TestCase):
         self.nwb_folder_path = self.tmpdir / "test_nwb"
         self.nwb_folder_path.mkdir()
         metadata = get_default_nwbfile_metadata()
-        metadata["NWBFile"].update(session_start_time=datetime.now().astimezone())
+        metadata["NWBFile"].update(session_start_time=datetime.now().astimezone(), session_id="test_automatic_upload")
         metadata.update(Subject=dict(subject_id="foo"))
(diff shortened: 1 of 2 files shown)


tests/test_minimal/test_tools.py
```diff
@@ -186,7 +186,7 @@ def setUp(self):
         self.nwb_folder_path = self.tmpdir / "test_nwb"
         self.nwb_folder_path.mkdir()
         metadata = get_default_nwbfile_metadata()
-        metadata["NWBFile"].update(session_start_time=datetime.now().astimezone())
+        metadata["NWBFile"].update(session_start_time=datetime.now().astimezone(), session_id="test_automatic_upload")
```
h-mayorquin, line 189: Should we add `neuroconv` to the session_id just for a little bit of better tracking?

Then I saw that the dandi set is called `NeuroConv SDK Testing` so probably not need. Maybe a timestamp could be useful?

h-mayorquin (approved):
LGTM

tests/test_minimal/test_tools.py
CodyCBakerPhD: A timestamp would be dangerous - it would result in complete spam of the dandiset (a new file every time the CI runs)

PR #200, opened 2022-11-10
Amend rate and fps warning in video interface

h-mayorquin:
For the conversion that I am working on (Murthy) I am transforming multiple videos. Their timestamps are 60 fps but when estimated with our `calculate_regular_series_rate` the estimate becomes 60.0000001 which is unequal to the 60 that the video context is extracting (the correct one). This raises a warning and stores the rate as 60.00001 which is an inaccuracy that arises from our floating point estimate. 

This PR solves this problem for by making the comparison with the estimated rates rounded to two decimals. Standard constant frame rate formats don't meaningfully differ by that amount. 

Also, I realized that the `VideoInterface` was using `DataChunkIterator` even when the conversion option `chunk_data` was False. I changed that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 ### Testing
 * Added a `session_id` to the test file for the `automatic_dandi_upload` helper function. [PR #199](https://github.com/catalystneuro/neuroconv/pull/199)
 
+### Fixes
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
LGTM

PR #208, opened 2022-11-23
Test new testing tools import for SI

CodyCBakerPhD:
## Motivation

Testing a small import fix for an upcoming dev branch change on SI caught by https://github.com/catalystneuro/neuroconv/actions/runs/3534188831/jobs/5930773743


CodyCBakerPhD:
Well this is annoying - due to a change in import structure for `generate_recording` and `generate_sorting` between the current dev branch of SI and previous release, if I make the change to NeuroConv dev that fixes this, then it breaks the suite against past releases.

Either way will have to bump the SI version here soon as next release is available, and just reroute the import dynamically for now I guess.

Smoother deprecation of this would be appreciated @alejoe91 @samuelgarcia

CodyCBakerPhD:
May not be necessary once Alessio reroutes this on SI. Maybe just live with the failing CI for the time being

Issue #185, opened 2022-10-19
Neuralynx metadata improvements


h-mayorquin:
During the discussion of #170 some things were left to-do. I am writing them here to keep track.

- [x] Add capability to `add_electrodes` to write collumns with boolean values (and remove the code from this on neuralynx)
- [x] Check whether `NeuralynxRecordingInterface`,  initialization or writing process fails when non `ncs` files are present in the `folder_path`.

JuliaSprenger:
Hi @h-mayorquin To clarify this: By the 2nd point you to test what happens if a session contains **only** non-`ncs` files, so no continuous data, right?

h-mayorquin:
Yes, I corrected the issue to reflect this. Hopefully it is more clear now.

CodyCBakerPhD:
> Add capability to add_electrodes to write collumns with boolean values (and remove the code from this on neuralynx)

To be clear, we cannot currently write a recording object with channel properties that are set to boolean values? (and not missing any data for any channels). I just wonder how this would work with append mode, like when you write one recording object that doesn't have that column defined, then write a second one that does define it - what value is used to auto-populate the rows of the previous recording in that case? Seems like we'd actually want a 3-state variable (`False, True, NaN`) but at that point perhaps it would be better to upcast it to float so we don't have to define our own standard for what a `NaN` value is in this case...

h-mayorquin:
> > Add capability to add_electrodes to write collumns with boolean values (and remove the code from this on neuralynx)
> 
> To be clear, we cannot currently write a recording object with channel properties that are set to boolean values? (and not missing any data for any channels). I just wonder how this would work with append mode, like when you write one recording object that doesn't have that column defined, then write a second one that does define it - what value is used to auto-populate the rows of the previous recording in that case? Seems like we'd actually want a 3-state variable (`False, True, NaN`) but at that point perhaps it would be better to upcast it to float so we don't have to define our own standard for what a `NaN` value is in this case...

Yes, we can't not, because we are missing default values. Good point about the nan issue.

CodyCBakerPhD:
Cleaning up some old issues

Item 1 was resolved in #217, item 2 seems the responsibility of neo/SI so I'll leave it up to @JuliaSprenger who would probably know best what would happen in that case

h-mayorquin:
The last point was about the folllowing:
[...]

PR #217, opened 2022-11-24
Write waveform_extractor from spikeinterface

### LINKED ISSUE #197: [Feature]: save SpikeInterface `WaveformExtractor` object

Add a function in `tools.spikeinterface` to `write_waveforms` from a `WaveformExtractor` object.

Yes. Currently there is no way to write mean waveforms and standard deviations using SpikeInterface>=v0.90 (maintained version)

Yes.

### Code of Conduct

alejoe91:
fix #197

## Motivation

`WaveformExtractor` is a base spikeinterface object that handles waveform computation and posptocessed data.

Added a function `neuroconv.tools.spikeinterface.write_waveforms()` to write this object. By default, only electrode info, sorted data + waveforms (mean + sd) are written, but optionally also raw traces can be written.

## How to test the behavior?
```
import spikeinterface.full as si
form neuroconv.tools.spikeinterface import write_waveforms

rec, sort = si.toy_example()
we = si.extract_waveforms(rec, sort, folder=None, mode="memory")

nwbfile = write_waveforms(we, nwbfile_path="test_waveforms.nwb")
print(nwbfile)
```

## Checklist


CodyCBakerPhD:
@alejoe91 Thanks a bunch! I'll be going through this in detail for the next few days. Thanks for the example files, too!

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -401,7 +401,7 @@ def add_electrodes(
         )
 
     # Find default values for custom (not schema defined) properties / columns already in the electrode table
-    type_to_default_value = {list: [], np.ndarray: np.array(np.nan), str: "", Real: np.nan}
+    type_to_default_value = {list: [], np.ndarray: np.array(np.nan), str: "", Real: np.nan, bool: False}
```
CodyCBakerPhD, line 404: What field is `bool` that requires this? Note that this type of change could potentially result in incorrect data, so we'd want to try to avoid it. Same reason we upcast `int` to `float` so we can designate `NaN` as 'unknown' rather than assuming some potentially incorrect default integer value

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1072,13 +1072,16 @@ def get_nspikes(units_table: pynwb.misc.Units, unit_id: int):
...
+    unit_ids: Optional[List[Union[str, int]]] = None,
     property_descriptions: Optional[dict] = None,
     skip_properties: Optional[List[str]] = None,
     skip_features: Optional[List[str]] = None,
     units_table_name: str = "units",
     unit_table_description: str = "Autogenerated by neuroconv.",
     write_in_processing_module: bool = False,
     write_waveforms: bool = False,
+    waveform_means: Optional[np.array] = None,
+    waveform_sds: Optional[np.array] = None,
```
CodyCBakerPhD, line 1084:
```suggestion
    waveform_means: Optional[np.ndarray] = None,
    waveform_sds: Optional[np.ndarray] = None,
```

or otherwise define generally as the `ArrayType` from `neuroconv.utils`

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1087,6 +1090,9 @@ def add_units_table(
     ----------
     sorting: SpikeInterfaceSorting
     nwbfile: NWBFile
+    unit_ids: list, optional
```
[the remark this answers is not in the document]
CodyCBakerPhD: Or I guess the `int` is forcibly upcast to `str` if that's what you use?

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1459,157 @@ def write_sorting(
...
+
+    # metrics properties (quality, template) are temporarily added as properties
+    # to be written in the Units table
+    tmp_properties = []
+    if waveform_extractor.is_extension("quality_metrics"):
+        qm = waveform_extractor.load_extension("quality_metrics").get_data()
+        for prop in qm.columns:
+            if prop not in sorting.get_property_keys():
+                sorting.set_property(prop, qm[prop])
+                tmp_properties.append(prop)
```
CodyCBakerPhD, line 1556: Lines not covered by tests: https://app.codecov.io/gh/catalystneuro/neuroconv/blob/write-waveforms/src/neuroconv/tools/spikeinterface/spikeinterface.py

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1459,157 @@ def write_sorting(
...
+            else:
+                recording = waveform_extractor.recording
+            if recording is not None:
+                add_electrodes_info(recording, nwbfile=nwbfile_out, metadata=metadata)
+            else:
+                print(
+                    "To add the electrode table, the waveform_extractor needs to have a recording attached. "
+                    "For recordingless mode, you can use the 'recording' argument."
+                    "The electrode table will not be added."
+                )
```
CodyCBakerPhD, line 1596: Lines not covered by tests - be sure to add some for conditions that raise/capture the assertions as well

CodyCBakerPhD, line 1596: I'd use a `warnings.warn` instead of a `print` to make sure it goes to `std.err` pipe (important for loggers or other filters/captures)

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1459,157 @@ def write_sorting(
...
+                    "For recordingless mode, you can use the 'recording' argument."
+                )
+            write_recording(waveform_extractor._recording, nwbfile_out, metadata, **write_electrical_series_kwargs)
+        else:
+            # we need add_electrodes_from_waveforms
+            if waveform_extractor._recording is None:
+                if recording is not None:
+                    assert isinstance(recording, BaseRecording), "'recording' needs to be a BaseRecording object."
+            else:
+                recording = waveform_extractor.recording
```
CodyCBakerPhD, line 1588: OK so this is the case where the user does not want to write the electrical series, but we still require a recording to be attached to the `WaveformExtractor` so we can use the electrode info?

Is it possible to have a `WaveformExtractor` that does not have either the `_recording` or `recording` attribute, or is at least one always `None`?

Are the two cases to support a previous version of SI or what else causes the distinction between the private and non-private version?

src/neuroconv/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: It's also a question if this ought to be a warning or an earlier assertion... It would seem odd to add waveforms, but not even have an electrodes table at all in the file (or worse, an incorrect electrodes table that only has items from a different probe)

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1159,26 +1170,29 @@ def add_units_table(
...
     properties_to_extract = [property for property in sorting_properties if property not in excluded_properties]
 
+    if unit_ids is not None:
+        checked_sorting = checked_sorting.select_units(unit_ids)
+    unit_ids = checked_sorting.unit_ids
+
     # Extract properties
     for property in properties_to_extract:
-        data = checked_sorting.get_property(property)
+        data = checked_sorting.get_property(property, ids=unit_ids)
```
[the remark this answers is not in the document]
alejoe91: not anymore, removing!

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1459,157 @@ def write_sorting(
...
+            units_table_name=units_name,
+            unit_table_description=units_description,
+            write_waveforms=False,
+            waveform_means=template_means,
+            waveform_sds=template_stds,
+        )
+
+    # rm tmp properties
+    for prop in tmp_properties:
+        sorting.delete_property(prop)
```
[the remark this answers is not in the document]
alejoe91: yes, that's a good idea

src/neuroconv/tools/spikeinterface/spikeinterface.py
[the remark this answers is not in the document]
alejoe91: not either or, but `unit_ids` can be either int or str (depending on the extractor). This list needs to be consistent with the sorting unit ids type

src/neuroconv/tools/spikeinterface/spikeinterface.py
alejoe91: Properties can be boolean (e.g., in our case we have a property `pass_qc` if the unit passes quality metrics). Currently, this is not supported (I believe it should though!)

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1159,26 +1170,29 @@ def add_units_table(
     excluded_properties = list(skip_properties) + ["contact_vector"]
     properties_to_extract = [property for property in sorting_properties if property not in excluded_properties]
 
+    if unit_ids is not None:
+        checked_sorting = checked_sorting.select_units(unit_ids)
+    unit_ids = checked_sorting.unit_ids
```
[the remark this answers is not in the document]
alejoe91: it's a property. The old extractor gets converted to a new API object, so it will work for old extractors too.

alejoe91:
> @alejoe91 Thanks for sharing the files. A couple things to note
> 
> 1. the examples files look like they used only the synthetic data with 4 units and 4 channels - I don't notice much of a difference with the multi-segment (I guess there ought to be a gap in the spike times perhaps? If so we might want to use the `obs_intervals` (see schema) to denote the times when the recording system was actively acquiring.

I think it's a good idea, but probably for another PR (since multi-segment was already present)

> 2. Also, neither of the example files look to have attached the recording or LFP, can you demo that as well?
I'll generate one with the recording. To add LFP, one would need to use the `write_recording()` function, since waveforms needs to be extracted from filtered recordings.

> 3. Can you try an example with a stubbed real SpikeGLX recording? With both AP and LF included
Same here, I'll send you an exported file with AP only, since LF has nothing to do with waveforms :)
> Schema for units: https://nwb-schema.readthedocs.io/en/latest/format.html#units

alejoe91:
Thank you guys! I applied most of the suggested changes!

@CodyCBakerPhD here are the updated files from the tests: 
- waveform_single_segment.nwb (with electrical series)
- waveform_multi_segment.nwb (without electrical series)
- waveform_subset.nwb (without electrical series)
- waveform_recless.nwb (with electrical series)
[waveforms_nwb.zip](https://github.com/catalystneuro/neuroconv/files/10168553/waveforms_nwb.zip)

I'll get back to you with a real NP example in the following days.

src/neuroconv/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Dev branch CI: https://github.com/catalystneuro/neuroconv/actions/runs/3631606377/jobs/6126460171

is fine with everything

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1169,24 +1184,23 @@ def add_units_table(
...
+        data_to_add["id"].update(data=unit_ids.astype("int"))
 
     units_table_previous_properties = set(units_table.colnames) - set({"spike_times"})
     extracted_properties = set(data_to_add)
     properties_to_add_by_rows = units_table_previous_properties | set({"id"})
     properties_to_add_by_columns = extracted_properties - properties_to_add_by_rows
 
     # Find default values for properties / columns already in the table
-    type_to_default_value = {list: [], np.ndarray: np.array(np.nan), str: "", Real: np.nan}
+    type_to_default_value = {list: [], np.ndarray: np.array(np.nan), str: "", Real: np.nan, bool: False}
```
CodyCBakerPhD, line 1203: See discussion above https://github.com/catalystneuro/neuroconv/pull/217/files#r1041188364

We used to support this a while back but it causes issues with append functionality, and the last thing we want to do is add incorrect data in those cases.

Please upcast `bool` types to `floats`, however inefficient that might seem, so that we can properly refer to missing values as `NaN`

@h-mayorquin Do you recall if you added any tests to the suite using boolean property types?

src/neuroconv/tools/spikeinterface/spikeinterface.py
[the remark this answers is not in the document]
CodyCBakerPhD: Was the `clone` command not working as intended? Given we're already making this PR dependent on SI dev branch, it would be fine to wait on fixes for it here

src/neuroconv/tools/spikeinterface/spikeinterface.py
h-mayorquin: No, I don't think so. I was thinking on reviewing this on #185 .

alejoe91: I see. Maybe casting to string would be better then? At least you still have 'True' and 'False'? With float you would completely lose the categorical aspect

src/neuroconv/tools/spikeinterface/spikeinterface.py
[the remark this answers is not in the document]
alejoe91: No, the problem is that clone reloads the data from the underlying files, so any in-memory changes would be lost. For example, if adding a property before calling the write function, that property would not be cloned. So temporary properties work better in this case

src/neuroconv/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: I'd be fine with strings, sure

alejoe91:
> @alejoe91 Though there isn't an explicit tutorial for doing this on the `Units` table, hopefully this ought to work from https://pynwb.readthedocs.io/en/stable/tutorials/domain/ecephys.html#sphx-glr-tutorials-domain-ecephys-py
> 
> ```python
> this_region = nwbfile.create_electrode_table_region(
>     region=list(range(electrode_counter)),  # reference row indices on the electrodes table starting from 0; will likely want to do an index lookup based on channel_name to find the right ones
>     description="The electrodes that detected waveforms for this unit.",
> )
> 
> # then when you add the unit
>     nwbfile.add_unit(
>         spike_times=spike_times, ..., waveform_mean=[1.0, 2.0, 3.0, 4.0, 5.0], electrodes=this_region
>     )
> ```

As discussed, we'll keep sparsity for a follow-up PR

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -318,14 +321,14 @@ def add_electrodes(
...
     else:
         channel_name_array = channel_ids.astype("str", copy=False)
         data_to_add["channel_name"].update(description="unique channel reference", data=channel_name_array, index=False)
-
-    # If the channel ids are integer keep the old behavior of asigning nwbfile.electrodes.id equal to channel_ids
-    if np.issubdtype(channel_ids.dtype, np.integer):
-        data_to_add["id"].update(data=channel_ids, index=False)
+        # If the channel ids are integer keep the old behavior of asigning nwbfile.electrodes.id equal to channel_ids
+        if np.issubdtype(channel_ids.dtype, np.integer):
+            data_to_add["id"].update(data=channel_ids, index=False)
```
alejoe91, line 331: @CodyCBakerPhD as discussed yesterday, passing the `channel_name` property overrides the default names assigned by `neuroconv`. However, if `channel_ids` are `int`, the `id` field of the rows is not consistent. In this case, we only set the `id` if the `channel_name` property is not set.

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1159,26 +1173,31 @@ def add_units_table(
...
-        unit_name_array = units_ids.astype("str", copy=False)
+        unit_name_array = unit_ids.astype("str", copy=False)
         data_to_add["unit_name"].update(description="Unique reference for each unit.", data=unit_name_array)
-
-    # If the channel ids are integer keep the old behavior of asigning table's id equal to unit_ids
-    if np.issubdtype(units_ids.dtype, np.integer):
-        data_to_add["id"].update(data=units_ids.astype("int"))
+        # If the channel ids are integer keep the old behavior of asigning table's id equal to unit_ids
+        if np.issubdtype(unit_ids.dtype, np.integer):
+            data_to_add["id"].update(data=unit_ids.astype("int"))
```
alejoe91, line 1200: @CodyCBakerPhD same here for units: passing the `unit_name` property overrides the default names assigned by `neuroconv`. However, if `unit_ids` are `int`, the `id` field of the rows is not consistent. In this case, we only set the `id` if the `unit_name` property is not set.

alejoe91:
@CodyCBakerPhD 

Here an example file from NP with the AP ElectricalSeries and units:
https://drive.google.com/drive/u/0/folders/1-RAlW0mxRCDOc-VAF0ai3VAvoMCL8Zr_

This is the code I used to generate it:

```
from nwbinspector.tools import make_minimal_nwbfile

num_seconds = 2
num_samples = int(num_seconds * recording.sampling_frequency)

# recording, sorting, and we are the "full" objects
recording_stub = recording.frame_slice(start_frame=0, end_frame=num_samples)
sorting_stub = sorting.frame_slice(start_frame=0, end_frame=num_samples)
we_stub = we
we_stub.sorting = sorting_stub

write_waveforms(we_stub, nwbfile=nwbfile_stub, write_electrical_series=True, recording=recording_stub)

with NWBHDF5IO("stub_np_waveforms.nwb", "w") as io:
    io.write(nwbfile_stub)
```

CodyCBakerPhD:
@alejoe91 

> Here an example file from NP with the AP ElectricalSeries and units:
https://drive.google.com/drive/u/0/folders/1-RAlW0mxRCDOc-VAF0ai3VAvoMCL8Zr_

Link results in 404

alejoe91:
> @alejoe91
> 
> > Here an example file from NP with the AP ElectricalSeries and units:
> > https://drive.google.com/drive/u/0/folders/1-RAlW0mxRCDOc-VAF0ai3VAvoMCL8Zr_
> 
> Link results in 404

Oups my bad! Wrong link! Here is the correct one: https://drive.google.com/file/d/1CNzwRtRiJQuRUq6TguuHsUGSYI9pfc6D/view?usp=sharing 

(also updated the comment)

CodyCBakerPhD:
Hoping #238 fixes the tests here

CodyCBakerPhD:
File insofar as this PR goes looks good - note that your custom electrode column `inter_sample_shift` does not have a description, curious what that field represents

Many of the (I can only suppose) 'newer' quality metrics on the units table also have no description - we can add/fix that in a follow-up, though, no problem.

alejoe91:
> File insofar as this PR goes looks good - note that your custom electrode column `inter_sample_shift` does not have a description, curious what that field represents

Neuropixels probe do not sample all channels at the same time, because they have more channels than ADCs. The `inter_sample_shift` is the "delay" of each channel within a sampling period, which can be corrected for using the [`phase_shift`](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/preprocessing/phase_shift.py) preprocessing step.


> Many of the (I can only suppose) 'newer' quality metrics on the units table also have no description - we can add/fix that in a follow-up, though, no problem.

Yes let's update default descriptions in a follow-up PR :)

CodyCBakerPhD:
> Neuropixels probe do not sample all channels at the same time, because they have more channels than ADCs. The inter_sample_shift is the "delay" of each channel within a sampling period, which can be corrected for using the [phase_shift](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/preprocessing/phase_shift.py) preprocessing step.

Oh my, that is news to me... So is the data in the `ElectricalSeries` in `acquisition` adjusted in this way?

alejoe91:
> > Neuropixels probe do not sample all channels at the same time, because they have more channels than ADCs. The inter_sample_shift is the "delay" of each channel within a sampling period, which can be corrected for using the [phase_shift](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/preprocessing/phase_shift.py) preprocessing step.
> 
> Oh my, that is news to me... So is the data in the `ElectricalSeries` in `acquisition` adjusted in this way?

No, the ElectricalSeries in acquisition always contains the raw data

CodyCBakerPhD:
@alejoe91 Tests are now failing due to the `has_recording` helper - do you think you could locally define that function here so we don't have to pin to SI dev branch? We're getting close to needing a new release of NeuroConv and I was hoping we could include this with it too

alejoe91:
Sure that's a good idea! I'll do it now :)

alejoe91:
> @alejoe91 Tests are now failing due to the `has_recording` helper - do you think you could locally define that function here so we don't have to pin to SI dev branch? We're getting close to needing a new release of NeuroConv and I was hoping we could include this with it too

Done ;)

tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -1009,6 +1028,173 @@ def test_write_units_table_in_processing_module(self):
...
+    def test_cleanup_tmp_properties(self):
+        # test that if write_waveforms fails, temporary properties are cleaned up
+        try:
+            write_waveforms(waveform_extractor=self.we_recless, nwbfile=self.nwbfile5, recording=None)
+        except:
+            sorting_properties = self.we_recless.sorting.get_property_keys()
+            self.assertNotIn("peak_to_valley", sorting_properties)
+            self.assertNotIn("amplitud_cutoff", sorting_properties)
+
+    def write_test_files(self):
```
h-mayorquin, line 1188: Are these being tested even if the name is not starting with test? 

Also, I am confused why you create five nwb files in the `setUp` but per test you really only use one file, correct?

tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -1009,6 +1028,173 @@ def test_write_units_table_in_processing_module(self):
...
+        self.base_sorting.set_property("test_bool", bool_property)
+        add_units_table(
+            sorting=self.base_sorting,
+            nwbfile=self.nwbfile,
+        )
+        self.assertIn("test_bool", self.nwbfile.units.colnames)
+        assert all(tb in ["False", "True"] for tb in self.nwbfile.units["test_bool"][:])
+
+
+@unittest.skipIf(
```
h-mayorquin, line 1056: Should we consider dropping python 3.7? I have the impression that is causing problems somewhere else and it will have its eol in six months. Seems like pointless to support this already to me.

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1344,22 +1367,23 @@ def _add_waveforms_to_units_table(
                     values=flatten_vals,
                     index=spikes_index,
                 )
-        else:
-            """
-            Currently (2022-04-22), spikeinterface does not support waveform extraction.
-            """
-            pass
+    else:
```
h-mayorquin, line 1370: Should we change this else here?

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1464,178 @@ def write_sorting(
...
         )
     return nwbfile_out
+
+
+def write_waveforms(
+    waveform_extractor: WaveformExtractor,
+    nwbfile_path: OptionalFilePathType = None,
+    nwbfile: Optional[pynwb.NWBFile] = None,
+    metadata: Optional[dict] = None,
+    recording: Optional[BaseRecording] = None,
```
h-mayorquin, line 1474: Minor but overwrite should go after metadata to keep to keep the signature consistent. That is, we should swap recording with overwrite here.

h-mayorquin (commented):
I am checking this to get an idea of the use of the waveform extraction. Some comments:

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1464,178 @@ def write_sorting(
...
+        Each string in this list that matches a spike feature will not be written to the NWBFile.
+    write_as: str (optional, defaults to 'units')
+        How to save the units table in the nwb file. Options:
+        - 'units' will save it to the official NWBFile.Units position; recommended only for the final form of the data.
+        - 'processing' will save it to the processing module to serve as a historical provenance for the official table.
+    units_name : str (optional, defaults to 'units')
+        The name of the units table. If write_as=='units', then units_name must also be 'units'.
+    units_description : str (optional)
+    """
+    if metadata is None:
```
h-mayorquin, line 1537:
```suggestion
    metadata =  metadata if metadata is not None else {}
```

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1464,178 @@ def write_sorting(
...
+        metadata = dict()
+    assert (
+        nwbfile_path is None or nwbfile is None
+    ), "Either pass a nwbfile_path location, or nwbfile object, but not both!"
+    if nwbfile is not None:
+        assert isinstance(nwbfile, pynwb.NWBFile), "'nwbfile' should be a pynwb.NWBFile object!"
+
```
h-mayorquin, line 1544: I think these assertions (1539 to 1543) are not necessary here, the `make_or_load_nwbfile` context manager already has those checks.

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1464,178 @@ def write_sorting(
...
+                tmp_properties.append(prop)
+
+    assert write_as in [
+        "units",
+        "processing",
+    ], f"Argument write_as ({write_as}) should be one of 'units' or 'processing'!"
+    if write_as == "units":
+        assert units_name == "units", "When writing to the nwbfile.units table, the name of the table must be 'units'!"
+    write_in_processing_module = False if write_as == "units" else True
+
```
h-mayorquin, line 1577: Similarly for those checks here (1570 to 1575) which are done in the `add_units_table` function.

h-mayorquin (commented):
I have also two big comments at the level of design:
1) Right now we have a strange structure in this module. On one hand, we have that the function `add_units_table` has an argument `write_waveforms` which now is only supposed to work with the legaly spikeextractors. On the other hand, we have this new method `write_waveforms` which works for the new spikeinterface but uses the `add_units_table` method but with `write_waveforms=False` (!). In brief, their design logic oppose each other. I think that the current form is better but we should probably come with a deprecation plan for the old functionality.
2) I don't think is a good idea to add the possibility of writing the electrical series with the `write_waveforms`. It goes against the zen on python with `There should be one-- and preferably only one --obvious way to do it`. If somebody wants to add the electrical series they can use the corresponding method `write_recording` with the recording that they already have. What's the point of adding a less straighforward way of doing the same? The drawbacks are increasing the cyclomatic complexity of the code further and adding even more arguments to a function that already has too many making it harder for end users.

CodyCBakerPhD:
> I think that the current form is better but we should probably come with a deprecation plan for the old functionality.

Already have - the previous forms will be deprecated

CodyCBakerPhD:
> I don't think is a good idea to add the possibility of writing the electrical series with the write_waveforms.

This is, however, the main draw to this function and the main use case @alejoe91 wants to get out of this. In the modern SI workflows, this allows the NWB file to be written from a single call at the end of the pipeline

h-mayorquin:
> > I don't think is a good idea to add the possibility of writing the electrical series with the write_waveforms.
> 
> This is, however, the main draw to this function and the main use case @alejoe91 wants to get out of this. In the modern SI workflows, this allows the NWB file to be written from a single call at the end of the pipeline

If this is the intended use case then we should replicated the same structure that we have for writing recording objects. That is, we can have an extra function `add_waveforms` (analogous to `add_electrical_series`) which has a single and clear defined purpose and then have this function `write_waveforms` (analogous to `write_recording` and `write_sorting`) which does everything that we think the end end-user wants to do with a single call. 

Decoupling in this way will make it easier for us to test and mantain instead of burying the functionality and use cases inside parameters of only one function. Plus, this has the benefit of being consistent with our module structure for both writing recording and sorting objects.

CodyCBakerPhD:
> If this is the intended use case then we should replicate the same structure that we have for writing recording objects. That is, we can have an extra function add_waveforms (analogous to add_electrical_series) which has a single and clear defined purpose and then have this function write_waveforms (analogous to write_recording and write_sorting) which does everything that we think the end end-user wants to do with a single call.

Yeah, fine by me. `write_*` functions are routers to multiple `add_*` calls as per whatever convenience is desired.

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1464,178 @@ def write_sorting(
...
+                if write_electrical_series_kwargs is None:
+                    write_electrical_series_kwargs = {}
+                write_recording(
+                    recording=recording, nwbfile=nwbfile_out, metadata=metadata, **write_electrical_series_kwargs
+                )
+            else:
+                # we need add_electrodes_from_waveforms
+                if not waveform_extractor_has_recording(waveform_extractor):
+                    if recording is not None:
+                        assert isinstance(recording, BaseRecording), "'recording' needs to be a BaseRecording object."
```
CodyCBakerPhD, line 1599: So silly and petty, I know, but you're SO close to 100% coverage on the new code additions for this PR - can you add a test for this assertion?

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1464,178 @@ def write_sorting(
...
+
+    assert write_as in [
+        "units",
+        "processing",
+    ], f"Argument write_as ({write_as}) should be one of 'units' or 'processing'!"
+    if write_as == "units":
+        assert units_name == "units", "When writing to the nwbfile.units table, the name of the table must be 'units'!"
+    write_in_processing_module = False if write_as == "units" else True
+
+    try:
```
CodyCBakerPhD, line 1578: @alejoe91 is it possible to encounter a failure betwen line 1568 and 1578 (here)? If so I'd extend the `try` range so the `finally` can trigger

src/neuroconv/tools/spikeinterface/spikeinterface.py
h-mayorquin: This block ends in the part that goes to `write_electrical_series`. I think that that is somwerhere deeper in the stack (probably in `add_electrical_series`) where the responsible for checking correct inputs should be checked.

alejoe91: sure!

tests/test_ecephys/test_tools_spikeinterface.py
alejoe91: the `write_*` function was done to save the created files for inspection. For the same reason, I made more files (so that I could save them with the proper names). I'll remove them and keep the usual design

src/neuroconv/tools/spikeinterface/spikeinterface.py
alejoe91: actually, those checks are in the `write_sorting`, not in the `add_units_table`

src/neuroconv/tools/spikeinterface/spikeinterface.py
alejoe91: ok

alejoe91:
> > If this is the intended use case then we should replicate the same structure that we have for writing recording objects. That is, we can have an extra function add_waveforms (analogous to add_electrical_series) which has a single and clear defined purpose and then have this function write_waveforms (analogous to write_recording and write_sorting) which does everything that we think the end end-user wants to do with a single call.
> 
> Yeah, fine by me. `write_*` functions are routers to multiple `add_*` calls as per whatever convenience is desired.

The `add_waveforms` doesn't make sense in my opinion, because waveforms need to be added to the units table. Therefore, the `write_waveforms()` function simply calls the `add_units_table()` adding the arguments to add waveform columns.

src/neuroconv/tools/spikeinterface/spikeinterface.py
alejoe91: I agree. Removing the assertion and using the `add_electrical_series` function instead of the `write_*`

tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -1009,6 +1028,173 @@ def test_write_units_table_in_processing_module(self):
...
+                nwbfile=self.nwbfile4,
+                recording=self.we_recless_recording,
+                write_electrical_series=True,
+            )
+
+    def test_cleanup_tmp_properties(self):
+        # test that if write_waveforms fails, temporary properties are cleaned up
+        try:
+            write_waveforms(waveform_extractor=self.we_recless, nwbfile=self.nwbfile5, recording=None)
+        except:
```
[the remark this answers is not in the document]
alejoe91: it does! The `write_waveforms` will trigger an exception because:
1. Waveform extractor is recordingless
2. `recording` is `None`

In that case we check that temporary properties are properly cleaned up

alejoe91:
> I have also two big comments at the level of design:
> 
> 1. Right now we have a strange structure in this module. On one hand, we have that the function `add_units_table` has an argument `write_waveforms` which now is only supposed to work with the legaly spikeextractors. On the other hand, we have this new method `write_waveforms` which works for the new spikeinterface but uses the `add_units_table` method but with `write_waveforms=False` (!). In brief, their design logic oppose each other. I think that the current form is better but we should probably come with a deprecation plan for the old functionality.

The old `write_waveforms` will be deprecated in the next release when we drop support for `spikeextractors` objects.

> 2. I don't think is a good idea to add the possibility of writing the electrical series in the `write_waveforms` function (the argument `write_electrical_series). It goes against the zen on python with `There should be one-- and preferably only one --obvious way to do it`. If somebody wants to add the electrical series they can use the corresponding method `write_recording` with the recording that they already have. What's the point of adding a less straighforward way of doing the same? The drawbacks are increasing the cyclomatic complexity of the code further and adding even more arguments to a function that already has too many making it harder for end users.

This is an important point. When writing waveforms, we have to be aware of the electrodes that were added to the table to correctly link the electrode region (coming in a follow up PR). This can be done automatically from the `write_waveforms` function because it has access to both the recording and the sorting object. If that was not the case, the user would need to:
1. `write_recording()`
2. get electrode table ids corresponding to written electrodes
3. pass this information to the `write_waveforms()` function for correct linking

I think that having a single function doing all this is a better design choice for the end-user.

src/neuroconv/tools/spikeinterface/spikeinterface.py
h-mayorquin: I stand corrected. I think they should be in the `add_units_table` to de-duplicate.

h-mayorquin:
@alejoe91 
> The add_waveforms doesn't make sense in my opinion, because waveforms need to be added to the units table. Therefore, the write_waveforms() function simply calls the add_units_table() adding the arguments to add waveform columns.

The `add_electrical_series` also calls `add_electrodes`. A builder function should write also all the infraestructure needed to write correctly:
- `add_electrical_series` calls `add_electrodes`
- `add_waveforms` calls  `add_units_table` and ``add_electrodes` ` (for correctly link the electrode region as you point in the other comment).

> This is an important point. When writing waveforms, we have to be aware of the electrodes that were added to the table to correctly link the electrode region (coming in a follow up PR). This can be done automatically from the write_waveforms function because it has access to both the recording and the sorting object. If that was not the case, the user would need to:

>write_recording()
>get electrode table ids corresponding to written electrodes
>pass this information to the write_waveforms() function for correct linking

Isn't the information you require contained in the elctrode table and not the electrical series? The function a `add_electrical_series` already uses this pattern:
https://github.com/catalystneuro/neuroconv/blob/53a75e9f4dfcda31a0f3fa25931e99a7e00da31e/src/neuroconv/tools/spikeinterface/spikeinterface.py#L674-L681

h-mayorquin:
@alejoe91 
> I think that having a single function doing all this is a better design choice for the end-user.

But we will have the two functions. If you want a full function that will do everything we can use `write_waveforms` for that. We also provide a builder function, `add_waveforms` that attaches the object to an in-memory nwbfile without any other side effect. We already follow this pattern for the recording and sorting pipelines, and in general it is a good design principle, to separate responsabilities into different pieces of the code:

For recording: writing and the single function pipeline is done in `write_electrical_series` while building the actual objects in an in-memory file is done by `add_electrical_series`, `add_electrodes` , etc .
For sorting: writing and the single fuction pipeline is provided by `write_sorting` whereas `add_units_table` builds the dynamic table in an in-memory file.

The adavantage is that we can separate building the object and attaching it to nwb from any other functionality that we desire to implement in the full pipeline (such as creating an nwbfile and writing it with the `make_or_load_nwbfile` context manager or also calling `add_electrical_series`). Better modularity, easier testing, code with less conditional structures and less parameters.

alejoe91:
@CodyCBakerPhD 

@h-mayorquin and I met and we refactored the `write_waveforms` function so that we have an additional `add_waveforms` helper function. Let us know what you think! 

Apart from it, I think it's ready to go n my side :)

tests/test_ecephys/test_tools_spikeinterface.py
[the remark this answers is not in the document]
CodyCBakerPhD: I know, but as a test it can pass 'succesfully' if, due to a potential future bug, that exception does not get raised, in which case no assertions on the cleanup are ever made. If you use `finally`, the cleanup is asserted independently of whether or not the exception raise itself was successful.

An even stronger test, if you want to keep this structure, would be to alias the exception like

```python
exception Exception as exc:
    exception_type = type(exc)
    exception_message = str(exc)
    ...
```

and then outside the `try/except` block, also assert that the exception was raised as you intended. Even with that approach, I'd still recommend

```python
try:
    # write waveforms with intentionally forced error
except Exception as exc:
    # grab exception details
finally:
    # assert cleanup - important thing is this line will ALWAYS run
#assert exception details match expectation
```

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1464,223 @@ def write_sorting(
...
+        How to save the units table in the nwb file. Options:
+        - 'units' will save it to the official NWBFile.Units position; recommended only for the final form of the data.
+        - 'processing' will save it to the processing module to serve as a historical provenance for the official table.
+    units_name : str (optional, defaults to 'units')
+        The name of the units table. If write_as=='units', then units_name must also be 'units'.
+    units_description : str (optional)
+    """
+    metadata = metadata if metadata is not None else dict()
+
+    # try:
```
CodyCBakerPhD, line 1640:
```suggestion
    metadata = metadata or dict()

```

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1436,3 +1464,223 @@ def write_sorting(
...
+        if waveform_extractor_has_recording(waveform_extractor):
+            recording = waveform_extractor.recording
+        assert recording is not None, (
+            "recording not found. To add the electrode table, the waveform_extractor "
+            "needs to have a recording attached or the 'recording' argument needs to be used."
+        )
+
+        if write_electrical_series:
+            if write_electrical_series_kwargs is None:
+                write_electrical_series_kwargs = {}
```
CodyCBakerPhD, line 1653:
```suggestion
            write_electrical_series_kwargs  = write_electrical_series_kwargs  or dict()
```

tests/test_ecephys/test_tools_spikeinterface.py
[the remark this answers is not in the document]
CodyCBakerPhD: Although, is this even still necessary since it seems as if you've returned to using a copy of the object rather than temporary properties?

CodyCBakerPhD (approved):

CodyCBakerPhD:
@alejoe91 Great, thanks for all the work on this

PR #218, opened 2022-11-24
Add MEArecRecordingInterface

### LINKED ISSUE #213: [New Format]: MEArec recording

MEArec recordings

https://mearec.readthedocs.io/en/latest/usage.html

### Existing APIs for format

Neo and SI

https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/mearec

Yes.

### Code of Conduct

CodyCBakerPhD:
fix #213 

## Motivation

Adding a MEArecRecordingInterface.

Interesting to note, data storage type is `float32` with microVolt units (twice the itemsize of the usual `int16`).

There's a fair amount of system configuration values in the file. Especially for waveforms and spikes (which will belong to a sister PR).

Attaching the recording specific metadata to the recording description as a JSON serialization, but we should really start thinking harder about a more general solution at the NWB level. Maybe an optional attribute to certain tables, called `config`, that itself can be a `DynamicTable`? The work @weiglszonja did on the `ndx-extract` extension where such things are explicitly defined in a schema as attributes of the tables looks the best but takes the most work and doesn't scale well to general cases. Will think about it.

It also has a probe device set automatically through SI - unsure how general this is or if it is something available only for the example file on GIN. But supporting ProbeInterface integration in NeuroConv has been on to-do list for a while (original motivation from a project at the User Days that the originator never had time to get around to finishing), so I'll start on that in a follow-up in the next month or so.

Sadly, in all that metadata, it does not appear that session start time is available.

Example output file: shared on slack b/c too big for GitHub (likely due to GIN data not being truncated below 10MB)

## How to test the behavior?

Test on GIN data added. Also for the metadata fetching.

## Checklist


src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -20,6 +20,7 @@ def __init__(self, verbose: bool = True, **source_data):
         self.recording_extractor = self.Extractor(**source_data)
         self.subset_channels = None
         self.verbose = verbose
+        self.es_key = None  # For automatic metadata extraction
```
CodyCBakerPhD, line 23: This is usually something that has to be manually overridden at the call to `run_conversion` to allow custom metadata. For automatically fetched electrical series metadata, I think it would help to have this automatically set and propagate without user input

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -180,7 +181,7 @@ def run_conversion(
             use_times=use_times,
             write_as=write_as,
             write_electrical_series=write_electrical_series,
-            es_key=es_key,
+            es_key=es_key or self.es_key,
```
CodyCBakerPhD, line 184: from https://github.com/catalystneuro/neuroconv/pull/218/files#r1031812255, of course they can always override it and that will take priority at this line

src/neuroconv/datainterfaces/ecephys/spikeinterface/sipickledatainterfaces.py
```diff
@@ -17,6 +17,7 @@ def __init__(self, file_path: FilePathType, verbose: bool = True):
         self.subset_channels = None
         self.source_data = dict(file_path=file_path)
         self.verbose = verbose
+        self.es_key = None
```
CodyCBakerPhD, line 20: Temporary fix here since this interface does not call `super`.

Also Alessio said they no longer use this and would be OK to remove, so I'll do that in a follow-up

bendichter:
I agree, maybe we could add a `config` optional arg to `Device`?

src/neuroconv/utils/json_schema.py
```diff
@@ -14,17 +14,25 @@
 
 
 class NWBMetaDataEncoder(json.JSONEncoder):
-    def default(self, o):
+    def default(self, obj):
```
CodyCBakerPhD, line 17: Opinion: `obj` slightly more descriptive than `o`. Can't use `object` as a name since that's a special Python built-in. It's also only used by `json` library as a positional argument so what we name it here shouldn't matter as long as it's easily readable

src/neuroconv/utils/json_schema.py
```diff
@@ -14,17 +14,25 @@
...
+        if isinstance(obj, np.generic):
+            return obj.item()
+
+        # Numpy-versions of various numeric types
+        if isinstance(obj, np.integer):
+            return int(obj)
+        if isinstance(obj, np.floating):
+            return float(obj)
+        if isinstance(obj, np.ndarray):
+            return obj.tolist()
```
CodyCBakerPhD, line 32: MEArec header had a bunch of values casted as `numpy` data types, which requires this decoder mapping to serialize

CodyCBakerPhD:
> I agree, maybe we could add a config optional arg to Device?

Hmm... well, another case would be what if they have multiple series from the same device but they changed various configuration parameters between the segments? Then whatever config we attach to the device would only apply to a subset of the series

That's why I'm, for the moment, attaching it to the `description` of the `ElectricalSeries` that the recorded config pairs with

bendichter:
I'm noticing some issues with the test files:

* `/acquisition/ElectricalSeries/channel_conversion` should be omitted if they are all equal to 1.0.
* `general/extracellular_ephys/electrodes`: `imp`, `location` `x`, `y` and `z` are now optional and should be omitted if no data is present.

CodyCBakerPhD:
> general/extracellular_ephys/electrodes: imp, location x, y and z are now optional and should be omitted if no data is present.

This is a global to-do item for all `RecordingInterfaces`. See Slack post about currently complications, but yes we should do it - no, it's not specific to this format or this PR.


> /acquisition/ElectricalSeries/channel_conversion should be omitted if they are all equal to 1.0.

Yeah, good catch. That one is very likely specific to this format, so will fix it here.

CodyCBakerPhD:
> Yeah, good catch. That one is very likely specific to this format, so will fix it here.

OK should be fixed now. Had to adjust a couple other ground truths for this new behavior.

bendichter (approved):
lgtm!

PR #221, opened 2022-11-28
[WIP]: Add MaxTwoRecordingInterface

### LINKED ISSUE #215: [New Format]: Maxwell

Maxwell

https://www.mxwbio.com/res/MaxOne_brochure.pdf but group structures of HDF5 file is not clear here

### Existing APIs for format

Neo and SI

https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/maxwell

Yes.

### Code of Conduct

CodyCBakerPhD:
fix #215 

WIP while considering best way to handle certain issues, like the HDF5 plugin stuff. Sometimes it also causes a segmentation fault during the test runs.

EDIT: splintered MaxOne into #222 - still encountering various segmentation fault issues with MaxTwo that are likely to take a while to pin down and resolve. Putting this on the back burner for now.

## Motivation

Adding a MaxOneRecordingInterface and MaxTwoRecordingInterface. They are two separate versions of the file format. The second one is like a nested version of the first but therefore has sufficiently different structure and input arguments to merit having a separate interface. In particular, the MaxTwo has multi-stream behavior much like Blackrock (though where Blackrock saves the different streams to different files, these are all combined into a single file) and thus following the precedence established there where the interface connects to only a single stream at a time (since each stream may actually contain different 'types' of data, like LFP or environmental electrodes) since and it is the users responsibility to identify which count for the `RecordingInterface`.

That's only the first difficulty. As indicated by the comments at https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/maxwellrawio.py#L214-L226, reading any data from these files requires setting up a specific HDF5 compression plugin ahead of time. Still thinking about the best way to handle that for the user (right now, assuming they set it up themselves ahead of time).

No rich metadata/annotations besides one little thing about the version of Maxwell used to run the recordings. No session start time.

Example output file: [MaxOneRecordingInterface_maxone.zip](https://github.com/catalystneuro/neuroconv/files/10099488/MaxOneRecordingInterface_maxone.zip) and [MaxTwoRecordingInterface_maxtwo-recording_name=rec0000-stream_name=well000.zip](https://github.com/catalystneuro/neuroconv/files/10099493/MaxTwoRecordingInterface_maxtwo-recording_name.rec0000-stream_name.well000.zip)

## How to test the behavior?

Test on GIN data added. Also spanning the entire set of recording session + stream IDs (SI only tests a single pair).

## Checklist


h-mayorquin:
> Adding our interest to this issue from the Braingeneers / UCSC lab. We are generating max2 6 and 24 well datasets. Our current process is to split max2 datasets to max1 format and use the Max1 converter.

Hi, I missed this somehow.

Do you have data that you could share so we can build a converter.

CodyCBakerPhD:
@h-mayorquin There is example data towards the bottom of the PR description; I swear I added it to the GIN suite as well

h-mayorquin:
> @h-mayorquin There is example data towards the bottom of the PR description; I swear I added it to the GIN suite as well

I confirmed, there is data on gin but did not find a commit associated to you.

PR #222, opened 2022-11-28
Add MaxOneRecordingInterface

### LINKED ISSUE #215: [New Format]: Maxwell

CodyCBakerPhD:
fix #215 

Splinter of #221 specific to MaxOne; MaxTwo is causing problems and there are higher priority items to work on

## Motivation

Adding a MaxOneRecordingInterface.

As indicated by the comments at https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/maxwellrawio.py#L214-L226, reading any data from these files requires setting up a specific HDF5 compression plugin ahead of time. Still thinking about the best way to handle that for the user (right now, assuming they set it up themselves ahead of time).

No rich metadata/annotations besides one little thing about the version of Maxwell used to run the recordings. No session start time.

Example output file: [MaxOneRecordingInterface_maxone.zip](https://github.com/catalystneuro/neuroconv/files/10099488/MaxOneRecordingInterface_maxone.zip)

## How to test the behavior?

Test on GIN data added.

## Checklist


CodyCBakerPhD:
And now this one is hitting a snag with the CI not being able to manage the HDF5 plugins... tests run fine locally however. Sending to backlog.

bendichter:
I take it a simple `pip install hdf5plugin` doesn't work?

CodyCBakerPhD:
> 

It's not a standard offering apparently - supposedly the SI class method should handle it but that's where I left it off: https://github.com/catalystneuro/neuroconv/pull/222/files#diff-366373c600f3f3c6f8ad5913adc768bd36084efebf3ac10737cd9581ffcfcebfR22

bendichter:
> It's not a standard offering apparently

What do you mean by that? I see that class method, but it looks like an unusual way to install a dependency. What happened when you tried to pip install it?

CodyCBakerPhD:
Ah, I recall now

This works fine locally when you set the environment variable for HDF5 plugins after doing the installation, problem is getting it to work in the CI

> I take it a simple pip install hdf5plugin doesn't work?

It does not. You need to install the specific plugin from the helper function https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/maxwellrawio.py#L229-L254 that pulls from the maxwell site

If you don't you get the informative error

```
Maxwell file format is based on HDF5.
The internal compression requires a custom plugin!!!
This is a big pain for the end user.
You, as a end user, should ask Maxwell company to change this.
Please visit this page and install the missing decompression libraries:
https://share.mxwbio.com/d/4742248b2e674a85be97/
Then, link the decompression library by setting the `HDF5_PLUGIN_PATH` to your
installation location, e.g. via
os.environ['HDF5_PLUGIN_PATH'] = '/path/to/cutum/hdf5/plugin/'
Alternatively, you can use the auto_install_maxwell_hdf5_compression_plugin() below
function that do it automagically.
```

Note all this only happens when you first try to *read* data from the IO - you can initialize everything fine, even get metadata, it's just the first time you call `get_traces` that it errors out in `h5py`

CodyCBakerPhD:
In particular, there's separate errors that occur in each condition

If you have `hdf5plugin` installed but not their specific compressor you get

```python
OSError: Can't read data (required filter 'mxw-data' is not registered)
```

whereas if you don't have `hdf5plugin` installed and don't have the environment variable set it simply tells you it can't read the block of data

CodyCBakerPhD:
But interestingly you don't need `hdf5plugin` installed to read the data, just their specific compressor plus the environment variable set.

This is a pretty odd format lol

bendichter:
> To get this through we could always just make it a linux-specific interface for the time being and try to resolve this later?

That sounds fine to me. Cloud-based deployments will be on Linux anyway

src/neuroconv/datainterfaces/ecephys/maxwell/maxonedatainterface.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Parameters
+        ----------
+        hdf5_plugin_path : string or Path, optional
+            Path to your systems HDF5 plugin library.
+            Uses the home directory by default.
+        download_plugin : boolean, default: True
+            Whether or not to force download of the decompression plugin.
+            It's a very lightweight install but does require an internet connection.
+            This is left as True for seamless passive usage and should not impact performance.
```
CodyCBakerPhD, line 34: Note that `neo` takes the same approach by default, for conveneience: https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/maxwellrawio.py#L229

If you swap the convention, then it introduces a workflow the user has to go through where they need to manage the extra installation and setup before instantiating the interface

CodyCBakerPhD: But there might be cases where the user wants or needs that control, hence why I'm including this as an argument where informed users may disable it by setting it to `False` and managing it the way they want

src/neuroconv/datainterfaces/ecephys/maxwell/maxonedatainterface.py
```diff
@@ -0,0 +1,84 @@
...
+        if system() != "Linux":
+            raise NotImplementedError(
+                "The MaxOneRecordingInterface has not yet been implemented for systems other than Linux."
+            )
+
+        if not os.environ.get("HDF5_PLUGIN_PATH", ""):
+            hdf5_plugin_path = hdf5_plugin_path or Path.home() / "hdf5_plugin_path_maxwell"
+            if not hdf5_plugin_path.exists():
+                self.auto_install_maxwell_hdf5_compression_plugin(hdf5_plugin_path=hdf5_plugin_path)
+            os.environ["HDF5_PLUGIN_PATH"] = str(hdf5_plugin_path)
```
CodyCBakerPhD, line 74: OK, this is one of the oddest formats I've ever worked with...

If this 5 lines are not exactly run like this, and at this level, the CI (even restricted to ubuntu only) runs into sporadic segmentation faults but otherwise does not complain as it usually would about not being able to decompress the values in the HDF5 dataset

bendichter (approved):

PR #225, opened 2022-11-28
improve docstring of SpikeGLXRecordingInterface

bendichter:


src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -48,10 +46,22 @@ def get_source_schema(cls):
     def __init__(
         self,
         file_path: FilePathType,
-        stub_test: Optional[bool] = False,
-        spikeextractors_backend: Optional[bool] = False,
```
bendichter, line 52: `Optional` should only be used when the arg can be `None`.

CodyCBakerPhD (approved):

PR #232, opened 2022-11-28
improve SpikeGLX gallery to include conversion of lf data

### LINKED ISSUE #227: [Documentation]: Add SpikeGLXLFPInterface to conversion gallery

As it stands, the SpikeGLX gallery only has the `SpikeGLXRecordingInterface`, which handles the ap.bin data. We should expand this to also handle the lf.bin data using the `SpikeGLXLFPInterface`.

Maybe.

### Code of Conduct

bendichter:
## Motivation

fix #227 

## Checklist


CodyCBakerPhD:
Usage of a `ConverterPipe` makes this a 'combined' example so should belong in that corner of the gallery. This portion of the gallery is for a single stream.

I'd say either add it to a separate column for supported LFP interfaces, or add a separate code block below the AP section showing the analogous set of commands for a standalone LFP conversion.

This also serves as the first example of a gallery for LFP interfaces, so lets make sure to arrive on a nice agreed-upon organization structure so it's really easy to add more later on

bendichter:
Do we really expect anyone to one to convert ap but not lf data for spikeGLX recordings though?

CodyCBakerPhD:
> Do we really expect anyone to one to convert ap but not lf data for spikeGLX recordings though?

Possible - what if they just want to run spike sorting on it (by, for instance, making use of DANDI storage and future possible automated workflows which would require it to be in `.nwb` instead of the original `.ap.bin`)? Then the LF portion wouldn't be used AFAIK

There's also the other way around -  they might want to convert just the LF portion to keep their NWB conversion lightweight. (I know we'd prefer they package both together, so feel free to include the LFP in the existing combination for SpikeGLX: https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/combinations/spikeglx_and_phy.html)

bendichter:
I was thinking about these gallery pages as guides to different data source formats, not to data interfaces specifically. Most data sources can be handled by a single data interface, but this would be an exception. While I would agree that there are reasons to package this data with only a single spikeGLX stream, I don't think that would be what I would recommend for most users. I'd like to gear this towards providing a good template for conversion in the most common and straightforward scenario of data publishing, in which case I'd really like to encourage users to package both the ap and lf streams. I'd rather create a full conversion where it's clear how you would comment out lines if you don't want to include both streams.

CodyCBakerPhD:
I see, that's fine then - but we should eventually combine other multi-modal formats in a similar way for consistency. Blackrock, AlphaOmega, Axona, Neuralynx, and Neuroscope all have similarly packaged streams for raw, lfp, sorted, (and when we have better waveform support, those as well) data.

bendichter:
Yeah, we have separated out raw and spike sorted for now, but maybe it would make more sense to combine them. Another option is to premake e.g. a `SpikeGLXConverterPipe` that takes in paths and has an arg `streams` that is be default `["ap", "lf"]` but can be changed to just `["ap"]` or just `["lf"]`.

CodyCBakerPhD:
Replaced by #424

PR #229, opened 2022-11-28
[WIP]: Adding start time alignment infrastructure

CodyCBakerPhD:
Still WIP, working/thinking on design and thorough testing

PR #237, opened 2022-12-04
[Sync IV] Integrate Temporal Alignment for ecephys Interfaces and TimeIntervals

CodyCBakerPhD:
Replace #229

## Motivation

Extension of #245 and #244 that integrates the tools developed and tested in those PRs into the common API for all data interfaces. The `BaseDataInterface` defines the methods abstractly but raises `NotImplementedErrors` so that they will have to be defined across child interfaces.

Introduces three synchronization methods for all interfaces:

1) shift of starting time
2) full replacement of timestamps with precisely synchronized timing information
3) inference of synchronized timestamps given the current unsynchronized timestamps within the interface and two references times allowing a mapping from the secondary timing system to the primary one

Still need to:
- [x] Copy the functionality to some behavior interfaces, like DLC used in the tests
- [x] Expand tests
- [x] Open [Sync V], the documentation PR

## How to test the behavior?
Tests are coming together harnessing the synthetic TTL tools established in #245 and #244 - will need to exhaustively check behavior across all supported interfaces, however.

## Checklist


src/neuroconv/basedatainterface.py
```diff
@@ -46,6 +49,30 @@ def get_metadata(self):
...
+        timestamps: ArrayType
+            A full vector of timestamps all temporal data in this interface.
+        """
+        if starting_time is not None or timestamps is not None:
+            assert starting_time != timestamps, "Specify either 'starting_time' or 'timestamps', but not both!"
+
+        if starting_time is not None:
+            self.starting_time = starting_time
+        elif timestamps is not None:
+            self.timestamps = np.array(timestamps)
```
bendichter, line 74: I'd actually rather we throw a NotImplemented error here. Otherwise we'll have a seemingly functional `synchronize` method on all children that doesn't actually do anything. It would be find to define a method like `synchronization_validation` that always gets called first and put this code there.

src/neuroconv/basedatainterface.py
```diff
@@ -46,6 +49,30 @@ def get_metadata(self):
...
+
+        Parameters
+        ----------
+        starting_time: float
+            The starting time for all temporal data in this interface.
+        timestamps: ArrayType
+            A full vector of timestamps all temporal data in this interface.
+        """
+        if starting_time is not None or timestamps is not None:
+            assert starting_time != timestamps, "Specify either 'starting_time' or 'timestamps', but not both!"
```
bendichter, line 69: why the assert statement? Why not just throw an error? Are you permitting `timestamps == starting_time` as valid? Why?

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -72,6 +71,15 @@ def get_metadata(self):
...
         return metadata
 
+    def synchronize(self, starting_time: Optional[float] = None, timestamps: Optional[ArrayType] = None):
+        if starting_time is not None or timestamps is not None:
+            assert starting_time != timestamps, "Specify either 'starting_time' or 'timestamps', but not both!"
+
+        if starting_time is not None:
+            self.recording_extractor.set_times(self.recording_extractor.frame_to_times() + starting_time)
+        elif timestamps is not None:
+            self.recording_extractor.set_times(timestamps)
```
bendichter, line 81: There's really no point in defining that abstract method if you're not even going to call it in child classes

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # Electrodes columns descriptions
+        metadata["Ecephys"]["Electrodes"] = [
+            dict(name="group_name", description="Name of the ElectrodeGroup this electrode is a part of."),
+        ]
+
+        metadata["Ecephys"]["ElectricalSeriesSync"] = dict(
+            name="ElectricalSeriesSync", description="Raw acquisition traces for the synchronization (nidq) data."
+        )
+        return metadata
```
bendichter, line 110: https://nwbinspector.readthedocs.io/en/dev/best_practices/time_series.html#subtypes

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -0,0 +1,137 @@
...
+            dLineList=channel,
+            meta=self._meta,
+        )
+        dig = np.squeeze(dig)
+        diff_dig = np.diff(dig.astype(int))
+
+        rising = np.where(diff_dig > 0)[0]
+        falling = np.where(diff_dig < 0)[0]
+
+        ttl_frames = np.concatenate((rising, falling))
```
bendichter, line 129: this line will fail if the recording finishes in the up position

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
CodyCBakerPhD: That's something we missed in the NeuroNexus review, then, too.

CodyCBakerPhD: This is quite unfortunate as well; there's a TON of metadata information about the configuration of the NIDQ device and since `TimeSeries` can't accept a `Device` link, I guess I'd be forced to dump all that info into the `TimeSeries.description` instead?

I guess I could also make a separate custom `DynamicTable` to serve as the metadata representation for this stuff - downside is it can't be linked to the series. Other suggestions?

For a more general approach though, it's also extremely intuitive for users (and even devs) to want to use something called `ElectricalSeries` to represent their data that is in units `Volts`; if the issue is less about the `ElectricalSeries` neurodata type, and more about how the `electrodes` table (not yet a neurodata type) is only supposed to specify electrode metadata for neurophysiological-related, then the best solution may simply be to make a proper `ElectrodeTable` type and follow it up with a separate child class that can apply to describing environmental electrodes and all their metadata.

CodyCBakerPhD:
Trying to re-trigger tests

CodyCBakerPhD:
OK fixed now, that was weird. It wasn't finding the workflow files so the deployment was failing, but no clear reason why it couldn't find them and why it can find them now 🤷‍♂️

bendichter (commented):
It's really coming together! Looks strong overall. See individual comments for suggestions

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -72,6 +72,19 @@ def get_metadata(self):
...
+
+    def synchronize_timestamps(self, timestamps: ArrayType):
+        self.recording_extractor.set_times(times=timestamps)
+
+    def synchronize_with_pulses(self, pulse_timestamps: ArrayType):
+        self.recording_extractor.set_times(
+            times=infer_timestamps_from_pulses(
+                relative_timestamps=self.recording_extractor.get_times(), pulse_timestamps=pulse_timestamps
+            )
+        )
```
bendichter, line 86:
```suggestion
    def synchronize_with_pulses(
        self,
        pulse_timestamps_parent: ArrayType,
        pulse_timestamps_child: ArrayType,
        timestamps_child: Optional[ArrayType] = None,
    ):
        if timestamps_child is None:
            timestamps_child = self.recording_extractor.get_times()
        self.recording_extractor.set_times(
            times=infer_timestamps_from_pulses(
                 pulse_timestamps_parent=pulse_timestamps_parent,
                 pulse_timestamps_child=pulse_timestamps_child
                timestamps_child=timestamps_child,
            )
        )
```

CodyCBakerPhD: Yeh, this one is still WIP

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -0,0 +1,144 @@
...
+        -------
+        rising_times: numpy.ndarray
+            The times of the rising TTL pulse; often used as the start time of the event.
+        falling_times: numpy.ndarray
+            The times of the falling TTL pulse; often used with the rising_times to perform decoding of digital words
+            which occur between the rise and fall.
+            Can also be used to track the duration of an on/off event.
+        """
+        # TODO: consider RAM cost of these operations and implement safer buffering version
+        rising_frames, _ = parse_rising_and_falling_frames_from_ttl(
```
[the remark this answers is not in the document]
CodyCBakerPhD: Moving dsicussion to [Sync II]

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -0,0 +1,144 @@
...
+            ]
+        metadata["Ecephys"].get("ElectrodeGroup", list()).extend(electrode_groups)
+
+        # Electrodes columns descriptions
+        metadata["Ecephys"]["Electrodes"] = [
+            dict(name="group_name", description="Name of the ElectrodeGroup this electrode is a part of."),
+        ]
+
+        metadata["Ecephys"]["ElectricalSeriesSync"] = dict(
+            name="ElectricalSeriesSync", description="Raw acquisition traces for the synchronization (nidq) data."
```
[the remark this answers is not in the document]
CodyCBakerPhD: Done

src/neuroconv/basedatainterface.py
```diff
@@ -46,6 +46,58 @@ def get_metadata(self):
...
+        ----------
+        timestamps: ArrayType
+            A full vector of timestamps all temporal data in this interface.
+        """
+        raise NotImplementedError(
+            "The protocol for synchronizing the timestamps of this interface has not been specified!"
+        )
+
+    @abstractmethod
+    def synchronize_with_pulses(self, pulse_timestamps: ArrayType):
```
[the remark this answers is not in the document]
CodyCBakerPhD: Renamed the method and input arguments - one thing to note is you may not be parsing any pulses on the NeuroConv side (we've seen several cases where you simply read a vector of the already-calculated reference times) so abstracted away from the notion these are tied to pulses.

Generally we simply have two sets of reference times, those from the primary system we're trying to sync the timestamps of this interface to, and those from the secondary system.

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
CodyCBakerPhD: I wouldn't say parent/child is the best way to refer to the relationship. They're systems acquiring in parallel, but we, or the experimenter, designates one to be the 'primary' timing system which then tracks reference times across the secondary systems.

I've adjusted the names here to reflect this primary/secondary notion, and also how if the timings from the secondary timing streams aren't reflective of exact event occurrences, they become 'reference times'.

tests/test_on_data/test_synchronization/test_interface_synchronization.py
```diff
@@ -0,0 +1,149 @@
...
+            self.nidq_interface.get_event_starting_times_from_ttl(
+                channel_name="nidq#XA2"  # the channel receiving pulses from the DLC system
+            ),
+        )
+
+        self.dlc_interface.synchronize_timestamps(timestamps=synchronized_dlc_timestamps)
+        self.behavior_interface.synchronize_between_systems(
+            primary_reference_timestamps=self.unsynchronized_secondary_timestamps,
+            secondary_reference_timestamps=synchronized_dlc_timestamps,
+        )
```
CodyCBakerPhD, line 75: This is the primary demonstration of how this kind of workflow would look like in an average conversion script using them

tests/test_on_data/test_synchronization/test_interface_synchronization.py
```diff
@@ -0,0 +1,149 @@
...
+
+        In this case, they simply store the timestamps in separate files and load them in during the conversion.
+        """
+        # For example, if the synchronization is performed ahead of time and timestamps stored in a CSV or similar file
+        # Then we just need to read in the pulse times instead of having to parse NIDQ signals
+        self.dlc_interface.synchronize_timestamps(timestamps=self.externally_synchronized_timestamps)
+        self.behavior_interface.synchronize_between_systems(
+            primary_reference_timestamps=self.unsynchronized_secondary_timestamps,
+            secondary_reference_timestamps=self.externally_synchronized_timestamps,
+        )
```
CodyCBakerPhD, line 129: This shows how the same setup works in exactly the same way even when the timestamps are given from an external file instead of explicit NIDQ parsing, due to this generalization away from pulses to timestamps (however they might have been parsed originally)

tests/test_on_data/test_synchronization/test_interface_synchronization.py
```diff
@@ -0,0 +1,149 @@
...
+        synchronized_dlc_timestamps = (
+            self.nidq_interface.get_event_starting_times_from_ttl(
+                channel_name="nidq#XA2"  # the channel receiving pulses from the DLC system
+            ),
+        )
+
+        self.dlc_interface.synchronize_timestamps(timestamps=synchronized_dlc_timestamps)
+        self.behavior_interface.synchronize_between_systems(
+            primary_reference_timestamps=synchronized_dlc_timestamps,
+            secondary_reference_timestamps=self.unsynchronized_secondary_timestamps,
```
CodyCBakerPhD, line 74: This is the primary demonstration of how this kind of workflow would look like in an average conversion script using them

tests/test_on_data/test_synchronization/test_interface_synchronization.py
```diff
@@ -0,0 +1,149 @@
...
+
+        In this case, they simply store the timestamps in separate files and load them in during the conversion.
+        """
+        # For example, if the synchronization is performed ahead of time and timestamps stored in a CSV or similar file
+        # Then we just need to read in the pulse times instead of having to parse NIDQ signals
+        self.dlc_interface.synchronize_timestamps(timestamps=self.externally_synchronized_timestamps)
+        self.behavior_interface.synchronize_between_systems(
+            primary_reference_timestamps=self.externally_synchronized_timestamps,
+            secondary_reference_timestamps=self.unsynchronized_secondary_timestamps,
+        )
```
CodyCBakerPhD, line 129: This shows how the same setup works in exactly the same way even when the timestamps are given from an external file instead of explicit NIDQ parsing, due to this generalization away from pulses to timestamps (however they might have been parsed originally)

tests/test_on_data/test_synchronization/test_interface_synchronization.py
```diff
@@ -0,0 +1,149 @@
...
+        cls.unsynchronized_dlc_timestamps = np.arange(
+            start=0.0,  # First DLC pulse sent when device turns on
+            stop=17.0,  # Was acquiring for 17 seconds
+            step=1 / 30.0,  # Sent a TTL pulse every on every frame at a rate of 30 fps
+        )
+        cls.synchronized_dlc_timestamps = np.arange(
+            start=3.23,  # DLC device starts acquiring 3.23 seconds after SpikeGLX
+            stop=20.23,
+            step=1 / 30 + 0.00001,  # Synchronization reveals slight drift compared with the SpikeGLX times
+        )
```
CodyCBakerPhD, line 24: For conversational reference to comments above.

Note this is not actually something we've seen with DLC before, I was just using that since it's an established behavioral interface in NeuroConv.

But the setup is taken from work we did with several custom interfaces in the Fee lab: https://github.com/catalystneuro/fee-lab-to-nwb/blob/main/src/fee_lab_to_nwb/happ_ecephys/happmotifinterface.py#L55-L67

Motif timings in a MATLAB file were given in the secondary time system of the audio recorder, and the audio stream had events identified and parsed (by them, timestamps recorded in a separate MATLAB file) so we had to estimate the true motif (and subsequent syllable) times in the primary system given both the IMEC and audio times as reference

tests/test_on_data/test_synchronization/test_interface_synchronization.py
```diff
@@ -0,0 +1,149 @@
...
+
+        self.dlc_interface.synchronize_timestamps(timestamps=synchronized_dlc_timestamps)
+        self.behavior_interface.synchronize_between_systems(
+            primary_reference_timestamps=self.nidq_interface.get_times(),
+            secondary_reference_timestamps=synchronized_dlc_timestamps,
+        )
+
+        assert_array_equal(x=self.dlc_interface.get_timestamps(), y=synchronized_dlc_timestamps)
+
+        expected_behavior_timestamps = np.array([8.83168, 10.53219, 12.93291])
```
CodyCBakerPhD, line 61: And then these are the expected timestamps for the behavioral events passed that are synchronized between the two systems

CodyCBakerPhD: (The relative unsynchronized times were `[5.6, 7.3, 9.7]`)

tests/test_on_data/test_synchronization/test_interface_synchronization.py
```diff
@@ -0,0 +1,149 @@
...
+            config_file_path=BEHAVIOR_DATA_PATH / "DLC" / "config.yaml",
+            subject_name="ind1",
+        )
+        self.dlc_interface.timestamps = self.unsynchronized_dlc_timestamps
+
+        self.behavior_interface = MockBehaviorEventInterface()
+
+    def test_nidq_interface_synchronized(self):
+        synchronized_dlc_timestamps = (
+            self.nidq_interface.get_event_starting_times_from_ttl(
```
CodyCBakerPhD, line 48: Could be `get_ttl_times`

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -72,6 +73,26 @@ def get_metadata(self):
...
+    def get_timestamps(self) -> np.ndarray:
+        return self.recording_extractor.get_times()
+
+    def synchronize_starting_time(self, starting_time: float):
+        self.recording_extractor.set_times(times=self.recording_extractor.get_times() + starting_time)
+
+    def synchronize_timestamps(self, synchronized_timestamps: ArrayType):
+        self.recording_extractor.set_times(times=synchronized_timestamps)
+
+    def synchronize_between_systems(
```
CodyCBakerPhD, line 85:
```suggestion
    def synchronize_by_interpolation(
```
?

src/neuroconv/basedatainterface.py
```diff
@@ -46,6 +47,74 @@ def get_metadata(self):
...
+            The starting time for all temporal data in this interface.
+        """
+        raise NotImplementedError(
+            "The protocol for synchronizing the start time of this interface has not been specified!"
+        )
+
+    @abstractmethod
+    def synchronize_timestamps(self, synchronized_timestamps: ArrayType):
+        """
+        Replace all timestamps for this interface with those syncrhonized with the primary system.
```
bendichter, line 83:
```suggestion
        Replace all timestamps for this interface with those synchronized with the primary system.
```

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -7,7 +7,8 @@
 from pynwb.ecephys import ElectrodeGroup
 
 from ...baseextractorinterface import BaseExtractorInterface
-from ...utils import get_schema_from_hdmf_class, get_base_schema, OptionalFilePathType
+from ...tools.signal_processing import synchronize_timestamps_between_systems
+from ...utils import get_schema_from_hdmf_class, get_base_schema, OptionalFilePathType, ArrayType
```
bendichter, line 11:
```suggestion
from ...utils import get_schema_from_hdmf_class, get_base_schema, FilePathType, ArrayType
```

bendichter: I'd like to phase out OptionalFilePathType in favor of Optional[FilePathType]

src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py
```diff
@@ -1,17 +1,20 @@
...
 from pynwb.device import Device
 from pynwb.ophys import ImagingPlane, TwoPhotonSeries
 
 from ...baseextractorinterface import BaseExtractorInterface
+from ...tools.signal_processing import synchronize_timestamps_between_systems
 from ...utils import (
     get_schema_from_hdmf_class,
     fill_defaults,
     get_base_schema,
     OptionalFilePathType,
```
bendichter, line 15:
```suggestion
    FilePathType,
```

src/neuroconv/tools/signal_processing.py
```diff
@@ -40,3 +42,31 @@ def parse_rising_and_falling_frames_from_ttl(trace: np.ndarray, num_bins: int =
...
+        A known set of timestamps in the primary system to be used as reference.
+    secondary_reference_timestamps : ArrayType
+        A known set of timestamps in the secondary system to be used as reference.
+
+    Returns
+    -------
+    synchronized_timestamps: numpy.ndarray
+        The interpolation of the unsynchronized_timestamps into the primary reference times.
+    """
+    return np.interp(x=unsynchronized_timestamps, xp=secondary_reference_timestamps, fp=primary_reference_timestamps)
```
bendichter, line 72: Since this is a one-liner, we could just get rid of this function and move `np.interp` to within the methods

CodyCBakerPhD:
@bendichter We had discussed changing the way we refer to these methods as 'temporal alignment' instead of 'synchronization' to avoid confusing it with the practice of interpolating data like MICrONS did.

Still agree with that? The three methods of this PR could then be

```
align_start_time
align_timestamps
align_by_interpolation
```

bendichter:
> @bendichter We had discussed changing the way we refer to these methods as 'temporal alignment' instead of 'synchronization' to avoid confusing it with the practice of interpolating data like MICrONS did.
> 
> Still agree with that? The three methods of this PR could then be
> 
> ```
> align_start_time
> align_timestamps
> align_by_interpolation
> ```

I'm not sure. Let's talk about it in our meeting

src/neuroconv/basedatainterface.py
```diff
@@ -46,6 +47,78 @@ def get_metadata(self):
...
+
+        Use this method if the unaligned timestamps of the data in this interface are not directly tracked by a primary
+        system, but are known to occur between timestamps that are tracked, then align the timestamps of this interface
+        by interpolating between the two.
+
+        An example could be a metronomic TTL pulse (e.g., every second) from a secondary data stream to the primary
+        timing system; if the time references of this interface are recorded within the relative time of the secondary
+        data stream, then their exact time in the primary system is inferred given the pulse times.
+
+        Must be in units seconds relative to the common 'session_start_time'.
```
CodyCBakerPhD, line 109: @alejoe91 As someone that often deals with this kind of stuff in experiments, how well does this explanation read for you? Any suggestions?

There will be a follow-up to this PR that focuses exclusively on documentation with some real examples, would you be OK with looking at that when the time comes?

src/neuroconv/tools/signal_processing.py
CodyCBakerPhD: Agreed

CodyCBakerPhD: (and done)

src/neuroconv/datainterfaces/ophys/basesegmentationextractorinterface.py
```diff
@@ -58,6 +60,26 @@ def get_metadata(self):
...
+        self, primary_reference_timestamps: ArrayType, secondary_reference_timestamps: ArrayType
+    ):
+        unsynchronized_timestamps = self.segmentation_extractor.get_times()
+        synchronized_timestamps = synchronize_timestamps_between_systems(
+            unsynchronized_timestamps=unsynchronized_timestamps,
+            primary_reference_timestamps=primary_reference_timestamps,
+            secondary_reference_timestamps=secondary_reference_timestamps,
+        )
+        self.segmentation_extractor.set_times(times=synchronized_timestamps)
+
```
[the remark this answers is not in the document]
CodyCBakerPhD: Turns out it's possible to describe the method purely in terms of the other two alignment methods plus the input arguments, so it's now a part of the base class so no repetition

CodyCBakerPhD:
Limiting the scope of this PR to just the `ecephys` interfaces plus the `CsvTimeIntervalsInterface`.

Deeplabcut and other behavioral interfaces will take some extra investigation to see how best to handle timestamps 

Tests are still being polished and debugged, should be ready for review by/on Monday

tests/test_on_data/test_synchronization/test_interface_synchronization.py
CodyCBakerPhD: Swapped from DLC to CsvTimeInterval's since that is actually a use case we've seen, or much closer to it anyway

CodyCBakerPhD:
Update: debugging took longer than expected as each new test case revealed something that needed a slight change to some components

As it stands now the features work in all the ways demonstrated by the `test_temporal_alignment` tests, however...

... there is a very mysterious issue that prevents Python 3.7 from properly aligning the behavior test interface using the interpolation method

I've combed through changelogs at the Python and Numpy levels and couldn't find anything that might be the cause

Unsure how much more time to invest in trying to understand the issue given we're expected to drop 3.7 soon anyway

CodyCBakerPhD:
- Drop Python 3.7

- Add retrieval for unaligned (or pre-aligned) timestamps vector

- In a follow-up PR, add `starting_time` support to bypass the wasted operation of generating a knowingly regular timestamp vector through SI/ROIExtractors (will require some minor rework to tools functions)

CodyCBakerPhD:
Added a method for retrieving the unaltered timestamps, calling it `get_original_timestamps` because we can't actually claim that they are 'unaligned' for any particular session (that's a determination to be made by the user) but we can guarantee that they are the version of the timestamps before calling any sequence of the alignment methods in this PR (which modify an in-place copy of those original timestamps)

tests/test_on_data/test_temporal_alignment/test_temporal_alignment_methods.py
```diff
@@ -0,0 +1,424 @@
...
+        )
+
+        # True stop times are not tracked, so estimate them from using the known regular trial length
+        self.trial_interface.align_timestamps(
+            aligned_timestamps=inferred_aligned_trial_start_timestamps + self.regular_trial_length, column="stop_time"
+        )
+
+        self.behavior_interface.align_by_interpolation(
+            aligned_timestamps=inferred_aligned_trial_start_timestamps,
+            unaligned_timestamps=self.unaligned_trial_start_times,
```
CodyCBakerPhD, line 106:
```suggestion
            unaligned_timestamps=self.trial_interface.get_original_timestamps(),
```

CodyCBakerPhD:
- [ ] add test for align start time

- [ ] investigate expected behavior time values (they seem off)

CodyCBakerPhD:
- Adjusted all tests to accurately portray usage of original timestamps when declared as 'unaligned'
- Fixed the behavior event ground truth values that seemed off
  - the problem was not an issue with the functionality, but rather the class setup didn't have the right trial alignment ground truth
- Added the requested functionality for the `TimeIntervalsInterface` to interpolate other columns when one is aligned
  - this is an optional, non-default behavior for safety
  - it is not uncommon to have TTL pulses tracking extra trial columns, so the default behavior is to assume the individual wants to align each column separately
  - it is a convenient method, but there are some things that can go wrong with it, such as how to handle the first or last value when beyond the unaligned range
    - it deals with this by using the `left`/`right` options of `np.interp` using the 'most recent' aligned difference to estimate the gap, but of course this may be wrong in some cases
    - this is likely a case where the linear regression approach could provide a better estimate

CodyCBakerPhD:
Also, here's a summary of the changelog, which is almost entirely new line additions...

\+ 91 lines to core `BaseDataInterface` defining these new alignment methods

\+ ~187 lines across each base, behavior, and some special testing interfaces (~17 lines per)  to define temporary methods that each raise `NotImplementedErrors` - this is needed because the new methods are abstract in the base class; and any new data interfaces implemented after the next release will also have to define these methods

\+ 70 lines to `TimeIntervalsInterface` with fully functional and tested implementations of alignment methods
\+ 50 new mock methods needed to test the features of this PR

\+ 527 demonstrative tests of the new functionality (more unit-like tests will be added in #283 and beyond)

src/neuroconv/basedatainterface.py
```diff
@@ -46,6 +47,91 @@ def get_metadata(self):
...
+        Parameters
+        ----------
+        unaligned_timestamps : numpy.ndarray
+            The timestamps of the unaligned secondary time basis.
+        aligned_timestamps : numpy.ndarray
+            The timestamps aligned to the primary time basis.
+        """
+        self.align_timestamps(
+            aligned_timestamps=np.interp(x=self.get_timestamps(), xp=unaligned_timestamps, fp=aligned_timestamps)
+        )
```
bendichter, line 133: I think the term "timestamps" is a bit overloaded here, because it refers to the sample times of the signal we are trying to align and to the times of the ttl pulses. What do you think about renaming:

unaligned_timestamps -> ttl_sent_times
aligned_timestamps -> ttl_received_times

or something like that

bendichter (approved):

src/neuroconv/basedatainterface.py
CodyCBakerPhD: `aligned_timestamps` is consistent with the keyword argument for `align_timestamps`, which is consistent with its usage in adjusting the `timestamps` field of the underlying object (when the object is a `TimeSeries`, at least)

Keep in mind the `timestamps` also may not be inferred only through TTL signals - it is also possible to decode the timings via more complicated digital words (compound TTL, each sent at different times and parsed together) or perhaps even some other mechanism (like recording the system time in UNIX seconds and saving it in a CSV)

So I'd say the argument names ought to be agnostic to the 'how' of the timings

If `timestamps` is too long, we can mimic SpikeInterface/ROIExtractors by just referring to it always as `times`

```
get_timestamps -> get_times
align_timestamps -> align_times
unaligned_timestamps -> unaligned_times
aligned_times -> aligned_times
```

in particular, `get_times` would be eventually refactored into that enhanced treatment of `starting_time` + `rate` vs. `timestamps` that will be eventually added to the write tools

CodyCBakerPhD:
Merging this to get it into the base - continuing discussion of further improvements in issues #303, #302, and #301

PR #241, opened 2022-12-07
add tutorial for tabular data

bendichter:
@CodyCBakerPhD, thoughts on this? Is there a way to simplify this?


CodyCBakerPhD (dismissed):

bendichter:
I'm flexible on the exact file format. I'd like a format that is easily exported from Google Sheets and Excel, because those are the apps I think users will be using to organize their data. Bonus points if the format is readable as plain text. Maybe csv I'd the best. They can probably use quotes to handle values with commas.

bendichter:
Google sheets can output in .tsv
image

But Excel only has .csv
image

bendichter:
@CodyCBakerPhD merging is blocked for me. I don't know why

CodyCBakerPhD (approved):

PR #242, opened 2022-12-07
Add BaseDataInterface API docs

CodyCBakerPhD:
Including the `BaseDataInterface` in API docs so `:py:class` references link to it.


bendichter:
I was wondering why that link wasn't working!

PR #246, opened 2022-12-11
Synchronize timestamps

bendichter:
@CodyCBakerPhD here's a draft of the function to synchronize timestamps with pulses

### DIFF
--- a/src/neuroconv/tools/signal_processing.py
+++ b/src/neuroconv/tools/signal_processing.py
@@ -44,6 +44,7 @@ def parse_rising_and_falling_frames_from_ttl(trace: np.ndarray, num_bins: int = 
 
 def syncrhonize_timestamps_from_pulses(pulse_sent_times, pulse_received_times, unsynched_timestamps):
     """
+    Use pulses to align timestamps to the clock of another system.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Function on my end has undergone updates since you made this branch, resulting in some conflicts.

A couple other things:

1) Can you demonstrate the intended input/output of this method with a series of tests against ground truths?
2) Can you point to a conversion where we've used this approach and how well it worked there?

bendichter:
@CodyCBakerPhD I don't have an example of a conversion where I have used this, I was basing it on our conversations about what we needed. Do you think this might not be the right approach?

CodyCBakerPhD:
> Do you think this might not be the right approach?

IDK if it's our place to determine if it's the 'right' approach or not for inferring such times given TTL info - as a statistical problem, there are infinitely many ways of going about it.

What differs in your approach from that of `synchronize_timestamps_between_systems` #237 seems to be the use of the intervals; it's hard for me to get a good grasp on what kind of input/output relation that would give when applied to some actual TTL signals, or how the estimated timing values would differ from the more basic interpolation, but that's something that could be well-communicated and explained via some thorough testing.

We can always offer multiple methods for timestamp inference given reference timings across systems, and accept/add new approaches in a modular way. Just need to think how that would look on the API level for interfaces using them to before sync (like an extra argument to `some_interface.synchronize_between_systems(...)` that specified the `method` they want to use.

bendichter:
@CodyCBakerPhD I'm pretty sure they are equivalent, but just using `np.interp` as you have is a better solution.

PR #245, opened 2022-12-09
[Sync I] Add testing tools for synthetic TTLs

CodyCBakerPhD:
## Motivation
Precursor to #244 and #237 that introduces a method for synthetically generating TTL pulses. Does not support digital words, only on/off intervals.

## How to test the behavior?
Extensively tested, and easy to add more cases if needed.

I was thinking about how best to test the output of this type of synthetic method, and landed on the most rigorous method of exact array assertion against a series of examples stored as ground truth.

Those ground truths are easily made via a helper function that generates the examples and freezes them by writing a very small NWBFile. Since that file might need to change faster than, for example, GIN data does (and can be generated entirely locally) - and is incredibly small - I think it would make sense to just store and package it with the normal testing suite.

This generator for the testing suite also has a very nice convenience function for generating `plotly` figure saved as a `.png` which shows the testing examples for quick and easy reference from the GitHub web interface (which could also be used then in the documentation as well when we add sections describing TTL types, if desired): https://github.com/catalystneuro/neuroconv/blob/add_testing_module/tests/test_minimal/test_testing/test_mock_ttl/example_ttl_reference.png

## Checklist
- [X] Have you thoroughly read our [Developer Guide](https://neuroconv.readthedocs.io/en/main/developer_guide.html)?
- [X] Have you ensured the PR description clearly describes the problem and solutions?
- [X] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/neuroconv/pulls) for the same change?
- [X] If this PR fixes an issue, is the first line of the PR description `fix #XXX` where `XXX` is the issue number?
- [X] Did you update the [CHANGLOG.md](https://github.com/catalystneuro/neuroconv/blob/main/CHANGELOG.md) file on your branch to describe your changes?

PR #244, opened 2022-12-09
[Sync II] Add ttl parsing tools

CodyCBakerPhD:
## Motivation
Extension of #245 that includes tools for parsing the TTL signal to identify the rising and falling frames. Precursor to sync methods of #237.

## How to test the behavior?
Added tests using the synthetic TTLs. Very easy to add new tests for different cases with this approach.

## Checklist

- [X] Have you thoroughly read our [Developer Guide](https://neuroconv.readthedocs.io/en/main/developer_guide.html)?
- [X] Have you ensured the PR description clearly describes the problem and solutions?
- [X] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/neuroconv/pulls) for the same change?
- [X] If this PR fixes an issue, is the first line of the PR description `fix #XXX` where `XXX` is the issue number?
- [X] Did you update the [CHANGLOG.md](https://github.com/catalystneuro/neuroconv/blob/main/CHANGELOG.md) file on your branch to describe your changes?

PR #247, opened 2022-12-11
[Sync III] SpikeGLXNIDQInterface

CodyCBakerPhD:
## Motivation

Extension of #245 and #244 that adds a `SpikeGLXNIDQInterface` which harnesses the TTL parsing tools for extracting event times from the channel traces, while also harnessing existing conversion tools for writing the raw trace data to the NWB file for provenance.

This conversion step is currently being done as an `ElectricalSeries` against the Best Practice regarding such data being neurophysiologic-specific instead of analog/digital/aux/environmental info - I'll start the discussion on the Inspector about how best to handle all that soon.

## How to test the behavior?

Need to add some tests specific to this interface usage - though, the data on GIN may not be usable for it, so it might have to be tested indirectly with the synthetic pulses through the mock interface.

## Checklist


src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -0,0 +1,144 @@
...
+        Useful for synchronizing multiple data streams into the common time basis of the SpikeGLX system.
+
+        Parameters
+        ----------
+        file_path: FilePathType
+            Path to .nidq.bin file.
+        stub_test: bool
+            Whether to shorten file for testing purposes. Default: False.
+        verbose: bool
+            Whether to output verbose text. Default: True.
```
bendichter, line 43:
```suggestion
        file_path : FilePathType
            Path to .nidq.bin file.
        stub_test : bool, default: False
            Whether to shorten file for testing purposes.
        verbose : bool, default: True
            Whether to output verbose text.
```

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -0,0 +1,144 @@
...
+        folder_path = Path(file_path).parent
+        super().__init__(folder_path=folder_path, stream_id=self.stream_id, verbose=verbose)
+        self.source_data.update(file_path=str(file_path))
+
+        self.meta = self.recording_extractor.neo_reader.signals_info_dict[(0, self.stream_id)]["meta"]
+
+    def get_metadata_schema(self):
+        metadata_schema = super().get_metadata_schema()
+        metadata_schema["properties"]["Ecephys"]["properties"].update(
+            ElectricalSeriesRaw=get_schema_from_hdmf_class(ElectricalSeries)
```
bendichter, line 56: This goes back to our conversation about what an ElectricalSeries should be

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -0,0 +1,144 @@
...
+        """Return a list of channel names as set in the recording extractor."""
+        return list(self.recording_extractor.get_channel_ids())
+
+    def get_event_starting_times_from_ttl(self, channel_name: str) -> np.ndarray:
+        """
+        Return the start of event times from the rising part of TTL pulses on one of the NIDQ channels.
+
+        Parameters
+        ----------
+        channel_name : int
```
bendichter, line 123:
```suggestion
        channel_name : str
```

bendichter: I'm not sure whether it should be an int or a str, but if it's an int then it shouldn't be called "name", and if it's a str then the docstring should indicate that

CodyCBakerPhD: Should be `str`

src/neuroconv/tools/signal_processing.py
```diff
@@ -24,7 +24,7 @@ def get_rising_frames_from_ttl(trace: np.ndarray, threshold: Optional[float] = N
     threshold = threshold if threshold is not None else np.mean(trace)
 
     sign = np.sign(trace - threshold)
-    diff = np.diff(sign)
+    diff = np.diff(sign, axis=0)
```
CodyCBakerPhD, line 27: Needed to add this due to odd issue with 1-D input returning an empty diff

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -0,0 +1,130 @@
...
+        Useful for synchronizing multiple data streams into the common time basis of the SpikeGLX system.
+
+        Parameters
+        ----------
+        file_path : FilePathType
+            Path to .nidq.bin file.
+        stub_test : bool
+            Whether to shorten file for testing purposes, default: False
+        verbose : bool
+            Whether to output verbose text, default: True
```
bendichter, line 37:
```suggestion
        verbose : bool, default: True
            Whether to output verbose text.
```

src/neuroconv/tools/signal_processing.py
CodyCBakerPhD: Or I guess it was a 2-D input with the second dimension being '1' instead of open.

Alternatively, could `squeeze` the trace on passing?

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -0,0 +1,130 @@
...
+        Read channel data from the NIDQ board for the SpikeGLX recording.
+
+        Useful for synchronizing multiple data streams into the common time basis of the SpikeGLX system.
+
+        Parameters
+        ----------
+        file_path : FilePathType
+            Path to .nidq.bin file.
+        stub_test : bool
+            Whether to shorten file for testing purposes, default: False
```
bendichter, line 35:
```suggestion
        stub_test : bool, default: False
            Whether to shorten file for testing purposes.
```

src/neuroconv/tools/testing/mock_interfaces.py
```diff
@@ -0,0 +1,64 @@
...
+            The number of seconds to simulate.
+            The default is 7.0 seconds.
+        ttl_times: list of list of floats, optional
+            The times within the `signal_duration` to trigger the TTL pulse for each channel.
+            The outer list is over channels, while each inner list is the set of TTL times for each specific channel.
+            The default generates 8 channels with periodic on/off cycle (which start in the 'off' state)
+            each of which is of length `ttl_duration` with a 0.1 second offset per channel.
+        ttl_duration: float, optional
+            How long the TTL pulse stays in the 'on' state when triggered, in seconds.
+            The default is 1 second.
```
bendichter, line 34:
```suggestion
        ttl_duration: float, default: 1.0
            How long the TTL pulse stays in the 'on' state when triggered, in seconds.
```

src/neuroconv/tools/testing/mock_interfaces.py
```diff
@@ -0,0 +1,64 @@
...
+        signal_duration: float, optional
+            The number of seconds to simulate.
+            The default is 7.0 seconds.
+        ttl_times: list of list of floats, optional
+            The times within the `signal_duration` to trigger the TTL pulse for each channel.
+            The outer list is over channels, while each inner list is the set of TTL times for each specific channel.
+            The default generates 8 channels with periodic on/off cycle (which start in the 'off' state)
+            each of which is of length `ttl_duration` with a 0.1 second offset per channel.
+        ttl_duration: float, optional
+            How long the TTL pulse stays in the 'on' state when triggered, in seconds.
```
bendichter, line 33:
```suggestion
        signal_duration : float, default: 7.0
            The number of seconds to simulate.
        ttl_times : list of list of floats, optional
            The times within the `signal_duration` to trigger the TTL pulse for each channel.
            The outer list is over channels, while each inner list is the set of TTL times for each specific channel.
            The default generates 8 channels with periodic on/off cycle (which start in the 'off' state)
            each of which is of length `ttl_duration` with a 0.1 second offset per channel.
        ttl_duration : float, optional
            How long the TTL pulse stays in the 'on' state when triggered, in seconds.
```

bendichter (approved):

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
CodyCBakerPhD: Yep, it's the least amount of work right now to just consider it another `ElectricalSeries`.

Would be a little extra work to setup a separate table from the main `nwbfile.electrodes` plus a `DynamicTableRegion` pointing to the custom one.

Otherwise, would lose device/group metadata information by writing directly to a `TimeSeries` instead.

tests/test_ecephys/test_mock_interfaces.py
```diff
@@ -0,0 +1,84 @@
...
+            }
+        }
+        self.assertDictEqual(d1=metadata["Ecephys"], d2=expected_ecephys_metadata["Ecephys"])
+
+        expected_start_time = datetime(2020, 11, 3, 10, 35, 10)
+        assert metadata["NWBFile"]["session_start_time"] == expected_start_time
+
+    def test_mock_run_conversion(self):
+        # TODO
+        pass
```
CodyCBakerPhD, line 84: @bendichter 

This test waiting on confirmation of https://github.com/catalystneuro/neuroconv/pull/247#discussion_r1064989056

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -352,6 +353,57 @@ class TestConverter(NWBConverter):
...
 
+    parameterized_aux_list = [
+        param(
+            data_interface=SpikeGLXNIDQInterface,
+            interface_kwargs=dict(file_path=str(DATA_PATH / "spikeglx" / "Noise4Sam_g0" / "Noise4Sam_g0_t0.nidq.bin")),
+        )
+    ]
+
+    @parameterized.expand(input=parameterized_recording_list, name_func=custom_name_func)
+    def test_aux_recording_extractor_to_nwb(self, data_interface, interface_kwargs, case_name=""):
```
CodyCBakerPhD, line 364: I had initially included this in the `test_recording_extractor_to_nwb` list but the behavior is slightly different.

Also had to debug one functionality in the Converter usage that differs from stand-alone interface conversion, left a comment next to that corresponding code

CodyCBakerPhD: Also this file is starting to get really cluttered so in a follow-up I think I might break each of these test functions into a separate file for readability

src/neuroconv/tools/signal_processing.py
bendichter: either way is fine. This will always be 1-d, right? You could also use .ravel()

CodyCBakerPhD: > This will always be 1-d, right?

As we use it internally, yes.

Technically someone could call the function directly and pass a `trace` with multiple dimensions; since we do not assert, check, or otherwise force it here.

Should I add such a check to enforce our intended usage of the function?

CodyCBakerPhD: `np.ravel` makes usage a bit clearer at no cost to performance

src/neuroconv/tools/signal_processing.py
```diff
@@ -21,9 +21,13 @@ def get_rising_frames_from_ttl(trace: np.ndarray, threshold: Optional[float] = N
     rising_frames : numpy.ndarray
         The frame indices of rising events.
     """
+    flattened_trace = np.ravel(trace)  # Shapes like (1, x, 1, 1) might result from slicing patterns and are allowed
+    if np.max(trace.shape) != flattened_trace.shape[0]:  # TODO: when 3.7 dropped, use math.prod to avoid overflow
+        raise ValueError(f"This function expects a one-dimensional array! Received shape of {trace.shape}.")
+
     threshold = threshold if threshold is not None else np.mean(trace)
```
bendichter, line 28:
```suggestion
    threshold = np.mean(trace) if threshold is None else threshold
```

slightly more intuitive syntax

bendichter (dismissed):

PR #248, opened 2022-12-12
Bump probeinterface version

CodyCBakerPhD:
#195 bypassed an issue between `packaging` and `probeinterface` regarding openephys versioning - looks like this has now been fixed, so testing it here


bendichter:
It looks like probeinterface dropped support for Python 3.7 2 months ago:

https://github.com/SpikeInterface/probeinterface/blob/3b7d17caa9b1c003d50a2c6c9a3fafd2972025a9/pyproject.toml#L11

PR #252, opened 2022-12-14
text parsers

bendichter:
## Motivation

Write parsers from text files

To do:
- [x] changelog
- [x] conversion gallery 

## Checklist


bendichter:
@CodyCBakerPhD before I spend time writing tests, what do you think of this layout?

src/neuroconv/datainterfaces/text/csv/requirements.txt
```diff
@@ -0,0 +1 @@
+pandas
```
CodyCBakerPhD, line 1: No need to include this explicitly - pandas comes free with PyNWB

src/neuroconv/datainterfaces/text/excel/requirements.txt
```diff
@@ -0,0 +1,3 @@
+openpyxl
+xlrd
```
CodyCBakerPhD, line 2: What are these two for?

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(
+        self,
+        file_path: FilePathType,
+        read_kwargs: Optional[dict] = None,
+        verbose: bool = True,
+        name: str = "trials",
+        description: Optional[str] = None,
+    ):
+        if read_kwargs is None:
+            read_kwargs = {}
```
CodyCBakerPhD, line 39:
```suggestion
        read_kwargs = read_kwargs  or dict()
```

src/neuroconv/datainterfaces/text/excel/requirements.txt
bendichter: pd.read_excel requires openpyxl for .xslx files and xlrd for .xls files

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def convert_df_to_time_intervals(df, name, description=None):
+    if description is None:
+        description = name
+    time_intervals = TimeIntervals(name, description)
+    if "start_time" not in df:
+        raise ValueError(f"df must contain a column named 'start_time'. Existing columns: {df.columns.to_list()}")
+    if "stop_time" not in df:
+        df["stop_time"] = np.nan
```
CodyCBakerPhD, line 19: So, this is a design question - do we want to enforce/require people to modify the column names in whatever file format they're using to conform with NWB column names?

Or, I was originally thinking one of the arguments to an interface like this could be `column_mapping` or `column_name_mapping` or anything better. Essentially a `Dict[str, str]` mapping the names from their lab-specific CSV format to how they want the names to be in NWB format (which we can enforce best practices like `_time` on). That way they don't have to comb through potentially hundreds of their own files changing things from how they're used to doing it and can instead just tell us what the mapping is to have it done seamlessly

bendichter: good idea. I'll add that

CodyCBakerPhD, line 19: Another design question - should we even allow a `TimeIntervals` mapping without some `stop_time` either available or mapping provided? It can sometimes be an indication that a `TimeIntervals` isn't the best representation of the data. Of course, still possible to be, but that leads to next suggestion...

...Or perhaps with an option to set the `stop_time` to be the the next consecutive `start_time` as we sometimes do when forming a trials table where there isn't really a definable stopping time except for when the next trial started?

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -0,0 +1,63 @@
...
+    for col in df:
+        if col not in ("start_time", "stop_time"):
+            time_intervals.add_column(col, col)
```
CodyCBakerPhD, line 22: Will also likely need to define some metadata structure for how to specify description fields for the custom columns

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class TimeIntervalsInterface(BaseDataInterface):
+    def __init__(
+        self,
+        file_path: FilePathType,
+        read_kwargs: Optional[dict] = None,
+        verbose: bool = True,
+        name: str = "trials",
+        description: Optional[str] = None,
```
CodyCBakerPhD, line 36: I'd include both the `name` and `description` for the actual table itself as well within the `metadata` dictionary, similar to how we map such properties in other interfaces. Usually the `__init__` for an interface just tells us how to open an I/O to allow metadata/data reading access to the source files

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -0,0 +1,63 @@
...
+        time_intervals.add_row(row.to_dict())
+
+    return time_intervals
+
+
+class TimeIntervalsInterface(BaseDataInterface):
+    def __init__(
+        self,
+        file_path: FilePathType,
+        read_kwargs: Optional[dict] = None,
```
CodyCBakerPhD, line 33: Yeah, `read_kwargs` is probably the best way to get around complexity in reading strangely formatted CSV files (multiple tabs, several header offsets, fill values, etc.)

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -0,0 +1,63 @@
...
+        name: str = "trials",
+        description: Optional[str] = None,
+    ):
+        if read_kwargs is None:
+            read_kwargs = {}
+        super().__init__(file_path=file_path)
+        self.verbose = verbose
+
+        self.df = self._read_file(file_path, **read_kwargs)
+        self.time_intervals = convert_df_to_time_intervals(self.df, name=name, description=description)
```
CodyCBakerPhD, line 43: This would be moved down to `run_conversion` to avoid potentially heavy initialization (which is supposed to be fully lazy I/O opening)

CodyCBakerPhD:
Some comments on micro-design, but macro-scale looks good

Not sure how I feel about the name "text" for this branch of data types - that might imply `.doc`/`.docx` files might also be supported (which, I mean, they technically could be I suppose since they can also contain Word tables lol)... can't think of anything better at the moment though

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
bendichter: done

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
bendichter: agreed

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
bendichter: ok, I moved the creation of the time_intervals object to the run_conversion method

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
bendichter: I moved the second down. I don't expect the first to take very long and I want to catch read errors early but I would be open to moving it down too.

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
bendichter: yeah that's what I was thinking

src/neuroconv/datainterfaces/text/excel/requirements.txt
```diff
@@ -0,0 +1,3 @@
...
+pandas
```
bendichter, line 3:
```suggestion
```

bendichter:
@CodyCBakerPhD I've moved this along. Would love to get your feedback when you have a moment

requirements-minimal.txt
```diff
@@ -9,3 +9,4 @@ pynwb>=1.4.0
 psutil>=5.8.0
 tqdm>=4.60.0
 dandi>=0.46.2
+pandas
```
CodyCBakerPhD, line 12: Also don't think we need it here at all either - it should be included with PyNWB

src/neuroconv/datainterfaces/text/csv/csvtimeintertervalsinterface.py
```diff
@@ -0,0 +1,9 @@
+import pandas as pd
+
+from ..timeintervalsinterface import TimeIntervalsInterface
+from ....utils.types import FilePathType, Optional
+
+
+class CsvTimeIntervalsInterface(TimeIntervalsInterface):
+    def _read_file(self, file_path: FilePathType, **read_kwargs):
+        return pd.read_csv(file_path, **read_kwargs)
```
CodyCBakerPhD, line 9: This structure starts to lend itself to a much more general base I/O class, similar to how the `GenericDataChunkIterator` requires specification of `_get_data` to tell it 'how to read the data' from something initialized during `__init__`

I can also foresee a case of building something like a `MatlabTimeIntervalsInterface` that defines the `_read_file` in terms of scipy or hdfstorage or just vanilla h5py commands

Not saying we need to do that in this PR, just thinking ahead

requirements-minimal.txt
bendichter: If we import a library, we should list it as a requirement, even if it is a requirement of another required repo. What if PyNWB removes pandas as a requirement? Our code would then break. Also, we may have certain version requirements for features of pandas that PyNWB does not have

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -0,0 +1,133 @@
...
+        If passed, rename subset of columns from key to value.
+    column_descriptions: dict, optional
+        Keys are the names of the columns (after renaming) and values are the descriptions. If not passed,
+        the names of the columns are used as descriptions.
+
+    Returns
+    -------
+    TimeIntervals
+
+    """
```
CodyCBakerPhD, line 45: Our usual organization would include the import for this function under `neuroconv.tools.text` in case someone wanted to use it as a handy helper independently of the data interfaces or other NeuroConv API

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -0,0 +1,133 @@
...
+from ...utils.dict import load_dict_from_file
+
+
+def convert_df_to_time_intervals(
+    df: pd.DataFrame,
+    name: str = "trials",
+    description: Optional[str] = None,
+    column_name_mapping: Dict[str, str] = None,
+    column_descriptions: Dict[str, str] = None,
+):
```
CodyCBakerPhD, line 23:
```suggestion
) -> TimeIntervals:
```

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -0,0 +1,133 @@
...
+
+        self.df = self._read_file(file_path, **read_kwargs)
+        self.time_intervals = None
+
+    def get_metadata(self):
+        return dict(
+            TimeIntervals=dict(
+                trials=dict(
+                    name="trials",
+                    description="trials",
```
CodyCBakerPhD, line 38: Can we get a richer default description for this? "Autogenerated by NeuroConv." is also acceptable to distinguish it from https://github.com/NeurodataWithoutBorders/pynwb/blob/0d42132b2bfd5dc4474e484953faeca46ed267ec/src/pynwb/file.py#L744

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -0,0 +1,133 @@
...
+                )
+            )
+        )
+
+    def get_metadata_schema(self):
+        fpath = os.path.join(os.path.split(__file__)[0], "timeintervals.schema.json")
```
CodyCBakerPhD, line 44: I appreciate the move towards pre-defined schemas for readability and easier portability to web form renderers - however, I don't see this file included in the changelog

You may have to add it explicitly since I think those are generally ignored by the `.gitignore` (meaning blanket `git add` calls will skip it even if it's a new file)

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -0,0 +1,133 @@
...
+        with make_or_load_nwbfile(
+            nwbfile_path=nwbfile_path, nwbfile=nwbfile, metadata=metadata, overwrite=overwrite, verbose=self.verbose
+        ) as nwbfile_out:
+            self.time_intervals = convert_df_to_time_intervals(
+                self.df,
+                column_name_mapping=column_name_mapping,
+                column_descriptions=column_descriptions,
+                **metadata["TimeIntervals"][tag],
+            )
+            nwbfile_out.add_time_intervals(self.time_intervals)
```
CodyCBakerPhD, line 66: Should we have conditions here such as when the tag or table name is `trials` or `epoch` to put it in the `nwbfile.trials` and `nwbfile.epochs`, respectively?

Also it seems `add_time_intervals` places it in a `nwbfile.intervals` container - are there any use cases where they might want to place it in either `acquisition` or a `processing` module?

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
bendichter: huh, how do you feel about removing "*.json" from the .gitignore file?

tests/test_text/test_time_intervals.py
```diff
@@ -0,0 +1,54 @@
...
+    nwbfile = interface.run_conversion(metadata=metadata, column_name_mapping=dict(condition="cond"))
+    assert nwbfile.intervals["trials"].colnames == ("start_time", "stop_time", "cond")
+
+
+def test_csv():
+    interface = CsvTimeIntervalsInterface(trials_csv_path)
+    metadata = interface.get_metadata()
+    metadata["NWBFile"] = dict(session_start_time=datetime.now().astimezone())
+    nwbfile = interface.run_conversion(metadata=metadata)
+    assert nwbfile.intervals["trials"]
```
CodyCBakerPhD, line 54: Nice units tests here - would also want to see a roundtrip NWB file write as well as an extra set of tests (write the `nwbfile` highlighted here, read it back and compare the time intervals)

src/neuroconv/tools/text.py
```diff
@@ -0,0 +1,64 @@
...
+        stop_time="Stop time of epoch, in seconds.",
+    )
+
+    if column_descriptions is None:
+        column_descriptions = default_column_descriptions
+    else:
+        column_descriptions = dict(default_column_descriptions, **column_descriptions)
+
+    if description is None:
+        description = name
```
CodyCBakerPhD, line 51: Feel like this is duplication of information - default table description could at the very least be some bit of extra information, like `"Autogenerated by NeuroConv."`. The interval usage could also make use of the `file_path` by writing something like `"This TimeIntervals table was converted from the source file {file_path}" using NeuroConv.` which leaves additional provenance of how that table came to be (either experimentally, or by way of NWB conversion)

src/neuroconv/datainterfaces/text/csv/csvtimeintertervalsinterface.py
bendichter: Yes, the idea is this structure would allow for reading many different kinds of formats. Hierarchical formats like HDF5 and .mat files might be tricky, but surely we could support some common patterns in the future. For now, I think supporting csv, xls, and xlsx is a good scope for this PR.

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
bendichter: ok, moved.

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
bendichter: I'm not crazy about putting "NeuroConv" into the data descriptors of the NWB file.

CodyCBakerPhD:
>@CodyCBakerPhD I've moved this along. Would love to get your feedback when you have a moment

Looking good so far, left some comments but nothing too major

src/neuroconv/tools/text.py
bendichter: I like the idea of including the source filepath in the description

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
bendichter: done

tests/test_text/test_time_intervals.py
bendichter: done

tests/test_text/test_time_intervals.py
```diff
@@ -0,0 +1,68 @@
...
+    metadata["NWBFile"] = dict(session_start_time=datetime.now().astimezone())
+    nwbfile = interface.run_conversion(metadata=metadata)
+    assert nwbfile.intervals["trials"].colnames == ("start_time", "stop_time", "condition")
+
+
+def test_excel_column_name_mapping():
+    interface = ExcelTimeIntervalsInterface(trials_xls_path)
+    metadata = interface.get_metadata()
+    metadata["NWBFile"] = dict(session_start_time=datetime.now().astimezone())
+    nwbfile = interface.run_conversion(metadata=metadata, column_name_mapping=dict(condition="cond"))
```
CodyCBakerPhD, line 47: Can you add a test that we can reference for people demonstrating a source file that does not have the `start_time`/`stop_time` fields and instead relies on the `column_name_mapping` to satisfy those requirements? I think that will actually be quite common

tests/test_text/test_time_intervals.py
```diff
@@ -0,0 +1,68 @@
...
+def test_csv_round_trip(tmp_path):
+    interface = CsvTimeIntervalsInterface(trials_csv_path)
+    metadata = interface.get_metadata()
+    metadata["NWBFile"] = dict(session_start_time=datetime.now().astimezone())
+    interface.run_conversion(nwbfile_path=tmp_path / "test.nwb", metadata=metadata)
+
+    with NWBHDF5IO(tmp_path / "test.nwb", "r") as io:
+        nwb_read = io.read()
+        assert nwb_read.trials.colnames == ("start_time", "stop_time", "condition")
+        assert_array_equal(nwb_read.trials["condition"][:], [1, 2, 3, 1, 3, 2, 2, 3, 1, 2, 3])
```
CodyCBakerPhD, line 68: Can you add a test where you manually override the metadata for the table name and description, and check that that works round-trip?

bendichter: sure

bendichter: done.

tests/test_text/test_time_intervals.py
bendichter: sure

src/neuroconv/tools/text.py
```diff
@@ -0,0 +1,61 @@
...
+import pandas as pd
+from pynwb.epoch import TimeIntervals
+
+
+def convert_df_to_time_intervals(
+    df: pd.DataFrame,
+    name: str = "trials",
+    description: str = "experimental trials",
+    column_name_mapping: Dict[str, str] = None,
+    column_descriptions: Dict[str, str] = None,
```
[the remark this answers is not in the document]
bendichter: ok, I changed these fields to "table_name" and "table_description"

bendichter:
@CodyCBakerPhD I think this might be ready

CodyCBakerPhD (approved):

PR #255, opened 2022-12-19
Handle writing multiple probes/streams data

### LINKED ISSUE #254: [Feature]: Handle multiple probes and electrode regions

Currently, if two `BaseRecording` or `BaseSorting` objects are added to the same nwbfile, the `add_*` functions only checks the `electrodes`/`units` tables for the existance of `channel_`/`unit_name`. However, two probes, belonging to different groups, could have:
1. the same `channel_name` (e.g. AP1, AP2, ...)
2. the same `unit_name` (output by the spike sorter)

The write functions should be able to handle writing data from multiple probes more flexibly.

Currently, the only way to add electrodes/units from different probes is to "manually" set the `channel_name`/`unit_name` properties. A more convenient and automated handling of multiple probes is needed.

Yes.

### Code of Conduct

alejoe91:
Fixes #254 

## Motivation

This PR allows one to write data from multiple probes (both recording, sorting, and waveforms) without needing to manually change `channel_name`/`unit_name` properties.

This is done by:
- checking existence of previous electrodes with both `channel_name` and `group` columns
- checking existence of previous units with both `unit_name` and `electrodes` columns
- automatically adding `electrodes` column when writing waveforms

## How to test the behavior?

See added [test](https://github.com/catalystneuro/neuroconv/blob/multi-probe/tests/test_ecephys/test_tools_spikeinterface.py#L1184)
```
"""This test that the waveforms are written to different electrode groups"""
# we write the first set of waveforms as belonging to group 0 
# (we is a WaveformExtractor, rec is the associated recording)
rec.set_channel_groups([0] * len(self.we_recless_recording.channel_ids))
write_waveforms(waveform_extractor=we, nwbfile=nwbfile,
                            write_electrical_series=False, recording=rec)
# now we set new channel groups to mimic a different probe and call the function again
rec.set_channel_groups([1] * len(rec.channel_ids))
write_waveforms(waveform_extractor=we, nwbfile=nwbfile,
                write_electrical_series=False, recording=rec)
# check that we have 2 groups
self.assertEqual(len(nwbfile.electrode_groups), 2)
self.assertEqual(len(np.unique(nwbfile.electrodes["group_name"])), 2)
# check that we have correct number of units
self.assertEqual(len(nwbfile.units), 2 * len(we.unit_ids))
# check electrode regions of units
for row in nwbfile.units.id:
    if row < len(we.unit_ids):
        self.assertEqual(we.nwbfile.units[row].electrodes.values[0], [0, 1, 2, 3])
    else:
        self.assertEqual(wenwbfile.units[row].electrodes.values[0], [4, 5, 6, 7])
```

## Checklist


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -326,9 +326,6 @@ def add_electrodes(
     else:
         channel_name_array = channel_ids.astype("str", copy=False)
         data_to_add["channel_name"].update(description="unique channel reference", data=channel_name_array, index=False)
-        # If the channel ids are integer keep the old behavior of asigning nwbfile.electrodes.id equal to channel_ids
-        if np.issubdtype(channel_ids.dtype, np.integer):
-            data_to_add["id"].update(data=channel_ids, index=False)
```
alejoe91, line 331: Keeping integers `channel_ids` as `id` is incompatible with new changes

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1195,9 +1204,6 @@ def add_units_table(
     else:
         unit_name_array = unit_ids.astype("str", copy=False)
         data_to_add["unit_name"].update(description="Unique reference for each unit.", data=unit_name_array)
-        # If the channel ids are integer keep the old behavior of asigning table's id equal to unit_ids
-        if np.issubdtype(unit_ids.dtype, np.integer):
-            data_to_add["id"].update(data=unit_ids.astype("int"))
```
alejoe91, line 1200: Keeping integers `unit_ids` as `id` is incompatible with new changes

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1551,6 +1572,8 @@ def add_waveforms(
                 sorting_copy.set_property(prop, tm[prop])
 
     add_electrodes_info(recording, nwbfile=nwbfile, metadata=metadata)
+    electrode_group_indices = get_electrode_group_indices(recording, nwbfile=nwbfile)
+    unit_electrode_indices = [electrode_group_indices] * len(sorting.unit_ids)
```
alejoe91, line 1576: This will also allow to automatically handle sparsity, when that is propagated to the `WaveformExtractor` object in SpikeInterface (see [PR](https://github.com/SpikeInterface/spikeinterface/pull/1152))

src/neuroconv/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Why is that? Couldn't you just cast the `channel_name` or `unit_name`  from an `int` to a string or something (I thought it already did this)? The table IDs must be unsigned integers, yes, but it can simply be a running count if needed, it's just for unique identification/indexing of rows

This seems to be the main culprit of the failing tests

alejoe91: Imagine we have two probes with the same `int` channel ids, but belonging to different electrode groups. In that case, we want to be able to add electrodes from both probes, since they are not the same elelectrodes, even if they have the same `int` channel ids.

CodyCBakerPhD: Yes, but what I'm saying is there's a disconnect for just this purpose between the `id` of the table (just a unique integer identifier) and the `channel_name`. I like the logic of comparing the `channel_name` together with `group_name` to disambiguate - what I don't see is how a string casting of that `int`-valued `channel_id` on that group to serve as the `channel_name` causes an issue with that.

In fact you might be able to bypass any direct logic for this by simply using `enforce_unique_id=True` from https://hdmf.readthedocs.io/en/stable/tutorials/plot_dynamictable_howto.html#adding-rows

alejoe91: Casting the `int`-channel_id to `str` doesn't cause the problem, but rather forcing the table `id` without using `enforce_unique_id=False`. Should we use `enforce_unique_id=False` when adding a channel/unit? IMO, since `channel_name` will already be a column, there is no point of forcing the `id` to be the `channel_id`, if `int`. But I'm happy to discuss!

CodyCBakerPhD: > but rather forcing the table id without using enforce_unique_id=False. Should we use enforce_unique_id=False when adding a channel/unit?

See the [Best Practice for Unique IDs](https://nwbinspector.readthedocs.io/en/dev/best_practices/tables.html#unique-ids). We should never use `enforce_unique_id=False` - what I'm saying is instead of specifying the `id` column manually, if you set `enforce_unique_id=True` it will automatically generate a unique `id` for each row.

What I'm saying is, we would treat the pair of `channel_name` (wherever it comes from) and `group_name` as the effective human-readable identifier of electrodes added via a recording extractor; but we also need the required ID field, too, but that doesn't have to match or correlate in any way to the pair of human-readable values.

alejoe91: ok, so if I understand correctly, we would set the `channel_name` column values as `f{group_name}-{channel_id}`? Ok for me :) 

Sorry for the misunderstanding!

CodyCBakerPhD: > ok, so if I understand correctly, we would set the channel_name column values as f{group_name}-{channel_id}?

That sounds as if it could work! I was imagining `channel_name=str(channel_id)` as one column and `group_name=str(electrode_group.name)` as the other (on the same row), which would make the `channel_name` column have non-unique elements, but unique pairs taken across the two `_name` columns - but just adding extra info to the `channel_name` might be more convenient

alejoe91: Actually, we already have the `channel_name` and `group_name` column and that's what is used to check if an electrode is already in the table: https://github.com/catalystneuro/neuroconv/blob/multi-probe/src/neuroconv/tools/spikeinterface/spikeinterface.py#L412-L433

I think this is a better approach, because we get to keep the original `channel_names`, which could come directly from the device (e.g. AP1... for spikeglx).

Maybe let me try to clarify: the behavior which was supported and in this new implementation should not be supported anymore is the following: 
```
# If the channel ids are integer keep the old behavior of asigning nwbfile.electrodes.id equal to channel_ids
if np.issubdtype(channel_ids.dtype, np.integer):
    data_to_add["id"].update(data=channel_ids, index=False)
```
Since now `channel_ids` can be non-unique in the electrodes table (if they belong to different groups), if we force the `id` to be the `channel_id` in case these are integers, we might have duplicate `id`, which of course we don't want to have. 
Basically, we have to move on to a new behavior :)

CodyCBakerPhD: I'm aware of all of this; but `id` is a required field of all tables, and its values must be unique. The `channel_ids`, for some time now, have been mapped instead for all practical purposes to `channel_name`. I'm saying, instead of mapping `channel_ids` to the `id` column at all, just use `enforce_unique_id=True` flag, don't include the `id` key when writing a row (i.e., in `data_to_add`), and it should take care of it automatically.

Let me know when the tests are passing and I can take a look at how you fixed it and go from there

tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -759,22 +776,6 @@ def test_row_matching_by_channel_name_with_new_property(self):
         self.assertListEqual(list(self.nwbfile.electrodes["channel_name"].data), expected_names)
         self.assertListEqual(list(self.nwbfile.electrodes["property"].data), expected_property_values)
 
-    def test_assertion_for_id_collision(self):
```
CodyCBakerPhD, line 762: Is there a test, either direct or indirect, that tests for this new non-collision for SI recordings?

alejoe91: Yes, this test: https://github.com/catalystneuro/neuroconv/blob/multi-probe/tests/test_ecephys/test_tools_spikeinterface.py#L1149

It adds two recordings with the same channel ids, but different groups.

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1237,9 +1279,12 @@ def add_units_table(
...
         if waveform_means is not None:
             unit_kwargs["waveform_mean"] = waveform_means[row]
-        if waveform_sds is not None:
-            unit_kwargs["waveform_sd"] = waveform_sds[row]
+            if waveform_sds is not None:
+                unit_kwargs["waveform_sd"] = waveform_sds[row]
+            if unit_electrode_indices is not None:
+                unit_kwargs["electrodes"] = unit_electrode_indices[row]
         units_table.add_unit(spike_times=spike_times, **unit_kwargs, enforce_unique_id=True)
+    added_unit_table_ids = units_table.id[-len(rows_to_add) :]
```
alejoe91, line 1287: @CodyCBakerPhD this is needed to keep old behavior of `write_waveforms`: as now table ids are not necessary coincident with unit ids, we need to pass the list of added unit ids to the `_add_waveforms_to_units_table()` function.

tests/test_on_data/setup_paths.py
```diff
@@ -22,16 +22,8 @@
...
-if not BEHAVIOR_DATA_PATH.exists():
-    pytest.fail(f"No folder found in location: {BEHAVIOR_DATA_PATH}!")
-
 ECEPHY_DATA_PATH = LOCAL_PATH / "ephy_testing_data"
-if not ECEPHY_DATA_PATH.exists():
-    pytest.fail(f"No folder found in location: {ECEPHY_DATA_PATH}!")
-
 OPHYS_DATA_PATH = LOCAL_PATH / "ophys_testing_data"
-if not OPHYS_DATA_PATH.exists():
-    pytest.fail(f"No folder found in location: {OPHYS_DATA_PATH}!")
```
alejoe91, line 27: moved exixtance checks to specific files

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -48,9 +49,16 @@
     MCSRawRecordingInterface,
 )
 
+# enable to run locally in interactive mode
+try:
+    from .setup_paths import ECEPHY_DATA_PATH as DATA_PATH
+    from .setup_paths import OUTPUT_PATH
+except:
+    from setup_paths import ECEPHY_DATA_PATH as DATA_PATH
+    from setup_paths import OUTPUT_PATH
```
alejoe91, line 58: this logic is done to allow to run tests cases in ipython, which is very convenient for debugging (relative imports are not allowed)

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -214,7 +222,7 @@ class TestConverter(NWBConverter):
                     folder_path=str(DATA_PATH / "openephysbinary" / "v0.4.4.1_with_video_tracking"),
                     spikeextractors_backend=spikeextractors_backend,
                 ),
-                case_name=f"spikeextractors_backend={spikeextractors_backend}",
+                case_name=f"spikeextractors_backend_{spikeextractors_backend}",
```
alejoe91, line 225: changed from `=` to `_` to allow to run tests interactively (having `=` in function name is not permitted)

tests/test_on_data/test_gin_ecephys.py
```diff
@@ -448,7 +456,11 @@ class TestConverter(NWBConverter):
             # NWBSortingExtractor on spikeinterface does not yet support loading data written from multiple segment.
             if sorting.get_num_segments() == 1:
                 nwb_sorting = NwbSortingExtractorSI(file_path=nwbfile_path, sampling_frequency=sf)
-                check_sorting_equal_si(SX1=sorting, SX2=nwb_sorting)
+                # In the NWBSortingExtractor, since unit_names could be not unique,
+                # table "ids" are loaded as unit_ids. Here we rename the original sorting accordingly
+                sorting_renamed = sorting.select_units(unit_ids=sorting.unit_ids,
+                                                       renamed_unit_ids=np.arange(len(sorting.unit_ids)))
```
alejoe91: this is a tricky one. The `spikeinterface.extractors.NWBSortingExtractor` used `id` as `unit_ids`. This is now required because `unit_name` might not be unique (we need to propagate the same change to recording extractors too). Since the original recording might have `str` unit_ids, we have to rename the original unit is to match the `nwb_sorting` object. The `check_*` function also checks for spike trains, so we are sure they are actually the same unit.

tests/test_on_data/test_gin_ecephys.py
h-mayorquin: How are you running your tests? 
I can run specific tests with the currente setup. Example:
```
pytest tests/test_on_data/test_gin_ecephys.py::TestEcephysNwbConversions
```

You need to be on the top level directory for this to work though.

tests/test_on_data/test_gin_ecephys.py
h-mayorquin: This works for me:
```
pytest tests/test_on_data/test_gin_ecephys.py::TestEcephysNwbConversions::test_recording_extractor_to_nwb_12_OpenEphys_spikeextractors_backend=True
```

Maybe you need some escape character on some terminals?

```
pytest tests/test_on_data/test_gin_ecephys.py::TestEcephysNwbConversions::test_recording_extractor_to_nwb_12_OpenEphys_spikeextractors_backend\=True
```

alejoe91: yes, but I wanted to be able to instantiate and call the test class in interactive mode, and the `=` makes it impossible

tests/test_on_data/test_gin_ecephys.py
alejoe91: I'd like to run them in python/ipython for debugging: e.g. from `tests/test_on_data/` folder:
```
python test_gin_ecephys.py
```

h-mayorquin: `python -m pytest tests/test_on_data/test_gin_ecephys.py` should work to run from python from that folder or any above.
 

How do you debug though? Do you put traces by hand with pdb?

tests/test_on_data/test_gin_ecephys.py
h-mayorquin: Can you show me a screenshot o the error that you are geeting (also to see what is your workflow for testing-debugging)?

tests/test_on_data/test_gin_ophys.py
```diff
@@ -18,7 +19,14 @@
     Suite2pSegmentationInterface,
 )
 
-from .setup_paths import OPHYS_DATA_PATH, OUTPUT_PATH
+# enable to run locally in interactive mode
+try:
+    from .setup_paths import OPHYS_DATA_PATH, OUTPUT_PATH
+except:
```
CodyCBakerPhD, line 25:
```suggestion
except InsertExceptionTypeHere:  # Interactive mode cannot support local imports
```

tests/test_on_data/test_gin_ecephys.py
CodyCBakerPhD: Requesting to add the specific error type to the catcher along with a comment at the line of exception explaining why interactivity needs it

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1218,10 +1247,23 @@ def add_units_table(
...
     properties_with_data = {property for property in properties_to_add_by_rows if "data" in data_to_add[property]}
     rows_in_data = [index for index in range(checked_sorting.get_num_units())]
-    rows_to_add = [index for index in rows_in_data if unit_name_array[index] not in unit_names_used_previously]
+    if not has_electrodes_column:
+        rows_to_add = [index for index in rows_in_data if unit_name_array[index] not in unit_names_used_previously]
+    else:
+        rows_to_add = []
+        for index in rows_in_data:
+            if unit_name_array[index] not in unit_names_used_previously:
+                rows_to_add.append(index)
```
CodyCBakerPhD, line 1260: Seems like this line is not covered by any tests - can you add one?

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1668,6 +1720,20 @@ def write_waveforms(
         )
 
 
+def get_electrode_group_indices(recording, nwbfile):
+    if "group_name" in recording.get_property_keys():
+        group_names = list(np.unique(recording.get_property("group_name")))
+    elif "group" in recording.get_property_keys():
+        group_names = list(np.unique(recording.get_property("group").astype(str)))
+    else:
+        group_names = None
```
CodyCBakerPhD, line 1729: Condition not tested - please add

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1668,6 +1720,20 @@ def write_waveforms(
         )
 
 
+def get_electrode_group_indices(recording, nwbfile):
+    if "group_name" in recording.get_property_keys():
+        group_names = list(np.unique(recording.get_property("group_name")))
```
CodyCBakerPhD, line 1725: Condition not tested - please add

tests/test_ecephys/test_tools_spikeinterface.py
CodyCBakerPhD: That test is pretty heavy integration-wise, already a lot going on in it. I'd feel a lot better about this one if you restored the original structure (write two electrodes manually, then call `add_electrode` as a unit test), just adjust the ground truth to show the new behavior (to not raise error on collision)

CodyCBakerPhD:
OK It's coming together - requested several more tests

Would still love to see an example file of multi-probe NP to see this in action

alejoe91:
> OK It's coming together - requested several more tests
> 
> Would still love to see an example file of multi-probe NP to see this in action

I have a full one without traces! Can I share that?

CodyCBakerPhD:
> I have a full one without traces! Can I share that?

Can start with that and then I'll request anything I can't find in it?

src/neuroconv/tools/spikeinterface/spikeinterface.py
alejoe91: done

tests/test_on_data/test_gin_ophys.py
alejoe91: done

src/neuroconv/tools/spikeinterface/spikeinterface.py
alejoe91: done

tests/test_on_data/test_gin_ecephys.py
CodyCBakerPhD: @alejoe91 To clarify, mostly out of curiosity, what is the command you used that caused this `=` vs `_` error?

src/neuroconv/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Thanks!

CodyCBakerPhD:
@alejoe91 So sorry... I appear to have broken something after trying to resolve a merge conflict - how did you want your tests to refer to the waveform recording less path here? https://github.com/catalystneuro/neuroconv/pull/255/files#diff-e4a9f87541cfa3dcbde126c9c4be15e73bb0a44267c2f2fc049f439d9e91e68dL1095-R1160

tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -1173,11 +1173,6 @@ def setUp(self):
             session_description="session_description1", identifier="file_id1", session_start_time=testing_session_time
         )
 
-    @classmethod
-    def tearDownClass(cls):
-        if cls.waveform_recordingless_path.is_dir():
-            shutil.rmtree(cls.waveform_recordingless_path)
-
```
alejoe91: @CodyCBakerPhD why did you remove this?

alejoe91:
> @alejoe91 So sorry... I appear to have broken something after trying to resolve a merge conflict - how did you want your tests to refer to the waveform recording less path here? https://github.com/catalystneuro/neuroconv/pull/255/files#diff-e4a9f87541cfa3dcbde126c9c4be15e73bb0a44267c2f2fc049f439d9e91e68dL1095-R1160

NP, I'll fix it

tests/test_ecephys/test_tools_spikeinterface.py
alejoe91: Oh just saw you're removing the tmpdir! :)

CodyCBakerPhD (approved):

PR #258, opened 2022-12-19
change FilePathType and FolderPathType from TypeVar to Union

bendichter:
## Motivation

See this [stackoverflow answer](https://stackoverflow.com/a/58903907). TypeVar enforces consistency across all uses of that typehint, whereas Union is just a simple union and does not enforce consistency. If we have two paths in a single function call, one should be allowed to be a Path while the other is a str, so Union is more appropriate.

## Checklist

### DIFF
--- a/src/neuroconv/utils/types.py
+++ b/src/neuroconv/utils/types.py
@@ -1,12 +1,11 @@
 """Authors: Luiz Tauffer, Cody Baker, Saksham Sharda and Ben Dichter."""
 from pathlib import Path
 from typing import Optional, Union
-from typing import TypeVar
 
 import numpy as np
 
-FilePathType = TypeVar("FilePathType", str, Path)
-FolderPathType = TypeVar("FolderPathType", str, Path)
+FilePathType = Union[str, Path]
+FolderPathType = Union[str, Path]
 OptionalFilePathType = Optional[FilePathType]
 OptionalFolderPathType = Optional[FolderPathType]
 ArrayType = Union[list, np.ndarray]


CodyCBakerPhD:
I believe this will trigger some errors in the tests; see https://github.com/catalystneuro/nwb-conversion-tools/pull/237 for when these types were introduced and why `Union` caused issues

bendichter:
@CodyCBakerPhD I can't tell from that convo why you can't just use `Union` for both

CodyCBakerPhD:
Surprised we don't have tests for this - there's a difference in behavior evident when you compare the code

```python
from neuroconv.datainterfaces import PhySortingInterface, SpikeGLXRecordingInterface

phy_schema = PhySortingInterface.get_source_schema()
spikeglx_schema = SpikeGLXRecordingInterface.get_source_schema()
```

between this branch and current `main`. Specifically, in this branch, lines https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/utils/json_schema.py#L109-L112 do not get triggered at all and so the JSON validation is slightly weaker (it only checks if the value passed is a `string`, whereas on `main` it tests that it is a `string` which represents either a `file` or a `directory` in order to prevent mistaken usage of `file_path` vs. `folder_path` inputs to an interface, which is what we seem to not have any tests for)

This is ultimately caused by how a simple `Union` does not allow distinction between references

```python
Type1 = Union[str, int]
Type2 = Union[str, int]

type_1 == type_2
>>> True
```

thus when you use either `Type1` or `Type2` as an annotation like

```python
def SomeFunc(
    some_input: Type1,
):
     ...
```

you lose the ability to distinguish if `some_input` is supposed to be `Type1` or `Type2`.

bendichter:
@CodyCBakerPhD what about something like
```python
from typing import Union

Type1 = Union[str, int]
Type1._name = "Type1"

Type2 = Union[str, int]
Type2._name = "Type2"
```

bendichter:
oh my

image

CodyCBakerPhD:
For understanding why any new hacky changes are necessary, can you elaborate on

> If we have two paths in a single function call, one should be allowed to be a Path while the other is a str, so Union is more appropriate.

with a code example of how `TypeVar` causes issues?

bendichter:
```python
FilePathType = TypeVar("FilePathType", str, Path)

def my_func(path1: FilePathType, path2:FilePathType):
    pass

my_func("path/to/file", Path("path/to/file"))  # <- this line fails if you are strictly enforcing typehints
```

Now this doesn't "fail" in the sense of causing the code not to run, because we are not currently strictly enforcing typehints, but for any machinery that expects typehints to be followed, this will be a violation.

bendichter:
This isn't causing any problems for me right now, and I see that I have opened a Pandora's box, so we can put this on the backburner until we come up with a less hacky way to deal with it

CodyCBakerPhD:
> This isn't causing any problems for me right now, and I see that I have opened a Pandora's box, so we can put this on the backburner until we come up with a less hacky way to deal with it

Anything involving the method signature stuff gets complicated quickly with this, but if this enforcements consistency could be a problem I can think on other solutions



> my_func("path/to/file", Path("path/to/file"))  # <- this line fails if you are strictly enforcing typehints

This is what I need to understand to determine if this can be a problem going forward - the code runs fine for me in Python 3.9

a) what do you mean by 'strictly enforcing typehints'?
b) How are you enforcing that, is it something in your IDE or is strict enforcement going to be a thing going into Python 3.11+?

I was always under the impression that annotations were just 'hints' and you'd have to use something like Pydantic to get strict enforcement of them - which of course, Pydantic is not good at all of the things we need for our source/metadata/option schema checking (that is, custom classes, including numpy derived types, and deeply nested dictionaries...) which is why we have these custom functions for inspecting signatures

CodyCBakerPhD:
I'm also unable to reproduce any errors from the original StackOverflow post from a couple years ago - maybe it *was* an issue with `typing.TypeVar` but is no longer?

bendichter:
> a) what do you mean by 'strictly enforcing typehints'?
> b) How are you enforcing that, is it something in your IDE or is strict enforcement going to be a thing going into Python 3.11+?

Right now, we don't enforce typehints, so any errors in our typehint specification won't be caught. It's actually tricky to do. I think you need to use mypy, or, as you pointed out, pydantic. In fact, I don't care much about validating inputs, I just want our typehints to accurately convey the function input requirements.

bendichter:
After digging into pydantic, I think it might make more sense to adopt the typehints they created for file and folder path

PR #264, opened 2022-12-22
Swap to Pydantic for schema validation

### LINKED ISSUE #261: [Feature]: replace signature_to_json_schema with pydantic

See demo:

```python
import json
from typing import List, Optional, Union

from pydantic.decorator import ValidatedFunction
from neuroconv.datainterfaces import SpikeGLXRecordingInterface

def print_json(dict_, indent=1):
    print(json.dumps(json.loads(dict_), indent=indent))


json_schema = ValidatedFunction(SpikeGLXRecordingInterface.__init__, None).model.schema_json()
print_json(json_schema)


def my_func(a: Optional[List[List[Union[str,int]]]] = None):
    pass

json_schema = ValidatedFunction(my_func, None).model.schema_json()
print_json(json_schema)
```

```json
{
 "title": "Init",
 "type": "object",
 "properties": {
  "self": {
   "title": "Self"
  },
  "v__duplicate_kwargs": {
   "title": "V  Duplicate Kwargs",
   "type": "array",
   "items": {
    "type": "string"
   }
  },
  "file_path": {
   "title": "File Path",
   "anyOf": [
    {
     "type": "string"
    },
    {
     "type": "string",
     "format": "path"
    }
   ]
  },
  "stub_test": {
   "title": "Stub Test",
   "default": false,
   "type": "boolean"
  },
  "spikeextractors_backend": {
   "title": "Spikeextractors Backend",
   "default": false,
   "type": "boolean"
  },
  "verbose": {
   "title": "Verbose",
   "default": true,
   "type": "boolean"
  },
  "args": {
   "title": "Args",
   "type": "array",
   "items": {}
  },
  "kwargs": {
   "title": "Kwargs",
   "type": "object"
  }
 },
 "required": [
  "self",
  "file_path"
 ],
 "additionalProperties": false
}
{
 "title": "MyFunc",
 "type": "object",
 "properties": {
  "a": {
   "title": "A",
   "type": "array",
   "items": {
    "type": "array",
    "items": {
     "anyOf": [
      {
       "type": "string"
      },
      {
       "type": "integer"
      }
     ]
    }
   }
  },
  "v__duplicate_kwargs": {
   "title": "V  Duplicate Kwargs",
   "type": "array",
   "items": {
    "type": "string"
   }
  },
  "args": {
   "title": "Args",
   "type": "array",
   "items": {}
  },
  "kwargs": {
   "title": "Kwargs",
   "type": "object"
  }
 },
 "additionalProperties": false
}
```

_No response_

Yes.

### Code of Conduct

CodyCBakerPhD:
fix #261

## Motivation

Showcase of how to use more rigorous Pydantic model validation instead of `jsonschema` validation.

Also moving it to the interface level, which will fix the current issue of interfaces not validating any schemas when used for stand-alone write to an NWB file.

This PR is only for `source`, subsequent PRs will extend to the other schemas.

## How to test the behavior?

Theoretically already tested but can add a few explicit tests to verify our understanding of the behavior going forward.

## Checklist

### DIFF
--- a/src/neuroconv/basedatainterface.py
+++ b/src/neuroconv/basedatainterface.py
@@ -32,38 +32,33 @@ class BaseDataInterface(ABC):
     source_data_to_validate: Optional[Dict[str, Any]] = None
 
     @classmethod
-    def get_source_model(cls, exclude: Optional[List[str]] = None):
+    def get_source_model(cls):
         """Infer the Pydantic model for the source_data from the `__init__` method signature (annotation typing)."""
-        exclude = exclude or list()
-        exclude.append("self")
-        return get_pydantic_model_from_method_signature(function=cls.__init__, exclude=exclude)
+        return get_pydantic_model_from_method_signature(function=cls.__init__, exclude=["self"])
 
     @classmethod
-    def get_source_schema(cls, exclude: Optional[List[str]] = None):
+    def get_source_schema(cls):
         """Infer the JSON schema for the source_data from the method signature (annotation typing)."""
-        return model_schema(cls.get_source_model(exclude=exclude))
+        return model_schema(cls.get_source_model())
 
     @classmethod
     def get_conversion_options_model(cls):
         """Infer the Pydantic model for the conversion_options from the `run_conversion` method signature (annotation typing)."""
-        return get_pydantic_model_from_method_signature(function=cls.run_conversion, exclude=["self", "nwbfile"])
+        return get_pydantic_model_from_method_signature(
+            function=cls.run_conversion, exclude=["self", "nwbfile", "metadata"]
+        )
 
     @classmethod
     def get_conversion_options_schema(cls):
         """Infer the JSON schema for the conversion options from the method signature (annotation typing)."""
         return model_schema(cls.get_conversion_options_model())
 
-    def _validate_source(self):
-        print(f"source_data: {self.source_data}")
-        print(f"source_data_to_validate: {self.source_data_to_validate}")
-        self.get_source_model().parse_obj(obj=self.source_data_to_validate or self.source_data)
-
-    def _validate_conversion_options(self, conversion_options: dict):
-        self.get_conversion_options_model().parse_obj(obj=conversion_options)
+    def validate_source(self, source_data_to_validate: dict):
+        self.get_source_model().parse_obj(obj=source_data_to_validate)
 
     def __init__(self, **source_data):
         self.source_data = source_data
-        self._validate_source()
+        self.validate_source(source_data_to_validate=self.source_data_to_validate or self.source_data)
 
     def get_metadata_schema(self):
         """Retrieve JSON schema for metadata."""
@@ -90,6 +85,9 @@ class BaseDataInterface(ABC):
     def get_conversion_options(self):
         """Child DataInterface classes should override this to match their conversion options."""
         return dict()
+
+    def validate_conversion_options(self, conversion_options: dict):
+        self.get_conversion_options_model().parse_obj(obj=conversion_options)
 
     @abstractmethod
     def run_conversion(
--- a/src/neuroconv/basedatainterface.py
+++ b/src/neuroconv/basedatainterface.py
@@ -1,12 +1,17 @@
 """Authors: Cody Baker and Ben Dichter."""
 from abc import abstractmethod, ABC
 import uuid
-from typing import Optional, Dict, Any
+from typing import Optional, Dict, Any, List
 
 from pynwb import NWBFile
 from pydantic.schema import model_schema
 
-from .utils import get_base_schema, get_schema_from_method_signature, get_pydantic_model_from_method_signature
+from .utils import (
+    get_base_schema,
+    get_schema_from_method_signature,
+    get_pydantic_model_from_method_signature,
+    FilePathType,
+)
 
 
 class BaseDataInterface(ABC):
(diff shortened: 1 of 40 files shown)


src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -39,13 +40,13 @@ class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
...
     def get_source_schema(cls):
-        source_schema = get_schema_from_method_signature(class_method=cls.__init__, exclude=["x_pitch", "y_pitch"])
+        source_schema = super().get_source_schema()
         source_schema["properties"]["file_path"]["description"] = "Path to SpikeGLX file."
         return source_schema
 
     def __init__(
         self,
-        file_path: FilePathType,
+        file_path: FilePath,
```
CodyCBakerPhD, line 49: This is the step that would have to be done to make more rigorous validation - it actually checks for file vs. format structure as well as existence on the different file systems.

src/neuroconv/basedatainterface.py
```diff
@@ -4,25 +4,36 @@
...
+    @classmethod
+    def get_source_model(cls):
+        """Infer the Pydantic model for the source_data from the method signature (annotation typing)."""
+        return ValidatedFunction(function=cls.__init__, config=None).model
+
     @classmethod
     def get_source_schema(cls):
         """Infer the JSON schema for the source_data from the method signature (annotation typing)."""
-        return get_schema_from_method_signature(cls.__init__, exclude=["source_data"])
+        return model_schema(cls.get_source_model())
```
CodyCBakerPhD, line 24: Maintaining the JSON view, even if not directly used any more, would still be good for future web form integration I assume

bendichter, line 24: yes, but we will need to adjust this a bit. Something like:

```python
json_schema = model_schema(model)

for prop in ["args", "kwargs", "self", "v__duplicate_kwargs"]:
    if prop in json_schema["properties"]:
        json_schema["properties"].pop(prop)
    
json_schema["required"].remove("self")
```

src/neuroconv/basedatainterface.py
```diff
@@ -4,25 +4,36 @@
...
 
     @classmethod
     def get_conversion_options_schema(cls):
         """Infer the JSON schema for the conversion options from the method signature (annotation typing)."""
         return get_schema_from_method_signature(cls.run_conversion, exclude=["nwbfile", "metadata"])
 
+    def _validate_source(self, source_data: dict):
+        self.get_source_schema().parse_obj(obj=source_data)
```
CodyCBakerPhD, line 32: This will also give more informative validation errors - currently, if there's a big `source_data` structure for a converter, it will report a validation error on the entire structure which can sometimes be so big as to completely fill the screen/console - more an issue with `metadata` than with `source`, but still conceptually possible

bendichter, line 32: don't you want to use the model here, not the schema?

CodyCBakerPhD: yeh

CodyCBakerPhD, line 32:
```suggestion
        self.get_source_model().parse_obj(obj=source_data)
```

CodyCBakerPhD:
OK it did indeed take a bit of tinkering to do properly.

It doesn't just come down to modifying the returned JSON schema, but the internal model schema as well, otherwise when it tries to validate it uses fields we might have wanted to exclude (the #1 example being 'self' when we apply this logic to any method of a class).

On top of that, while there are some easy methods for copying or generating instances of Pydantic models from other instantiated models, the use of the method from `pydantic.decorator` is fundamentally built to operate as we expect - as a class method - as such the `model` from the example code throughout discussion associated with this PR is never actually fully instantiated and fails such attempts to perform simple copies.

Instead, I do a small override of the part of the `__init__` for the class we're using (https://github.com/pydantic/pydantic/blob/a96794ddc1d8b9101bb8dc4e1f7d1e57b0ec03c0/pydantic/decorator.py#L60) to filter out the `fields` (which annoyingly is a purely intrinsic variable throughout that class) that we want to skip. For now I'm keeping the `super().__init__` call there but it is duplicated/unnecessary computation so might replace it with minimal setting of expected attributes.

A couple of tests are still failing here and there but they are all informative, reflective of how this integration at the level of the `BaseDataInterface` makes initialization much stricter (and I would say, safer). I'll continue debugging those

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -3,9 +3,10 @@
 import json
 
 from pynwb.ecephys import ElectricalSeries
+from pydantic import FilePath
```
CodyCBakerPhD, line 6: Opinion: should each interface call something like

i) `from pydantic import FilePath`

or

ii) `from ....utils import FIlePath` (and the `pydantic` import being done in `neuroconv.utils.json_schema` and exposed to `neuroconv.utils`

?

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
```diff
@@ -57,6 +58,7 @@ def __init__(
             assert fallback_sampling_frequency is not None, assert_msg
             sampling_frequency = fallback_sampling_frequency
 
+        self.keyword_aliases = dict(fallback_sampling_frequency="sampling_frequency")
```
CodyCBakerPhD, line 61: This is what I could come up with when the input arguments to an interface don't 100% correspond to the way we initialize them with SI and store the record as `self.source_data`

src/neuroconv/utils/json_schema.py
```diff
@@ -48,6 +53,64 @@ def get_base_schema(tag=None, root=False, id_=None, **kwargs) -> dict:
...
+        Can be useful when skipping reserved names such as 'self' in methods of a class.
+
+    Returns
+    -------
+    model : instance of a pydantic model
+    """
+    # The default model of the parent 'ValidatedFunction' includes many fields we do not need
+    # but typical model-copying methods only work on fully instantiated models (which also triggers validation),
+    # whereas the model of this decorator class works more as a class method
+    class CustomValidatedFunction(ValidatedFunction):
```
CodyCBakerPhD, line 75: I will say, due to the underdocumented nature of this function on Pydantic, I don't actually know if it's intended to be used generally. It seems to still be undergoing active development, and many values we might want to control (including more detailed `configs`) don't seem to be exposed or propagate very well.

Just something to keep in mind if we decide we want to offload all this to another library

src/neuroconv/utils/json_schema.py
```diff
@@ -48,6 +53,64 @@ def get_base_schema(tag=None, root=False, id_=None, **kwargs) -> dict:
...
+                fields=fields,
+                takes_args=False,
+                takes_kwargs=True,
+                config=dict(extra=Extra.allow),  # extras usually show up as fixed values passed into neo/SI
+            )
+
+    return CustomValidatedFunction(function=function, exclude=exclude).model
+
+
+def get_schema_from_pydantic_model(model: BaseModel) -> dict[str, Any]:
```
CodyCBakerPhD, line 99: Opinion: should we

i) simply import `from pydantic.schema import model_schema` in `BaseDataInterface`

or

ii) keep all schema stuff defined in the `utils` like I do here, and the `BaseDataInterface` instead imports `..utils import get_schema_from_pydantic_model`

tests/test_minimal/test_converter.py
```diff
@@ -49,17 +52,17 @@ class TestNWBConverterAndPipeInitialization(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         class InterfaceA(BaseDataInterface):
-            def __init__(self, **source_data):
-                super().__init__(**source_data)
+            def __init__(self):
```
CodyCBakerPhD, line 55: One change, that is probably good to have to added required specificity - any child of the `BaseDataInterface` must explicitly define all keyword arguments and attach annotation typing to them (or else override the new class method for getting the pydantic model for the source using `exclude` on the thing you want to skip validation of)

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
CodyCBakerPhD: Actually it's not going to work in more complicated cases like SpikeGLX where we take a `file_path` and alias it to a `folder_path` - thinking on a better solution

src/neuroconv/basedatainterface.py
```diff
@@ -1,28 +1,64 @@
...
 class BaseDataInterface(ABC):
-    """Abstract class defining the structure of all DataInterfaces."""
+    """
+    Abstract class defining the structure of all DataInterfaces.
+
+    Attributes
+    ----------
+    source_data : dict
+        Structure of input arguments used in upstream initialization.
+    source_data_to_validate : dict, default: None
```
bendichter, line 25:
```suggestion
    source_data_to_validate : dict, optional
```

This is the correct way to annotate when the default is None

src/neuroconv/datainterfaces/ecephys/spikegadgets/spikegadgetsdatainterface.py
```diff
@@ -18,9 +22,9 @@ def get_source_schema(cls):
 
     def __init__(
         self,
-        file_path: FilePathType,
-        gains: OptionalArrayType = None,
-        probe_file_path: OptionalFilePathType = None,
+        file_path: FilePath,
+        gains: List[float] = None,
```
bendichter, line 26:
```suggestion
        gains: Optional[List[float]] = None,
```

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
bendichter: I'd rather i. It's simpler and clearer to the user that we are relying on pydantic

src/neuroconv/utils/json_schema.py
```diff
@@ -48,6 +53,68 @@ def get_base_schema(tag=None, root=False, id_=None, **kwargs) -> dict:
...
+
+    Parameters
+    ----------
+    model : any pydantic model
+
+    Returns
+    -------
+    json_schema : dict
+    """
+    return model_schema(model)
```
bendichter, line 115: I don't think this function is necessary. We can just call `model_schema`, right?

CodyCBakerPhD:
replaced by #1016

PR #268, opened 2022-12-24
Fix DLC tests (from 12/23/2022 release)

CodyCBakerPhD:
Latest release of `deeplabcut` (yesterday) made `tensorflow` an extra dependency - while this is overall a good thing, since we don't need it at all for `dlc2nwb`, they haven't yet released the version of `dlc2nwb` which [contains a safer import](https://github.com/DeepLabCut/DLC2NWB/blob/main/dlc2nwb/utils.py#L16-L17) of the DLC package (though at the loss of recording the version provenance, it seems...)

I'll ask them to cut a new release and once they do I'll just replace this temporary addition with the up-pin on `dlc2nwb`. But for now, tests are breaking due to DLC attempting to import the uninstalled dependency: https://github.com/catalystneuro/neuroconv/actions/runs/3769847940/jobs/6409325021

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Upcoming
 
+### Fixes
+* Temporarily hotfixed the `tensorflow` dependency after the release of `deeplabcut==2.3.0`. [PR #268](https://github.com/catalystneuro/neuroconv/pull/268)
+
+
+
 ### Pending deprecation
 
 * Added `DeprecationWarning`s for `spikeextractors` objects in `neuroconv.tools.spikeinterface`. [PR #266](https://github.com/catalystneuro/neuroconv/pull/266)
--- a/src/neuroconv/datainterfaces/behavior/deeplabcut/requirements.txt
+++ b/src/neuroconv/datainterfaces/behavior/deeplabcut/requirements.txt
@@ -1,1 +1,2 @@
 dlc2nwb>=0.2
+tensorflow  # temporary until next release of dlc2nwb


CodyCBakerPhD:
Requested release: https://github.com/DeepLabCut/DLC2NWB/issues/17

bendichter:
thanks for tackling this, @CodyCBakerPhD. When are we attempting to import tensorflow?

CodyCBakerPhD:
> If the new version of deeplabcut does not require tensorflow, then why are they importing it in the package __init__?

*We* aren't  - *they* are:

We [import `dlc2nwb`](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py#L76)
-> current release of [`dlc2nwb` imports `deeplabcut`](https://github.com/DeepLabCut/DLC2NWB/blob/077e42d48a819840278775cd8b71293c51f5f185/dlc2nwb/utils.py#L8) (this would be fixed if they released the [current `main` branch](https://github.com/DeepLabCut/DLC2NWB/blob/main/dlc2nwb/utils.py#L17))
-> and `deeplabcut.__init__` unsafely imports `tensorflow` despite it being now [an 'optional' dependency](https://github.com/DeepLabCut/DeepLabCut/blob/main/setup.py#L58)

_why_ that is, is a good question.

bendichter:
-> and deeplabcut.__init__ unsafely imports tensorflow despite it being now an 'optional' dependency

It seems like we should say something about this

CodyCBakerPhD:
> It seems like we should say something about this

OK, raising an issue over there now.

CodyCBakerPhD:
Raised: https://github.com/DeepLabCut/DeepLabCut/issues/2104

Not sure how quickly any of this will move on their side though, I'd favor this hotfix for now just to keep our CI healthy (`main` is currently failing as is)

bendichter (approved):

Issue #202, opened 2022-11-21
[Bug]: `write_recording()` fails with `NWBZarrIO` backend


alejoe91:
### What happened?

When trying to write a `recording` object using the `NWBZarrIO` backend from `hdmf-zarr`, I get the following error related to the `len` of the `SpikeInterfaceRecordingDataChunkIterator`.

The same `nwbfile` does not produce any error when writing to HDF5.

@oruebel @rly I think this might be due to the way `neuroconv` handles the data iteration, but tagging you here in case it's a Zarr backend issue.

### Steps to Reproduce

```python
from pynwb import NWBFile
from hdmf_zarr.nwb import NWBZarrIO
from datetime import datetime
from neuroconv.tools.spikeinterface import write_recording

import spikeinterface.full as si


# create toy objects
rec, sort = si.toy_example()

# instantiate nwbfile
nwb_metadata = dict(session_description="toy_example", identifier="tpy", session_start_time=datetime.now())
nwbfile = NWBFile(**nwb_metadata)

# write recording
write_recording(rec, nwbfile=nwbfile)

# write to Zarr
with NWBZarrIO('test_recording_zarr.nwb', "w") as io:
    io.write(nwbfile)
```

### Traceback

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File ~/anaconda3/envs/nwb/lib/python3.9/site-packages/hdmf_zarr/backend.py:904, in ZarrIO.__list_fill__(self, parent, name, data, options)
    903 try:
--> 904     dset[:] = data  # If data is an h5py.Dataset then this will copy the data
    905 # For compound data types containing strings Zarr sometimes does not like wirting multiple values
    906 # try to write them one-at-a-time instead then

File ~/anaconda3/envs/nwb/lib/python3.9/site-packages/zarr/core.py:1373, in Array.__setitem__(self, selection, value)
   1372 else:
-> 1373     self.set_basic_selection(pure_selection, value, fields=fields)

File ~/anaconda3/envs/nwb/lib/python3.9/site-packages/zarr/core.py:1468, in Array.set_basic_selection(self, selection, value, fields)
   1467 else:
-> 1468     return self._set_basic_selection_nd(selection, value, fields=fields)

File ~/anaconda3/envs/nwb/lib/python3.9/site-packages/zarr/core.py:1772, in Array._set_basic_selection_nd(self, selection, value, fields)
   1770 indexer = BasicIndexer(selection, self)
[...]

PR #272, opened 2023-01-03
Standardize option names

CodyCBakerPhD:
## Motivation
Compression and iteration options have varied quite a bit over the years - this PR standardizes them into a single common structure (dictionaries `compression_options` and `iterator_options`) used within the tools and data interfaces. Those dictionaries contain all the information needed for someone to specify what type of compression/iteration to use. They consistent of two primary key/value pairs: `method` (the old `compression` or `iterator_type`) and `method_options` (the old `compression_opts` and `iterator_opts`).

A follow-up PR deals with docstrings, so don't focus on that here. This PR is just about the code itself; deprecation warnings and aliasing the inputs to the new standard.

## How to test the behavior?
Tests have been adjusted to reflect any changes in behavior, though for the most part is back-compatible until final deprecation and removal of the arguments.

## Checklist


bendichter:
Would this support dynamically loaded filters?

CodyCBakerPhD:
> Would this support dynamically loaded filters?

It would at least support whatever we previously supported through our other options. I can't say entirely for sure b/c we don't test it and I haven't tried to.

If it isn't currently supported, this framework at least could very well allow it (at least I don't see why not): the `method` would be the filter from `hdf5plugin` and `method_options` are potentially whatever key/value pairs it takes.

Discussion also applies to the future support for a Zarr backend, where the `method` could well be an actual `numcodecs` class to be passed into the `ZarrDataIO`.

bendichter:
I think I prefer the format of having `compression_type` and `compression_options` as separate arguments. Maybe we could change `compression_opts` to `compression_kwargs` for consistency with other places

CodyCBakerPhD:
@h-mayorquin @weiglszonja As primary users of the NeuroConv package, what are your opinions about this from any of the times you've had to specify these types of options in your conversions?

CodyCBakerPhD:
For reference, also tracked down the original suggestion for the `*_options` structure to this PR with @h-mayorquin: https://github.com/catalystneuro/neuroconv/pull/116, though I'm sure we also had in-person discussion on the topic.

> I think I prefer the format of having compression_type and compression_options as separate arguments.

The main point against that is as @h-mayorquin points out, the [code smell of too many parameters](https://en.wikipedia.org/wiki/Code_smell#:~:text=Too%20many%20parameters%3A%20a%20long,a%20more%20clean%2Dcut%20way.).

I'd want to follow a similar structure for data interface `stub_options` as well - if we used separate arguments as you suggestion then each `run_conversion` call would have 6 pairwise inter-dependent arguments as opposed to just 3 that enforce a common nesting structure.

> I think this is simpler for the user and nested dicts are more difficult to communicate ...

We're already utilizing such argument structure across this specific project (`source_data`, `metadata`, `conversion_options`), so if anything it would be more consistent with the API design of NeuroConv, and so long as it's well-documented it's not that much harder to communicate as opposed to having inter-dependent arguments.

I'd say it is also easier to communicate via YAML specifications since both linked arguments would be under the same heading

Example (a)

```yaml
compression_options:
  method: gzip
  method_options: 4
iterator_options:
  method: v2
  method_options:
    buffer_gb: 5.0
```

as opposed to example (b)

```yaml
compression_type: gzip
compression_options: 4
iterator_type: v2
iterator_options:
  buffer_gb: 5.0
```
which additionally allows for the possibility of scrambling these options, which could lead to hard-to-debug things like

```yaml
conversion_options:
  SomeInterface:
    compression_options: 4
    iterator_options:
      buffer_gb: 5.0
    some_other_options: ...
    ...
    compression_type: gzip
    even_more_options: ...
    ...
    iterator_type: v1  # would actually cause an error, but can you easily see how to fix it?
```

whereas in (a) the dependent arguments remain intentionally linked and cannot be disentangled.


> ...and validate.

After looking into Pydantic more I don't actually think it would be difficult to setup validation for this using Pydantic schemas for the common options. That's even aside from discussions about the JSON side of it.

h-mayorquin:
My personal vote goes for the simplification.

I find function signatures with a lot of arguments distracting and more intimidating at first glance. I also like keeping things semantically organized so everything related to compression is in one object (in this case a dict called `compression_options`) and everything related to iterative write in one object.  At an abstract level, I think this last preference can be defened by appealing to code cohesion or invoking the "code smell" of too many parameters. From a practical stand-point I feel that is easier to make changes and document with fewer keywords.

I think that some of the @bendichter concerns above can be solved by not nesting . Juse use a  one level dictionary  with the following structure: `compression_options=dict(method="method_name",  **method_kwargs)`. To be explicit, we should use:
`compression_options=dict(method="zip", gzip_level=8)` instead of `compression_options=dict(method="zip", method_options=dict(gzip_level=8))`.

CodyCBakerPhD:
> Just use a one level dictionary with the following structure: `compression_options=dict(method="method_name",  **method_kwargs)`

Hm... That might just work (and might even play nicer with validation), good idea!

I'll play around with it in a splinter PR from this base

h-mayorquin:
Kind of parallel to the central point but a good point to discuss now that @bendichter as mentioned dynamic filters and the following:

>. Maybe we could change compression_opts to compression_kwargs for consistency with other places and other APIs

The wrapper [H5DataIO](https://hdmf.readthedocs.io/en/stable/hdmf.backends.hdf5.h5_utils.html#hdmf.backends.hdf5.h5_utils.H5DataIO) has a `compression_opts` arguments already (that only works for gzip right now). We might want to have a different name for our argument at the level of neuroconv as there are things we want to modify (.e.g the `allow_plugin_filters` for dynamic filters for example) that are not properly `conversion_opts`. If we don't go for the simplification proposed here by @CodyCBakerPhD at least we should do as @bendichter suggests and go for `compression_kwargs`.

h-mayorquin:
Finally, a lot of the arguments that I made above focus more on the perspective of maintance but we are also making a library that should be simple and usable.

I feel that modifying iterative write and compression strategies is an advance use of our library. We don't want our users to have to modify this most of the time. From that perspective, I am not too concerned about using a dictionary vs a flat sructure (passing just all the arguments at the top level).  However, what if they want, what would be easier for them? Given that this is difficult to answer empirically maybe at least we can follow some standard way of doing this by other libraries. Unortunately, there is not a lot of consitency in the scientific stack. Some relevant examples:

Scipy favors the current design in their [optimize module](https://docs.scipy.org/doc/scipy/tutorial/optimize.html). That is, their functions take a name (the strategy) and a set of arguments corresponding to them as a different keyword.

Scikit-learn module on the [Gaussian Process ](https://scikit-learn.org/stable/modules/gaussian_process.html) uses the "all information in one object" design. They define a kernel with all the relevant arguments and pass that object.

Pandas uses a very similar strategy to what @CodyCBakerPhD is proposing here for their [to_csv](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html) function for handling compression options but use more of a flat passing of parameters in their [to_sql](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_sql.html) function.

I really don't know what is better from a usability perspective, scientists tend to use flat inputs in their scripts in my experience (spikeinterface looks like that) where the top level function just takes all the arguments of everything in a flat way. I guess we definitley don't want to go there.

bendichter:
We should support this:

https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/h5dataio.html#dynamically-loaded-filters

bendichter:
@h-mayorquin I agree, given the downstream APIs, it makes sense to keep it as `compression_opts`. 

> that only works for gzip right now

Is that true? It shouldn't be. See [PyNWB docs on DLFs](https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/h5dataio.html#dynamically-loaded-filters).

This naming convention is also used by hdf5plugin:

image

bendichter:
> They define a kernel with all the relevant arguments and pass that object. They use the strategy of one object.

I'd prefer this to a large dictionary. Classes have machinery for communicating and validating available args.

bendichter:
> Pandas uses a very similar strategy to what @CodyCBakerPhD is proposing here for their [to_csv](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html) function for handling compression options (one object strategy for compression)

Not really, `compression` takes a flat dict, not a nested dict.

CodyCBakerPhD:
> Not really, compression takes a flat dict, not a nested dict.

I'm fine with a flat dict, too. I think he means its similar in how its a single argument that is a dict instead of unpacking each key-value pair in that dict as additional arguments

bendichter:
> I'm fine with a flat dict, too. I think he means its similar in how its a single argument that is a dict instead of unpacking each key-value pair in that dict as additional arguments

well no one is advocating for that

h-mayorquin:
> > Pandas uses a very similar strategy to what @CodyCBakerPhD is proposing here for their [to_csv](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html) function for handling compression options (one object strategy for compression)
> 
> Not really, `compression` takes a flat dict, not a nested dict.

I was constrating pandas strategy to handle compression to the one that we have right now in `neuroconv` and the one that they have in `scipy.optimze`.

In Pandas they have:
```
compression={'method': 'gzip', 'compresslevel': 1, 'mtime': 1}.
```
That is, the function uses a single dictionary that has both the method name and the parameters instead of having both the method (compression) name and the arguments (compresion_opts) as different parameters of the function.

bendichter:
I like "compression" and "compression_opts". This is how it is done in: PyNWB, HDMF, hdf5plugin, and h5py.

h-mayorquin:
> > They define a kernel with all the relevant arguments and pass that object. They use the strategy of one object.
> 
> I'd prefer this to a nested dict. Classes have machinery for communicating and validating available args.

@bendichter 
I think at this point neither @CodyCBakerPhD or me want a nested dictionary. The strategies that I was constrating were the one where both `compression` and  `compression_opts` are used as separate keywords (as current neuroconv and scipy.optimize do) vs passing everything in a single object as a single keyword (in the examples above, scikit-learn does it with a class and pandas with a flat dict). That is, one vs two arguments for the specification of the compression strategies.

Ideally, I would like this to be a single object called whose esponsability is to instantiate the `HFDataIO` appropiately (the name of the keyword does not matter). The reason being, as you mention in the quote above, is that you can implement checks and relationships between the parameters in a coheisve way However, because this is more work and because most of those checks are done at the hdmf level I was fine with settling with a flat dict (which is like a very primitive data class in this context) but still a single object. That way if more checks are needed in our side or the definition of the strategy becomes more complex  we only need to change the dict for a class instead of changing both `compression` and `compression_opts` for an object.

> I like "compression" and "compression_opts". This is how it is done in: PyNWB, HDMF, hdf5plugin, and h5py.
Fair enough, keeping consistency with the other libraries in our ecosystem is an important argument.

CodyCBakerPhD:
> I like "compression" and "compression_opts". This is how it is done in: PyNWB, HDMF, hdf5plugin, and h5py.

We alias many arguments from upstream packages for enhanced user readability (in particular, expanding things like `_opts` into `_options` for less confusion over abbreviations), such as the `file_path` alias from neo/SI `filename`, `path`, and `nwbfile_path` alias from PyNWB `path` and h5py `name`. Point being, upstream argument names aren't always the best and consistency with those isn't something we tend to worry about since NeuroConv is intended to be the most user-friendly downstream tool (maybe one day, with the GUI, the only tool anyone has to use).

Two further points below, which count as separate issues but would be heavily influenced by these decisions...

<details>
  <summary>Generalizing this to be compatible with Zarr compression</summary>

Another point to how even that choice of names is going to break down once we add Zarr support to NeuroConv (impending per request #202) - though there isn't much documentation for the `ZarrDataIO` yet, judging from its [usage in the tests](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/tests/unit/test_io_zarr.py) and [its docval](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/src/hdmf_zarr/utils.py#L173-L177) it doesn't look like it follows the same argument conventions (`compressor: bool or Codec` instead of `compression: bool or str or hdf5plugin.Filter`).

It also doesn't seem to allow options outside of the `Codec` object, which instead of that fancy `dict` casting instead must retrieve config via

```python
from numcodecs import Blosc

compressor = Blosc(cname='zstd', clevel=3, shuffle=Blosc.BITSHUFFLE)
compressor.get_config()
>>> {'id': 'blosc', 'cname': 'zstd', 'clevel': 3, 'shuffle': 2, 'blocksize': 0}
```

which takes us back to the single-argument flat-dictionary approach that @h-mayorquin and I now support in favor; the aliasing for these cases would be `compression_options["method"] == compressor.get_config()["id"]` for Zarr and
`compression_options["method"] == dict(hdf5plugin.SomeFilter)["compression"]` for HDF5.


Now, the discussion about how to handle extra `**method_options` requires some more thought here. First, enabling dynamically loaded filters for NeuroConv can be done even on `main` by simply setting `allow_plugin_filters=True` by default, which I hope shouldn't harm anything even if the user never wants to use anything other than default. But that would allow them to pass either the `compression=filter_id` + `compression_opts=dict(hdf5plugin.SomeFilter)["compression_opts"]` or `compression=hdf5plugin.SomeFilter` directly if using NeuroConv via the API.

</details>



<details>
  <summary>YAML representation of compression options</summary>

Another thing to think about as we make efforts to make things more and more YAML compatible in line with the upcoming GUI project, that second dynamically loaded filter usage  wouldn't be possible. It would also be annoying to ask the user to look up the integer ID for the filter, so we're already looking at doing some kind of aliasing on our end for convenience.

That is, what could work right now on `main` if we make that small aforementioned change, the following would work in YAML

```yaml
conversion_options:
  SomeInterface:
    compression: 307
    compression_opts: 4
```

but it's rather unreadable. (What type of filter is that? What is the option name and what does a value of 4 mean?)

Though note this too would break if you wanted to specify more than one `compression_opts` (such as  the `tuple` that comes out of `dict(hdf5plugin.SomeFilter)["compression_opts"]` without adding some extra YAML machinery due to [YAML difficulty in parsing Python tuples](https://stackoverflow.com/questions/39553008/how-to-read-a-python-tuple-using-pyyaml).

My point is it would be far more convenient to design this in one of two other ways

(i)

```yaml
compression_options:  # this could also sit at a global level in the YAML if done this way
  method: gzip
  level: 4
```

whose keys beyond `method` are dynamically passed, or

(ii)

the original suggestion, whose first level is static and only the `method_options` level changes for each `method`

```yaml
compression_options:  # this could also sit at a global level in the YAML if done this way
  method: blosc
  method_options:  # not required for all `method`s
    cname: zstd
    clevel: 3  # note we could do either 'clevel' or 'level' for consistency
    shuffle: Bit  # Small alias from the Enum classes for this, which is BITSHUFFLE
```

which now that I look at it again does actually have an additional benefit of allowing separate specification of another compression-related field, the [Zarr filters](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/src/hdmf_zarr/utils.py#L178-L181) which could be specified like

```yaml
compression_options:  # this could also sit at a global level in the YAML if done this way
  filter: Delta
  filter_options:  # OK, not many filters have extra options (at least ones that we might want to suggest people use) but still possible
    ...
  method: blosc  # Note extra small aliasing for Zarr capitalization
  method_options:  # not required for all `method`s
    cname: blosclz
    clevel: 4
    shuffle: byte  # Small alias from the Enum classes for this, which is BYTESHUFFLE
```

</details>

CodyCBakerPhD:
> I think at this point neither @CodyCBakerPhD or me want a nested dictionary.

I may be rethinking that w.r.t. the Zarr compression filters actually - they would get passed into the `DataIO` but they are also separate entities from the compression method itself. See bottom of part 2 of above post for YAML detailing.

What I absolutely would not want as we go forward into that case, is to add more and more arguments to the outer level of `conversion_options`. Like, I would not want `compression`, `compression_opts`, `filter`, `filter_opts`.

Another important thing to remember is this entire discussion isn't just about standardizing compression options, but also iterator options and soon after 'stub' options, which all follow this similar pattern. It's inflates our arguments to have `stub`, `stub_options`, `compression`, `compression_opts`, `iterator_type`, `iterator_opts`, `filter`, `filter_opts`. Seems much nicer, even safer as per points of https://github.com/catalystneuro/neuroconv/pull/272#issuecomment-1370154477 to keep each pair contained in a respective object.

h-mayorquin:
@CodyCBakerPhD yes, I did not see your last point when I wrote, if you want to mantain the hierarchical structure in yaml then you needed the same in the signature.  

> Another point to how even that choice of names is going to break down once we add Zarr support to NeuroConv (impending per request https://github.com/catalystneuro/neuroconv/issues/202) - though there isn't much documentation for the ZarrDataIO yet, judging from its [usage in the tests](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/tests/unit/test_io_zarr.py) and [its docval](https://github.com/hdmf-dev/hdmf-zarr/blob/dev/src/hdmf_zarr/utils.py#L173-L177) it doesn't look like it follows the same argument conventions (compressor: bool or Codec instead of compression: bool or str or hdf5plugin.Filter).

Yep, maybe the [ZarrDataIO](https://hdmf-zarr.readthedocs.io/en/latest/hdmf_zarr.utils.html#hdmf_zarr.utils.ZarrDataIO)  shows that we are trying to follow the specific signature of `H5DataIO` too much, on the other hand, maybe `ZarrDataIO` should be similar to `H5DataIO`? not sure.

> First, enabling dynamically loaded filters for NeuroConv can be done even on main by simply setting allow_plugin_filters=True by default,

Can't we just make `allow_plugin_filters=True` when `compression` (or `compression_method`) is one of the the ids ( blosc, bitshuffle, bzip2, FCIDECOMP, lz4, ZFP, zstd)?

CodyCBakerPhD:
> Can't we just make allow_plugin_filters=True when compression (or compression_method) is one of the the ids ( blosc, bitshuffle, bzip2, FCIDECOMP, lz4, ZFP, zstd)?

It appears from the [HDMF code](https://github.com/hdmf-dev/hdmf/blob/2a42e4194853b856fe8b13a121d326b3d3d91de6/src/hdmf/backends/hdf5/h5_utils.py#L611-L613) that `allow_plugin_filters` isn't even used if the argument is a string and is one of the defaults that don't require `hdf5plugin` (note `h5py_filters.encode=('gzip', 'lzf', 'shuffle', 'fletcher32', 'scaleoffset')`) so it seems we could set it true without loss of generality. Otherwise, if anything, I'd flip it to True if it's _not_ in the default offerings, which I mostly point out as it is at least conceptually possible that they could add additional filters to `hdf5plugin` in the future (or there could even be custom plugins like the ones holding up #222 and #221: https://github.com/catalystneuro/neuroconv/actions/runs/3830900364/jobs/6519358809) so it might be annoying for us to support a static list.

There also seem to be other [more nuanced sets of conditions](https://github.com/hdmf-dev/hdmf/blob/2a42e4194853b856fe8b13a121d326b3d3d91de6/src/hdmf/backends/hdf5/h5_utils.py#L617-L618) for whether or not the `H5DataIO` can actually use the filter beyond that point, so I say we may as well offload that logic to HDMF.

PR #273, opened 2023-01-03
Dynamic Docstring: imported docstrings

CodyCBakerPhD:
## Motivation
We have many docstrings over the repo that repeat themselves multiple times. Each time we type it fresh in a new function, we run the risk of adding typos and other differences. Each time we copy/paste it we run the risk of propagating tiny differences that existed from the copy source. If we ever need to adjust some small part of the docstring, we have to remember to propagate that change everywhere.

This becomes an annoyance. I'm proposing two PRs using two different approaches to dealing with this.

This PR defines a `docstring.py` file much like our `requirements.txt` nesting, where they exist at the most common outer-most level for which they apply. Example, the new `compression_options` and `iterator_options` have docstrings that apply to all `tools` and ``datainterfaces` and thus exist at the `src/neuroconv/` level. If there were a docstring common only to recording interfaces, it would be defined at `src/neuroconv/datainterfaces/ecephys`.

The injection into the docstrings is similar to the sister PR, but I did take the extra step to make a nice util to handle a couple extra parsing tasks and reduce lines of code during injection. The imports and parsing of the upstream docstring is much easier in this PR, as it becomes just a normal Python import in this case.

## Checklist


bendichter:
I like this idea, but what about this:

```python
import func1

def func2(arg1):
    """"
    {func1.arg1}
    """"
    func1(arg1)
```

bendichter:
Let's discuss in https://github.com/catalystneuro/neuroconv/issues/278 then reopen when we settle on an approach (and if we even want to do this)

PR #260, opened 2022-12-20
Numpy docstring

bendichter:
## Motivation

Improve numpy docstrings and call signatures throughout. None of these changes should alter code function.

## Checklist

- [ ] Have you thoroughly read our [Developer Guide](https://neuroconv.readthedocs.io/en/main/developer_guide.html)?
- [ ] Have you ensured the PR description clearly describes the problem and solutions?
- [ ] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/neuroconv/pulls) for the same change?
- [ ] If this PR fixes an issue, is the first line of the PR description `fix #XXX` where `XXX` is the issue number?
- [ ] Did you update the [CHANGLOG.md](https://github.com/catalystneuro/neuroconv/blob/main/CHANGELOG.md) file on your branch to describe your changes?

PR #274, opened 2023-01-03
Dynamic Docstring: Numpydoc

CodyCBakerPhD:
## Motivation
We have many docstrings over the repo that repeat themselves multiple times. Each time we type it fresh in a new function, we run the risk of adding typos and other differences. Each time we copy/paste it we run the risk of propagating tiny differences that existed from the copy source. If we ever need to adjust some small part of the docstring, we have to remember to propagate that change everywhere.

This becomes an annoyance. I'm proposing two PRs using two different approaches to dealing with this.

This PR uses `numpydoc.docscrape` together with the assumption that all functions in our repo adhere to Numpy doc styles (thanks to #260) to automatically parse the docstring from one common source and inject it into other docstrings.

This does require us to designate one class/function to serve as the docstring source. The docstring we want to replicate will only be written once in that location and then 'read' and copied programmatically by all other docstrings for all other functions that is the common argument.

## Checklist


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -823,6 +831,11 @@ def add_electrical_series(
         ecephys_mod.data_interfaces["LFP"].add_electrical_series(es)
 
 
+add_electrical_series.__doc__ = add_electrical_series.__doc__.format(
+    compression_options_docstring=compression_options_docstring, iterator_options_docstring=iterator_options_docstring
+)
```
bendichter, line 836:
```suggestion
add_electrical_series.__doc__ = add_electrical_series.__doc__.format(
    compression_options_docstring=compression_options_docstring, iterator_options_docstring=iterator_options_docstring
)
```

I'm sure this works, but it's a bit counter-intuitive to anyone who is not expecting the doc to be constructed dynamically. Would it instead be possible to do this as 

```
from .docstrings import compression_docstring, iterator_options_docstring

def my_func(my_arg: int):
    """
    my_arg: int
        desc
    """ + compression_docstring + iterator_options_docstring
    pass
```

CodyCBakerPhD: Alas, the actual specifcatiton of docstrings at time of function definition must be a raw string literal, and cannot perform string operations on the same line.

Simple example to play with:

```python
def test_fun():
    """
    Testing123
    """ + "abc" + "def"
    pass

>>> test_fun?  # or help(test_fun), or print(test_fun.__doc__)
```

Thus modifications must take place after function definition, as done here and sister PR #273 (though I prefer most the util function in #273)

CodyCBakerPhD: Note in the example above, there is no docstring because the string operations nullified it. Remove the '+' temporarily to see the `Testing123` docstring

h-mayorquin: Isn't this a use case for a decorator?

bendichter: hmm thwarted

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -16,16 +16,35 @@
...
 
 SpikeInterfaceRecording = Union[BaseRecording, RecordingExtractor]
 SpikeInterfaceSorting = Union[BaseSorting, SortingExtractor]
 
+# Definitions of re-used docstrings
+compression_options_docstring = "\n\t".join(
+    next(
+        filter(
+            lambda x: "compression_options" in x[0],
+            NumpyDocString(BaseRecordingExtractorInterface.run_conversion.__doc__)["Parameters"],
```
h-mayorquin, line 35: Nice way to avoid parsing the string more than once per keyword.

h-mayorquin (commented):
This is a good idea, I have been thinking about this for a while.

One complain that I read about this type of solutions is that it makes it harder for someone to read the docstrings as they navigate the code (using a code editor I mean). What do you think of this?

Another option is to just add a test to our testing suite that iterates through all the docstrings and throws an error if not all the arguments matches (that is, if an argument as a different docstring in two different functions).
Pros: 
- We avoid increasing the complexity in a code base and keep the complexity only in the test suit.
- Tests do still show when navigating the code base in an editor.
Cons:
- Does not avoid the work of having the propagate manually the changes.
- May might be difficult to produce clear enough errors so developers can see how to correct the data string.

I don't feel very strong about this, just throwing some comments in the air.

bendichter:
> One complain that I read about this type of solutions is that it makes it harder for someone to read the docstrings as they navigate the code (using a code editor I mean). What do you think of this?

@h-mayorquin that's a good point..

bendichter:
> Another option is to just add a test to our testing suite that iterates through all the docstrings and throws an error if not all the arguments matches (that is, if an argument as a different docstring in two different functions).

I'd worry that we'd constrain ourselves too much with this approach. We might want to alter the docstring for two args that have the same name.

CodyCBakerPhD:
> Isn't this a use case for a decorator?

You know my feelings on decorators 🤣 especially after some recent errors from the NWB Inspector caused by 'too much decorator magic' I do try to avoid them as much as possible since they do a very good job at hiding what they actually do, which can make it very hard and time consuming to fix them if they cause a problem.

However, I'd be willing to do it here since it's only altering docstrings and not actual code behavior of the function (beyond calls to things like `help`)


> One complain that I read about this type of solutions is that it makes it harder for someone to read the docstrings as they navigate the code (using a code editor I mean). What do you think of this?
> @h-mayorquin that's a good point..

Yeah, it's my one reservation about this as well, which is why I prefer the import structure from #274 more since you can, from either an editor/IDE or the web interface, just navigate the import structure back to where its commonly defined (or from web interface, just click the 'reference' tab and should lead you straight to it).

Results in a little more work if you're visually scanning raw code, but it pays off via interactive mode. Note these are the docstrings rendered in the API docs as well, so it really is only when you're reading code at the raw level that it becomes an issue

h-mayorquin:
Using the automatic propagation methodology from a single source that you are proposing here I have the following remark. I don't see why we need to place the ground truth for an argument docstring in an specific function. I think it is better to just write the docstrings directly somewhere else like you do in #273 with the `docstrings.py` file. That way, you avoid having to do the extra computation of parsing at run time while keeping the docstring still accesible by click through navigation. Is there any downside to this that I am not seeing?

CodyCBakerPhD:
> We might want to alter the docstring for two args that have the same name.

Examples of this include `file_path`, which has a different description in every data interface, and `metadata`, which has a different description to inform the usage depending on the tool/interface.

Cases like that would not make sense to standardize like this.

But can we think of a case for extremely common arguments like `nwbfile`, `nwbfile_path`, `*_options` where we might want to alter them specific to a particular function? Their use is uniform across all our cases in all the tools/interfaces, so that's the case I'm focusing on here

h-mayorquin:
@CodyCBakerPhD 
> You know my feelings on decorators  especially after some recent errors from the NWB Inspector caused by 'too much decorator magic' I do try to avoid them as much as possible since they do a very good job at hiding what they actually do, which can make it very hard and time consuming to fix them if they cause a problem.

Yeah, I share those fellings, this is a recluctant recommendation on my side which is maybe a good argument in favor (don't use them unless YOU REALLY need them).  

> Yeah, it's my one reservation about this as well, which is why I prefer the import structure from https://github.com/catalystneuro/neuroconv/pull/273 more since you can, from either an editor/IDE or the web interface, just navigate the import structure back to where its commonly defined (or from web interface, just click the 'reference' tab and should lead you straight to it).

Great we seem to agree on this.

> Examples of this include file_path, which has a different description in every data interface, and metadata, which has a different description to inform the usage depending on the tool/interface. 

Right, I am assuming that we will only test for some of them to be unique. I am also expecting some exceptions even with that restriction but I have not think this through.

CodyCBakerPhD:
> Using the automatic propagation methodology from a single source that you are proposing here I have the following remark. I don't see why we need to place the ground truth for an argument docstring in an specific function.

It's also ambiguous and relatively arbitrary what you choose that function to be. Here in #274 I chose that to be the `BaseRecordingExtractor` but it also applies to `ophys` interfaces and `tools.roiextractors`, which makes for an awkward function import if pulling from a NumpyDoc.

This is all specific to the conversation about standardizing common docstrings like for `*_options`, another trajectory we're discussing in the broader issue #278 is for interface-specific arguments to 'inherit' parts of their docstring from upstream so that if SI ever changes those docstrings, we don't have to think about it and propagate it ourselves here on NeuroConv.

bendichter:
Let's discuss in #278 then reopen when we settle on an approach (and if we even want to do this)

PR #276, opened 2023-01-03
Use Pydocstyle to automatically check for Numpy docstring compliance

CodyCBakerPhD:
## Motivation

Testing this out to see how well it checks our docstrings for Numpy compliance. Probably doesn't get too specific about sub-standards like how to specify default values.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,3 +10,8 @@ repos:
     hooks:
     -   id: black
         exclude: ^docs/
+-   repo: https://github.com/PyCQA/pydocstyle
+    rev: 5.11.4
+    hooks:
+    -   id: pydocstyle
+        exclude: ^docs/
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
The D104's here also point to something I've been meaning to ask for a while: historically, NeuroConv has used top-level docstrings to indicate authorship, e.g., the first line of every `.py` file is `"""Authors: ..."""` but IMO that information is also really easy to glean (and goes into more line-by-line detail) from the git blame (which maybe we didn't know about back then).

I like how the NWB Inspector follows a more 'summary'-like convention where the top-level docstring instead gives a quick description of what that sub-module contains, e.g.: https://github.com/NeurodataWithoutBorders/nwbinspector/blob/dev/src/nwbinspector/register_checks.py#L1

bendichter:
@CodyCBakerPhD I agree, I always use git blame and don't find those lines to be that useful

Issue #280, opened 2023-01-06
[Feature]: Standardizing Conversion Config Options


CodyCBakerPhD:
### What would you like to see added to NeuroConv?

@bendichter 

Summary of conversation in PR #272.

I've broken it into several expandable sections as each point relates to multiple comments.

I've linked to original comments whenever they contain more information than the summary.

Conversations about this began back in #116 between @h-mayorquin and I.

### The Problem

It has become annoying to compile various fields in what we refer to as the `conversion_options` of `run_conversion` calls. These overlap with their corresponding parts of the `neuroconv.tools`.

<details>
 <summary>They tended to evolve differently which lead to some splits in how identical fields are specified between certain types of interfaces</summary><br />

In our ecephys data interfaces and tools we have

```yaml
conversion_options:
  stub_test:
  compression:
  compression_opts:
  iterator_type:
  iterator_opts:
```

In our ophys data interfaces and tools we have

```yaml
conversion_options:
  stub_test:
  stub_frames:
  compression_options:  # takes a field 'compression' to specify type
  iterator_type:
  iterator_options:
```

the behavior interfaces can be all over the place in which convention they choose, as raised in an older issue #42. <br />

</details>

<details>
 <summary>NeuroConv is not always consistent with external APIs (nor should it be)</summary><br />

From https://github.com/catalystneuro/neuroconv/pull/272#issuecomment-1370331255 in response to https://github.com/catalystneuro/neuroconv/pull/272#issuecomment-1370261918

- NeuroConv, like SpikeInterface, is a universal API for data ingestion from multiple sources.

- However, upstream methods that do similar things can have different call signatures across different libraries.

- For consistency, SpikeInterface and NeuroConv alias upstream `kwargs` with common references.
  - Example: SpikeInterface aliases Neo RawIO kwargs to  `file_path` and `folder_path` + `stream_name`
  - Example: Neo aliases most of it's dependencies to `filename` and `dirname`
  - Example: NeuroConv maps SI/ROIExtractors/Neo(icephys)/other behavior dependencies (like OpenCV) to `file_path` and `folder_path`

- Even the name of `DataIO` arguments like `compression`/`compression_opts` change between HDF5 and Zarr.

</details>

<details>
 <summary>Readability of argument names is important </summary><br />

- Abbreviations in argument names (and variable names in general) can actually make code unnecessarily hard to read.

[...]

PR #281, opened 2023-01-08
Implement dynamic filters compression with config / strategy class 

h-mayorquin:
This a specific implementation of an idea that is discussed in #280. Instead of using `compression` and `compression_opts` to specify the compression configuration / strategy we use a full fledge class, `HDF5CompressionStrategy`, which is in charge of two things:

1) Validating the conversion options.
2) Wrapping the data object into the appropiate instance of `H5DataIO` so this process can be encapsulated in our pipelines.

To showcase an advantage of this method I have added dynamic filter support and some tests. In line with the defaults of `H5DataIO`,  gzip is used when no `compresion_strategy` is provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Upcoming
 
+### Features
+* Implement compression strategy WIP [PR #281](https://github.com/catalystneuro/neuroconv/pull/281)
+
+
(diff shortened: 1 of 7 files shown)


h-mayorquin:
I remember that we have chosen another strategy so I am closing this.

PR #285, opened 2023-01-25
add links to request and build data interface from conversion gallery

bendichter:


weiglszonja (approved):
The link works and form looks good/clear to me, not sure how strict we are with `CHANGELOG.md` update in this case with small changes.

bendichter:
I don't think the changelog is necessary for small changes to the docs like this

CodyCBakerPhD:
> not sure how strict we are with CHANGELOG.md update in this case with small changes
> I don't think the changelog is necessary for small changes to the docs like this

Yep, that's why it's not a 'required' check in order to merge with approval. So for small stuff, perfectly fine to ignore it

The CI itself has proven invaluable as a reminder, though, as seen by https://github.com/NeurodataWithoutBorders/nwbinspector/pull/335 whose changelog will have to be updated after the fact because no one remembered at the time

PR #234, opened 2022-11-28
support Python 3.11

bendichter:
## Motivation

fix #192

## Checklist

- [ ] Have you thoroughly read our [Developer Guide](https://neuroconv.readthedocs.io/en/main/developer_guide.html)?
- [ ] Have you ensured the PR description clearly describes the problem and solutions?
- [ ] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/neuroconv/pulls) for the same change?
- [ ] If this PR fixes an issue, is the first line of the PR description `fix #XXX` where `XXX` is the issue number?
- [ ] Did you update the [CHANGLOG.md](https://github.com/catalystneuro/neuroconv/blob/main/CHANGELOG.md) file on your branch to describe your changes?

PR #288, opened 2023-01-28
SpikeGLXLFP Improvements

### LINKED ISSUE #287: [Bug]: SpikeGLXLFPInterface saves data with wrong name
### What happened?

SpikeGLXLFPInterface saves a timeseries with the name "ElectricalSeriesRaw" despite metadata suggesting it would save as "ElectricalSeriesLFP".

```python
import os
from neuroconv.datainterfaces import SpikeGLXLFPInterface

ECEPHY_DATA_PATH = "/Users/bendichter/data/neuroconv_testing_data/ephy_testing_data"

folder_path = f"{ECEPHY_DATA_PATH}/spikeglx/Noise4Sam_g0"
dir_name = os.path.split(folder_path)[1]
lf_path = os.path.join(folder_path, f"{dir_name}_imec0", f"{dir_name}_t0.imec0.lf.bin")

lf_inter = SpikeGLXLFPInterface(lf_path)
metadata = lf_inter.get_metadata()
print(metadata)
lf_inter.run_conversion("lf_out5.nwb", metadata=metadata)
```

metadata:
```
{'NWBFile': {'session_description': 'Auto-generated by neuroconv',
  'identifier': '4b617739-050b-4c6b-ab51-093a3ffa64a8',
  'session_start_time': datetime.datetime(2020, 11, 3, 10, 35, 10)},
 'Ecephys': {'Device': [{'name': 'Neuropixel-Imec',
    'description': '{"probe_type": "0", "probe_type_description": "NP1.0", "flex_part_number": "NP2_FLEX_0", "connected_base_station_part_number": "NP2_QBSC_00"}',
    'manufacturer': 'Imec'}],
  'ElectrodeGroup': [{'name': 's0',
    'description': 'a group representing shank s0',
    'location': 'unknown',
    'device': 'Neuropixel-Imec'}],
  'Electrodes': [{'name': 'shank_electrode_number',
    'description': '0-indexed channel within a shank.'},
   {'name': 'group_name',
    'description': 'Name of the ElectrodeGroup this electrode is a part of.'},
   {'name': 'contact_shapes', 'description': 'The shape of the electrode'}],
  'ElectricalSeriesLFP': {'name': 'ElectricalSeriesLFP',
   'description': 'LFP traces for the processed (lf) SpikeGLX data.'}}}
```

This metadata would suggest that the name of the ElectricalSeries will be "ElectricalSeriesLFP" but it is not. It is "ElectricalSeriewsRaw".

image

### Steps to Reproduce

```python
see above
```

### Traceback

_No response_

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.7

### Package Versions

_No response_

### Code of Conduct

CodyCBakerPhD:
fix #287 

## Motivation

There were two parts to the two separate problems noted in #287

1)

Very interesting bug related to the lazy loading mechanism for `SpikeGLXLFPInterface`, which inherits from the `SpikeGLXRecordingInterface` and thus shares any class variables set in the parent.

If a `SpikeGLXRecordingInterface` is initialized first in the same environment, then the `SpikeGLXLFPInterface` class inherits the extractor type set at instantiation of the other interface; if the LFP interface is used on its own or prior to the recording interface, the lazy loading cannot set the extractor type since there is no extractor by the name of `SpikeGLXLFP`. 

The correct solution always ought to have always been to set the `ExtractorName` since the interface name doesn't match it exactly.

What is interesting, and tricky to do in the testing suite, is that this was not detected until now because when dealing with SpikeGLX, the ap/lf bands tend to always be paired together, and as a matter of habit the ap recording is usually written first.

2)

The metadata not being used was something lost when we added stand-alone interface writing for all interfaces.

Specifically, `run_conversion` didn't know to use `es_key="ElectricalSeriesLFP"`, which made it use the defaults for raw recordings instead.

Technically, if you do the same workflow in the issue but with `ConverterPipe` or `NWBConverter` with a single data interface, the metadata would be used correctly because the converter `run_conversion` methods make use of a pre-function called `get_conversion_options`. The interfaces stand-alone conversions do not use that pre-function.

I don't think I'm crazy about having a `get_conversion_options` method that *must* be injected properly to have an intended affect. This PR removes it for SpikeGLX and absorbs the functionality into default functionalities of `run_conversion`.


bendichter (dismissed):
Thanks looks great!

src/neuroconv/datainterfaces/behavior/deeplabcut/requirements.txt
```diff
@@ -1,2 +1,2 @@
-dlc2nwb>=0.2
+dlc2nwb==0.2
```
CodyCBakerPhD, line 1: Latest release of dlc2nwb causes some secondary dependence issues in the CI

Nothing to worry about, but they're fixed in #234 

Pinning here so CI can pass

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -181,10 +179,44 @@ def get_device_metadata(self) -> dict:
...
+            starting_time=starting_time,
+            write_as=write_as or "raw",
+            write_electrical_series=write_electrical_series,
+            es_key=es_key or "ElectricalSeriesRaw",
+            compression=compression,
+            compression_opts=compression_opts,
+            iterator_type=iterator_type,
+            iterator_opts=iterator_opts,
+        )
+
```
bendichter, line 213: If you were to just adjust the default args to "raw" and "ElectricalSeriesRaw" the behavior would be more clear from the function signature. Is there an advantage to doing it this way?

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -205,6 +237,34 @@ def get_metadata(self) -> dict:
...
+            stub_test=stub_test,
+            starting_time=starting_time,
+            write_as=write_as or "raw",
+            write_electrical_series=write_electrical_series,
+            es_key=es_key or "ElectricalSeriesLFP",
+            compression=compression,
+            compression_opts=compression_opts,
+            iterator_type=iterator_type,
+            iterator_opts=iterator_opts,
+        )
```
bendichter, line 270: same question as above

tests/test_on_data/test_gin_ecephys/test_lfp.py
```diff
@@ -69,7 +87,10 @@ class TestConverter(NWBConverter):
         recording = converter.data_interface_objects["TestLFP"].recording_extractor
         with NWBHDF5IO(path=nwbfile_path, mode="r") as io:
             nwbfile = io.read()
-            nwb_lfp_electrical_series = nwbfile.processing["ecephys"]["LFP"]["ElectricalSeriesLFP"]
+            if expected_write_module == "raw":
+                nwb_lfp_electrical_series = nwbfile.acquisition["ElectricalSeriesLFP"]
+            else:
+                nwb_lfp_electrical_series = nwbfile.processing["ecephys"]["LFP"]["ElectricalSeriesLFP"]
             nwb_lfp_unscaled = nwb_lfp_electrical_series.data[:]
             nwb_lfp_conversion = nwb_lfp_electrical_series.conversion
```
bendichter, line 95: Do you think it might make sense to switch to an object-oriented test structure where we have an abstract Interface test and all the other interfaces inherit from it? That would make it easier to have specialized behavior for each interface, would allow us and would allow us to separate the interface tests into individual files that could handle testing the data I/O as well as the extraction of metadata.

bendichter: I know that would be a decent amount of refactoring work, but I think increasing the modularity there would allow us to scale more easily. For instance, a DataInterface being contributed from an external library would be able to have a test that inherits from the InterfaceTest class.

bendichter: obviously, this would be in a separate PR

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
CodyCBakerPhD: > Is there an advantage to doing it this way?

Not really, happy to change the default.

It should also be changed at a more general level anyway (may require a little more debugging for other interfaces, though) since this is a problem for any such interface

CodyCBakerPhD: Well, half generalized anyway. The LFP Interface can be adjusted that way at the top level, but the recording triggers some failures so I'll do it in a separate PR

tests/test_on_data/test_gin_ecephys/test_lfp.py
CodyCBakerPhD: I'll respond to that idea on the corresponding issue

bendichter (approved):

Issue #290, opened 2023-01-30
[Feature]: reorganize SpikeGLX interfaces


bendichter:
### What would you like to see added to NeuroConv?

Right now we have:

ap -> `SpikeGLXRecordingInterface`
lf -> `SpikeGLXLFPInterface`
nidq -> `SpikeGLXNIDQInterface`

This scheme is useful because each interface uses the appropriate extractor in the standard way.

The main problem with this layout is that it is too difficult for a user to create a conversion that includes all 3. Only 1 of them is the "recording" interface, but from a NeuroConv perspective they really all should be, because they are all data produced by a SpikeGLX recording. The ap is the stream that generally goes into spike sorting pipelines, but that doesn't make it more of a recording than lf or nidq. It often makes sense to differentiate between a recording and an LFP stream because often LFP is derived from the raw unfiltered stream, but that is not the case here. In this case, I think having a "recording" that does not include all three is misleading.

I would like to propose the following change:

ap -> `SpikeGLXAPInterface`, which creates an ElectricalSeriesAP (renaming the current SpikeGLXRecordingInterface)
lf -> `SpikeGLXLFInterface`, which creates an ElectricalSeriesLF (renaming the current SpikeGLXLFPInterface)
nidq -> `SpikeGLXNIDQInterface` (this is unchanged)

all three together -> `SpikeGLXRecordingInterface`

The `SpikeGLXRecordingInterface` would take as input the entire directory of the spikeglx recording and flags for what streams to include, defaulting to including all of them. It would automatically infer the paths of the ap, lf, and nidq streams and would load the data and metadata and do the conversion very similarly to how an NWBConverter works. If there is something non-standard, like a different folder structure for the spikeglx data, a user can always go down to the individual interfaces. The conversion gallery would continue to show the `SpikeGLXRecordingInterface` in its default configuration, which would convert all three streams.

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

CodyCBakerPhD:
> ap -> SpikeGLXAPInterface / lf -> SpikeGLXLFInterface

[...]

PR #292, opened 2023-01-30
Add SpikeGLXConverter

CodyCBakerPhD:
Draft of functionally-related ideas from #290 

After looking at the code code for `NWBConverter`, if we structure the proposed special class this way it may not require much modification to the core structure, other than removing the sub-nesting for `source_data` and `conversion_options`

`conversion_options` will be difficult no matter which approach we take, if you want to maintain control over each individual data stream's customizable options

The `run_conversion` displayed here is very nearly the same as the base `NWBConverter` and would dispatch the calls in same way at the outer level (a converter containing this converter) as if the sub converter was an interface; the only difference being customization of interface-level conversion options


CodyCBakerPhD:
This one is about halfway done

The special converter will successfully deliver an NWB file with the contents, but metadata needs polishing

Also an issue comes up when putting it inside another NWBConverter due to a conflict in the way the method `get_conversion_options_schema` works (class method in NWBConverter, instance method in ConverterPipe, which this is); pretty sure it's easy to fix, can try finishing it up when I get back from break

CodyCBakerPhD:
@bendichter Assuming CI agrees with me, this should be ready for review

CodyCBakerPhD:
Just tested for single-probe multi-band right now

Will add tests and some suspected metadata enhancements for multi-probe in a follow up

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
```diff
@@ -0,0 +1,108 @@
...
+                # To avoid warning/error turing write
+                # TODO: When PyNWB supports other more proper AUX electrode types, remove
+                interface.recording_extractor.set_property(key="shank_electrode_number", values=[np.nan] * num_channels)
+                interface.recording_extractor.set_property(key="contact_shapes", values=[np.nan] * num_channels)
+            data_interfaces.update({stream: interface})
+
+        super().__init__(data_interfaces=data_interfaces, verbose=verbose)
+
+    def get_conversion_options(self) -> dict:
+        return dict()
```
bendichter, line 87:
```suggestion
```

bendichter: do we need this?

tests/test_on_data/test_format_converters/test_spikeglx_converter.py
```diff
@@ -0,0 +1,69 @@
...
+        self.assertNWBFileStructure(nwbfile_path=nwbfile_path)
+
+    def test_in_converter_pipe(self):
+        spikeglx_converter = SpikeGLXConverter(folder_path=SPIKEGLX_PATH / "Noise4Sam_g0")
+        converter_pipe = ConverterPipe(data_interfaces=[spikeglx_converter])
+
+        nwbfile_path = self.tmpdir / "test_spikeglx_converter_in_converter_pipe.nwb"
+        converter_pipe.run_conversion(nwbfile_path=nwbfile_path, overwrite=True)
+
+        self.assertNWBFileStructure(nwbfile_path=nwbfile_path)
```
bendichter, line 69: Will this work in NWBConverter?

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
CodyCBakerPhD: Yes, yes we do - as I discovered while debugging

TBH I don't think the method itself is needed anywhere and should probably be removed entirely, but I'm playing by the rules in this PR

tests/test_on_data/test_format_converters/test_spikeglx_converter.py
CodyCBakerPhD: It will not, and after thinking about it it shouldn't be expected to

This is the first time it's come up, but an `NWBConverter `simply cannot, by its design, take a data interface or subconverter that is not designed around the notion of having all information available ahead of time via class methods

That was the point of the `ConverterPipe`, where information is collected and added on a dynamic basis

Specifically, the `get_schema`s and related methods would need to be able to know the `folder_path` that contains the spikeglx sessios, and needs to know how many streams you intend to load, which determines the metadata structure around how many of which streams you might have, etc. But all the `get_schema` class methods are designed to be agnostic to the input arguments of the `__init__`, so this use case doesn't make sense

In general, subconverters like this made from a `ConverterPipe` can only be run from a unifying `ConverterPipe` unless we start changing internal design  a bit more radically

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
bendichter: ok, in that case can you indicate this with a comment in the code?

CodyCBakerPhD: Done

tests/test_on_data/test_format_converters/test_spikeglx_converter.py
CodyCBakerPhD: **Unless** - we prohibit control over sub-stream specification and the role of the `SpikeGLXConverter` is always by default to load in as many streams as possible - and through the process we keep the conversion options fixed so they cannot be specified for individual streams and only apply globally to all the IMEC/NIDQ 

Then it might be possible

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
```diff
@@ -0,0 +1,107 @@
...
+from pynwb import NWBFile
+
+from .spikeglxdatainterface import SpikeGLXRecordingInterface, SpikeGLXLFPInterface
+from .spikeglxnidqinterface import SpikeGLXNIDQInterface
+from ....nwbconverter import ConverterPipe
+from ....tools.nwb_helpers import make_or_load_nwbfile
+from ....utils import get_schema_from_method_signature, FolderPathType
+
+
+class SpikeGLXConverter(ConverterPipe):
```
bendichter, line 15:
```suggestion
class SpikeGLXConverterPipe(ConverterPipe):
```

src/neuroconv/converters/__init__.py
```diff
@@ -0,0 +1 @@
+from ..datainterfaces.ecephys.spikeglx.spikeglxconverter import SpikeGLXConverter
```
bendichter, line 1:
```suggestion
from ..datainterfaces.ecephys.spikeglx.spikeglxconverter import SpikeGLXConverterPipe
```

bendichter:
@CodyCBakerPhD would it be possible to use the `get_streams` class method to generate the source and conversion options schemas on-the-fly?

bendichter:
something like

```python
def source_conversion_options_schema():
    conversion_options_schema = dict()
    streams = cls.get_streams
    for stream in streams:
        stream_type = get_stream_type(stream)
        if stream_type in ("ap", "lf"):
            conversion_options_schema[stream] = SpikeGLXRecordingInterface.get_conversion_schema()
        else:
            conversion_options_schema[stream] = SpikeGLXNIDQInterface.get_conversion_schema()

```

bendichter:
@CodyCBakerPhD can we get #298 through first?

bendichter:
See https://drive.google.com/drive/u/1/folders/1L9QJOav5VRxxo_qEH4LvtpO7zSbGsRn4 for examples of multi-probe data

CodyCBakerPhD:
I think the Svoboda one is NP V2 as well right?

The GIN dataset has even more diverse examples for V1 probes: https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/spikeglx/multi_trigger_multi_gate

Again, that will take some time to fully support the wide range of patterns - will leave that for a follow-up PR

Initial PR here is only debugged for multi-band single-probe + IMEC

bendichter:
> Again, that will take some time to fully support the wide range of patterns - will leave that for a follow-up PR

OK, fair enough. Let's integrate #298 and merge this, then I will work on generalizing to multi-probe, since that will be needed for the Svoboda conversion

bendichter:
I see what you mean. You could do

```python
@classmethod
def get_streams(folder_path):
    ...
```

but then you'd still need `get_conversion_options_schema()` to take as input a `folder_path` argument.

Is it really buying is anything to have `get_conversion_options_schema()` as a class method? `get_source_schema()` makes sense as a class method because it provides the information you need before initialization, but we could easily ask for conversion options after we've already initialized. What do you think?

CodyCBakerPhD:
> Is it really buying us anything to have get_conversion_options_schema() as a class method?

I can't think of any benefit; it's always validated after instantiation anyhow, and its role in the workflow occurs after the metadata schema, which itself was never a class method either

Keeping source schema as class does always make sense, even here because it's simply the session's `folder_path` and nothing more

CodyCBakerPhD:
Depends on #353 + add a test that it works in a NWBConverter and YAML spec

CodyCBakerPhD:
@bendichter This is working now within NWBConverters thanks to those changes to the class methods

Example files uploaded on Slack

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 * Added `PlexonSortingInterface` to support plexon spiking data. [PR #316](https://github.com/catalystneuro/neuroconv/pull/316)
 * Changed `SpikeGLXRecordingInterface` to accept either the AP or LF bands as file paths. Each will automatically set the correseponding `es_key` and corresponding metadata for each band or probe. [PR #298](https://github.com/catalystneuro/neuroconv/pull/298)
 * The `OpenEphysRecordingInterface` redirects to `OpenEphysLegacyRecordingInterface` for legacy format files instead of raising NotImplementedError. [PR #349](https://github.com/catalystneuro/neuroconv/pull/349)
+* Added a `SpikeGLXConverter` for easy combinatin of multiple IMEC and NIDQ data streams. [PR #292](https://github.com/catalystneuro/neuroconv/pull/292)
```
bendichter, line 23:
```suggestion
* Added a `SpikeGLXConverter` for easy combination of multiple IMEC and NIDQ data streams. [PR #292](https://github.com/catalystneuro/neuroconv/pull/292)
```

bendichter:
@CodyCBakerPhD sorry this review has taken so long

1. Now that we changed `get_conversion_options_schema` to an instance method instead of a class method, we should be able to change this from a `ConverterPipe` to an `NWBConverter`.
2. Can we adjust this so that the rows of the electrodes table matches the number of electrodes, not the number of channels?
3. Can we make this multi-probe?

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
```diff
@@ -0,0 +1,107 @@
...
+from ....tools.nwb_helpers import make_or_load_nwbfile
+from ....utils import FolderPathType, get_schema_from_method_signature
+
+
+class SpikeGLXConverterPipe(ConverterPipe):
+    """Primary conversion class for handling multiple SpikeGLX data streams."""
+
+    @classmethod
+    def get_source_schema(cls):
+        source_schema = get_schema_from_method_signature(method=cls.__init__, exclude=["streams"])
```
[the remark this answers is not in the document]
bendichter: oh, because you need the folder path to get the streams. Is there any way we can provide a user with the option to select specific streams?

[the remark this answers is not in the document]
CodyCBakerPhD: See the function `get_streams` just below this one.

I suppose it could be included in the source schema as a list of strings, but what I was really imagining for this on the GUIDE side (and in general) is it really actually ought to be a `Literal` of the number of actual choices for streams (so a dropdown/enum rather than freeform)

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
```diff
@@ -0,0 +1,107 @@
...
+
+        with make_or_load_nwbfile(
+            nwbfile_path=nwbfile_path,
+            nwbfile=nwbfile,
+            metadata=metadata,
+            overwrite=overwrite,
+            verbose=self.verbose,
+        ) as nwbfile_out:
+            for interface_name, data_interface in self.data_interface_objects.items():
+                data_interface.run_conversion(nwbfile=nwbfile_out, metadata=metadata)
```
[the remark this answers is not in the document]
CodyCBakerPhD: There isn't since #412

CodyCBakerPhD:
Only remaining issue is 

> Can we make this multi-probe?

Would you like me to test that out here or in a follow-up?

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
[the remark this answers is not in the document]
bendichter: I suppose in jsonschema this would translate to:

```json
{
  "type": "array",
    "items": {
      "enum" : streams
  },
  "uniqueItems": true
}
```

I'm not sure it would be easy to express the uniqueness in the function signature, so we might need to construct that part of the jsonschema manually

CodyCBakerPhD:
Will raise issues for two follow ups

1) extension for channels table, integration here
2) multi-probe metadata fixes

bendichter (approved):

.github/workflows/deploy-tests.yml
```diff
@@ -29,7 +29,7 @@ jobs:
   run-tests:
     needs: assess-file-changes
     if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
-    uses: catalystneuro/neuroconv/.github/workflows/testing.yml@main
+    uses: catalystneuro/neuroconv/.github/workflows/testing.yml@spikeglx_converter
```
bendichter, line 32: Is this part of getting around the streaming issues?

CodyCBakerPhD: We don't test streaming here in NeuroConv - the issues were with DANDI upload and Globus imports (both rely on `boto` at various levels)

But yes, relaxing the version on AWS CLI step of the testing workflows was a part of resolving the spontaneous issues on this PR

#420 should still happen though to avoid any kind of sudden live service interruption in the workflows going into the future

bendichter: can we revert this now?

CodyCBakerPhD: Done

bendichter (approved):

PR #294, opened 2023-02-03
Rename existing OpenEphys interface to `OpenEphysBinaryRecordingInterface`

weiglszonja:
## Motivation

Rename `OpenEphysRecordingInterface` to `OpenEphysBinaryRecordingInterface` to reflect this interface is for Open Ephys binary format (".dat" files). 
The refactor for interface name is useful when we add support for legacy format in #295.

## How to test the behavior?

Existing tests should pass

## Checklist


CodyCBakerPhD:
I would still keep an `OpenEphysRecordingInterface` present on this PR so we don't immediately break back-compatibility.

For this PR specifically, it can simply be a wrapper around or direct router to the `OpenEphysBinaryRecordingInterface` until we need to route between the two upcoming version types.

weiglszonja:
@CodyCBakerPhD  Did you mean something like this, e.g. in `datainterfaces/__init__.py`:
```
from .ecephys.openephys.openephysbinarydatainterface import (
    OpenEphysBinaryRecordingInterface,
    OpenEphysSortingInterface,
)
OpenEphysRecordingInterface = OpenEphysBinaryRecordingInterface
```

CodyCBakerPhD:
Huh, I don't think I've ever tried that...

If it works, it works - if not I was just imagining

```python
class OpenEphysRecordingInterface(OpenEphysBinaryRecordingInterface):
    ExtractorName = "OpenEphysBinaryExtractor"
```

Actually, now I think of it, you'll have to do that with the `ExtractorName` override or the lazy import mechanism won't work

CodyCBakerPhD (approved):

PR #295, opened 2023-02-03
Add interface for OpenEphys legacy format (.continuous files)

weiglszonja:
## Motivation

Add support for OpenEphys legacy format (.continuous files). Uses `OpenEphysLegacyRecordingExtractor` from `spikeinterface`. When the recording stream is not specified the channel stream is chosen when available.
  When channel stream is not available the `stream_name` must be specified. 

## How to test the behavior?
Added:
`tests/test_on_data/test_metadata/test_openephyslegacy_metadata.py`
`tests/test_on_data/test_gin_ecephys/test_openephyslegacy.py`
Also added the interface in `tests/test_on_data/test_gin_ecephys/test_raw_recordings.py`

## Checklist


CodyCBakerPhD:
Is there anything of note in the metadata?

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,78 @@
...
+        import pyopenephys
+
+        metadata = super().get_metadata()
+
+        folder_path = self.source_data["folder_path"]
+        fileobj = pyopenephys.File(foldername=folder_path)
+        session_start_time = fileobj.experiments[0].datetime
+
+        metadata["NWBFile"].update(session_start_time=session_start_time)
+        return metadata
```
weiglszonja, line 78: @CodyCBakerPhD fetching the `session_start_time` could be done the same way as for the binary interface, so ideally this should go to the redirect class in #296, but not sure this is possible when that class only returns the interfaces, there is no "inheritance" there.. but maybe i'm missing something there.

CodyCBakerPhD: Nah, you'd want it to be present at the lowest level in case anyone wanted to explicitly use one lower level interface or the other

The convenient router will merely serve as that; a decision between which interface to instantiate based on which version it auto-detects. Doesn't have to do anything extra beyond that

bendichter:
Looks great so far! Can you incorporate data tests using the test data on gin?

https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/openephys

weiglszonja:
> Is there anything of note in the metadata?

I checked the header and annotations from the reader, but didn't find anything of note there (only channel names, ids, file origin..) 
I added a metadata test for extracting `session_start_time`.

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -43,15 +45,22 @@ def __init__(
...
         all_annotations : bool, default: False
         stub_test : bool, default: False
         verbose : bool, default: True
         """
         self.RX = OpenEphysLegacyRecordingExtractor
+
+        self.folder_path = folder_path
+        if stream_name is None and stream_id is None:
+            stream_channels = self._get_stream_channels()
+            stream_id = list(stream_channels["id"])[0]
```
weiglszonja, line 62: Notes from @bendichter >
if  "CH" is present in signal streams then choose it, if not possible can return list of possible stream ids

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+            When the recording stream is not specified the channel stream ("CH") is chosen if available.
+            When channel stream is not available the stream_id or stream_name must be specified.
+        stream_name : str, default: None
+            The name of the recording stream.
+        block_index : int, default: 0
+        all_annotations : bool, default: False
+        stub_test : bool, default: False
+        verbose : bool, default: True
+        """
+        self.RX = OpenEphysLegacyRecordingExtractor
```
CodyCBakerPhD, line 59: Why is this necessary? Setting the `ExtractorName` above ought to be the only thing necessary

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
+"""Authors: Szonja Weigl, Cody Baker."""
+from typing import Optional
+from warnings import warn
+
+from neo.rawio import OpenEphysRawIO
+from spikeinterface.extractors import OpenEphysLegacyRecordingExtractor
+
```
CodyCBakerPhD, line 7: Need to either move these into the local scope of the relevant class methods, or you should be able to remove the `spikeinterface.extractors` import entirely (handled by lazy load mechanism of all extractor-based interfaces)

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+        if stub_test:
+            self.subset_channels = [0, 1]
+
+    def _get_stream_ids(self):
+        """Returns the identifier of signal streams for this folder."""
+        reader = OpenEphysRawIO(self.folder_path)
+        reader.parse_header()
+        signal_streams = reader.header["signal_streams"]
+        return list(signal_streams["id"])
```
CodyCBakerPhD, line 90: TBH I've wanted this functionality as a class method for a while - it's super convenient to be able to retrieve the info when you need it (which is inevitably _before_ you try initializing the interface)

Would prefer at the SpikeInterface level of course, like the one I [added to the IblStreamingRecording](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/extractors/iblstreamingrecording.py#L67-L68), but we can always start here and move our way up

What do you think?

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+        stream_name : str, default: None
+            The name of the recording stream.
+        block_index : int, default: 0
+        all_annotations : bool, default: False
+        stub_test : bool, default: False
+        verbose : bool, default: True
+        """
+        self.RX = OpenEphysLegacyRecordingExtractor
+
+        self.folder_path = folder_path
```
CodyCBakerPhD, line 61: This will be recorded automatically in the `self.source_data` dictionary so no need to set it explicitly

CodyCBakerPhD: Also, the only place it's used is in the private method that I'd like to see as a classmethod (and would therefore take `folder_path` as an input instead of as an attribute) so I think you can remove this

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+        Parameters
+        ----------
+        folder_path : FolderPathType
+            Path to OpenEphys directory.
+        stream_id : str, default: None
+            The identifier of the recording stream.
+            When the recording stream is not specified the channel stream ("CH") is chosen if available.
+            When channel stream is not available the stream_id or stream_name must be specified.
+        stream_name : str, default: None
+            The name of the recording stream.
```
CodyCBakerPhD, line 53: Seems a little redundant to take both of these - we usually take one and infer the other since there's a one-to-one mapping, or am I missing something about this case?

That is, how would these arguments interact if I specified both?

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+        block_index : int, default: 0
+        all_annotations : bool, default: False
```
CodyCBakerPhD, line 55: We usually just set this to `True` by default if it gives us more metadata from the source format - if not, then I wouldn't include it either way

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+        ----------
+        folder_path : FolderPathType
+            Path to OpenEphys directory.
+        stream_id : str, default: None
+            The identifier of the recording stream.
+            When the recording stream is not specified the channel stream ("CH") is chosen if available.
+            When channel stream is not available the stream_id or stream_name must be specified.
+        stream_name : str, default: None
+            The name of the recording stream.
+        block_index : int, default: 0
```
CodyCBakerPhD, line 54: Is this required by `neo`? Or I thought SpikeInterfaces usage of neo could accept `None`, in which case all blocks are loaded as multi-segment, which our `tools.spikeinterface` functions are capable of handling

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+    ):
+        """
+        Initialize reading of OpenEphys legacy recording (.continuous files).
+        See :py:class:`~spikeinterface.extractors.OpenEphysLegacyRecordingExtractor` for options.
+
+        Parameters
+        ----------
+        folder_path : FolderPathType
+            Path to OpenEphys directory.
+        stream_id : str, default: None
```
bendichter, line 48:
```suggestion
        stream_id : str, optional
```

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+            The identifier of the recording stream.
+            When the recording stream is not specified the channel stream ("CH") is chosen if available.
+            When channel stream is not available the stream_id or stream_name must be specified.
+        stream_name : str, default: None
```
bendichter, line 52:
```suggestion
        stream_name : str, optional
```

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
bendichter: yes, I'd be happy with this as a non-private class method.

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+        source_schema["properties"]["folder_path"][
+            "description"
+        ] = "Path to directory containing OpenEphys legacy files."
+        return source_schema
+
+    def __init__(
+        self,
+        folder_path: FolderPathType,
+        stream_id: Optional[str] = None,
```
bendichter, line 33:
```suggestion
        stream_id: str = "CH",
```

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+        Parameters
+        ----------
+        folder_path : FolderPathType
+            Path to OpenEphys directory.
+        stream_id : str, default: None
+            The identifier of the recording stream.
+            When the recording stream is not specified the channel stream ("CH") is chosen if available.
+            When channel stream is not available the stream_id or stream_name must be specified.
+        stream_name : str, optional
+            The name of the recording stream.
```
bendichter, line 53: What is the difference between stream_id and stream_name?

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+        if stream_name is None and stream_id is None:
+            stream_ids = self._get_stream_ids()
+
+            assert "CH" in stream_ids, (
+                "The identifier or name of recording stream must be specified when the "
+                "channel stream cannot be found in the data. The stream IDs in this data: "
+                f"{', '.join(stream_ids)}"
+            )
+
+            stream_id = "CH"
```
bendichter, line 71:
```suggestion
        available_stream_ids = self._get_stream_ids()
        if stream_id not in available_stream_ids:
            raise ValueError(
                f"The stream_id '{stream_id}' cannot be found in the data. The stream IDs in this data: "
                f"{', '.join(available_stream_ids)}"
            )
```

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
bendichter: It is better to inform the user of the default up front

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
CodyCBakerPhD: We prefer to use `stream_name` in NeuroConv

They form a 1:1 mapping, but the `stream_id` is 'more like' an index whereas `stream_name` is more human-readable

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
CodyCBakerPhD: Actually looks like the neo base in SI has a class method for it, so can simply harness that and return the `stream_name`: https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/extractors/neoextractors/neobaseextractor.py#L139-L147

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+
+        super().__init__(
+            folder_path=folder_path,
+            stream_name=stream_name,
+            block_index=block_index,
+            all_annotations=all_annotations,
+            verbose=verbose,
+        )
+
```
weiglszonja, line 59: @CodyCBakerPhD To harness [get_streams](https://github.com/SpikeInterface/spikeinterface/blob/aa4dfebafcce94b71461fc397a8ed4e69c8ddb53/spikeinterface/extractors/neoextractors/neobaseextractor.py#L139-L147) from SI, I can do that after `super().init` like:

```suggestion
        super().__init__(
            folder_path=folder_path,
            stream_name=stream_name,
            block_index=block_index,
            all_annotations=all_annotations,
            verbose=verbose,
        )
        
        stream_names = self.recording_extractor.get_streams(folder_path)

```
but this way we wouldn't know in advance what streams are in the data, or I am still missing a point here?

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
...
+        return source_schema
+
+    def __init__(
+        self,
+        folder_path: FolderPathType,
+        stream_id: Optional[str] = None,
+        stream_name: Optional[str] = None,
+        block_index: Optional[int] = 0,
+        all_annotations: Optional[bool] = False,
+        stub_test: bool = False,
```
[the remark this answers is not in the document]
weiglszonja: Thanks! removed

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,103 @@
+"""Authors: Szonja Weigl, Cody Baker."""
+from typing import Optional
+from warnings import warn
+
+from neo.rawio import OpenEphysRawIO
+from spikeinterface.extractors import OpenEphysLegacyRecordingExtractor
```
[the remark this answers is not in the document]
weiglszonja: Thanks, removed!

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
weiglszonja: Thanks, removed!

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
CodyCBakerPhD: > but this way we wouldn't know in advance what streams are in the data

Right, so if it's a class method

```python
@classmethod
def get_stream_names(cls, folder_path: FolderPathType) -> List[str]:
    from spikeinterface.extractors import OpenEphysLegacyExtractor

    return OpenEphysLegacyExtractor.get_streams(folder_path=folder_path)
```

Then the user can call it ahead of time to see what's available, then we can also double-check during `__init__`

```python
def __init__(self, ..., stream_name):
    available_strreams = self.get_stream_names(folder_path=folder_path)
    assert stream_name in available_streams, f"The selected stream '{stream_name}' is not in the available streams '{available_strreams}'!"
```

which is usually a more informative error than you might get if you just passed it to SI

weiglszonja: Ah i see, thanks for the clarification!

CodyCBakerPhD:
@weiglszonja This is really coming together!

Just need resolution on if we can just keep `all_annotations=True` all the time and remove `block_index` from the keywords (`None` all the time)

Also could you share an example stub NWB file to see what it looks like?

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,83 @@
...
+
+from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
+from ....utils import FolderPathType, get_schema_from_method_signature
+
+
+class OpenEphysLegacyRecordingInterface(BaseRecordingExtractorInterface):
+    """Primary data interface for converting legacy Open Ephys data (.continuous files).
+    Uses :py:class:`~spikeinterface.extractors.OpenEphysLegacyRecordingExtractor`."""
+
+    ExtractorName = "OpenEphysLegacyRecordingExtractor"
```
bendichter, line 12: Is this line needed?

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,83 @@
...
+        )
+        source_schema["properties"]["folder_path"][
+            "description"
+        ] = "Path to directory containing OpenEphys legacy files."
+        return source_schema
+
+    def __init__(
+        self,
+        folder_path: FolderPathType,
+        stream_name: Optional[str] = "Signals CH",
```
bendichter, line 29: Is this really optional? Can it be None?

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,83 @@
...
+
+        Parameters
+        ----------
+        folder_path : FolderPathType
+            Path to OpenEphys directory.
+        stream_name : str, default: "Signals CH"
+            The name of the recording stream.
+            When the recording stream is not specified the channel stream is chosen if available.
+            When channel stream is not available the name of the stream must be specified.
+        block_index : int, default: None
```
bendichter, line 46:
```suggestion
        block_index : int, optional
```

tests/test_on_data/test_gin_ecephys/test_raw_recordings.py
```diff
@@ -37,6 +37,9 @@
     OpenEphysBinaryRecordingInterface,
 )
 
+from neuroconv.datainterfaces.ecephys.openephys.openephyslegacydatainterface import OpenEphysLegacyRecordingInterface
```
bendichter, line 40: This should be imported the same way all the other interfaces are

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
CodyCBakerPhD: I would be interested to know as well - it's possible that this is needed in order for the lazy import to work, but also possible that the router triggers the correct sub-call at the lower level

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,73 @@
...
+            Path to OpenEphys directory.
+        stream_name : str, optional
+            The name of the recording stream.
+            When the recording stream is not specified the channel stream is chosen if available.
+            When channel stream is not available the name of the stream must be specified.
+        verbose : bool, default: True
+        """
+
+        available_streams = self.get_stream_names(folder_path=folder_path)
+        stream_name = stream_name or "Signals CH"
```
bendichter, line 44: I prefer to assign the value as a default of the function signature instead of using this pattern.

bendichter, line 44:
```suggestion
        stream_name: str = "Signals CH",
        verbose: bool = True,
    ):
        """
        Initialize reading of OpenEphys legacy recording (.continuous files).
        See :py:class:`~spikeinterface.extractors.OpenEphysLegacyRecordingExtractor` for options.

        Parameters
        ----------
        folder_path : FolderPathType
            Path to OpenEphys directory.
        stream_name : str, defulat: "Signals CH"
            The name of the recording stream.
        verbose : bool, default: True
        """

        available_streams = self.get_stream_names(folder_path=folder_path)
```

Is there something wrong with this solution?

CHANGELOG.md
```diff
@@ -1,12 +1,15 @@
 # Upcoming
 
 ### Features
+
+
```
bendichter, line 5:
```suggestion
```

bendichter (changes requested):
see comment. Looks good otherwise.

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
CodyCBakerPhD: I believe it can be `None` in SI, indicating to load whatever stream is available if there is only one stream (and will fail with error otherwise)

bendichter: That works too

src/neuroconv/datainterfaces/ecephys/openephys/openephyslegacydatainterface.py
```diff
@@ -0,0 +1,70 @@
...
+    ):
+        """
+        Initialize reading of OpenEphys legacy recording (.continuous files).
+        See :py:class:`~spikeinterface.extractors.OpenEphysLegacyRecordingExtractor` for options.
+
+        Parameters
+        ----------
+        folder_path : FolderPathType
+            Path to OpenEphys directory.
+        stream_name : str, defulat: "Signals CH"
```
bendichter, line 36:
```suggestion
        stream_name : str, default: "Signals CH"
```

bendichter (approved):

CodyCBakerPhD (approved):

PR #296, opened 2023-02-03
Redirect OpenEphys interface based on file format (binary or legacy)

weiglszonja:
## Motivation

Add abstract class that redirects to an OpenEphys interface based on file format;
For "legacy" format (.continuous files) the interface redirects to `OpenEphysLegacyRecordingInterface`. #295 
For "binary" format (.dat files) the interface redirects to `OpenEphysBinaryRecordingInterface`. #294 

## How to test the behavior?
```
```

## Checklist

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
@@ -1,0 +1,57 @@
+from abc import ABC
+from pathlib import Path
+from typing import Optional
+
+from .openephysdatainterface import OpenEphysBinaryRecordingInterface
+from .openephyslegacydatainterface import OpenEphysLegacyRecordingInterface
+from ....utils import FolderPathType
+
+
+class OpenEphysRecordingInterface(ABC):
+    """Abstract class that defines which interface class to use for a given Open Ephys recording."""
+
+    def __new__(
+        cls,
+        folder_path: FolderPathType,
+        interface_kwargs: Optional[dict] = None,
+        stub_test: bool = False,
+        verbose: bool = True,
+    ):
+        """
+        Abstract class that defines which interface class to use for a given Open Ephys recording.
+        For "legacy" format (.continuous files) the interface redirects to OpenEphysLegacyRecordingInterface.
+        For "binary" format (.dat files) the interface redirects to OpenEphysBinaryRecordingInterface.
+
+        Parameters
+        ----------
+        folder_path : FolderPathType
+            Path to OpenEphys directory (.continuous or .dat files).
+        interface_kwargs : dict, default: None
+            The keyword arguments to propagate to the recording interface depending on the format.
+            For "legacy" format use "stream_id" to define which recording stream to convert.
+            For "binary" format use "experiment_id" or "recording_id".
+        stub_test : bool, default: False
+        verbose : bool, default: True
+        """
+        self = super().__new__(cls)
+        self.folder_path = folder_path
+        self.interface_kwargs = interface_kwargs or dict()
+
+        if any((Path(self.folder_path).rglob("*.continuous"))):
+            return OpenEphysLegacyRecordingInterface(
+                folder_path=self.folder_path,
+                stub_test=stub_test,
+                verbose=verbose,
+                **self.interface_kwargs,
+            )
+
+        elif any((Path(self.folder_path).rglob("*.dat"))):
+            return OpenEphysBinaryRecordingInterface(
+                folder_path=self.folder_path,
+                stub_test=stub_test,
+                verbose=verbose,
+                **self.interface_kwargs,
+            )
+
+        else:
+            raise AssertionError("The data must be in 'legacy' (.continuous) or in 'binary' (.dat) format.")


weiglszonja:
Yes, that was my plan here and  also to rename the file to `openephysdatainterface.py`

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
```diff
@@ -0,0 +1,57 @@
...
+        self.folder_path = folder_path
+        self.interface_kwargs = interface_kwargs or dict()
+
+        if any((Path(self.folder_path).rglob("*.continuous"))):
+            return OpenEphysLegacyRecordingInterface(
+                folder_path=self.folder_path,
+                stub_test=stub_test,
+                verbose=verbose,
+                **self.interface_kwargs,
+            )
```
CodyCBakerPhD, line 46: If you swapped this to a `raise NotImplementedError(...)` then we could actually merge this PR as is - what do you think @weiglszonja?

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
```diff
@@ -0,0 +1,57 @@
+from abc import ABC
+from pathlib import Path
+from typing import Optional
+
+from .openephysdatainterface import OpenEphysBinaryRecordingInterface
+from .openephyslegacydatainterface import OpenEphysLegacyRecordingInterface
+from ....utils import FolderPathType
+
+
+class OpenEphysRecordingInterface(ABC):
```
CodyCBakerPhD, line 10: Does this actually have to be an abstract class? What do we gain by doing that (since not using any abstract methods?)

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
```diff
@@ -0,0 +1,57 @@
...
+    """Abstract class that defines which interface class to use for a given Open Ephys recording."""
+
+    def __new__(
+        cls,
+        folder_path: FolderPathType,
+        interface_kwargs: Optional[dict] = None,
+        stub_test: bool = False,
```
CodyCBakerPhD, line 17: `stub_test` is a conversion option

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
```diff
@@ -0,0 +1,57 @@
...
+from ....utils import FolderPathType
+
+
+class OpenEphysRecordingInterface(ABC):
+    """Abstract class that defines which interface class to use for a given Open Ephys recording."""
+
+    def __new__(
+        cls,
+        folder_path: FolderPathType,
+        interface_kwargs: Optional[dict] = None,
```
CodyCBakerPhD, line 16: This is an interesting approach... normally, this would be `**interface_kwargs` at the top level just as it is when it passes them off after routing

CodyCBakerPhD: Actually, yeah, I'd prefer that... Otherwise the user has to say `interface_kwargs=dict(their_intended_kwargs)` which is not as seamless as just taking and passing them along

The idea being that the `OpenEphysRecordingInterface` should behave exactly like either interface as far as the users can tell, but it figures out which version of the format to use under the hood

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
```diff
@@ -0,0 +1,57 @@
...
+        For "binary" format (.dat files) the interface redirects to OpenEphysBinaryRecordingInterface.
+
+        Parameters
+        ----------
+        folder_path : FolderPathType
+            Path to OpenEphys directory (.continuous or .dat files).
+        interface_kwargs : dict, default: None
+            The keyword arguments to propagate to the recording interface depending on the format.
+            For "legacy" format use "stream_id" to define which recording stream to convert.
+            For "binary" format use "experiment_id" or "recording_id".
```
bendichter, line 32: why? SI usually calls this param a stream_id. Is that the convention they use here?

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
bendichter: This is going to be tricky to get a good json schema for because we are not exposing the form of the input params. Would it be possible to homogenize the input args of the two different classes? What is the purpose of having one as "stream_id" and the other as  "experiment_id"?

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+        Abstract class that defines which interface class to use for a given Open Ephys recording.
+        For "legacy" format (.continuous files) the interface redirects to OpenEphysLegacyRecordingInterface.
+        For "binary" format (.dat files) the interface redirects to OpenEphysBinaryRecordingInterface.
+
+        Parameters
+        ----------
+        folder_path : FolderPathType
+            Path to OpenEphys directory (.continuous or .dat files).
+        interface_kwargs : dict, default: None
```
bendichter, line 29:
```suggestion
        interface_kwargs : dict, optional
```

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
CodyCBakerPhD: Well, that too... good point on the schema, I forgot that would have problems with that approach anyway

After looking into the [extractor](https://github.com/SpikeInterface/spikeinterface/blob/master/spikeinterface/extractors/neoextractors/openephys.py#L92-L95), the `experiment_id` has a similar role to multi-block, which SI ought to be able to handle as multi-segment, so we can safely ignore that argument here

I can't find a `recording_id` argument either

So our keywords for this interface are `folder_path`, `stream_name`, and that's it actually. No need to handle or pass anything else from what I see

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
weiglszonja: No, I based it on the EXTRACT extractor that redirects. I changed it to inherit from `BaseRecordingExtractorInterface`, is that correct?

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
weiglszonja: Thanks, removed.

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
weiglszonja: I just did that, after the Legacy interface is merged I'll open a follow up for changing this.

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
weiglszonja: I see why this was incorrect, the original intention was to propagate all the arguments that the two interface would need. I changed it now to take `folder_path` and `stream_name`.

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
weiglszonja: Resolving this as inteface_kwargs are removed

src/neuroconv/datainterfaces/ecephys/openephys/openephysroutingdatainterface.py
CodyCBakerPhD: Hmm I can't recall why we did that back then either hehe

Technically, with proper routing in `__new__`, it wouldn't produce any different behavior either way

It just also wouldn't make use of any the `ABC` behavior either

Was just curious if there was something I was overlooking

CodyCBakerPhD (approved):

CodyCBakerPhD:
@weiglszonja Looks great, thanks!

PR #298, opened 2023-02-03
refactor spikeGLX

### LINKED ISSUE #290: [Feature]: reorganize SpikeGLX interfaces

bendichter:
## Motivation
fix #290 

SpikeGLXRecordingInterface can now be used for ap and lf. It assigns es_key based on the filepath


CodyCBakerPhD:
OK so this approach still allows manual override if necessary, just shifting it from the `conversion_options` to the `__init__` to that the reference between `get_metadata` and `run_conversion` are in agreement

Works for me, let me know when tests are passing for a final look

One question is how to handle the conversion gallery, since this is the first time we've deprecated a popular interface...

After the next release that includes this PR, anyone running the current gallery will get a deprecation warning - not the best thing for a user to see upon their first attempts using our package

Then again, if we update the gallery now and they try it with a previous version of NeuroConv, the interface won't even exist so it'll be an import error...

Only thing I can think of is versioning the gallery/docs to make it clear what's on dev vs. most recent released versions

bendichter:
I realized that we should map to names of ElectricalSeries that allow for multi probes. For right now I am simply using the stream_id, e.g. the ElectricalSeries for ap 0 and lf 0 are called "imec0.ap" and the "imec0.lf". This solves the problem but may not be the best choice for readability. I suppose I would be open to changing to "ElectricalSeries_imec0.ap" or similar to keep with best practices

bendichter:
@CodyCBakerPhD I moved so much logic up into BaseRecordingExtractorInterface that I don't even need different types for processing the AP and LF streams, so I removed `SpikeGLXInterface` altogether and just made it a copy of `SpikeGLXRecordingInterface` for backward compatibility.

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -38,18 +35,17 @@ def add_recording_extractor_properties(recording_extractor) -> None:
...
 
 class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
-    """Primary data interface class for converting the high-pass (ap) SpikeGLX format."""
+    ExtractorName = "SpikeGLXRecordingExtractor"
 
     @classmethod
     def get_source_schema(cls) -> dict:
         source_schema = get_schema_from_method_signature(class_method=cls.__init__, exclude=["x_pitch", "y_pitch"])
-        source_schema["properties"]["file_path"]["description"] = "Path to SpikeGLX file."
+        source_schema["properties"]["file_path"]["description"] = "Path to SpikeGLX ap.in or lf.bin file."
```
bendichter, line 43:
```suggestion
        source_schema["properties"]["file_path"]["description"] = "Path to SpikeGLX ap.bin or lf.bin file."
```

bendichter:
@CodyCBakerPhD this PR does 3 things:
1. It deletes some automated BlackRock behavior that put certain ns files in processing rather than acquisition. I don't think the automated decision making was necessarily right and the user can use the "write_as" arg to control that manually.
2. It makes `es_key` an attribute of the `BaseRecordingExtractorInterface`. It defaults to "ElectricalSeries", which can be overridden as an arg in the `__init__`. The es_key arg is no longer available for `run_conversion`.
3. `SpikeGLXLFPInterface` has been deleted. This can now be done entirely with `SpikeGLXRecordingExtractor`, which infers whether it is an ap or lf stream based on the input path.


I know this is a lot for 1 PR. I was just trying to accomplish 3 but then 1 and 2 ended up being necessary. I can separate these out into individual PRs if that will make them easier for you to review.

bendichter:
nvm looks like a lot of bugs just cropped up

bendichter:
depends on #318

bendichter:
@CodyCBakerPhD thanks for helping me push these through. This is the next big one.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
 * The `OpenEphysRecordingInterface` now redirects to legacy or binary interface depending on the file format.
   It raises NotImplementedError until the interface for legacy format is added. [PR #296](https://github.com/catalystneuro/neuroconv/pull/296)
 * Added `PlexonSortingInterface` to support plexon spiking data. [PR #316](https://github.com/catalystneuro/neuroconv/pull/316)
+* Move `es_key` to the `__init__` of the `BaseRecordingExtractorInterface` class. It is no longer possible to use
+  this argument in the `run_conversion` method. [PR #318](https://github.com/catalystneuro/neuroconv/pull/318)
```
bendichter, line 17:
```suggestion
```

CHANGELOG.md
```diff
@@ -16,6 +16,9 @@
   It raises NotImplementedError until the interface for legacy format is added. [PR #296](https://github.com/catalystneuro/neuroconv/pull/296)
 * Added the `OpenEphysLegacyRecordingInterface` to support Open Ephys legacy format (`.continuous` files). [PR #295](https://github.com/catalystneuro/neuroconv/pull/295)
 * Added `PlexonSortingInterface` to support plexon spiking data. [PR #316](https://github.com/catalystneuro/neuroconv/pull/316)
+* Move `es_key` to the `__init__` of the `BaseRecordingExtractorInterface` class. It is no longer possible to use
+  this argument in the `run_conversion` method. [PR #318](https://github.com/catalystneuro/neuroconv/pull/318)
```
bendichter, line 20:
```suggestion
```

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglx_utils.py
```diff
@@ -62,3 +63,34 @@ def fetch_stream_id_for_spikelgx_file(file_path: FilePathType) -> str:
...
+
+    if "imDatBsc_pn" in meta:
+        metadata_dict.update(connected_base_station_part_number=meta["imDatBsc_pn"])
+
+    description_string = "no description"
+    if metadata_dict:
+        description_string = json.dumps(metadata_dict)
+    device_metadata = dict(name="Neuropixel-Imec", description=description_string, manufacturer="Imec")
+
+    return device_metadata
```
CodyCBakerPhD, line 96: Did you mean to delete this as a method in the main class? https://github.com/catalystneuro/neuroconv/blob/51c254bf28952f8d4da24070bafab1d7af552433/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py#L118-L148

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -146,7 +148,9 @@ def get_device_metadata(self) -> dict:
         return device
 
 
+# include for backwards compatibility
 class SpikeGLXLFPInterface(SpikeGLXRecordingInterface):
+    pass
```
CodyCBakerPhD, line 153: Did you want to eventually deprecate it altogether? If so just need a warning here

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglx_utils.py
bendichter: yes, I did. I think it got back in there with merges. I'll remove it again

CodyCBakerPhD:
Two small comments otherwise LGTM

CodyCBakerPhD (dismissed):

CodyCBakerPhD (approved):

PR #313, opened 2023-02-15
update globbing and tests to use regex for the match finding

bendichter:
## Motivation

@CodyCBakerPhD take a look at this and let me know what you think. This approach changes things around quite a bit to rely more on Python's regular expressions library. The approach here is to take the f-string notation and transform it into a regex expression, and then proceed from there using regex. I think moving into regex would have some advantages, as it would technically allow any user that knows regex to perform more advanced pattern matches.

## Checklist


bendichter:
Example usage:

```python
>>> fstring_pattern = "sub-{subject_id}/sub-{subject_id}/ses-{session_id}"
>>> pattern = generate_regex_from_fstring(fstring_pattern)
>>> print(pattern)
sub-(?P<subject_id>.+)/sub-(?P=subject_id)/ses-(?P<session_id>.+)

>>> import re
>>> re.match(pattern, "sub-001/sub-001/ses-abc")
{'subject_id': '001', 'session_id': 'abc'}
```

bendichter:
Translation of the regex expression sub-(?P<subject_id>.+)/sub-(?P=subject_id)/ses-(?P<session_id>.+)


* `(?P<subject_id>.+)`: Match the pattern `.+` and assign it to the group name "subject_id".
* `(?P=subject_id)` find another match for "subject_id : with the same pattern rule as the first instance.
* `(?P<session_id>.+)`: Match the pattern `.+` and assign it to the group name "session_id"

bendichter:
tests are failing due to escape character issue I have not yet figured out

CodyCBakerPhD:
Basic approach seems sound

In the YAML implementation, people will be able to specify either an f-string pattern or a precise regex pattern (if they know it)?

I'm not expecting the majority of users to know regex, but yes if they do its good to have the option for greater customization

bendichter:
@CodyCBakerPhD yeah, exactly. Would you be able to help with debugging on Windows? I don't have a Windows machine handy. It should just be a matter of properly handling backslashes.

CodyCBakerPhD:
> Would you be able to help with debugging on Windows? I don't have a Windows machine handy. It should just be a matter of properly handling backslashes.

Yeah I remember something like that popping up somewhere before

src/neuroconv/utils/globbing.py
```diff
@@ -1,90 +1,71 @@
...
-        # variable_names = ["x", "y", "z"]
-        # separators = ["a/", "b", "/c"", ""]
+
+    >>> fstring_pattern = "sub-{subject_id}/sub-{subject_id}/ses-{session_id}"
+    >>> pattern = generate_regex_from_fstring(fstring_pattern)
+    >>> print(pattern)
+    sub-(?P<subject_id>.+)/sub-(?P=subject_id)/ses-(?P<session_id>.+)
+
+    >>> import re
+    >>> re.match(pattern, "sub-001/sub-001/ses-abc")
```
bendichter, line 30:
```suggestion
    >>> re.match(pattern, "sub-001/sub-001/ses-abc").groupdict()
```

bendichter:
@CodyCBakerPhD actually, don't bother. I am working on a more robust solution now.

bendichter:
It might be better to use the parse library for this

h-mayorquin:
> It might be better to use the parse library for this

I remember that me and @CodyCBakerPhD discussed parse when implementing this. I remember that he pointed out some edge case that was not working in case you decide to be that way maybe the previous discussion is useful:

https://github.com/catalystneuro/nwb-conversion-tools/pull/457

bendichter:
Thanks for the pointer, @h-mayorquin !

bendichter:
I've been looking into this. At first, I was in favor of switching to `parse` because I thought it could parse arbitrary date formats. Then when I tried it, it was more limited than I realized. I opened an issue here: https://github.com/r1chardj0n3s/parse/issues/150

If they respond I'd like to consider reopening the issue of using parse.

PR #314, opened 2023-02-17
simplify round trip tests for RecordingExtractorDataInterface

bendichter:
## Motivation

These tests test the same functionality but are a bit simpler

## Checklist

### DIFF
--- a/tests/test_on_data/test_gin_ecephys/test_raw_recordings.py
+++ b/tests/test_on_data/test_gin_ecephys/test_raw_recordings.py
@@ -1,7 +1,6 @@
 import unittest
 import pytest
 import itertools
-from pathlib import Path
 from datetime import datetime
 from platform import python_version
 from sys import platform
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
> These tests test the same functionality but are a bit simpler

They do not have the same functionality

Look no further than the fact that `NWBConverter` and `ConverterPipe` validate schemas, whereas stand-alone interface write does not

These tests currently ensure that all `RecordingInterface`s can be used within an `NWBConverter`, and that calling `run_conversion` or other functions on that converter behave as expected

Your changes here test that they perform stand-alone interface operations, including conversion (which we should indeed test), in which case they should be separate tests

h-mayorquin (approved):
This simplification makes sense. Defining the TestConverter class here was boilerplate that we can now avoid.

EDIT. I see that I was too confident. I saw the changes and did not look at the results.

CodyCBakerPhD:
replaced by #355 ?

PR #309, opened 2023-02-10
[DO NOT MERGE]: Remove all spikeextractors usage

CodyCBakerPhD:
## Motivation

With the new release of SpikeInterface, we can finally fully deprecate all `spikeextractors` usage.

Still a number of items to do...

- PR #267 will need to happen first
- Remove the tests for `spikeextractors`
- Check the interfaces for any specific usage of `spikeextractors`

## Checklist

- [X] Have you thoroughly read our [Developer Guide](https://neuroconv.readthedocs.io/en/main/developer_guide.html)?
- [X] Have you ensured the PR description clearly describes the problem and solutions?
- [X] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/neuroconv/pulls) for the same change?
- [X] If this PR fixes an issue, is the first line of the PR description `fix #XXX` where `XXX` is the issue number?
- [X] Did you update the [CHANGLOG.md](https://github.com/catalystneuro/neuroconv/blob/main/CHANGELOG.md) file on your branch to describe your changes?

PR #324, opened 2023-02-22
simplify fetching of Extractor object

bendichter:
## Motivation

We really don't need to do all that fancy `__getattribute__` and `__new__` logic. The solution provided here works the same and will allow us to us the `ABC` metaclass in `BaseExtractorInterface` and downstream classes.

## How to test the behavior?
no change to user-interface. When building a an `ExtractorDataInterface`, call `self.get_extractor()` instead of `self.Extractor`.

## Checklist


CodyCBakerPhD:
Fine with this if you are

Represents a reversion in opinion from class attribute approach, as recently reinforced in https://github.com/catalystneuro/neuroconv/pull/74#discussion_r949468328 and originating from https://github.com/catalystneuro/nwb-conversion-tools/issues/54#issuecomment-663249122

Changing from attribute + magic functionality to just a function is more explicit, less complex, and thus more zen


> no change to user-interface. When building a an ExtractorDataInterface, call self.get_extractor() instead of self.Extractor.

That's gonna be a backward incompatible change - but we're doing so many of those in the next release its probably fine

There are some times when the attribute is actually used and expected but they tend to be internal usage so we ought to be able to catch them all


> will allow us to us the ABC metaclass in BaseExtractorInterface and downstream classes.

What is your motivation here? What functionality does `ABC` give if not utilizing any `abstractmethods`?

CodyCBakerPhD:
PS: did you mean for this branch to point to #309? What with all the spikeextractors removals

I can change base if so

bendichter:
@CodyCBakerPhD 

> Changing from attribute + magic functionality to just a function is more explicit, less complex, and thus more zen

yup, that's what I was thinking :-). It also provides the extractor as a classmethod, so it can be accessed before initialization, even in cases where it is not provided explicitly.

> What is your motivation here? What functionality does ABC give if not utilizing any abstractmethods?

This will communicate to users and developers that they should not be creating `BaseExtractorInterface` objects and it allows us to add abstract methods later if we want to.

I also considered creating a `getter` so we could keep the `self.Extractor` syntax, but that would require us to rename the class attribute `Extractor` so it would be backwards incompatible in that way. I thought using an explicit method instead of a getter was a bit more explicit, making it clear to users that there might be some shenanigans in fetching the `Extractor` class.

This caused some tests to fail, but they were all for spikeextractors_backend cases, so I just removed those modes of those specific DataInterfaces and those tests.

bendichter:
depends on #329

CodyCBakerPhD (approved):

PR #325, opened 2023-02-23
Add codespell

CodyCBakerPhD:
Testing to see if the mighty magic @yarikoptic has been sharing around works for us

Hoping it works much better than my hidden spell checker workflow


CodyCBakerPhD:
OK that is a much more reasonable list of misspellings to deal with

I'll fix them when I get some time after the conference or during travel

PR #328, opened 2023-02-23
Fix OpenCV mac tests

CodyCBakerPhD:
Fresh release of OpenCV yesterday appears to break our mac workflow on all Python versions.

Until investigating further, setting upper bound for just that OS


bendichter:
@CodyCBakerPhD it looks like we still have a failing osx test for py3.11

bendichter (approved):

PR #339, opened 2023-02-26
Conditional doctest

CodyCBakerPhD:
Trying to fix the doctests by conditioning on python version for Spike2

### DIFF
--- a/.github/workflows/doctests.yml
+++ b/.github/workflows/doctests.yml
@@ -63,4 +63,4 @@ jobs:
 
 
       - name: Run doctests
-        run: pytest docs
+        run: pytest docs --doctest-glob=*.rst --doctest-module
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD:
None of the existing 'skipif' functionality for doctest or pytest-sphinx seem to work with our setup

`pytest-sphinx` + `testcode` + `:skipif:` comes close but doesn't inherit or allow setup of the testing namespace and is only itself capable of single-line calls (due to using `eval` instead of `exec`)

bendichter:
It was a good effort. Maybe we can put the check explicitly in the gallery somehow?

CodyCBakerPhD:
> Maybe we can put the check explicitly in the gallery somehow?

That's what I recommend; just ask if the python version is < 3.10 before initializing the interface and running the conversion. Or don't doctest it at all, either way

PR #340, opened 2023-02-26
use literals

### LINKED ISSUE #96: Use typing.Literals

bendichter:
fix #96


tests/test_minimal/test_utils/test_json_schema_utils.py
```diff
@@ -38,7 +38,7 @@ class A:
         def __init__(self, a: int, b: float, c: Union[Path, str], d: bool, e: str = "hi"):
             pass
 
-    schema = get_schema_from_method_signature(A.__init__)
+    schema = get_schema_from_method_signature(A.__init__, exclude=["self"])
```
bendichter, line 41:
```suggestion
    schema = get_schema_from_method_signature(A.__init__)
```

src/neuroconv/utils/json_schema.py
```diff
@@ -76,54 +79,56 @@ def get_schema_from_method_signature(class_method: classmethod, exclude: list =
...
+            raise NotImplementedError(
+                f"The annotation type of '{param}' in function '{class_method}' is not implemented! "
+                "Please request it to be added at github.com/catalystneuro/nwb-conversion-tools/issues "
+                "or create the json-schema for this method manually."
+            )
+        if param.default is param.empty:
+            input_schema["required"].append(param_name)
+        elif param.default is not None:
+            args_spec[param_name].update(default=param.default)
+        input_schema["properties"] = dict_deep_update(input_schema["properties"], args_spec)
```
bendichter, line 131: This was actually a really small change. The large number of lines is due to the fact that I changed

```python
for param_name in params:
    if param_name not in exclude:
          ...
```

to 

```python
for param_name in params:
    if param_name in exclude:
        continue
     ...
```

I prefer this way because it is less indented

tests/test_behavior/test_audio_interface.py
```diff
@@ -89,8 +90,14 @@ def test_get_metadata(self):
...
-        expected_error_message = "Acoustic series can be written either as 'stimulus' or 'acquisition'."
-        with self.assertRaisesWith(exc_type=AssertionError, exc_msg=expected_error_message):
+        expected_error_message = """'test' is not one of ['stimulus', 'acquisition']
+
+Failed validating 'enum' in schema['properties']['Audio']['properties']['write_as']:
+    {'default': 'stimulus', 'enum': ['stimulus', 'acquisition']}
+
+On instance['Audio']['write_as']:
+    'test'"""
+        with self.assertRaisesWith(exc_type=jsonschema.exceptions.ValidationError, exc_msg=expected_error_message):
```
bendichter, line 100: I am taking the Literal values and turning them into "enum" entries in the json-schema, so if you try to enter a value that is not one of these, it will be caught by the json schema validation.

src/neuroconv/utils/json_schema.py
```diff
@@ -3,8 +3,11 @@
 import json
 import inspect
 from datetime import datetime
+from typing import _LiteralGenericAlias
```
bendichter, line 6: I should have known importing a private class would bite me in the butt. `typing._LiteralGenericAlias` isn't available in Python 3.8. I'll need to work out a more robust solution.

CodyCBakerPhD: Was this resolved? All tests are green after all

src/neuroconv/utils/json_schema.py
CodyCBakerPhD: Thanks for explaining

CodyCBakerPhD (approved):

PR #344, opened 2023-02-27
try out regex in typehints using pydantic axonadatainterface.py

bendichter:
@CodyCBakerPhD what are your thoughts on doing something like this to formalize file suffixes? I programmed the json schema creation so that it expresses the pattern. I'm sure if we end up switching to generating it with pydantic it will continue to work


bendichter:
See open issue [here](https://github.com/pydantic/pydantic/issues/5121)

CodyCBakerPhD:
@h-mayorquin This particular PR, as I see it, is less about parsing it for any iteration / metadata retrieval purposes but rather a way to combine the `pyandtic.FilePath` annotation with constrained `typing.Literal`; that is, our acceptable file paths for the given type would be anything that is (a) an existing file at that path on the system and (b) also follows the suffix structure we would expect for that format.

That said we'll have to be careful about such stringent typing for some formats; I've seen a couple over the years that can have nearly arbitrary suffixes on the binary blob part as long as the header is next to it and has all the info needed to memmap

bendichter:
Actually here's another way to accomplish the same thing that's a lot simpler and does not rely on pydantic

CodyCBakerPhD:
replaced by suggested suffixes across the interfaces, which highlights file selection in GUIDE and makes suggestions in the documentation but does not stop users from having slightly unexpected versions of a given suffix

PR #352, opened 2023-03-02
Add convenient lookup tables for interfaces based on name/modality

CodyCBakerPhD:
Could be of use to @garrettmflynn to simplify things like https://github.com/catalystneuro/nwb-guide/commit/0f17ef6f75fec9549be4af2c8b58c7498cfe47c1#diff-bfe22e422f7a3669b04d7ddfc8b73491eda63907f0a4da680bb6494f98f15d69R11


src/neuroconv/datainterfaces/__init__.py
```diff
@@ -132,3 +132,22 @@
     CsvTimeIntervalsInterface,
     ExcelTimeIntervalsInterface,
 ]
+
+interface_lookup = {interface.__name__: interface for interface in interface_list}
```
CodyCBakerPhD, line 136: @garrettmflynn Or would you prefer this to be called just `interfaces`?

garrettmflynn: If we aren't using this internally, it might be justified to keep this responsibility with the end-user.

garrettmflynn: You might just want to disregard this comment. My thinking is not quite pythonic—so I think I'm way off the mark here

src/neuroconv/datainterfaces/__init__.py
```diff
@@ -132,3 +132,22 @@
...
+ecephys_recording_interfaces = {
```
garrettmflynn, line 137: Would it be possible to export a single dictionary here (`interface_categories`) that has these variables as keys without the _interface suffix? This would be how I'd design it myself—though I'm not sure of any tradeoffs that come in Python

src/neuroconv/datainterfaces/__init__.py
CodyCBakerPhD: It's mostly that I look at a line like https://github.com/catalystneuro/nwb-guide/commit/0f17ef6f75fec9549be4af2c8b58c7498cfe47c1#diff-bfe22e422f7a3669b04d7ddfc8b73491eda63907f0a4da680bb6494f98f15d69R11 and see it as being overly complex for downstream usage

Happy to make easier internal references for anything like that - the job of NeuroConv is to make everyone's lives easier, after all

src/neuroconv/datainterfaces/__init__.py
CodyCBakerPhD: Are you thinking a double-nested structure?

```
{
    ecephys_recordings: {"SpikeGLX": SpikeGLXRecordingInterface, ...},
    icephys_recordings: {"Abf": AbfInterface},
    behavior: {"Video": VideoInterface, ...},
    ....
}
```

Also what do you think of the approach where the "<modality>Interface" is stripped?

src/neuroconv/datainterfaces/__init__.py
garrettmflynn: Now that I understand what's actually happening, I think `interfaces` is the way to go

src/neuroconv/datainterfaces/__init__.py
garrettmflynn: At least for NWB GUIDE, the idea would be to send all names with associated categories to the client. Then receive a list of chosen names that can be accessed with the `interfaces` dict and processed using another function.

CodyCBakerPhD: Can you give an example of what the GUIDE would request, as you're imagining it right now?

garrettmflynn: Double-nested would be what I'd go with. Also apologies for the hodge-podge of messages, I'd assumed that Github Issues would populate with new responses without a refresh...

CodyCBakerPhD: No worries, this is going great! I'll get that double nested ready in a second

CodyCBakerPhD:
@garrettmflynn Allright, checkout

```python
from neuroconv.datainterfaces import interfaces_by_category

interfaces_by_category
```

and let me know if that meets your needs

garrettmflynn:
Looks good to me! Here's the change I had to make to NWB GUIDE: 

```
interfaces = {}
interface_info = {}

for category in interfaces_by_category:
    for name in interfaces_by_category[category]:
        if (name not in interface_info.keys()): 
            interface = interfaces_by_category[category][name]
            interfaces[interface.__name__] = interface
            interface_info[name] = {'tags': [], 'name': interface.__name__}
        interface_info[name]['tags'].append(category)
```

From: 
```
interface_keys = list(filter(lambda key: 'Interface' in key, dir(ndi)))
interface_info = list(map(lambda key: {
    'name': key,
    'category': os.path.basename(os.path.dirname(os.path.dirname(inspect.getfile(getattr(ndi, key))))),
}, interface_keys))

interfaces = dict(zip(interface_keys, map(lambda key: getattr(ndi, key), interface_keys)))
```

If there's a more pythonic way to get this done, let me know!

garrettmflynn (approved):
Works in practice in the NWB GUIDE repo. Looks good to merge!

PR #355, opened 2023-03-03
modular unittest.TestCase classes for DataInterfaces

### LINKED ISSUE #289: [Feature]: increasing modularity of tests

Right now, we have one test for all interfaces and we use `parameterized` to iterate over all of the interfaces. This worked great at first, but now that we have 36+ supported interfaces it has become a bit cumbersome. Having one identical test for all DataInterface classes makes it difficult to build in customized behavior for any individual class. This structure also makes it difficult for externally contributed DataInterface classes to use our test structure.

At its current size, I think an object-oriented approach may be more appropriate. We would have one `DataInterfaceTest` abstract class that defines the I/O tests that are the same testing for all `DataInterface` classes and would allow for individual child classes to implement specialized tests for each interface. This would have a few benefits over the current structure:
1. We could move the metadata extraction tests to methods of these classes, unifying tests for each DataInterface into a single class.
2. All tests for a given format, including the I/O and metadata tests, would only require one read operation.
3. We would have the freedom to move tests that correspond to different source formats into different files.
4. If the `DataInterfaceTest` class is implemented in a `neuroconv.testing` module, DataInterfaces contributed from an external repository would be able to import it and use it to build their own tests.
5. We are accumulating a lot of flow control logic in the current pattern and I think this would just be a lot more readable.

Any downsides to this besides the obvious labor in refactoring?

_No response_

Yes.

### Code of Conduct

bendichter:
fix #289

This creates an abstract unittest class for testing each interface that does the following:

* allows for multiple "cases" (different input paths and configurations)
* validates source schema against json schema
* validates conversion schema against json schema
* validates metadata schema against json schema
* instantiates interface
* validates metadata against metadata schema
* validates extracted metadata against expectations
* runs conversion
* reads data and checks it against in-memory interface

The beauty of doing tests this way is that someone could write a RecordingInterface in a completely different library and all they would need to do to test it is:

```python
# test_interface.py
from unittest import TestCase

from neuroconv.tools.testing.data_interface import AbstractRecordingExtractorTest

from .my_library import MyRecordingInterface

class TestMyRecordingInterface(AbstractRecordingInterfaceTest, TestCase):
    data_interface_cls = MyRecordingInterface
    data_interface_kwargs = dict(file_path="path/to/data")
    save_directory = OUTPUT_PATH

```

This could really pave the way for users to submit custom interfaces from their on repositories.

Also, this allows us to incorporate metadata tests with the reading of the interface instead of in a different part of the test suite.


tests/test_on_data/test_sorting_interfaces.py
```diff
@@ -0,0 +1,87 @@
...
+    from .setup_paths import ECEPHY_DATA_PATH as DATA_PATH
+    from .setup_paths import OUTPUT_PATH
+except ImportError:
+    from setup_paths import ECEPHY_DATA_PATH as DATA_PATH
+    from setup_paths import OUTPUT_PATH
+
+
+class TestAxonRecordingInterface(SortingExtractorInterfaceTestMixin, TestCase):
+    data_interface_cls = PhySortingInterface
+    interface_kwargs = (dict(folder_path=str(DATA_PATH / "phy" / "phy_example_0")),)
```
bendichter, line 26:
```suggestion
    interface_kwargs = dict(folder_path=str(DATA_PATH / "phy" / "phy_example_0"))
```

tests/test_on_data/test_sorting_interfaces.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def check_extracted_metadata(self, metadata: dict):
+        assert metadata["NWBFile"]["session_start_time"] == datetime(2015, 8, 31, 0, 0)
+
+
+class TestPhySortingInterface(SortingExtractorInterfaceTestMixin, TestCase):
+    data_interface_cls = PhySortingInterface
+    interface_kwargs = dict(
+        folder_path=str(DATA_PATH / "phy" / "phy_example_0"),
+    )
```
bendichter, line 77:
```suggestion
    interface_kwargs = dict(folder_path=str(DATA_PATH / "phy" / "phy_example_0"))
```

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 
 ### Testing
 * The tests for `automatic_dandi_upload` now follow up-to-date DANDI validation rules for file name conventions. [PR #310](https://github.com/catalystneuro/neuroconv/pull/310)
+* modularize tests for DataInterface classes using `unittest.TestCase` [PR #355]((https://github.com/catalystneuro/neuroconv/pull/355))
```
bendichter, line 25:
```suggestion
* modularize tests for DataInterface classes using `unittest.TestCase` [PR #355](https://github.com/catalystneuro/neuroconv/pull/355)
```

bendichter:
@CodyCBakerPhD I'm not sure why codespell is failing. What are the words that are setting it off? How do I add them to an ignore list?

bendichter:
(this can wait until after Cosyne)

CodyCBakerPhD:
> I'm not sure why codespell is failing. 

You can view it in the workflow: click on 'Details': https://github.com/catalystneuro/neuroconv/actions/runs/4336597805/jobs/7572002186

And scroll down to error: `Error: ./tests/test_on_data/test_recording_interfaces.py:75: Pyton ==> Python`

Codespell is so magical that I've not seen a case yet where a true 'ignore' is needed.

You can fix this locally by running `codespell -w` (`-w` auto-fixes easy change slike this), I plan to add that capability to the workflow eventually so it's more like black. Won't work in all cases, it can fail when a misspelling has multiple candidate corrections.

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -0,0 +1,163 @@
...
+    BaseSegmentationExtractorInterface,
+)
+
+
+def json_serial(obj):
+    """JSON serializer for objects not serializable by default json code"""
+
+    if isinstance(obj, (datetime, date)):
+        return obj.isoformat()
+    raise TypeError("Type %s not serializable" % type(obj))
```
CodyCBakerPhD, line 35: I'd say just import the `NWBMetaDataEncoder` from `neuroconv.utils` : https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/utils/json_schema.py#L16

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -0,0 +1,163 @@
...
+
+        self.check_extracted_metadata(metadata)
+
+    def run_conversion(self, nwbfile_path: str):
+        metadata = self.interface.get_metadata()
+        metadata["NWBFile"].update(session_start_time=datetime.now().astimezone())
+        self.interface.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, metadata=metadata)
+
+    @abstractmethod
+    def check_read(self, nwbfile_path: str):
```
CodyCBakerPhD, line 72: I'd recommend making this more specific as `check_roundtrip_read` since that's really what's being tested

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -0,0 +1,163 @@
...
+        pass
+
+    def check_extracted_metadata(self, metadata: dict):
+        pass
```
CodyCBakerPhD, line 76: OK, not required...

But this is what someone would override any time they want to make assertions on extracted metadata structure?

CodyCBakerPhD:
Idea looks quite sound, and quite beneficial

To make the sequence of changes more manageable, could you break up this large PR into the following bite-size pieces?

PR 1) The base Mixin class

This first one could be approved and merged quickly, the others could all be separate (unchained) for modularity 

PR 2) Implementation on recording
PR 3) Implementation on sorting
PR 4) Implementation on imaging
PR 5) Implementation on segmentation
PR 6) Implementation on behavior

I really just want to be careful and make sure during review that we aren't loosing anything along the way, and with a refactor this big it would be relatively easy to drop a line or two might be important

bendichter:
@CodyCBakerPhD that breakdown seems reasonable. I'll keep this one up for reference

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Oh, actually the docstring for the method might be even better - that way the user can run `.check_extracted_metadata?` and it can instruct them what to do (and maybe show a small example of what such an assertion could include)

CodyCBakerPhD:
> that breakdown seems reasonable

Thanks, I know it's a lot of work but it helps tremendously for review

PR #359, opened 2023-03-08
Modular interface tests recording

bendichter:


CodyCBakerPhD:
A couple conflicts now due to the inactivation of the maxwell test

CodyCBakerPhD (approved):

CodyCBakerPhD:
Some ubuntu-only tests are causing irregular stalls in the CI (full tests just keep on running for forever until terminated by GitHub)

I've seen this kind of behavior before but I don't believe we discovered the true underlying cause (i.e., which format is responsible)

CodyCBakerPhD (approved):

PR #363, opened 2023-03-09
Disable DANDI test temporarily (again)

CodyCBakerPhD:
CI breaking again due to failure on DANDI server

Good to catch so we can warn them and get a fix ASAP

But for years this test has run without hardly ever causing issues, wonder what's been going on lately


CodyCBakerPhD:
The better long-term solution might just be to turn this one test into a separate check run outside of the global pytest, kind of like [how we treat imports](https://github.com/catalystneuro/neuroconv/blob/temp_disable_dandi_test_again/tests/imports.py)

and then make the result of the test a non-requirement for merging, so we can be aware of upload issues as they happen (since that's pretty important for us and user base), but not hold up our own dev cycle

bendichter (approved):

CodyCBakerPhD:
> it looks like the MaxOneRecordingInterface is having problems for Python 3.11 

Great, yet another intermittent issue then - looks like rerunning worked the second time 🤨

PR #376, opened 2023-03-17
define base metadata schema

### LINKED ISSUE #14: Centralize definitions for metadata schemas

bendichter:
fix #14

* Define base metadata schema in separate file
* Use constraints for Subject that follow DANDI best practices
* Fetch this base schema for `NWBConverter` and for `BaseDataInterface`
* modify existing yaml conversion files to follow new stricter subject requirements
* rmv `get_schema_for_NWBFile`

Future:
* Make subject required
* Add util to remove "required" from json schema dicts recursively if `run_conversion` is writing to an NWB file in append mode.
*


src/neuroconv/schemas/base_metadata_schema.json
```diff
@@ -0,0 +1,171 @@
...
+        }
+      }
+    },
+    "Subject": {
+      "type": "object",
+      "required": [
+        "subject_id",
+        "sex",
+        "species"
+      ],
```
CodyCBakerPhD, line 114: I know this gets a little tricky with the 'either/or' logic with `age` _or_ `date_of_birth` - would it be totally out there to consider _both_ as requirements and auto-fill the other when one is specified?

Not for NWB files in general of course, just through NeuroConv/GUIDE automation

src/neuroconv/schemas/base_metadata_schema.json
```diff
@@ -0,0 +1,171 @@
...
+          "description": "Terms to search over",
+          "type": "array",
+          "items": {
+            "title": "keyword",
+            "type": "string"
+          }
+        },
+        "experiment_description": {
+          "type": "string",
+          "description": "general description of the experiment"
```
CodyCBakerPhD, line 24: It's relatively minor, but...

Stylistically there are divergences in the descriptions for these fields - some are just a bunch of lower case text with no punctuation, some are full proper sentences.

I'd recommend we standardize these to use full proper sentence structure; a part of the eventual plan for the GUIDE could include having a little 'info' hover button or something next to each field which describes what it means, and I'm fairly sure the JSON field descriptions are the way that info would be communicated

src/neuroconv/schemas/base_metadata_schema.json
```diff
@@ -0,0 +1,171 @@
...
+          "description": "A unique text identifier for the file. If one is not provided it will be randomly assigned"
+        },
+        "institution": {
+          "type": "string",
+          "description": "Institution(s) where experiment is performed"
+        },
+        "lab": {
+          "type": "string",
+          "description": "Lab where experiment was performed"
+        },
```
CodyCBakerPhD, line 49: @garrettmflynn Though this PR is mostly a NeuroConv thing, this part can factor into the GUIDE as well

In particular, I recall that the institution and lab are things that we extract early on in a separate manner - nonetheless, this information should then copied over to each metadata form for each session propagating from that data set; mostly because, like the publication DOI, these are not things that change over sessions or even experiments

CodyCBakerPhD:
Attn: @garrettmflynn 

This is a PR I've been alluding to us needing for a little while - this will influence the way the forms show up by default before any interaction/modification on behalf of the GUIDE

src/neuroconv/schemas/base_metadata_schema.json
```diff
@@ -0,0 +1,171 @@
...
+            "type": "string"
+          }
+        },
+        "slices": {
+          "type": "string",
+          "description": "Description of slices, including information about preparation thickness, orientation, temperature and bath solution."
+        },
+        "source_script": {
+          "type": "string",
+          "description": "Script file used to create this NWB file."
```
CodyCBakerPhD, line 84: Does JSON allow an 'examples' sub-field? It could be really useful info to display via the GUIDE

Like, is this field supposed to be the relative path to a file? And how would it differ from the `source_script_file_name`?

src/neuroconv/schemas/base_metadata_schema.json
```diff
@@ -0,0 +1,171 @@
...
+      "properties": {
+        "age": {
+          "type": "string",
+          "description": "The age of the subject. The ISO 8601 Duration format is recommended, e.g., ‘P90D‘ for 90 days old, ‘P2W‘ for 2 weeks old, ‘P18Y‘ for 18 years old.",
+          "pattern": "^P((\\d+Y)?(\\d+M)?(\\d+D)?(T(\\d+H)?(\\d+M)?(\\d+S)?))|(\\d+W)?$"
+        },
+        "age__reference": {
+          "type": "string",
+          "description": "Age is with reference to this event. Can be ‘birth’ or ‘gestational’.",
+          "default": "birth"
```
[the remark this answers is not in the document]
CodyCBakerPhD: Ah yes, I see it below. `enum`.

src/neuroconv/schemas/base_metadata_schema.json
```diff
@@ -0,0 +1,171 @@
...
+      ],
+      "additionalProperties": false,
+      "properties": {
+        "age": {
+          "type": "string",
+          "description": "The age of the subject. The ISO 8601 Duration format is recommended, e.g., ‘P90D‘ for 90 days old, ‘P2W‘ for 2 weeks old, ‘P18Y‘ for 18 years old.",
+          "pattern": "^P((\\d+Y)?(\\d+M)?(\\d+D)?(T(\\d+H)?(\\d+M)?(\\d+S)?))|(\\d+W)?$"
+        },
+        "age__reference": {
+          "type": "string",
```
CodyCBakerPhD, line 123:
```suggestion
          "type": "string",
          "enum": ["birth", "gestational"],
```

src/neuroconv/schemas/base_metadata_schema.json
```diff
@@ -0,0 +1,171 @@
...
+            "F",
+            "U",
+            "O"
+          ],
+          "default": "U"
+        },
+        "species": {
+          "type": "string",
+          "description": "Species of subject. Use latin name.",
+          "pattern": "^[A-Z][a-z]+ [a-z]+"
```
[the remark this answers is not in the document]
garrettmflynn: Right now inputs are not. Though it'd be a trivial addition to the form generator to include a validation function that flags invalid inputs.

src/neuroconv/schemas/base_metadata_schema.json
bendichter:
```suggestion
          "description": "General description of the experiment."
```

 Here is the description for the NWB Schema. I think you are right that they should be standardized and hold meaningful details, but I think we should ideally do it at that level.

src/neuroconv/schemas/base_metadata_schema.json
bendichter: Yes. they should be filled as default values which the user can then change session-by-session if they want

src/neuroconv/schemas/base_metadata_schema.json
bendichter: Yeah, it's not really clear from the schema

src/neuroconv/schemas/base_metadata_schema.json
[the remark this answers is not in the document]
bendichter: @garrettmflynn that would be great. Make sure if you build a regex checker that if allows an empty string of the field is optional.

CodyCBakerPhD (approved):

PR #377, opened 2023-03-17
extracting and reshaping metadata from dynamic paths

bendichter:
Add `expand_paths`, which finds paths that match a format spec and extracts metadata


src/neuroconv/utils/path_expansion.py
```diff
@@ -0,0 +1,104 @@
...
+
+def _unddict(d):
+    """Turn a ddict into a normal dictionary"""
+    return {key: _unddict(value) for key, value in d.items()} if isinstance(d, defaultdict) else d
+
+
+class AbstractPathExpander(abc.ABC):
+    @abc.abstractmethod
+    def __init__(self, *args, **kwargs):
+        pass
```
bendichter, line 37:
```suggestion
```

src/neuroconv/utils/path_expansion.py
```diff
@@ -0,0 +1,102 @@
...
+                        if "session_id" in metadata:
+                            out[key]["metadata"]["NWBFile"]["session_id"] = metadata["session_id"]
+                        if "subject_id" in metadata:
+                            out[key]["metadata"]["Subject"]["subject_id"] = metadata["subject_id"]
+        return list(_unddict(out).values())
+
+
+class LocalPathExpander(AbstractPathExpander):
+    def __init__(self):
+        pass
```
bendichter, line 97:
```suggestion
```

src/neuroconv/tools/google_drive.py
CodyCBakerPhD, line 1: Strongly suggest removing this Google Drive implementation (or at least separating it to a different PR) so that this PR is only about iterating local paths

src/neuroconv/utils/path_expansion.py
```diff
@@ -0,0 +1,100 @@
...
+
+def _unddict(d):
+    """Turn a ddict into a normal dictionary"""
+    return {key: _unddict(value) for key, value in d.items()} if isinstance(d, defaultdict) else d
+
+
+class AbstractPathExpander(abc.ABC):
+    def extract_metadata(self, folder, format_: str):
+        for filepath in self.list_directory(folder):
+            result = parse(format_, filepath)
```
CodyCBakerPhD, line 37: After playing around with this both in Python and GUIDE, I think it would be a better user experience if we constrain our f-string language to always use "/" to indicate file path separators independent of the OS; this is easy to correct in the format passed into `parse` like so

```suggestion
class AbstractPathExpander(abc.ABC):
    def extract_metadata(self, folder, format_: str):
        format_ = format_.replace("/", os.sep)  # our f-string uses '/' to communicate os-independent separators
        for filepath in self.list_directory(folder):
            result = parse(format_, filepath)
```

src/neuroconv/utils/path_expansion.py
```diff
@@ -0,0 +1,100 @@
...
+            if result:
+                yield filepath, result.named
+
+    @abc.abstractmethod
+    def list_directory(self, folder):
```
CodyCBakerPhD, line 42: Recommend rich annotation typing for abstract methods to communicate the input-output expected of child classes (which they can of course override as needed, but they should also annotate that respectively)

```suggestion
    def list_directory(self, folder: Union[FilePathType, FolderPathType]) -> Tuple[str]:
```

tests/test_minimal/test_utils/test_expand_paths.py
```diff
@@ -0,0 +1,75 @@
...
+        for session_id in ("101", "102"):
+            Path.mkdir(base / f"sub-{subject_id}" / f"session_{session_id}")
+            (base / f"sub-{subject_id}" / f"session_{session_id}" / "abc").touch()
+            (base / f"sub-{subject_id}" / f"session_{session_id}" / "xyz").touch()
+
+    # run path parsing
+    out = expander.expand_paths(
+        dict(
+            aa=dict(folder=base, file_path=os.path.join("sub-{subject_id:3}", "session_{session_id:3}", "abc")),
+            bb=dict(folder=base, file_path=os.path.join("sub-{subject_id:3}", "session_{session_id:3}", "xyz")),
```
CodyCBakerPhD, line 38: Note that if https://github.com/catalystneuro/neuroconv/pull/377/files#r1165963109 is accepted, this can simply become

```suggestion
            aa=dict(folder=base, file_path="sub-{subject_id:3}/session_{session_id:3}/abc"),
            bb=dict(folder=base, file_path="sub-{subject_id:3}/session_{session_id:3}/xyz"),
```

which is a more readable explicit test case

src/neuroconv/utils/path_expansion.py
```diff
@@ -0,0 +1,100 @@
...
+    >>> a = dict(b=dict(c=dict(d=5)))
+
+    becomes
+    >>> a = _ddict()["b"]["c"]["d"] = 5
+
+    It becomes particularly useful when modifying an existing hierarchical dictionary,
+    because the next level is only created if it does not already exist.
+
+    """
+    return defaultdict(_ddict)
```
CodyCBakerPhD, line 26: Would you prefer to get #404 in before this so you can use the `DeepDict` class here instead? I'd be in favor of that

CodyCBakerPhD:
@bendichter Mind if I move this these classes to `tools` instead of `utils`?

They used to rely purely on regex (so included in Python standard) but the dependence on `parse` here means it relies on an external library

src/neuroconv/tools/path_expansion.py
```diff
@@ -0,0 +1,77 @@
...
+    def extract_metadata(self, folder, format_: str):
+        format_ = format_.replace("\\", os.sep)  # Actual character is a single back-slash; first is an escape for that
+        format_ = format_.replace("/", os.sep)  # our f-string uses '/' to communicate os-independent separators
+        for filepath in self.list_directory(folder):
+            result = parse(format_, filepath)
+            if result:
+                yield filepath, result.named
+
+    @abc.abstractmethod
+    def list_directory(self, folder: Union[FilePath, DirectoryPath]) -> Iterable[str]:
```
CodyCBakerPhD, line 22: @bendichter To make this argument more specific (as well as making the surrounding code more explicit about how it treats that variable) I was thinking about renaming it

1) `root`
2) `base`
3) `root_folder`
4) `base_folder`

Any preference?

CodyCBakerPhD, line 22: Also, should this be only a `DirectoryPath`? Iteration fails silently (no return values) if you specify a `FilePath`, perhaps an error would be clearer?

CodyCBakerPhD: Or even

5. `root_directory` 
6. `base_directory`

to be consistent with the method name

src/neuroconv/tools/path_expansion.py
```diff
@@ -0,0 +1,82 @@
...
```
bendichter, line 22:
```suggestion
    def list_directory(self, folder: DirectoryPath) -> Iterable[str]:
```

src/neuroconv/tools/path_expansion.py
```diff
@@ -0,0 +1,82 @@
...
+                        if "session_id" in metadata:
+                            out[key]["metadata"]["NWBFile"]["session_id"] = metadata["session_id"]
+                        if "subject_id" in metadata:
+                            out[key]["metadata"]["Subject"]["subject_id"] = metadata["subject_id"]
+        return list(dict(out).values())
+
+
+class LocalPathExpander(AbstractPathExpander):
+    def list_directory(self, folder: Union[FilePath, DirectoryPath]) -> Iterable[str]:
+        return (str(path.relative_to(folder)) for path in Path(folder).rglob("*"))
```
bendichter, line 82:
```suggestion
    def list_directory(self, base_directory: DirectoryPath) -> Iterable[FilePath]:
        return (str(path.relative_to(base_directory)) for path in Path(folder).rglob("*"))
```

src/neuroconv/tools/path_expansion.py
```diff
@@ -0,0 +1,82 @@
...
+    def extract_metadata(self, folder, format_: str):
+        format_ = format_.replace("\\", os.sep)  # Actual character is a single back-slash; first is an escape for that
+        format_ = format_.replace("/", os.sep)  # our f-string uses '/' to communicate os-independent separators
+        for filepath in self.list_directory(folder):
+            result = parse(format_, filepath)
+            if result:
+                yield filepath, result.named
+
+    @abc.abstractmethod
+    def list_directory(self, folder: DirectoryPath) -> Iterable[FilePath]:
```
bendichter, line 22:
```suggestion
    def list_directory(self, base_directory: DirectoryPath) -> Iterable[FilePath]:
```

src/neuroconv/tools/path_expansion.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class AbstractPathExpander(abc.ABC):
+    def extract_metadata(self, folder, format_: str):
+        format_ = format_.replace("\\", os.sep)  # Actual character is a single back-slash; first is an escape for that
+        format_ = format_.replace("/", os.sep)  # our f-string uses '/' to communicate os-independent separators
+        for filepath in self.list_directory(folder):
+            result = parse(format_, filepath)
+            if result:
+                yield filepath, result.named
```
bendichter, line 19:
```suggestion
    def extract_metadata(self, base_directory: DirectoryPath, format_: str):
        format_ = format_.replace("\\", os.sep)  # Actual character is a single back-slash; first is an escape for that
        format_ = format_.replace("/", os.sep)  # our f-string uses '/' to communicate os-independent separators
        for filepath in self.list_directory(base_directory):
            result = parse(format_, filepath)
            if result:
                yield filepath, result.named
```

src/neuroconv/tools/path_expansion.py
```diff
@@ -0,0 +1,82 @@
...
+            The base folder whose contents will be iterated recursively.
+
+        Yields
+        -------
+        sub_paths : iterable of strings
+            Generator that yields all sub-paths of file and folders from the common root `folder`.
+        """
+        pass
+
+    def expand_paths(self, source_data_spec: dict) -> List[DeepDict]:
```
bendichter, line 38:
```suggestion
    def expand_paths(self, source_data_spec: Dict[str,dict]) -> List[DeepDict]:
```

src/neuroconv/tools/path_expansion.py
CodyCBakerPhD: #6 was chosen

CodyCBakerPhD: Annotation updated - error can be added later if usage becomes a problem

CodyCBakerPhD (dismissed):

PR #383, opened 2023-03-27
[Cloud Deployment I]: Main dockerfile

CodyCBakerPhD:
# TODO

- [x] add a test of the docker invocation ~~that ensures the testing suite runs completely using only the docker container~~
- [x] contemplate a GitHub Action for automatically building the containers over a matrix of architectures, python versions, and even release versions


dockerfiles/neuroconv_dockerfile
CodyCBakerPhD, line 1: We can comment on files now BTW

If you noticed this is under an outer folder, I intend to add several other docker files that integrate our base image with data transfer services like AWS and rclone

These have been working well in early prototyping, and is much easier than trying to get their Python SDKs installed and setup in the same environment as the rest of NeuroConv, which also helps us keep things modular

dockerfiles/rclone_auto_config_dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM rclone/rclone:latest
+CMD printf "[$DRIVE_NAME]\ntype = drive\nscope = drive\ntoken = {\"access_token\":\"$TOKEN\",\"token_type\":\"Bearer\",\"refresh_token\":\"$REFRESH_TOKEN\",\"expiry\":\"$EXPIRY\"}" > /mnt/data/rclone.conf \
```
CodyCBakerPhD, line 2: Another possibility might be to `echo` the direct file stream as in the neuroconv auto config image

But there was one or two minor things preventing it from working right away

.github/workflows/auto-release-docker.yml
```diff
@@ -0,0 +1,32 @@
...
+          registry: ghcr.io
+          username: ${{ github.MY_GITHUB_USERNAME }}
+          password: ${{ secrets.MY_GITHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          push: true  # Push is a shorthand for --output=type=registry
+          tags:
+            - ghcr.io/CatalystNeuro/NeuroConv:latest
+            - ghcr.io/CatalystNeuro/NeuroConv:0.3.1  # TODO, figure out how to auto-set this
```
luiztauffer, line 30: @CodyCBakerPhD something like this could work I think
```
      - name: Extract version
        id: extract-version
        run: |
          version=$(sed -n 's/version = "\(.*\)"/\1/p' ./src/neuroconv/version.py)
          echo "packageversion=${version}" >> "$GITHUB_ENV"
      - name: Build and push
        uses: docker/build-push-action@v3
        with:
          push: true
          tags:
            - ghcr.io/CatalystNeuro/NeuroConv:latest
            - ghcr.io/CatalystNeuro/NeuroConv:${{ env.packageversion }}
```

luiztauffer: of course, we would have to create a `version.py` file

CodyCBakerPhD: Neat idea - could probably also leverage a one-liner Python `-c` call like

```
version=$( python -c "from nwbinspector import get_package_version; print(get_package_version(package_name="neuroconv"))" )
echo "packageversion=${version}" >> "$GITHUB_ENV"
```

dockerfiles/neuroconv_developer_build_auto_config
```diff
@@ -0,0 +1,8 @@
+FROM python:3.10
```
luiztauffer, line 1: @CodyCBakerPhD 
usually changing to a -slim tag (e.g. `python:3.10.12-slim`) can substantially cut the final image size
slim images do come with some limitations in libraries, but these tend to be unnecessary for production images, I think it's worth a try

docs/developer_guide/docker_images.rst
```diff
@@ -0,0 +1,94 @@
...
+
+.. code:
+
+    docker pull ghcr.io/catalystneuro/neuroconv:latest
+
+and can then run the entrypoint (equivalent to the usual CLI usage) on a YAML specification file (named `your_specification_file.yml`) with
+
+.. code:
+
+    docker run -it --volume /your/local/drive/:/desired/alias/of/drive/ ghcr.io/catalystneuro/neuroconv:latest neuroconv /desired/alias/of/drive/your_specification_file.yml
```
bendichter, line 64: Thanks @CodyCBakerPhD, and sorry this has taken be so long to review. This looks great, but as you note in the comment, I think this needs testing. Could you at least create or point to a YAML file that can be used as input for this docker run command?

.github/workflows/auto-publish-docker-dev-branch.yml
CodyCBakerPhD, line 1: Oh dang, I totally forgot I already added this workflow

Needed to update the credentials but appears to be working: https://github.com/catalystneuro/neuroconv/pkgs/container/neuroconv/182375280?tag=dev

.github/workflows/auto-publish-docker-dev-branch.yml
```diff
@@ -0,0 +1,36 @@
...
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ secrets.DOCKER_UPLOADER_USERNAME }}
+          password: ${{ secrets.DOCKER_UPLOADER_PASSWORD }}
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          push: true  # Push is a shorthand for --output=type=registry
+          tags: ghcr.io/catalystneuro/neuroconv:dev
```
CodyCBakerPhD, line 34:
```suggestion
          tags: ghcr.io/catalystneuro/neuroconv:main
```

One question is how we want the current `main` branch to be tagged - should this be `dev`, `main`, or something else?

@weiglszonja @bendichter @luiztauffer

.github/workflows/auto-publish-docker-dev-branch.yml
```diff
@@ -0,0 +1,36 @@
+name: Upload Docker Image of Latest Developer Branch to GHCR
+
+on:
+  schedule:
+    - cron: "0 16 * * *"  # Daily at noon EST
+  workflow_dispatch:
+  push:
```
CodyCBakerPhD, line 7:
```suggestion
```
This was just to get the ball rolling (can't do manual triggers via `workflow_dispatch` until it runs at least once)

.github/workflows/auto-publish-docker-dev-branch.yml
CodyCBakerPhD: `latest` is another standard keyword

.github/workflows/auto-release-docker.yml
CodyCBakerPhD, line 1: This one might take some follow-up debugging to see if it tags right

dockerfiles/neuroconv_developer_build_auto_config
CodyCBakerPhD, line 1:
```suggestion
FROM python:3.11.7-slim
```

dockerfiles/neuroconv_developer_build_dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM python:3.10
```
CodyCBakerPhD, line 1:
```suggestion
FROM python:3.11.7-slim
```

dockerfiles/neuroconv_latest_release_dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM python:3.10
```
CodyCBakerPhD, line 1:
```suggestion
FROM python:3.11.7-slim
```

.github/workflows/auto-publish-docker-dev-branch.yml
luiztauffer: @CodyCBakerPhD it might be useful to have one at least two: one for dev and one for main?
```
tags: ghcr.io/catalystneuro/neuroconv:${{ github.ref_name }}
```

although I agree that, when built from main, the standard tag should be `latest`

dockerfiles/neuroconv_latest_release_dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM python:3.11.7-slim
+LABEL org.opencontainers.image.source=https://github.com/catalystneuro/neuroconv
```
weiglszonja, line 2: @CodyCBakerPhD I'm finally checking out the docker files, sorry for the delay!

I run 
```
docker build -f neuroconv_latest_release_dockerfile -t neuroconv_latest_release . 
```
but had an error for `psutil`
```python-traceback
53.10 Failed to build psutil numcodecs
53.10 ERROR: Could not build wheels for psutil, numcodecs, which is required to install pyproject.toml-based projects
```
but [this](https://github.com/giampaolo/psutil/issues/2192#issuecomment-1396079848) solved it for me:

```suggestion
LABEL org.opencontainers.image.source=https://github.com/catalystneuro/neuroconv
RUN apt update && apt install musl-dev python3-dev -y
```

dockerfiles/neuroconv_latest_release_dockerfile
```diff
@@ -0,0 +1,4 @@
...
+RUN pip install neuroconv[full]
+CMD ["python3 -m"]
```
weiglszonja, line 4: I also had an error after running
```
docker run -it neuroconv_latest_release
```
```python-traceback
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "python3 -m": executable file not found in $PATH: unknown.
ERRO[0000] error waiting for container: context canceled
```
So I changed this to:
```suggestion
CMD ["python3",  "-c", "import neuroconv"]
```

.github/workflows/auto-release-docker.yml
CodyCBakerPhD: This was resolved in a way reliant on the GitHub tags for each release

dockerfiles/neuroconv_developer_build_auto_config
CodyCBakerPhD: Swapped all to slim

.github/workflows/auto-publish-docker-dev-branch.yml
CodyCBakerPhD: Debugged now

.github/workflows/auto-publish-docker-dev-branch.yml
CodyCBakerPhD: latest releases should have two tags now: `latest` and the exact version

New releases should swap the `latest` tag to the newer one automatically but the exact version should still stick around forever

Current `main` branch is built daily or on manual dispatch with tag `dev`

dockerfiles/neuroconv_latest_release_dockerfile
CodyCBakerPhD: The point of a CMD is to define the prepend to the entrypoint - since we want this docker image, when run as a container, to effectively emulate an entire functioning Python environment, I'm trying to design it in such a way that it is pretty much just like calling the CLI from a normal Python environment

That is, in a typical Python NeuroConv environment you would do

```
neuroconv my_yaml_file.yml
```

now with the docker we want

```
docker run <other docker required stuff> neuroconv my_yaml_file.yml
```

While it is conceptually possible to just rely on implicit entrypoint definition from successful installation and association inside the docker container, it is much safer apparently to call `python -m` directly on the module

> CMD ["python3",  "-c", "import neuroconv"]

This simply verifies that the module is importable, it does not route the CLI - I don't even think it would facilitate pretending of a command sent to the container either

Anyway, I think the problem was `python` vs. `python3` usage. This seems to be working now, and I added a full demo if you want to check that out and see if that works for you

dockerfiles/neuroconv_latest_release_dockerfile
CodyCBakerPhD: I can confirm this issue on my M1 but not on anything else

Thanks for the fix, we'll see if it harms the linux or other platform builds at all

dockerfiles/neuroconv_latest_release_dockerfile
CodyCBakerPhD: Yep, seems fine

CodyCBakerPhD:
OK this is finally ready for full review @weiglszonja 

- Running the docker-based CLI tests on push to every PR is overkill since nothing about the invocation or code inside the container would be updated. Set to daily CRON instead. Evidence of docker tests currently passing: https://github.com/catalystneuro/neuroconv/actions/runs/8395655155/job/22995388956
- All current package tags can be seen from https://github.com/catalystneuro/neuroconv/pkgs/container/neuroconv/versions

weiglszonja (approved):
Thanks @CodyCBakerPhD, looks good to me!

PR #384, opened 2023-03-27
[Cloud Deployment IIIb]: AWS batch deployment helper

CodyCBakerPhD:
Example prototype for easy deployment of a docker image on AWS Batch


CodyCBakerPhD:
(various changes particular to your compute environments will probably be necessary)

This also uses EFS instead of EBS, which is more expensive per allocation but allows completely dynamic sizing in that way

src/neuroconv/tools/data_transfers.py
```diff
@@ -349,3 +351,154 @@ def automatic_dandi_upload(
             rmtree(path=nwb_folder_path)
         except PermissionError:  # pragma: no cover
             warn("Unable to clean up source files and dandiset! Please manually delete them.")
+
+
+def submit_aws_batch_job(
+    job_name: str,
+    docker_container: str,
```
bendichter, line 358:
```suggestion
    docker_container: str = ghcr.io/catalystneuro/rclone_auto_config:latest,
```

tests/test_minimal/test_tools/s3_tools.py
```diff
@@ -46,3 +47,37 @@ def test_estimate_total_conversion_runtime():
...
+def test_submit_aws_batch_job():
+    job_name = "test_submit_aws_batch_job"
+    docker_image = "ubuntu:latest"
+    command = "echo 'Testing NeuroConv AWS Batch submission."
+
+    submit_aws_batch_job(
+        job_name=job_name,
+        docker_image=docker_image,
+        command=command,
+    )
```
CodyCBakerPhD, line 61: What this is currently testing is simply that the calls via `boto3` (including the final submission) do not fail (which would have raised errors)

They do not guarantee the job itself succeeded, since that is separate from the submission

In a follow-up, I plan to leverage the helper function for table status updates to actually get these minimal tests to flag their completion at runtime, and then add an assertion (likely with some delay or repeat w/ timeout) that the status table was updated to 'completed' by the end of it all

CodyCBakerPhD:
Evidence of AWS tests passing in CI: https://github.com/catalystneuro/neuroconv/actions/runs/10044953517/job/27761154548?pr=384

I'm disabling this hereafter from running on push to pull requests and limiting the matrix to reduce needless AWS costs

Otherwise, we only need to check it about once a week to ensure the `boto3` API doesn't deprecate structures we utilize here

tests/test_minimal/test_tools/s3_tools.py
h-mayorquin: OK.

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,283 @@ def estimate_s3_conversion_cost(
     cost_gb_m = 0.08 / 1e3  # $0.08 / GB Month
     cost_mb_s = cost_gb_m / (1e3 * 2.628e6)  # assuming 30 day month; unsure how amazon weights shorter months?
     return cost_mb_s * total_mb_s
+
+
+def submit_aws_batch_job(
+    *,
+    job_name: str,
```
h-mayorquin, line 43: Can we make this optional and generate one automatically?

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,283 @@ def estimate_s3_conversion_cost(
...
+    docker_image: str,
+    command: Optional[str] = None,
+    job_dependencies: Optional[List[Dict[str, str]]] = None,
+    region: str = "us-east-2",
```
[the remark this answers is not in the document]
CodyCBakerPhD: It's more like it changes so much over time I wouldn't want to manually maintain a list

I could do something similar to what I do [here](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py#L36) though where I can try to find some `boto3` API call to fetch currently available regions

src/neuroconv/tools/data_transfers/_aws.py
CodyCBakerPhD: An internal `jobId` is automatically generated and is part of the info dict returned by this function; that is the internal one used by AWS operations

The `job_name` is a user/human readable tag related to the intent/usage/reference of that ID, so I think the user should specify it how they want

Otherwise the most we would do I think is utilize the `docker_image` as the base name and add a truncated UUID on the end to make it unique; but users would HAVE to grab the `info` dict returned to know how to find the name, which could be hard if they have lots of jobs running at once

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,283 @@ def estimate_s3_conversion_cost(
...
+    Parameters
+    ----------
+    job_name : str
+        The name of the job to submit.
+    docker_image : str
+        The name of the Docker image to use for the job.
+    command : str, optional
+        The command to run in the Docker container.
+        Current syntax only supports a single line; consecutive actions should be chained with the '&&' operator.
+    job_dependencies : list of dict
```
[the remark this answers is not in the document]
CodyCBakerPhD: The only thing I'm cautious of here is offering too much information about the CURRENT `boto3` API given how much it can change even over the course of a year

My approach here is just to expose the ability to pass the arguments; perhaps we can simply link out to `boto3` documentation and tell users to reference that for exact syntax?

Most of the specifics for our own usage though will be handled in the following PR(s)

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,283 @@ def estimate_s3_conversion_cost(
...
+    docker_image: str,
+    command: Optional[str] = None,
+    job_dependencies: Optional[List[Dict[str, str]]] = None,
+    region: str = "us-east-2",
+    status_tracker_table_name: str = "neuroconv_batch_status_tracker",
+    iam_role_name: str = "neuroconv_batch_role",
+    compute_environment_name: str = "neuroconv_batch_environment",
+    job_queue_name: str = "neuroconv_batch_queue",
+    job_definition_name: Optional[str] = None,
+    minimum_worker_ram_in_gb: float = 4.0,
```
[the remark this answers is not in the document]
CodyCBakerPhD: It's for our own API consistency, since all the iterator and backend configuration are in bytes

The reason for the RAM specification here is also in reference to iterative read/write usage through the iterators, so probably best that they both be specified in the same basis

We can go over this more in the meeting

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,283 @@ def estimate_s3_conversion_cost(
...
+        id=submission_id, job_name=job_name, submitted_on=datetime.now().isoformat(), status="submitted"
+    )
+    table.put_item(Item=table_submission_info)
+
+    info = dict(job_submission_info=job_submission_info, table_submission_info=table_submission_info)
+    return info
+
+
+def update_table_status(
+    *, submission_id: str, status: str, status_tracker_table_name: str = "neuroconv_batch_status_tracker"
```
CodyCBakerPhD, line 287:
```suggestion
    *, submission_id: str, status: str, status_tracker_table_name: str = "neuroconv_batch_status_tracker", region: str = "us-east-2",
```

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,283 @@ def estimate_s3_conversion_cost(
...
+    Intended for use by the running job to indicate its completion.
+
+    Parameters
+    ----------
+    submission_id : str
+        The random hash that was assigned on submission of this job to the status tracker table.
+    status : str
+        The new status value to update.
+    status_tracker_table_name : str, default: "neuroconv_batch_status_tracker"
+        The name of the DynamoDB table to use for tracking job status.
```
CodyCBakerPhD, line 301:
```suggestion
    status_tracker_table_name : str, default: "neuroconv_batch_status_tracker"
        The name of the DynamoDB table to use for tracking job status.
    region : str, default: "us-east-2"
        The AWS region to use for the job.
```

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,283 @@ def estimate_s3_conversion_cost(
...
+    """
+    import boto3
+
+    aws_access_key_id = os.environ.get("AWS_ACCESS_KEY_ID")
+    aws_secret_access_key = os.environ.get("AWS_SECRET_ACCESS_KEY")
+
+    dynamodb_resource = boto3.resource(
+        service_name="dynamodb",
+        region_name=region,
```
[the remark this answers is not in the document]
CodyCBakerPhD: Good catch

This helper isn't actually tested or used until the follow-up, so I could have also removed from here altogether

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,285 @@ def estimate_s3_conversion_cost(
...
+    command : str, optional
+        The command to run in the Docker container.
+        Current syntax only supports a single line; consecutive actions should be chained with the '&&' operator.
+    job_dependencies : list of dict
+        A list of job dependencies for this job to trigger. Structured as follows:
+        [
+            {"jobId": "job_id_1", "type": "N_TO_N"},
+            {"jobId": "job_id_2", "type": "SEQUENTIAL"},
+            ...
+        ]
```
CodyCBakerPhD, line 75:
```suggestion
        ]
        
        Refer to the boto3 API documentation for latest syntax:
        https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/batch/client/submit_job.html
```

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,289 @@ def estimate_s3_conversion_cost(
...
+    docker_image: str,
+    command: Optional[str] = None,
+    job_dependencies: Optional[List[Dict[str, str]]] = None,
+    region: str = "us-east-2",
+    status_tracker_table_name: str = "neuroconv_batch_status_tracker",
+    iam_role_name: str = "neuroconv_batch_role",
+    compute_environment_name: str = "neuroconv_batch_environment",
+    job_queue_name: str = "neuroconv_batch_queue",
+    job_definition_name: Optional[str] = None,
+    minimum_worker_ram_in_gb: float = 4.0,
```
CodyCBakerPhD, line 53:
```suggestion
    minimum_worker_ram_in_gib: int = 4,
```

Maybe you're right though, since it does determine all kinds of allocation stuff in the cloud

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,289 @@ def estimate_s3_conversion_cost(
...
+    # Ensure job definition exists
+    # By default, keep name unique by incorporating the name of the container
+    job_definition_docker_name = docker_image.replace(":", "_")
+    job_definition_name = job_definition_name or f"neuroconv_batch_{job_definition_docker_name}"
+    current_job_definitions = [
+        definition["jobDefinitionName"] for definition in batch_client.describe_job_definitions()["jobDefinitions"]
+    ]
+    resource_requirements = [
+        {
+            "value": str(int(minimum_worker_ram_in_gb * 1e3 / 1.024**2)),
```
CodyCBakerPhD, line 220:
```suggestion
            "value": str(int(minimum_worker_ram_in_gib * 1024)),
```

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,289 @@ def estimate_s3_conversion_cost(
...
+    job_queue_name : str, default: "neuroconv_batch_queue"
+        The name of the job queue to use for the job.
+    job_definition_name : str, optional
+        The name of the job definition to use for the job.
+        Defaults to f"neuroconv_batch_{ name of docker image }",
+        but replaces any colons from tags in the docker image name with underscores.
+    minimum_worker_ram_in_gb : int, default: 4.0
+        The minimum amount of base worker memory required to run this job.
+        Determines the EC2 instance type selected by the automatic 'best fit' selector.
+        Recommended to be several GB to allow comfortable buffer space for data chunk iterators.
```
CodyCBakerPhD, line 93:
```suggestion
    minimum_worker_ram_in_gib : int, default: 4
        The minimum amount of base worker memory required to run this job.
        Determines the EC2 instance type selected by the automatic 'best fit' selector.
        Recommended to be several GiB to allow comfortable buffer space for data chunk iterators.
```

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,289 @@ def estimate_s3_conversion_cost(
...
+    current_compute_environments = [
+        environment["computeEnvironmentName"]
+        for environment in batch_client.describe_compute_environments()["computeEnvironments"]
+    ]
+    if compute_environment_name not in current_compute_environments:
+        batch_client.create_compute_environment(
+            computeEnvironmentName=compute_environment_name,
+            type="MANAGED",
+            state="ENABLED",
+            computeResources={
```
CodyCBakerPhD, line 187: TODO: enable spot and expose % as keyword argument

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,289 @@ def estimate_s3_conversion_cost(
...
+                "instanceRole": "ecsInstanceRole",
+                "securityGroupIds": ["sg-851667c7"],
+                "instanceTypes": ["optimal"],
+            },
+        )
+
+    # Ensure job queue exists
+    current_job_queues = [queue["jobQueueName"] for queue in batch_client.describe_job_queues()["jobQueues"]]
+    if job_queue_name not in current_job_queues:
+        batch_client.create_job_queue(
```
CodyCBakerPhD, line 202: TODO: expose argument for job state timeouts (when in 'runnable' not 'running' stalls out due to incompatibility of resource specification combined with spot %

src/neuroconv/tools/data_transfers/_aws.py
CodyCBakerPhD: default should be `100%` for minimum delay, but some users (such as us for big non-priority data churning) might want to manually set it lower to save costs

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,289 @@ def estimate_s3_conversion_cost(
...
+        dependsOn=job_dependencies,
+        jobDefinition=job_definition_name,
+        jobName=job_name,
+        containerOverrides=container_overrides,
+    )
+
+    # Update DynamoDB status tracking table
+    submission_id = submission_id or str(uuid4())
+    table_submission_info = dict(
+        id=submission_id, job_name=job_name, submitted_on=datetime.now().isoformat(), status="submitted"
```
CodyCBakerPhD, line 278: TODO: include jobID from job submissio info for good measure

src/neuroconv/tools/data_transfers/_aws.py
```diff
@@ -30,3 +36,289 @@ def estimate_s3_conversion_cost(
...
+        raise ValueError(
+            f"There is already a job named '{job_name}' running in the queue! "
+            "If you are submitting multiple jobs, each will need a unique name."
+        )
+
+    # Set environment variables to the docker container as well as optional commands to run
+    job_dependencies = job_dependencies or []
+    container_overrides = dict(
+        # Set environment variables
+        environment=[
```
CodyCBakerPhD, line 258: TODO: pass AWS credentials as well for status table updating

src/neuroconv/tools/data_transfers/_aws.py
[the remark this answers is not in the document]
h-mayorquin: Yeah, a link would be fine. Or just "refer for "x" on the boto3 documentation to see more" if we don't want to deal with possible dead links.

src/neuroconv/tools/data_transfers/_aws.py
h-mayorquin: Explained on the meeting. Thanks.

src/neuroconv/tools/data_transfers/_aws.py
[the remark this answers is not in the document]
h-mayorquin: Probably not wroth it. I think this is fine and self-explanatory.

src/neuroconv/tools/data_transfers/_aws.py
CodyCBakerPhD: Actually after looking more into it, I recall that we don't intend to immediately support this because the processes can be easily interrupted by anyone else paying for on-demand, and these are such long processes usually that the risk of losing the instance (and thus any money spent to that point) is too high

src/neuroconv/tools/data_transfers/_aws.py
CodyCBakerPhD: Looking further into this, it seems all timeout conditions are for the event that the job exits `RUNNABLE` state and enters `RUNNING`. How long it stays in that state is entirely up to how much data we expect to process and how fast we expect it to be processed

So I'm OK with setting a somewhat large value here just for the sake of safety, but it doesn't solve the problem of jobs not deploying (or the problem of how to track that easily from our tools)

src/neuroconv/tools/data_transfers/__init__.py
```diff
@@ -1,14 +1,6 @@
 """Collection of helper functions for assessing and performing automated data transfers."""
 
-from ._aws import estimate_s3_conversion_cost
 from ._dandi import automatic_dandi_upload
 from ._globus import get_globus_dataset_content_sizes, transfer_globus_content
-from ._helpers import estimate_total_conversion_runtime
```
CodyCBakerPhD, line 6: TBH these functions were probably never used by anyone so OK with removing them. Haven't even reviewed if they are still correct or not

CodyCBakerPhD:
I broke up some of the more repetitive bits blocks of code into private helpers and added docstrings to enhance long-term readability and maintainability

Also moved the new functions into a distinct submodule since this isn't only about 'data transfers' anymore

CodyCBakerPhD:
@h-mayorquin OK did a thorough deep dive with the new enhanced testing structure and fixed some issues from the running image, updated the subnets (these may have to be checked every few months-years) and documented some edge cases I discovered in the relevant docstrings

This should be ready for a second pass now

CodyCBakerPhD:
Evidence of enhanced tests passing: https://github.com/catalystneuro/neuroconv/actions/runs/10242963418/job/28333513602?pr=384

CodyCBakerPhD:
Breakdown of PR size:

+43 lines for a new distinct GitHub Action specific to these tests
+158 lines of tests, including thorough verification of final AWS job status
+640 lines of source code for the helper function and its private components; about 1/3 of this is docstrings recording the nuances of each particular step

src/neuroconv/tools/aws/_dynamodb.py
```diff
@@ -0,0 +1,42 @@
...
+    status : str
+        The new status value to update.
+    status_tracker_table_name : str, default: "neuroconv_batch_status_tracker"
+        The name of the DynamoDB table to use for tracking job status.
+    region : str, default: "us-east-2"
+        The AWS region to use for the job.
+    """
+    import boto3
+
+    aws_access_key_id = os.environ["AWS_ACCESS_KEY_ID"]
```
h-mayorquin, line 29: missing the import os

src/neuroconv/tools/aws/_submit_aws_batch_job.py
```diff
@@ -0,0 +1,640 @@
...
+        return None
+
+    with open(file=default_config_file_path, mode="r") as io:
+        all_lines = io.readlines()
+    lines_with_region = [line for line in all_lines if "region" == line[:7]]
+
+    if len(lines_with_region) != 1:
+        return None
+
+    region = region_lines[0].split("=")[1].strip()
```
h-mayorquin, line 222: `lines_with_region` instead of `region_lines` here?

src/neuroconv/tools/aws/_submit_aws_batch_job.py
```diff
@@ -0,0 +1,640 @@
...
+        time.sleep(10)
+        compute_environment_request = batch_client.describe_compute_environments(
+            computeEnvironments=[compute_environment_name]
+        )
+        compute_environment_response = compute_environment_request["computeEnvironments"]
+        compute_environment_status = (
+            compute_environment_response[0]["status"] if len(compute_environment_response) == 1 else ""
+        )
+
+    if len(job_definition_response) != 1 and compute_environment_status != "VALID":
```
h-mayorquin, line 414: is this undefined?

src/neuroconv/tools/aws/_dynamodb.py
```diff
@@ -0,0 +1,42 @@
+"""Helper functions for operations on DynamoDB tables."""
+
```
CodyCBakerPhD, line 2:
```suggestion
import os
```

src/neuroconv/tools/aws/_submit_aws_batch_job.py
CodyCBakerPhD, line 222:
```suggestion
    region = lines_with_region[0].split("=")[1].strip()
```

src/neuroconv/tools/aws/_submit_aws_batch_job.py
```diff
@@ -0,0 +1,640 @@
...
+        aws_secret_access_key=aws_secret_access_key,
+    )
+
+    # Get the tracking table and IAM role
+    table = _create_or_get_status_tracker_table(
+        status_tracker_table_name=status_tracker_table_name,
+        dynamodb_client=dynamodb_client,
+        dynamodb_resource=dynamodb_resource,
+    )
+    role_info = _create_or_get_iam_role(iam_role_name=iam_role_name, iam_client=iam_client)
```
h-mayorquin, line 134: Let's cal this iam_role_info

tests/test_minimal/test_tools/aws_tools.py
```diff
@@ -0,0 +1,158 @@
...
+    job_dependencies = [{"jobId": job_submission_info_1["jobId"], "type": "SEQUENTIAL"}]
+    job_info_2 = submit_aws_batch_job(
+        job_name=job_name_2,
+        docker_image=docker_image,
+        commands=commands_2,
+        job_dependencies=job_dependencies,
+    )
+
+    # Wait for AWS to process the jobs
+    time.sleep(120)
```
h-mayorquin, line 109: Isn't this brittle? Can we query directly to see if the job is concluded?

h-mayorquin (commented):
Some more comments.

src/neuroconv/tools/aws/_submit_aws_batch_job.py
```diff
@@ -0,0 +1,640 @@
...
+    Also note that registration is strict; once created, a job definition is tagged with ':< revision index >' and
+    any subsequent changes create a new revision index (and technically a new definition as well).
+
+    Parameters
+    ----------
+    job_definition_name : str
+        The name of the job definition to use for the job.
+    docker_image : str
+        The name of the Docker image to use for the job.
+    minimum_worker_ram_in_gib : int
```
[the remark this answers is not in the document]
CodyCBakerPhD: Nope - there's a comment in the code about the serialization. It really should be an int always. I could relax and make it `MiB`, but I really think the conversion from `GiB` is a lot more convenient since you will always want RAM in that schedule

tests/test_minimal/test_tools/aws_tools.py
CodyCBakerPhD: Technically yes

If you want something better we could do the same retry mechanism as seen in source code

But this is such a simple job and has not had trouble so far just waiting a static amount of time

Whichever you prefer

src/neuroconv/tools/aws/_submit_aws_batch_job.py
CodyCBakerPhD: Done

src/neuroconv/tools/aws/_submit_aws_batch_job.py
```diff
@@ -0,0 +1,640 @@
...
+) -> str:  # pragma: no cover
+    """
+    Ensure that the job definition exists in AWS Batch and return the official ARN.
+
+    Automatically generates a job definition name using the docker image, its tags, and worker configuration.
+    The creation date is also appended if the docker image was not tagged or was tagged as 'latest'.
+
+    Note that registering a job definition does not require either a compute environment or a job queue to exist.
+
+    Also note that registration is strict; once created, a job definition is tagged with ':< revision index >' and
```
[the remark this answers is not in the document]
h-mayorquin: TODO: For me.

h-mayorquin (approved):
TODO for me in another PR. Improve the string about the ARN changes.

PR #388, opened 2023-03-29
Update ecephys requirements.txt

bendichter:
I had some issues with the SortingExtractorInterface until I updated spikeinterface


CodyCBakerPhD:
Were the issues anything that we could detect in a CI?

bendichter:
actually I was running 0.96.0 before, so maybe the listed version would have been fine.

I don't think we need to make any changed to the tests if the latest release works as expected

PR #391, opened 2023-03-31
Make session start time required

CodyCBakerPhD:
Looks like this wasn't required on the metadata schema side

### DIFF
--- a/src/neuroconv/schemas/base_metadata_schema.json
+++ b/src/neuroconv/schemas/base_metadata_schema.json
@@ -10,7 +10,7 @@
     "NWBFile": {
       "type": "object",
       "additionalProperties": false,
-      "required": ["session_start_time"]
+      "required": ["session_start_time"],
       "properties": {
         "keywords": {
           "description": "Terms to search over",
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Should we also include things like `identifier` here even though they're most often easy to generate by default?

bendichter (dismissed):

CodyCBakerPhD:
Oh, should we also make `session_id` required? Since that's what's used (in addition to other fallback fields) to name the files on DANDI

src/neuroconv/schemas/base_metadata_schema.json
```diff
@@ -10,6 +10,7 @@
     "NWBFile": {
       "type": "object",
       "additionalProperties": false,
+      "required": ["session_start_time"]
```
bendichter, line 13:
```suggestion
      "required": ["session_start_time"],
```

CodyCBakerPhD:
@bendichter ping

bendichter (approved):

PR #395, opened 2023-04-04
Add stub tests for multisegment extractor

h-mayorquin:
At the moment the stubbing machinery is not working with multi segment extractors:

See:
https://github.com/TheChymera/neuralynx_nwb/issues/9

The problem is that `frame_slice`, the method that we use to stub in spikeinterface, does not support multi segment recording objects. It is not clear to me yet where this should be fixed, most likely at the spikeinterface level but I need to discuss with @alejoe91.

In the meanwhile this PR does two things:
1) It adds a new generator interface for testing ecephys interfaces. Most of the tests that we have work only at the recorder level so this will allow us to test behavior at the interface level. Right now, it probably needs some work but at this stage it works for testing his behavior and we can build from that as needed.
2) Tests for stubing and mult-segment stubing. 

EDIT: OK, I talked with @alejoe91 and he does not want the `frame_slice` to handle multi-segments as he does not believe that the use-case is enough to justify the increase in API complexity. Therefore, I am implementing the logic for stubbing here.


h-mayorquin:
All is pasing. Dev tests are failing for reasons unrelated to this.

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -110,14 +110,20 @@ def subset_recording(self, stub_test: bool = False):
...
-        if self.subset_channels is not None:
-            kwargs.update(channel_ids=self.subset_channels)
-        recording_extractor = self.recording_extractor.frame_slice(start_frame=0, end_frame=end_frame)
+        from spikeinterface.core.segmentutils import ConcatenateSegmentRecording
+
+        max_frames = 100
+
+        recording_extractor = self.recording_extractor
+        number_of_segments = recording_extractor.get_num_segments()
+        recording_segments = [recording_extractor._select_segments([index]) for index in range(number_of_segments)]
```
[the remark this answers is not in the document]
h-mayorquin: Thanks!

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -110,14 +110,20 @@ def subset_recording(self, stub_test: bool = False):
...
+
+        recording_extractor = self.recording_extractor
+        number_of_segments = recording_extractor.get_num_segments()
+        recording_segments = [recording_extractor.select_segments([index]) for index in range(number_of_segments)]
+        end_frame_list = [min(max_frames, segment.get_num_frames()) for segment in recording_segments]
+        recording_segments_stubbed = [
+            segment.frame_slice(start_frame=0, end_frame=end_frame)
+            for segment, end_frame in zip(recording_segments, end_frame_list)
+        ]
+        recording_extractor = ConcatenateSegmentRecording(recording_segments_stubbed)
```
CodyCBakerPhD, line 125: Thanks, this segment looks great!

src/neuroconv/datainterfaces/ecephys/generatorrecordinginterface.py
```diff
@@ -0,0 +1,33 @@
...
+        self,
+        num_channels=4,
+        sampling_frequency=30_000.0,
+        durations=[1.0],
+        seed=0,
+        verbose=True,
+        es_key: str = "ElectricalSeries",
+    ):
+        from spikeinterface.core.generate import generate_recording
+
```
CodyCBakerPhD, line 20: I am rather against having this be an actual DataInterface stored in the main listing - can you move it to the `tools.testing.mock_interfaces` and name it `MockRecordingInterface` instead?

CodyCBakerPhD:
Looks good for the most part - one comment about where to put the testing interface, once that is resolved it should be good to go

Thanks for the tests as well!

src/neuroconv/datainterfaces/ecephys/generatorrecordinginterface.py
h-mayorquin: What do you mean in the main listing?

This would not show in datainterfaces because is not in /src/neuroconv/datainterfaces/__init__.py`, correct?

I put it here because it depends on spikeinterface and I remember that we wanted to segragate interfaces by dependencies. That said, having it in `tools.testing.mock_interfaces` makes sense to me as well if you think that's better.

As for the naming, `MockRecordingInterface` works for me. One important aspect is that after the next version of spikeinterface is out this will be a proper generator. That is, it will have a recorder with as much data as desired but loaded lazily. This can be used to tests chunked writing and memory tests in general. That's why I put the name Generator in the name.

See:
https://github.com/SpikeInterface/spikeinterface/issues/1413

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
h-mayorquin: Thanks! : D

src/neuroconv/datainterfaces/ecephys/generatorrecordinginterface.py
CodyCBakerPhD: By 'listing' I mean anyone perusing the package structure (either via deep imports or on GitHub) of the ecephys interfaces, which are supposed to be format/file type specific

That all sounds great, the parameters of the generation should definetely be exposed to the `__init__` so we can make a `MockRecordingInterface` take on any behavior we wish to test to mimic a real `RecordingInterface`

If you want to name it `MockGenereratorRecordingInterface` or `GeneratorRecordingInterface`, they should be fine - I'd just favor `Mock` starting the title due to precedence of the other recently added cases for things like this

h-mayorquin: Makes sense. Let's go with the simplest option and later when the new version I will point out the property in the docstring.

CodyCBakerPhD (approved):

CodyCBakerPhD:
Amazing, thanks!

PR #398, opened 2023-04-05
move make_or_load_nwbfile logic to BaseDataInterface.

bendichter:
Problem: All `DataInterface.run_conversion` calls are supposed to handle the cases where a filepath is input, or an NWBFile object is input, and handle `overwrite`. This logic is essentially duplicated across all of these methods. This is a problem because: 1) it's duplication of logic that can be centralized 2) it's a lot of extra work when creating a data interface 3) data interfaces sometimes skip this step, assuming an NWBFile input, which breaks the api standardization across the classses.

Solution: move `make_or_load_nwbfile` logic to `BaseDataInterface.run_conversion`, and call `BaseDataInterface._run_conversion` within that context. This allows us to write a `BaseDataInterface._run_conversion` method for new interface classes instead of `BaseDataInterface.run_conversion`. `_run_conversion` is much simpler because it always expects an `NWBFile` object as a required argument and does not need to handle a possible path argument.

For backwards compatibility, DataInterfaces can also continue to define a `run_conversion` method just as before.


src/neuroconv/basedatainterface.py
```diff
@@ -20,12 +24,16 @@ def get_source_schema(cls):
...
+    def __init__(self, verbose: bool = True, **source_data):
         self.source_data = source_data
+        self.verbose = verbose
 
     def get_conversion_options_schema(self):
         """Infer the JSON schema for the conversion options from the method signature (annotation typing)."""
-        return get_schema_from_method_signature(self.run_conversion, exclude=["nwbfile", "metadata"])
+        try:
+            return get_schema_from_method_signature(self._run_conversion, exclude=["nwbfile", "metadata"])
+        except ValueError:
```
CodyCBakerPhD, line 35: I'd change to excepting only the `NotImplementedError`; if there's a separate error in the schema inference (but private method is defined), wouldn't want to fall back to the wrong signature

CodyCBakerPhD:
Basic approach seems fine

I could also see this being a non-private method `.add_to_nwbfile(nwbfile=an_in_memory_nwbfile, ...)` that specifically differs from `.run_conversion(...)` in that an already existing in-memory `NWBFile` is a required field

CodyCBakerPhD:
Though I might suggest waiting until after I finish the last round of temporal alignment stuff and the next release to avoid causing too many merge conflicts - several of the behavior interfaces are undergoing slight refactoring of their `run_conversion` methods to integrate better with the new alignment methods

CodyCBakerPhD:
@bendichter I think #455 essentially replaces this - OK to close this branch?

PR #283, opened 2023-01-23
[Sync V] Integrate Temporal Alignment for remaining non-behavior Interfaces

CodyCBakerPhD:
## Motivation

Follow-up to #237 specific to ophys interfaces

Also starts adding holistic per-interface, per-method unit tests instead of just relying on the demonstrative integration tests added in #237

Next follow-up will be other behavior interfaces

## Checklist

- [X] Have you thoroughly read our [Developer Guide](https://neuroconv.readthedocs.io/en/main/developer_guide.html)?
- [X] Have you ensured the PR description clearly describes the problem and solutions?
- [X] Have you checked to ensure that there aren't other open or previously closed [Pull Requests](https://github.com/catalystneuro/neuroconv/pulls) for the same change?
- [X] If this PR fixes an issue, is the first line of the PR description `fix #XXX` where `XXX` is the issue number?
- [X] Did you update the [CHANGLOG.md](https://github.com/catalystneuro/neuroconv/blob/main/CHANGELOG.md) file on your branch to describe your changes?

PR #400, opened 2023-04-06
[Sync VIa] Icephys Temporal Alignment

CodyCBakerPhD:
This follows after #283, adding start time alignment functionality to icephys (which does not currently support timestamps at all; a separate TODO in a follow-up)


CodyCBakerPhD:
### TODO

- [ ] add tests

CodyCBakerPhD:
Similar to the audio interface, timestamps not yet supported here in general so alignment is done only w.r.t. starting times

Also built off of ordered `file_paths` approach, so designed alignment calls to match the approach of #402  and #403 

Added two basic tests of alignment but further integration of the DataInterfaceMixIn will require a little bit more work in a follow-up

bendichter (approved):

bendichter (approved):

Issue #301, opened 2023-02-07
[Feature]: Reach agreement on temporal alignment method/argument names


CodyCBakerPhD:
### What would you like to see added to NeuroConv?

From discussions following #237, the usage of the word `timestamps` is a bit overloaded.

Specifically, `timestamps` ought to solely refer to the time vector of the associated `TimeSeries` of each interface, but it currently also applies to the time vectors used in `align_by_interpolation` which may be parsed from TTL or other digital signals.

Also, specific to `align_by_interpolation`, a description of what the function does in words could be "A digital signal is sent from a secondary system and received by the primary system to align the timing differences between the two clocks."

The keyword arguments for the function should be chosen to best describe the role of the inputs

Current: `unaligned_timestamps, aligned_timestamps`
Alternative #1: `ttl_sent_times, ttl_recieved times`
Alternative #2: `unaligned_times, aligned_times`
Alternative #3: `sent_times, recieved_times`
Alternative #3: `secondary_times, primary_times`

For alternative #2, keep in mind that the timing information does not need to be from TTL pulses - it could be inferred from an algorithm similar to how MICrONS does it - or it could be parsed from a non-TTL digital signal, such as audio events

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

CodyCBakerPhD:
From meeting, this is fine for now

CodyCBakerPhD:
> you claim that align for setters is a good name because the only way to align timestamps is to actually set them.

Not entirely, and I entirely embrace the core logic of your argument

My reasoning is that _a sequence of code written by the user_ would be read by anyone looking at the code or running it _as an affirmation that the timestamps are now aligned_ after calling these methods (or that the motor in the car is repaired or that the dinner is cooked) - of course, this does not mean that they are [in any sense, perhaps it was an identity mapping `interface.align_timestamps(aligned_timestamps=interface.get_timestamps())`], or that they are set correctly (which is something only the experimenter would actually know)

The method and argument names are simply there to communicate the role in the workflow - the goal of which is the align timing information - not necessarily to communicate what exactly that function does or how it works

> For example...

[...]

PR #402, opened 2023-04-07
[Sync VIc] Audio Temporal Alignment

CodyCBakerPhD:
This follows after https://github.com/catalystneuro/neuroconv/pull/283, adding start time alignment functionality to the Audio interface

It also incorporates the new mixin testing class for Audio


CodyCBakerPhD:
Needs a little more debugging to refactor the current `conversion_options` approach

src/neuroconv/schemas/time_series_model.py
```diff
@@ -0,0 +1,42 @@
...
+    name: str = Field(description="The name of this TimeSeries dataset.")
+    description: str = Field(default="No description.", description="Description of this TimeSeries dataset.")
+    unit: str = Field(description="The base unit of measurement (should be SI unit).")
+    resolution: float = Field(
+        default=-1.0, description="The smallest meaningful difference (in specified unit) between values in data."
+    )
+    comments: str = Field(default="No comments.", description="Human-readable comments about this TimeSeries dataset.")
+    continuity: Optional[Literal["continuous", "instantaneous", "step"]] = Field(
+        default=None, description=_continuity_description
+    )
```
CodyCBakerPhD, line 30: First example of what our metadata schema(s) would look like in Pydantic

src/neuroconv/schemas/time_series_model.py
```diff
@@ -0,0 +1,42 @@
...
+    # Other fields explicitly excluded from metadata control are:
+    #   - data
+    #   - rate
+    #   - starting_time
+    # Since these are all set by internal NeuroConv operations
+
+
+if __name__ == "__main__":
+    with open(file=Path(__file__).parent / "time_series_schema.json", mode="w") as fp:
+        json.dump(obj=json.loads(TimeSeriesBaseModel.schema_json()), fp=fp, indent=4)
```
CodyCBakerPhD, line 42: First example of how it's easy to regenerate the matching `.json` file from running the file with the Pydantic model in it (so we can have both files available to easily read and reference, but only one - the pydantic model - would need to be hard coded)

CodyCBakerPhD:
A couple things to note...

1) `timestamps` are not currently supported in the interface, need to add that in follow-up

2) included the metadata refactor here since the `starting_time` was actually being included in the metadata field (as well as a conversion option), which we do not want it to be

3) removed the untested duplicate metadata feature and will now raise an error if that occurs informing the user to not pass duplicated metadata per audio file

CHANGELOG.md
```diff
@@ -31,7 +32,7 @@
...
 * Added basic temporal alignment methods to ecephys, ophys, and icephys DataInterfaces. These are `get_timestamps`, `align_starting_time`, `align_timestamps`, and `align_by_interpolation`. Added tests that serve as a first demonstration of the intended uses in a variety of cases. [PR #237](https://github.com/catalystneuro/neuroconv/pull/237) [PR #283](https://github.com/catalystneuro/neuroconv/pull/283) [PR #400](https://github.com/catalystneuro/neuroconv/pull/400)
-* Added `expand_paths`. [PR #377](https://github.com/catalystneuro/neuroconv/pull/377)
+* Added `expand_paths`. [PR #377](https://github.com/catalystneuro/neuroconv/pull/377)* Added basic temporal alignment methods to the AudioInterface. `align_starting_time` is split into `align_starting_times` (list of times, one per audio file) and `align_global_starting_time` (shift all by a scalar amount). `align_by_interpolation` and other timestamp-based approaches is not yet implemented for this interface. [PR #402](https://github.com/catalystneuro/neuroconv/pull/402)
```
bendichter, line 35:
```suggestion
* Added `expand_paths`. [PR #377](https://github.com/catalystneuro/neuroconv/pull/377)
* Added basic temporal alignment methods to the AudioInterface. `align_starting_time` is split into `align_starting_times` (list of times, one per audio file) and `align_global_starting_time` (shift all by a scalar amount). `align_by_interpolation` and other timestamp-based approaches is not yet implemented for this interface. [PR #402](https://github.com/catalystneuro/neuroconv/pull/402)
```

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
```diff
@@ -107,21 +91,74 @@ def get_metadata(self) -> dict:
...
-            "Unable to retrieve timestamps for this interface! Define the `get_timestamps` method for this interface."
+            "The VideoInterface operates on a list of file paths; to reduce ambiguity, please choose "
+            "between `align_global_starting_time` (shift starting time of each video by the same value) "
+            "and `align_starting_times` (specify a list of values to use in shifting the starting time for each video)."
         )
 
-    def align_timestamps(self, aligned_timestamps: np.ndarray):
-        raise NotImplementedError(
-            "The protocol for synchronizing the timestamps of this interface has not been specified!"
+    def align_global_starting_time(self, global_starting_time: float, stub_test: bool = False):
```
bendichter, line 109: I'd really like to keep the consistent form of the API if we can. I think if a single float is input it's pretty clear that every video should be shifted by that float in seconds, and we can use `align_starting_time` for this without worrying. Cody, what is your concern with doing that? How are you worrying that will be misused? I don't understand why there is a need for `align_global_starting_time`.

CodyCBakerPhD: So a class like this should have two methods named `align_starting_time(starting_time=...)` and `align_starting_times(starting_times=...)` and the presence of pluralization is sufficient to avoid any confusion about what each method does? And will never lead to any minor mistakes?

My 2 cents, a 1 character difference is hard to notice visually; I can foresee a less verbose approach as being much harder to debug than having an explicit method name that makes it clear how it applies to a scalar rather than a vector

bendichter: I agree, separating these long function names by a single character could cause confusion. What if we instead change `align_starting_times` to something else, like `align_each_starting_time`. That would allow us to keep a consistent API for `align_starting_time` and would make it easy to differentiate between the functions. Does that satisfy your concerns?

CodyCBakerPhD: It's better, but the word 'each' isn't very descriptive...

What about defining the base method to be `align_global_starting_time` since that is truly what it does in _all_ interfaces, not just the ones with multiple segments or objects?

And then for the recording and sorting interfaces that have multiple segments, have it be called `align_segment_starting_times`?

And then for the audio and video interfaces that have multiple files (that may not be 'segments', but rather a contiguous stream getting concatenated smoothly) call it `align_file_starting_times`? Or I guess we could still get away with calling it `align_segment_starting_times` in that case, they just happen to be segments that are evenly partitioned from the full stream...

bendichter: I'd be happy with `align_starting_time` and `align_segment_starting_times`. I'm not a big fan of the word "global" here, because it makes me think that this might somehow have to do with a global variable, or it would somehow be applied to all interfaces.

CodyCBakerPhD: `align_segment_starting_times` it is then

Adjusted it in the audio PR, might add some follow-ups to improve naming consistency for multi-segment recording/sortings once everything is in `main`

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
```diff
@@ -70,14 +45,23 @@ def __init__(self, file_paths: list, verbose: bool = False):
...
+        self._segment_starting_times = None
 
     def get_metadata_schema(self) -> dict:
         metadata_schema = super().get_metadata_schema()
-        time_series_metadata_schema = get_schema_from_hdmf_class(TimeSeries)
-        metadata_schema["properties"]["Behavior"] = get_base_schema(tag="Behavior")
+
+        time_series_metadata_schema_path = (
+            Path(__file__).parent.parent.parent.parent / "schemas" / "time_series_schema.json"
+        )
```
bendichter, line 59: Is there a reason we are using this over a standard get_hdmf_schema call?

CodyCBakerPhD: Same reason we're centralizing everything from #14, makes it magnitudes easier to read and debug if we have a file that we can look at without having to generate it code every time

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -126,7 +127,7 @@ def check_align_starting_time_internal(self):
         expected_timestamps = unaligned_timestamps + starting_time
         assert_array_equal(x=aligned_timestamps, y=expected_timestamps)
 
-    def check_align_starting_time_external(self):
+    def check_align_timestamps_internal(self):
```
h-mayorquin, line 130: Can you add a short string to these methods? I can't understand what they are doing.

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
```diff
@@ -106,21 +90,63 @@ def get_metadata(self) -> dict:
...
-        )
+    def get_timestamps(self) -> Optional[np.ndarray]:
+        raise NotImplementedError("The AudioInterface does not yet support timestamps.")
+
+    def align_timestamps(self, aligned_timestamps: List[np.ndarray]):
+        raise NotImplementedError("The AudioInterface does not yet support timestamps.")
+
+    def align_starting_time(self, starting_time: float):
+        """
+        Align all starting times for all videos in this interface relative to the common session start time.
```
h-mayorquin, line 103: Videos?

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
```diff
@@ -161,26 +183,35 @@ def run_conversion(
...
         if any_duplicated_audio_names:
             raise ValueError("There are duplicated file names in the metadata!")
 
-        starting_times = _check_starting_times(starting_times=starting_times, metadata=audio_metadata)
+        if self._number_of_audio_files > 1 and self._segment_starting_times is None:
+            raise ValueError(
+                "If you have multiple audio files, then you must specify each starting time by calling "
+                "'.align_segment_starting_times(segment_starting_times=...)'!"
+            )
+        starting_times = self._segment_starting_times or [0.0]
```
h-mayorquin, line 205: Isn't this considred bad practice? Testing for None as a boolean? What is your take on that?

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
```diff
@@ -106,21 +90,63 @@ def get_metadata(self) -> dict:
...
-        raise NotImplementedError(
-            "Unable to retrieve timestamps for this interface! Define the `get_timestamps` method for this interface."
-        )
+    def get_timestamps(self) -> Optional[np.ndarray]:
+        raise NotImplementedError("The AudioInterface does not yet support timestamps.")
+
+    def align_timestamps(self, aligned_timestamps: List[np.ndarray]):
+        raise NotImplementedError("The AudioInterface does not yet support timestamps.")
+
+    def align_starting_time(self, starting_time: float):
```
h-mayorquin, line 101: I feel that the term `starting_time` is overloaded here (and maybe in other instances of this?). 

As in, you are referring to `starting_time` as the time when this stream / device / recorder started acquiring data with respect to the start of the session. That is, this `starting_time` is relative to the session. But, if I understand correctly, you also use the term `starting_time` for the segments relative to the stream / device / recorder. That is, those `starting_times` are relative to the themselves.

I kind of feel that we should have the concept of session time here. That said, you probably have thought more about this. How do you feel? am I missing something?

h-mayorquin (commented):
As requested, @CodyCBakerPhD . I took a look at this today. This week is my first encounter with temporal alignment so I will need to revisit this. 

One thing that I find is that I feel that the name `align_timestamps` kind of overpromises. It is actually just a setter for (and I guess this is where the term comes from) the correct times. However, most of the work of buidling the correct times still has to be done by the user outside of this framwork. In other words, I feel we are forcing them to implement this setters and getters and we don't do much for them.

My preliminary thinking is that  for things that inherit for the yet to be merged `BaseTemporalAlignmentInterface` this work should be done automatically in `run_conversion`. This is probably difficult but in the simplest case it should just require the relative time of the stream with respect to the session. Then we can just the setter and getter infraestructure that you built to do that automatically. 

This is very preliminary and most likely wrong but I wanted to put it out there to get feedback.

CodyCBakerPhD:
Thanks for all the comments

> One thing that I find is that I feel that the name align_timestamps kind of overpromises. It is actually just a setter for (and I guess this is where the term comes from) the correct times. However, most of the work of buidling the correct times still has to be done by the user outside of this framwork.

Feel free to reopen issue #301 with these points to further discuss the alignment method names; within the context of this PR we must treat it as if it is already established by [Sync I] - [Sync VI]

I'd also say though that setting the timestamps _is all it takes to align them_, given that the user is the _only person who knows what the aligned timestamps are_. There is no magic that I am aware of that we can implement to just 'do it for them'.

We do offer tool methods for extracting TTL pulses, and NIDQ interface methods for extracting those from a specified channel, which might be as close as you can get to that; but many labs can store their timing info in something as simple as an external `.csv` or `.npy` file, so we have to be really generic with how that information gets passed to the interface; which takes the form of a simple getter and setter

We also offer 2 non-abstract alignment methods that are defined in terms of the 3 abstract ones that do the more complicated approaches such as aligning by interpolation and shifting the relative start time irrespective of the timing mode



>  In other words, I feel we are forcing them to implement this setters and getters and we don't do much for them.

To be clear, 'forcing them to implement setters and getters' only applies to custom interfaces, which can be built around any API, which may or may not have it's own analogous treatment of timing information

Any core interface in NeuroConv, such as the one updated in this PR, are extensively modified so that the `run_conversion` method may run under the assumption that attached timing information is in fact correct, whereas previously these various types of alignment were implemented ad-hoc as conversion options which made the conversion code unnecessarily complicated and not generic


> My preliminary thinking is that for things that inherit for the yet to be merged BaseTemporalAlignmentInterface this work should be done automatically in run_conversion. This is probably difficult but in the simplest case it should just require the relative time of the stream with respect to the session. Then we can just the setter and getter infraestructure that you built to do that automatically.

Kind of touched on this above, but it simplifies the `run_conversion` step by _a lot_ if it can safely assume that existing timing information is aligned so that alignment is considered a 'pre-step' of running the conversion

In particular, it is often not just the 'relative time of the stream with respect to the session' (a shift to the starting time) but whether or not there are also exact timestamps known for each frame of time in that streams data (very common for video, behavior events, and ophys imaging); that means every run conversion method would have had to encounter an outer product product of timing modes with alignment strategies on top of it's own logic for assembling into NWB

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
CodyCBakerPhD: > As in, you are referring to starting_time as the time when this stream / device / recorder started acquiring data with respect to the start of the session.

That is the intention, yes.


> But, if I understand correctly, you also use the term starting_time for the segments relative to the stream / device / recorder. That is, those starting_times are relative to the themselves.

Can you point to where you're looking for the segment specific stuff? Some internal code might be irrelative to the session start time, but the methods should generally be w.r.t. the session start time


> I kind of feel that we should have the concept of session time here. That said, you probably have thought more about this. How do you feel? am I missing something?

Kind of tricky, because knowing the exact session start time as, say, a `datetime` would not be helpful if we knew the `datetime` of each interface (recorded on separate device, whose clocks are entirely out of sync). Instead, all that can be known by alignment is the relative time as it occurs on a centralized device, usually as sent by TTLs from a secondary system

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Looking back over the mixin methods, there's a lot that could be improved...

Would you be open to working together on a follow-up that cleans up and enhances those core testing methods?

Note that in this PR, we really don't use the mixins in the core testing suite and instead it's all in the `test_behavior/test_audio_interface` which uses generated files (since that's where the majority of tests were prior to this PR anyway)

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
h-mayorquin: I don't think it is exactly that. I think is a combination of two things:
1) https://en.m.wikipedia.org/wiki/Short-circuit_evaluation
2) How python defines truthiness of certain objects 

That's how the [documentation](https://docs.python.org/3/library/stdtypes.html#truth-value-testing) of python talks about it and therein likes the usual critique. The expression works because None evaluates to false and then x = None or y returns y. The problem is that for python other stuff like `0`, `[]` and `""` have values of False. So if you really wanted to test for None only you would have a problem.

Here it is not a problem because for 0, [], and None values of `self._segment_starting_times` it makes sense that they default to [0] as you do. But I was just wondering if you considered that, I think it is good to be aware of.

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
h-mayorquin: I think that I was confused because I was thinking in multisegments objects from spikeinterface. Those objects have starting times `t_start` for every segment and I think they are relative to the first segment. But now that I think about it, I am not sure this is even true for all modalities in spikeinterface.

h-mayorquin:
> Thanks for all the comments

Thank you @CodyCBakerPhD for your elaborated reply. I need to understand this for the conversion that I will be doing now and this discussion is helping me learn this new part of the library.
 
> > One thing that I find is that I feel that the name align_timestamps kind of overpromises. It is actually just a setter for (and I guess this is where the term comes from) the correct times. However, most of the work of buidling the correct times still has to be done by the user outside of this framwork.
> 
> Feel free to reopen issue #301 with these points to further discuss the alignment method names; within the context of this PR we must treat it as if it is already established by [Sync I] - [Sync VI]


Yes, I agree with this. This discussion should be in a different issue.

> Kind of touched on this above, but it simplifies the `run_conversion` step by _a lot_ if it can safely assume that existing timing information is aligned so that alignment is considered a 'pre-step' of running the conversion

Yes, having a method to set the timestamps of streams is a prerequiste for anything. I can see how this simplifies everything by a lot.
 
> In particular, it is often not just the 'relative time of the stream with respect to the session' (a shift to the starting time) but whether or not there are also exact timestamps known for each frame of time in that streams data (very common for video, behavior events, and ophys imaging); that means every run conversion method would have had to encounter an outer product product of timing modes with alignment strategies on top of it's own logic for assembling into NWB

I guess this is what I am confused. If you have the `session_start_time` and the `stream_start_time` isn't it just a matter of
`correct_timestamps` = `oringinal_timestamps_in_stream` + (`stream_start_time` - `session_start_time`)?. But you are telling me about that this would not be enough even if the simplest cases. Let's discuss this outside of here!

src/neuroconv/tools/testing/data_interface_mixins.py
h-mayorquin: Yes, let's, I will schedule something with you.

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
CodyCBakerPhD: > I think it is good to be aware of.

Fair enough


> But I was just wondering if you considered that

I did yeah, that's also why I like that behavior of Python. If any type of nullish states occur (`None` from not specified, empty from who knows what, or `0`/`0.0` but not wrapped in a list), use the fallback value; if not a nullish state then trust it was set correctly and checked by other assertions

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
CodyCBakerPhD: >  I am not sure this is even true for all modalities in spikeinterface.

Uhh I hope so lol, can you confirm that we can trust that? lol

We used that mechanism for our temporal alignment methods here for multi-segment recording/sorting interfaces

I essentially sought to mimic it here viewing each separate file as a different 'segment'

CodyCBakerPhD:
> If you have the session_start_time and the stream_start_time isn't it just a matter of
correct_timestamps = oringinal_timestamps_in_stream + (stream_start_time - session_start_time)?. But you are telling me about that this would not be enough even if the simplest cases. Let's discuss this outside of here!

Yep, feel free to grab a time on my calendly

Ironically it was the same strategy I used in the very first PR for temporal alignment but it was quickly revealed that the strategy would not apply broadly across the synchronization styles

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
h-mayorquin: Thanks for sharing your view!

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
h-mayorquin: That's my belief as well but I don't know. I will confirm with Sam in our next meeting.

CHANGELOG.md
```diff
@@ -33,19 +34,19 @@
 * The `OpenEphysLegacyRecordingInterface` is now extracts the `session_start_time` in `get_metadata()` from `Neo` (`OpenEphysRawIO`) and does not depend on `pyopenephys` anymore. [PR #410](https://github.com/catalystneuro/neuroconv/pull/410)
 * Added `expand_paths`. [PR #377](https://github.com/catalystneuro/neuroconv/pull/377)
 * Added basic temporal alignment methods to ecephys, ophys, and icephys DataInterfaces. These are `get_timestamps`, `align_starting_time`, `align_timestamps`, and `align_by_interpolation`. Added tests that serve as a first demonstration of the intended uses in a variety of cases. [PR #237](https://github.com/catalystneuro/neuroconv/pull/237) [PR #283](https://github.com/catalystneuro/neuroconv/pull/283) [PR #400](https://github.com/catalystneuro/neuroconv/pull/400)
+* Added `expand_paths`. [PR #377](https://github.com/catalystneuro/neuroconv/pull/377)
```
bendichter, line 37:
```suggestion
```

CodyCBakerPhD:
@h-mayorquin Just touching base on this

I think the plan is to get final approval on this, get it into main, and then work together on a final assessment of the feature across interfaces, including improving the testing and documentation

How does that sound?

PR #403, opened 2023-04-11
[Sync VId]: Add video alignment

CodyCBakerPhD:
This adds start time alignment functionality to the Video interface

It also incorporates the new mixin testing class for Video


CodyCBakerPhD:
OK, this interface was a complete mess to refactor to the new starting time + timestamps alignment approach, but it's working now

Similar to #402, removed the magic resolution of duplicate input paths and make it raise an explicit error

I'd like to do a follow-up eventually to simplify the VideoInterface by splitting it into separate interfaces to make the code and tests more modular; one interface for internal mode, one for external mode, and the VideoInterface takes that choice as as argument to `__init__` then routes to the corresponding interface

This is largely because each usage has a different approach to using the metadata and very different write procedures, especially with respect to interactions with starting time, rate, and timestamps

Also, another follow-up should push to completely integrate the `test_gin_behavior.py` into the new mixins

src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
```diff
@@ -212,78 +311,58 @@ def run_conversion(
...
-            for j, (image_series_kwargs, file_list) in enumerate(zip(videos_metadata_unique, file_paths_list)):
-                with VideoCaptureContext(str(file_list[0])) as vc:
-                    fps = vc.get_video_fps()
-                    max_frames = stub_frames if stub_test else None
-                    extracted_timestamps = vc.get_video_timestamps(max_frames)
-                    video_timestamps = (
-                        starting_times[j] + extracted_timestamps if timestamps is None else timestamps[:max_frames]
+            if external_mode:
+                image_series_kwargs = videos_metadata[0]
+                if self._number_of_files > 1 and starting_frames is None:
```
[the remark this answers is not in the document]
CodyCBakerPhD: > We've been using starting frames wrong. See the [NWB core schema](https://github.com/NeurodataWithoutBorders/nwb-schema/blob/19c4bbb36effb1454253072f70ebf2c83f2ea095/core/nwb.image.yaml#L94-L104).

What do you mean? The schema description is how I've understood it to be specified, though we don't actually check if it is equal to the running counter of the video lengths (minus the last video)


> Can we automatically compute these from the videos using openCV?

Of course, but then the entire operation becomes non-lazy - the current approach with external mode only opens a single frame to fetch the sampling rate, and so is very fast.

Also, this PR doesn't mess with that part of the interface, so I'd add it to the list of follow-ups to do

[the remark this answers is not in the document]
CodyCBakerPhD: As per meeting: agreed to add this functionality in a follow-up

tests/test_on_data/test_behavior_interfaces.py
```diff
@@ -69,3 +72,88 @@ def test_conversion_as_lone_interface(self):
...
+    def check_read_nwb(self, nwbfile_path: str):
+        with NWBHDF5IO(path=nwbfile_path, mode="r", load_namespaces=True) as io:
+            nwbfile = io.read()
+            video_type = Path(self.test_kwargs["file_paths"][0]).suffix[1:]
+            assert f"Video: video_{video_type}" in nwbfile.acquisition
+
+    def check_align_starting_time_internal(self):
+        pass  # disabled since this interface follows a different strategy
+
+    def check_align_global_starting_time(self):
```
bendichter, line 97: Same note as [here](https://github.com/catalystneuro/neuroconv/pull/402/files#r1186466052)

CodyCBakerPhD: Adjusted the name for consistency in the Video PR

src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
```diff
@@ -212,78 +309,58 @@ def run_conversion(
...
+
+                image_series_kwargs.update(format="external", external_file=file_paths)
+
+                if timing_type == "starting_time and rate":
+                    starting_time = self._segment_starting_times[0] if self._segment_starting_times is not None else 0.0
+                    with VideoCaptureContext(file_path=str(file_paths[0])) as video:
+                        rate = video.get_video_fps()
+                    image_series_kwargs.update(starting_time=starting_time, rate=rate)
+                elif timing_type == "timestamps":
+                    image_series_kwargs.update(timestamps=np.concatenate(self._timestamps))
```
bendichter, line 355:
```suggestion
                    image_series_kwargs.update(timestamps=np.concatenate(self._timestamps))
                else:
                    raise ValueError(f"Unrecognized timing_type: {timing_type}")
```

bendichter (approved):

PR #404, opened 2023-04-12
add DeepDict

bendichter:


CodyCBakerPhD:
Re: 'try to fix test error in Python 3.11'

This failure is currently affecting `main` as of todays daily runs (yesterdays ran fine).

Trying to pin down what changed between yesterday and today; the only package I can see that was released in that time frame is PyNWB 2.3.2

We would have caught issues with the dev branch tests, but I think the cause might have been the changes made in this PR that was merged just prior to release: https://github.com/NeurodataWithoutBorders/pynwb/pull/1672/files

bendichter:
The errors look to me like an issue with h5py.Dataset

src/neuroconv/utils/dict.py
```diff
@@ -198,3 +198,24 @@ def dict_deep_update(
...
+
+class DeepDict(dict):
+    """A defaultdict of defaultdicts"""
+
+    def __getitem__(self, item):
+        try:
+            return dict.__getitem__(self, item)
+        except KeyError:
+            self[item] = DeepDict()
+            return self[item]
```
CodyCBakerPhD, line 211: This looks very useful generally; only thing, could you add some unit tests that showcase the base expected usage of this class? Such as

1) test that demonstrates commands like

```python
my_deep_dict["key1"]["key2"]["key3"] = some_value
```

2) can keys at any level be any hashable type? I.e.,

```python
my_deep_dict["key1"][2][(3,)] = some_value
```

3) Demonstration of the `print` (indirect call to `repr`) of an instance of such a deep dict (one that is say, 3 levels deep with around ~6 items injected at various levels

4) optional, but encourage at least one test showcasing raising of the `KeyError` at a deep level

src/neuroconv/utils/dict.py
```diff
@@ -198,3 +198,24 @@ def dict_deep_update(
             dict_to_update[key_to_update] = update_values
 
     return dict_to_update
+
+
+class DeepDict(dict):
+    """A defaultdict of defaultdicts"""
```
CodyCBakerPhD, line 204: Would it make sense, perhaps not in this PR, but in a follow-up; to define a `.deep_update` method to this class which is essentially our current `deep_dict_update` or the one from FastAPI?

bendichter: yes, I think it would. I can put that in here. Also, I could work on a better __init__

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -487,7 +487,7 @@ def get_roi_pixel_masks(self, roi_ids: Optional[ArrayLike] = None) -> List[np.nd
         plane_segmentation = plane_segmentations[self.plane_segmentation_name]
 
         true_pixel_masks = _generate_casted_test_masks(num_rois=self.num_rois, mask_type="pixel")
-        assert_array_equal(plane_segmentation["pixel_mask"], true_pixel_masks)
+        assert_array_equal(plane_segmentation["pixel_mask"][:], true_pixel_masks)
```
bendichter, line 490:
```suggestion
        assert_array_equal(plane_segmentation["pixel_mask"], true_pixel_masks)
```

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -518,7 +518,7 @@ def get_roi_pixel_masks(self, roi_ids: Optional[ArrayLike] = None) -> List[np.nd
         plane_segmentation = plane_segmentations[self.plane_segmentation_name]
 
         true_voxel_masks = _generate_casted_test_masks(num_rois=self.num_rois, mask_type="voxel")
-        assert_array_equal(plane_segmentation["voxel_mask"], true_voxel_masks)
+        assert_array_equal(plane_segmentation["voxel_mask"][:], true_voxel_masks)
```
bendichter, line 521:
```suggestion
        assert_array_equal(plane_segmentation["voxel_mask"], true_voxel_masks)
```

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -613,7 +613,7 @@ def get_roi_pixel_masks(self, roi_ids: Optional[ArrayLike] = None) -> List[np.nd
...
```
bendichter, line 616:
```suggestion
        assert_array_equal(plane_segmentation["voxel_mask"], true_voxel_masks)
```

tests/test_ophys/test_tools_roiextractors.py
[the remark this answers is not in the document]
bendichter: done

tests/test_minimal/test_utils/test_dict.py
```diff
@@ -0,0 +1,36 @@
...
+    def test_getitem(self):
+        self.assertEqual(self.dd["a"]["b"]["c"], 42)
+
+    def test_missing_key(self):
+        dd = DeepDict()
+        self.assertIsInstance(dd["non_existent"], DeepDict)
+
+    def test_to_dict(self):
+        expected = self.data
+        self.assertEqual(self.dd.to_dict(), expected)
```
CodyCBakerPhD, line 21: Is this equivalent to `dict(my_deep_dict)`? I assume that would be the more common calling case - I also assumed it would need to be called `__to_dict__` or something for that to work

CodyCBakerPhD: Upon inspecting the `OrderedDict` class, it's just called `__dict__`. Not a bad thing to keep `.to_dict()` as is, but I'd recommend adding `__dict__` and a test that `dict(my_deep_dict)` works for convenience

src/neuroconv/utils/dict.py
bendichter: 1: done
2: done
3: done
4: the function is refactored so this is no longer relevant

tests/test_minimal/test_utils/test_dict.py
bendichter: ok, done

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -123,13 +123,13 @@ def run_conversion(
         metadata: dict
             metadata info for constructing the nwb file (optional).
         overwrite: bool, optional
-            Whether or not to overwrite the NWBFile if one exists at the nwbfile_path.
+            Whether to overwrite the NWBFile if one exists at the nwbfile_path.
         """
-        base_metadata = self.get_metadata()
-        metadata = dict_deep_update(base_metadata, metadata)
+        _metadata = self.get_metadata()
+        _metadata.deep_update(metadata)
```
CodyCBakerPhD, line 129: I would highly recommend reverting all these integrations within the interfaces - that can be a next step, but currently the `dict_deep_update` does have occasionally very different behavior from `deep_update` on a `DeepDict`, particularly with respect to `list` type values

src/neuroconv/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
```diff
@@ -124,7 +124,7 @@ def get_metadata(self) -> dict:
         metadata["Ecephys"].update(NeuroScopeRecordingInterface.get_ecephys_metadata(xml_file_path=xml_file_path))
         session_start_time = get_session_start_time(str(xml_file_path))
         if session_start_time is not None:
-            metadata = dict_deep_update(metadata, dict(NWBFile=dict(session_start_time=session_start_time)))
+            metadata["NWBFile"]["session_start_time"] = session_start_time
```
CodyCBakerPhD, line 127: This type of integration is great though, really showcases how much nicer the new class is to work with

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
bendichter: ok, but they were failing. I suppose a simpler solution would be to have `dict_deep_update` cast any DeepDicts to dicts.

bendichter:
@CodyCBakerPhD 

latest changes:

* Adds `DeepDict.__deepcopy__` which fixes the issue with `dict_deep_update`
* Removes the places where `dict_deep_update` was replaced, except for the straightforward places were a single nested metadata element is adjusted.
* `DeepDict.deep_update` can take a `dict` or kwargs, similar to`dict.update`.

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
CodyCBakerPhD: Yeah, I wouldn't be surprised if there aren't any explicit tests out there for append vs. extend usages of the previous deep update; will need to carefully examine where the assumptions and uses of that pop up in practice

Or, better solution, just starting requiring everything in NeuroConv metadata to be a dict instead of list and then we never need to worry about it

tests/test_minimal/test_utils/test_dict.py
CodyCBakerPhD: > ok, done

Not seeing this in the code or tests

tests/test_minimal/test_utils/test_dict.py
```diff
@@ -0,0 +1,58 @@
...
+        update_data = {"a": {"b": {"d": 55}, "e": {"f": 66}}, "g": {"h": 77}}
+        self.dd.deep_update(**update_data)
+        expected = {"a": {"b": {"c": 42, "d": 55}, "e": {"f": 66}}, "g": {"h": 77}}
+        self.assertEqual(dict(self.dd), expected)
+
+    def test_deepcopy(self):
+        dd2 = deepcopy(self.dd)
+        dd2["a"]["b"]["c"] = 0
+        self.assertEqual(dd2["a"]["b"]["c"], 0)
+        self.assertEqual(self.dd["a"]["b"]["c"], 42)
```
CodyCBakerPhD, line 58: 👍 good idea

tests/test_minimal/test_utils/test_dict.py
CodyCBakerPhD: Oh, nevermind, I see the test now

bendichter: The `__dict__` magic method is removed. `DeepDict.to_dict()`was working properly, but `dict(DeepDict)` was not being applied recursively. Apparently `dict()` does not call `__dict__` in this case? I couldn't quite figure it out.

CodyCBakerPhD (approved):

PR #405, opened 2023-04-12
Trying to mix 3.11 tests on main

CodyCBakerPhD:
CI failing lately; seeing if it's due to recent PyNWB changes


CodyCBakerPhD:
@bendichter A quick temporary fix to restore the CI

Will raise a separate issue to try to determine the source cause with PyNWB

bendichter (approved):

Issue #406, opened 2023-04-12
[Bug]: PyNWB 2.3.2 on Python 3.11 causes dataset shaping issue in ophys tests


CodyCBakerPhD:
### What happened?

Main CI ran fine on all platforms and python versions on 4/11/2023 prior to PyNWB 2.3.2 release.
Dev CI also ran fine on that day, indicating the `dev` branch of `pynwb` was fine that day as well. Though our dev CI does not run on 3.11... might just expand that so we can detect this kind of thing earlier

Logs available https://github.com/catalystneuro/neuroconv/actions/runs/4680669338/jobs/8292313477?pr=404

Speculation is has something to do with the shape of the `h5py.Dataset`

### Steps to Reproduce

```python
Logs available https://github.com/catalystneuro/neuroconv/actions/runs/4680669338/jobs/8292313477?pr=404
```

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.7

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

PR #407, opened 2023-04-12
[CI]: expand dev test python versions

CodyCBakerPhD:
To detect the problems similar to #406 going into the future


weiglszonja (commented):
The CHANGELOG.md update is missing but otherwise LGTM, we should catch these in time. Are the tests taking longer now that we're testing for four Python versions?

CodyCBakerPhD:
> Are the tests taking longer now that we're testing for four Python versions?

They're run in parallel so will take same total amount of time if all resources start at the same time

It will use up slightly more resources which might indirectly slow down multiple simultaneous PRs, but I'd rather be more thorough in that regard

weiglszonja (approved):

CodyCBakerPhD:
GitHub Actions down this morning, testing to see if working yet

PR #408, opened 2023-04-13
Docstring improvements to excel interface

bendichter:


src/neuroconv/datainterfaces/text/excel/exceltimeintervalsinterface.py
```diff
@@ -7,5 +9,27 @@
...
+            Passed to pandas.read_excel()
+        verbose : bool, default: True
+        """
+        read_kwargs = read_kwargs or dict()
+        super().__init__(file_path=file_path)
+        self.verbose = verbose
+
+        self._read_kwargs = read_kwargs
+        self.dataframe = self._read_file(file_path, **read_kwargs)
+        self.time_intervals = None
```
CodyCBakerPhD, line 32: This appears to be duplicated line for line from the parent - can you not simply call `super().__init__(file_path=file_path, read_kwargs=read_kwargs, verbose=verbose)`? If the only purpose is to give this a better non-inherited docstring

bendichter, line 32:
```suggestion
        super().__init__(file_path=file_path, read_kwargs=read_kwargs, verbose=verbose)
```

bendichter: good point

CodyCBakerPhD (approved):

PR #410, opened 2023-04-14
Get `session_start_time` from Neo reader for `OpenEphysLegacyRecordingInterface`

weiglszonja:
`OpenEphysLegacyRecordingInterface` doesn't need to rely on `pyopenephys` to get the start time of the recording.
This information is already in `self.recording_extractor.neo_reader` under `raw_annotations`.


CodyCBakerPhD:
LGTM; the neo parser doesn't even use it either (https://github.com/NeuralEnsemble/python-neo/blob/2d63e18fb9a75131a2e54e32323cda23e2557356/neo/rawio/openephysrawio.py#L545)

Which makes me wonder, do we need the `pyopenephys` package at all then? Or can we remove it from the requirements for this interface?

weiglszonja:
@CodyCBakerPhD I think the `OpenEphysBinaryRecordingInterface` would still need it because the [openephysbinaryrawio](https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/openephysbinaryrawio.py) wouldn't have this "date_created" field like [here](https://github.com/NeuralEnsemble/python-neo/blob/2d63e18fb9a75131a2e54e32323cda23e2557356/neo/rawio/openephysrawio.py#L288).

CodyCBakerPhD:
@weiglszonja Can you check if it has any equivalent in the neo annotations? If not then sure, I guess we have to keep it even just for the Binary version

weiglszonja:
@CodyCBakerPhD I checked but didn't find it in the binary io reader.

CodyCBakerPhD (approved):

CodyCBakerPhD:
> @CodyCBakerPhD I checked but didn't find it in the binary io reader.

@weiglszonja Thanks for checking!

PR #413, opened 2023-04-18
[Sync VIf] Fix sorting interface alignment

CodyCBakerPhD:
The previous implementation was very naive; I've refactored the SortingInterface to now properly utilize the various inter-compatabilities that arise when fetching spike times from an extractor for single vs. multi segment and those with and without attached recordings


src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -72,14 +74,72 @@ def get_metadata_schema(self) -> dict:
...
+        raise NotImplementedError(
+            "Unable to fetch original timestamps for a SortingInterface since it relies upon an attached recording."
+        )
+
+    def get_timestamps(self) -> Union[np.ndarray, List[np.ndarray]]:
+        if not self.sorting_extractor.has_recording():
+            raise NotImplementedError(
+                "In order to align timestamps for a SortingInterface, it must have a recording "
+                "object attached to it! Please attach one by calling `.register_recording(recording_interface=...)`."
+            )
```
bendichter, line 90: sorting extractors don't have sampling rate that can be used for this?

CodyCBakerPhD: They have a sampling rate, but they don't have a notion of timestamps as a vector without a recording extractor

Technically the sorting extractor design is based on which frames have spiking activity and that gets mapped into a 'time' in a number of ways depending on the attached recording, which is the object that really knows about the time a certain frame occurs

bendichter: Would it be reasonable to just assume constant sampling rate if no recording extractor is present? Most electrophysiology recordings are quite regular, right?

CodyCBakerPhD: See this relevant snippet in SI: https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/basesorting.py#L115-L122

Yes, the default behavior of any sorting extractor will simply use the sampling frequency (which is a required property) to map the frames to times, up to the shift value per-segment that is handled below by the `align_starting_time` method.

I don't really see any point in having a method like `my_sorting_extractor.get_timestamps()` that, without an attached recording, only generates a `np.arange` vector from the sampling frequency. Can you give an example of how that vector would ever be used?

bendichter: Maybe it's not necessary. So if you tried to warp the timestamps of a sorting extractor, what would happen? Let's say you wanted to use the interpolation method. Your spike sorting system sends pulses to a central system and you want to align timestamps to that central system. Is that possible with the current implementation? Would the timestamps be generated on the fly? I suppose even in the interpolation case you would be able to determine the timings algebraically so it may not even be necessary to ever compute an entire array of pre-warp timestamps.

CodyCBakerPhD: > So if you tried to warp the timestamps of a sorting extractor, what would happen?

You would warp the timestamps of the recording interface, which serves as the 'ground truth' for the spike timings of the sorting extractor (which otherwise only knows the frame indices at which spikes were identified, and without an attached recording would simply map those based on the sampling frequency)

Critically, this is why I require an attached recording interface - because whatever temporal alignment you applied, or would apply in order to align the recording interface, it will have the same exact affect on the spike times derived from that recording


>  Is that possible with the current implementation?

Yes, all alignment methods that work on recording interfaces apply to the sorting interfaces with this PR



> Would the timestamps be generated on the fly?

Depends on the format. Some formats might record segment shift times, so that would get passed along. Some formats might have irregular timestamps set automatically. Some might be adjusted manually through our alignment methods. Some might generate on the fly.

Point being, you don't actually know until you initialize the recording interface on that instance of data; and in terms of design here it's agnostic - it simply relies on the existing SpikeInterface machinery to decide when and how to do all that

bendichter: So if you had no recording interface, you could not do this warping? For instance, if you have a Phy results from a NP recording that you are trying to warp, you could not do that unless you also have a SpikeGLX file? 

> Some formats might have irregular timestamps set automatically.

Could you clarify what you mean by this? If I recall correctly, SortingExtractors record spike times as sample integers, so knowing the precise sample times is essential for knowing spike times. That means we should either be able to make the assumption that samples are regular, or we would need to store the individual timestamps within the sorting extractor. Either way, we should be able to recover timestamps without needing a recording extractor, right?

CodyCBakerPhD: > So if you had no recording interface, you could not do this warping? For instance, if you have a Phy results from a NP recording that you are trying to warp, you could not do that unless you also have a SpikeGLX file?

The term 'warping' is new here; I've listed below what is possible with and without an attached recording. Essentially, anything involving irregular timestamps has to go through a recording, but anything you want to do under the assumption of regular timestamps wouldn't.

### (1) Without a recording attached

We can...

(1a) Define starting times of all segments, including a constant shift of each

(1b) Retrieve the spike timing of the sample indices, which are inferred regularly using the sampling frequency


### (2) With a recording attached

We can...

(2a) Define starting times of all segments, including a constant shift of each; though it uses a different mechanism than (1a) above

If the attached recording _does not_ have a timestamps vector set, then we can

(2b) Retrieve the spike timing of the sample indices, which are inferred regularly using the sampling frequency in the case when the recording extractor did not load or otherwise have set a timestamps vector, which can occasionally occur for certain formats

Note that some formats can set timestamps automatically during their  `__init__`. Otherwise we can set it ourselves using the RecordingInterface alignment methods

If the attached recording has a timestamps vector set, then we can

(2c) Retrieve the spike timing of the sample indices, which index the timestamps vector of the recording

(2d) Interpolate the spike timing of the sample indices, which index the timestamps vector of the recording, which is interpolated from the rule that would apply to the recording timestamps

CodyCBakerPhD: > Could you clarify what you mean by this?

Example, the NeuralynxRecordingExtractor applied to the version collected by the Brody lab a couple years ago. Timestamps parsed and loaded at time of `__init__` (a rather expensive operation too), and very slightly irregular. No clue how the mechanism itself worked on the hardware level.


>  If I recall correctly, SortingExtractors record spike times as sample integers, so knowing the precise sample times is essential for knowing spike times.

Yes, exactly. That is how the SI SortingExtractors work. However...

> That means we should either be able to make the assumption that samples are regular, or we would need to store the individual timestamps within the sorting extractor.

That is where SI separates things; the SortingExtractor does not itself store the timing information as it is considered a property of the associated RecordingExtractor; so instead, you store the individual timestamps in a RecordingExtractor and then attach the RecordingExtractor to the SortingExtractor and voila.

bendichter: I'm not a big fan of separating information this way, because it requires us to attach a RecordingExtractor and therefore requires us to have potentially large raw data files present in order to do timestamps interpolation, when that really should not be required. We should be able to apply the interpolation method without a RecordingExtractor present. If we don't have the timestamps available from the RecordingExtractor, we should be able to infer them from the SortingExtractor.

bendichter: I don't anticipate interpolation of a sorting extractor to be very common, and I suppose needing to interpolate a sorting extractor and not having a recording extractor present would be even less common. If it's really difficult to handle this use-case I suppose it would be fine to document this in a comment or NotImplemented message and put this on the back-burner to prioritize other features

bendichter (approved):

PR #390, opened 2023-03-31
add `BrukerTiffImagingInterface`

weiglszonja:
Add support for Bruker 2P imaging data which consists of:
- individual TIFF files in OME-TIF format (.ome.tif files) where each TIFF file corresponds to a single frame
- metadata in XML format (.xml and .env files)

TODO:
- [x] https://github.com/catalystneuro/roiextractors/pull/220
- [x] [Upload](https://gin.g-node.org/CatalystNeuro/ophys_testing_data/src/main/imaging_datasets) Bruker test data and add unittests for interface metadata

PR #418, opened 2023-04-19
Extend type array condition in `get_schema_from_hdmf_class`

weiglszonja:
While working on #390 I noticed that it's not possible to add `grid_spacing` `ImagingPlane` metadata in `get_metadata() `because it returns :

`jsonschema.exceptions.ValidationError: Additional properties are not allowed ('grid_spacing' was unexpected)`

The metadata schema for ImagingPlane https://github.com/catalystneuro/neuroconv/blob/1357319ffe454fda82824dfec6af6d0f2854ed3f/src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py#L49 doesn't contain `grid_spacing` and other array type metadata if they are not of `Iterable` type:
https://github.com/catalystneuro/neuroconv/blob/1357319ffe454fda82824dfec6af6d0f2854ed3f/src/neuroconv/utils/json_schema.py#L205-L208

I added two tests, but if this may lead to hidden changes in other parts of the schema, I'm open for better solutions for this.


CodyCBakerPhD:
This is yet another case of something that would have been much easier to catch and debug if the schemas were explicitly written out instead of implicitly generated from docval

Happy to get it in as a patch needed for Bruker, but we should keep looking for better options going forward that make life easier for us here

CodyCBakerPhD (approved):

PR #420, opened 2023-04-19
Separate live service testing workflows

CodyCBakerPhD:
Very annoying `boto` error coming somewhere in both DANDI and Globus specific to Mac (all Python versions): https://github.com/catalystneuro/neuroconv/actions/runs/4746446753/jobs/8431451075?pr=292

Proposing to separate out the tests that run using live services to make these a non-requirement for merging PRs that are unrelated to those functionalities


.github/workflows/live-service-testing.yml
```diff
@@ -0,0 +1,45 @@
+name: Live service testing
+on:
+  schedule:
+    - cron: "0 16 * * *"  # Daily at noon EST
```
bendichter, line 4: So call this daily, not with PRs?

CodyCBakerPhD: Like the main tests, it's both - its the 'workflow_call' trigger that gets called from the main deployment that triggers on PRs

bendichter: oh ok

bendichter (dismissed):

CodyCBakerPhD:
Issue was solved on #292, but I still think this is a good idea going forward just for the sake of our sanity (such as when the DANDI tests broke a while back)

bendichter (approved):

PR #424, opened 2023-04-24
Add SpikeGLXConverter to gallery

CodyCBakerPhD:
replaces #232


CodyCBakerPhD:
ophys doctests failing because `main` has not yet made a cache with the new testing data, seeing if I can force it

docs/conversion_examples_gallery/recording/spikeglx.rst
```diff
@@ -7,7 +7,41 @@ Install NeuroConv with the additional dependencies necessary for reading SpikeGL
...
+
+.. code-block:: python
+
+    >>> from datetime import datetime
+    >>> from dateutil import tz
+    >>> from pathlib import Path
+    >>> from neuroconv.converters import SpikeGLXConverterPipe
+    >>>
+    >>> folder_path = f"{ECEPHY_DATA_PATH}/spikeglx/Noise4Sam_g0"
+    >>> converter = SpikeGLXConverterPipe(folder_path=folder_path)
```
CodyCBakerPhD, line 26: Allowing verbose here because it's cleaner to handle with ELLIPSIS as an output field in the last step

docs/conversion_examples_gallery/recording/spikeglx.rst
```diff
@@ -7,7 +7,41 @@ Install NeuroConv with the additional dependencies necessary for reading SpikeGL
...
+    >>>
+    >>> # Extract what metadata we can from the source files
+    >>> metadata = converter.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> session_start_time = metadata["NWBFile"]["session_start_time"].replace(tzinfo=tz.gettz("US/Pacific"))
+    >>> metadata["NWBFile"].update(session_start_time=session_start_time)
+    >>>
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> nwbfile_path = output_folder / "my_spikeglx_session.nwb"
```
CodyCBakerPhD, line 35: Two reasons I had to do this - a) having two doctests in the same file led to a file collision where one tried to write in append mode; b) I got feedback at the NeuroDataShare workshop that the f-string approach was more confusing than having an explicit file name as the example input - if this looks good I can also do something similar for the rest of the gallery so it all looks consistent

docs/conversion_examples_gallery/recording/spikeglx.rst
```diff
@@ -7,7 +7,41 @@ Install NeuroConv with the additional dependencies necessary for reading SpikeGL
...
+    >>> converter.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
+    root pynwb.file.NWBFile at ...
+
+
+
+Single-stream
+~~~~~~~~~~~~~
+
+Defining a 'stream' as a single band on a single NeuroPixels probe, we can convert either an AP or LF SpikeGLX stream to NWB using
```
h-mayorquin, line 44: Is band the right word here? Not that I find a better one. Maybe this is very clear for people that work making this recordings but is not for me. Other possibilites: modalities, recoding mode.

docs/conversion_examples_gallery/recording/spikeglx.rst
h-mayorquin: I see. Yes, maybe it makes more sense to have the paths in this way. I do have the feeling that users are more likely to be aware of writing paths with slashes (i.e. paths/look/like/this) than with f-strings. This means that even if the following is more explicit, f"{your_file_path}, the fact that they don't know f-strings will be distracting there. My only minor grip is that users of windows make a mistake with the reverse slash (/ vs \) but we are  splitting hairs here. 

I am fine with both, so up to you and Ben.

h-mayorquin: Another hacky solution for this kind of problem is to profit from unicode characters to create something that is visually identical to `path_to_save_nwb_file`. It should be possible to use for example 𝚎: U+1D792 (MATHEMATICAL SANS-SERIF ITALIC SMALL E) at the to go the append with duplicate file problem go away.

Just kidding.

h-mayorquin (approved):
This looks good to me.

docs/conversion_examples_gallery/recording/spikeglx.rst
CodyCBakerPhD: Band is the word I hear most for talking about AP vs LF, i.e., 'AP band'; I don't actually know though lol

Maybe something for @alejoe91 to weigh in on

CodyCBakerPhD:
> My only minor grip is that users of windows make a mistake with the reverse slash (/ vs ) but we are splitting hairs here.

Which is exactly why I love `pathlib` so much - the `/` operator from Python maps onto whatever their `os.sep` happens to be, so all they need to do is copy and paste our code and it should 'just work'


> It should be possible to use for example 𝚎: U+1D792 (MATHEMATICAL SANS-SERIF ITALIC SMALL E) at the to go the append with duplicate file problem go away.

... wow 😆 did not even cross my mind, but I'll have to remember that one lol

PR #430, opened 2023-05-02
Set upper numpy bound for Python < 3.11

CodyCBakerPhD:
Testing a fix that was just revealed by failing tests on #428; zero clue why it happened this way, the [daily tests ran perfectly find 6 hours earlier](https://github.com/catalystneuro/neuroconv/actions/runs/4863353553/jobs/8670942626), and the only major difference I can spot between the two CI installations is the numpy version was 1 patch lower in the dailies.


CodyCBakerPhD:
@bendichter Needed to unblock #428 and prevent tomorrows dailies from failing

bendichter:
sorry for the delay on this. These types of things can be approved by other members, not be personally

bendichter (approved):

PR #433, opened 2023-05-05
Add new interfaces to API docs

CodyCBakerPhD:


bendichter:
Could we also keep this alphabetical within section? Is there a way to automate this so we don't have to remember to add them to the docs every time?

CodyCBakerPhD:
Alphabetized the `ophys` section

`automodule` alphabetized contents automatically, but then we can't control the section titles as nicely (which is why we apply it to each format as a sub-module)

bendichter (approved):

PR #436, opened 2023-05-08
Add Zarr backend for electrical series

alejoe91:
Added option to write electrical series to Zarr. 
This is just a proposal, but I'm happy to discuss.

This is briefly what I've added/changed:
- unify compression parameters to `compression_kwargs`
- add a `backend` parameter (`hdf5` -default- or `zarr`) to use either the `H5DataIO` or the `ZarrDataIO` class.

### TODO
- [ ] Add and modify tests
- [ ] Propagate to data interfaces

### DIFF
--- a/src/neuroconv/tools/spikeinterface/spikeinterface.py
+++ b/src/neuroconv/tools/spikeinterface/spikeinterface.py
@@ -518,7 +518,7 @@ def add_electrical_series(
     compression_kwargs: Optional[dict] = None,
     iterator_type: str = "v2",
     iterator_opts: Optional[dict] = None,
-    backend: Literal["hdf5", "zarr"] = "hdf5"
+    backend: Literal["hdf5", "zarr"] = "hdf5",
 ):
     """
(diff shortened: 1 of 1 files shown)


bendichter:
The arg syntax of `compression` and `compression_opts` is currently mirroring [`h5py.create_dataset`](https://docs.h5py.org/en/stable/high/group.html#h5py.Group.create_dataset) and [all of the filters in the hdf5plugin library](https://github.com/h5py/h5py/blob/bc2e0bba6bff8e33f0619b17b5126cdf03d0cf7e/h5py/_hl/filters.py#L115-L119). I am not 100% opposed to changing it, but we would need a pretty good reason. Would it be possible to represent Zarr compressors and filters with a similar syntax, or is this structure not rich enough to allow for important features in Zarr?

alejoe91:
Thanks for the discussion pointer! I'll comment there ;)

CodyCBakerPhD:
This PR has been replaced by the larger backend chain of PRs

PR #437, opened 2023-05-09
Update `write_waveforms` to write sparse waveforms

alejoe91:
From SpikeInterface 0.97, one can extract waveforms in sparse mode. This patch updates the `write_waveforms` function to support this format.

I also added a `force_dense` option to pad the waveforms with zeros and make them dense.
While in principle we can create nwbfile objects with sparse waveforms (and respective electrode indices), note that writing this to file is currently not supported, since it will require the `waveform_mean` and `waveform_sd` columns to be indexed (see relared issue on `pynwb`: https://github.com/NeurodataWithoutBorders/pynwb/issues/1693)

I also added another option to support an additional use case:
A user might want to write waveforms that have been computed after preprocessing and removing noisy channels, but at the same time add the full recording. If this is the case, there is a mismatch between the `recording.channel_ids` and `waveform_extractor.channel_ids`. This PR also fixes this by re-indexing the `electrode_indices` accordingly.


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1275,9 +1276,35 @@ def add_waveforms(
             if prop not in sorting_copy.get_property_keys():
                 sorting_copy.set_property(prop, tm[prop])
 
+    # find electrode table region corresponding to the recording
```
h-mayorquin, line 1279: I just realized that recording above is optional but this call will fail if recording is None.

h-mayorquin (commented):
I will take a look at the tests once they are passing.

I have two main comments:
1) I believe that the machinery to transform dense to sparse and viceversa should be in spikeinterface. My take is that they should be methods of the sparsity class. If we leave this logic here and there where this is needed as we are doing right now and you decide to change the sparsity in the future then we need to go and make a lot of changes. If instead the sparsity definitions and transformations are together then this is less likely to happen.

2) Contraty to most tests in this module, the tests here are done for writing instead of appending.  Given we are discussing this https://github.com/catalystneuro/neuroconv/issues/441 at least the new tests that we add now should be for appending to the object.

alejoe91:
@CodyCBakerPhD we can close this. Ragged waveforms meand is on its way on pynwb, so we'll pick this up when that is merged

PR #438, opened 2023-05-09
Propagate `write_as`, `units_name`, `units_description` to `BaseSortingExtractorInterface` `run_conversion` method

weiglszonja:
From a sorting interface, it is not possible currently to write units other than the default units table, with the default description. I'd like to propagate `write_as`, `units_name`, `units_description` to `BaseSortingExtractorInterface` `run_conversion` method to be able to modify them e.g. add units to processing with a custom table name, and a description.


bendichter:
lgmt

CodyCBakerPhD (approved):

PR #440, opened 2023-05-09
add path expansion docs

bendichter:


docs/index.rst
```diff
@@ -13,7 +13,7 @@ of proprietary formats to the Neurodata Without Borders (NWB) standard.
 
 Features:
 
-* Reads data from 40 popular neurophysiology data formats and writes to NWB using best practices.
+* Reads data from 42 popular neurophysiology data formats and writes to NWB using best practices.
```
CodyCBakerPhD, line 16: Where did these other 2 come from? We recently bumped from 38 due to Bruker and MicroManager

bendichter, line 16:
```suggestion
* Reads data from 40 popular neurophysiology data formats and writes to NWB using best practices.
```

bendichter:
> This section is a good description of how to use the API for this, but isn't quite as user-friendly for the people that will get routed here from the NWB GUIDE
> 
> Could you add a second section in a follow-up PR that is even more high level and shows several examples using screenshots (or another visual descriptor) of folders containing files following naming conventions taken from real world cases and what their corresponding f-string patterns look like

Agreed. I'll work on this

PR #442, opened 2023-05-10
[Sync VII] Define BaseTemporalAlignmentInterface

### LINKED ISSUE #439: Should alignments methods be abstract?
Currently, the `BaseDataInterface` requires child classes to implement methods related to alignment, as outlined in the following link:

https://github.com/catalystneuro/neuroconv/blob/5c6b5d982bc896c5ec05b9cc417f5040e4d967ef/src/neuroconv/basedatainterface.py#L44-L90

This however,  breaks compatibility with previous conversion code, as behavior interfaces often did not implement these methods, causing them to fail under the current contract.

However, the second point could be seen as a "feature, not a bug" situation. As @CodyCBakerPhD suggests, the intention may be to encourage developers to think carefully about these issues and implement empty methods with a pass statement if no further action is needed. Both points have their merits.

To summarize, the current approach has its pros and cons:

Pros: 
* Encourages developers to think about alignmen from the start.
* Can be bypassed easily by defining an empty method.
Cons:
* Breaks compatibility with previous code.
* Introduces more frictions in development.  It increases development time by imposing additional requirements on developers. In other words, anyone creating a new interface now needs to implement these methods and consider their implications, which incurs an upfront cost.

@CodyCBakerPhD points out that past conversions should have requirements that allow them to run with those specific requirements. However, it is uncertain how thorough the project has been in implementing these requirements, leaving some doubt as to their effectiveness.

CodyCBakerPhD:
fix #439 by making it a pure feature; a new interface with the new methods that existing interfaces now use exclusively, but maintains back-compatibility of previously created custom data interfaces


CodyCBakerPhD:
Dang, GitHub actions having trouble all day long now. First time I've seen that...

h-mayorquin (commented):
This looks good to me. Now it is opt-in instead of opt-out by overiding methods. Something that I don't see is the warning that @bendichter mentioned on the issue.

CodyCBakerPhD:
> Something that I don't see is the warning that @bendichter mentioned on the issue.

Changelog is already somewhat large for a minimal change, I was thinking of adding and testing the warning in a follow-up

tests/imports.py
```diff
@@ -30,15 +30,18 @@ def test_top_level(self):
...
+            "baseextractorinterface",
             "nwbconverter",
             "tools",  # Attached to namespace by NWBConverter import
             "utils",  # Attached to namespace by NWBconverter import
             # Exposed attributes
             "NWBConverter",
             "ConverterPipe",
             "BaseDataInterface",
-            "TemporallyAlignedDataInterface",
-            "basedatainterface",  # Imported by ConverterPipe as a type-hint
```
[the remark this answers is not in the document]
CodyCBakerPhD: I think it's from intentional importing not just type-hints now, since I did adjust the core `__init__` to expose these 3 base data interfaces

So when the BaseDataInterface gets imported by one of the submodules (the all lower-case ones) the submodule itself gets set in the module `__dict__` as well

CodyCBakerPhD:
@h-mayorquin OK this seems ready now

src/neuroconv/baseextractorinterface.py
```diff
@@ -2,11 +2,11 @@
 from abc import ABC
 from typing import Optional
 
-from .basedatainterface import BaseDataInterface
+from .basetemporalalignmentinterface import BaseTemporalAlignmentInterface
 from .tools import get_package
 
 
-class BaseExtractorInterface(BaseDataInterface, ABC):
```
h-mayorquin, line 9: I also think it is too early to make this contract. Are we sure that all the objects from spikeinterface and roiextractors will be alignable with our methods? If not, then we are overcommiting. I think we should separate this.

h-mayorquin (commented):
I read this. It was useful for me to start thinking about the issue. This looks good to me but I want to hear what you and @bendichter think about  forcing everything that inherits from `BaseExtractor` to implement time alignment methods.

src/neuroconv/baseextractorinterface.py
CodyCBakerPhD: These are already done

Audio is the final interface to refactor to use these methods

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
```diff
@@ -44,7 +44,7 @@ def _check_starting_times(starting_times: list, metadata: List[dict]) -> list:
     return starting_times
 
 
-class AudioInterface(BaseDataInterface):
```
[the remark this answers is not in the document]
CodyCBakerPhD: Fair enough, adjusted

src/neuroconv/baseextractorinterface.py
h-mayorquin: OK. I also could not think why someone that is not us would like to inherit from this class so this makes sense. I still have the lingering feeling that adding this capablities to the interfaces should be done through composition but this probably does not matter. Just leaving the thought.

h-mayorquin (approved):
This looks good to me.

src/neuroconv/baseextractorinterface.py
CodyCBakerPhD: What do you mean by composition? It's not too late to change anything yet

PR #443, opened 2023-05-10
Fix text characters in schema

CodyCBakerPhD:
This ended up being the source of the issue with the GUIDE rendering of those fields @garrettmflynn


garrettmflynn (approved):
Looks great. Thanks for finding this!

PR #446, opened 2023-05-11
Update `CellExplorerSortingInterface` to new format

h-mayorquin:
As we have working on https://github.com/catalystneuro/buzsaki-lab-to-nwb/issues/55 we have discussed with one of the members of the Buzsaki lab which shared the new specification with us. In the new specification the samping_rate is no longer present in a file `sessionInfo` as it used to be. I am updating both this interface and the sorting extractor in spikeinterface to read the sampling frequency when is available in the same file as the new format stipulates

https://github.com/SpikeInterface/spikeinterface/pull/1628

Note quite suere if I am setting the requirements right to test.


CodyCBakerPhD:
Biggest comment will be similar to the extractor - I would try to maintain it as a file path style unless there is a significant reason to swap to folder; the only difference I'm aware of ATM is the different type of header file from previous versions

h-mayorquin:
Yeah, mabye you are right, @CodyCBakerPhD .

I am also thinking on loading the properties in spikeinterface instead of doing that here.

CodyCBakerPhD:
> I am also thinking on loading the properties in spikeinterface instead of doing that here.

I would be OK with that too, good idea. Kind of similar to what I did for the `StreaingIblSortingExtractor` (or w/e it's called)

h-mayorquin:
> > I am also thinking on loading the properties in spikeinterface instead of doing that here.
> 
> I would be OK with that too, good idea. Kind of similar to what I did for the `StreaingIblSortingExtractor` (or w/e it's called)

Going to do that later though. Once we are checking the data of Huszar conversion in detail. I am hoping I will be more familiar with the data and I will be able to make a better design decision at that point.

CodyCBakerPhD:
@h-mayorquin Could you send me an example file for this? Just for a sanity check

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
```diff
@@ -34,7 +34,7 @@ def __init__(self, file_path: FilePathType, verbose: bool = True):
 
         hdf5storage = get_package(package_name="hdf5storage")
 
-        super().__init__(spikes_matfile_path=file_path, verbose=verbose)
+        super().__init__(file_path=file_path, verbose=verbose)
```
CodyCBakerPhD, line 37: Oh, also, issue a deprecation warning on NeuroConv side

h-mayorquin: Good idea. 

Here is the file:
https://www.dropbox.com/s/q4i4dsnk7rqh09h/Peter_MS22_180629_110319_concat.spikes.cellinfo.mat?dl=0

Maybe you can drop the other columns to make it smaller so we can add it to the gin tests?

h-mayorquin: Wait, I did something possibly dumb here. Can you check out the new code and see if this is what you intended? You already had the right input `file_path` and that is only handled internally.

CodyCBakerPhD: Sorry, I mean the converted NWB file from using this interface (possibly on that file, but also maybe other GIN files too)

h-mayorquin: What did you mean about the deprecation though? What should be deprecated?

h-mayorquin:
Not necessary any more after this is done inside spikeinterface now.

PR #450, opened 2023-05-17
improve path expansion tutorial

bendichter:

PR #456, opened 2023-05-23
[Documentation] Add usage examples to path expansion documentation

### LINKED ISSUE #454: [Documentation]: Add examples of path expansion

See comment chain from https://github.com/catalystneuro/neuroconv/pull/450#discussion_r1198100306

Currently assigned to @felixp8 to officially type up and integrate my proposed examples (and I guess try running the code and making sure it works as expected, lol)

No.

### Code of Conduct

felixp8:
fix #454 

Added examples following https://github.com/catalystneuro/neuroconv/pull/450#discussion_r1198100306 as requested in https://github.com/catalystneuro/neuroconv/issues/454. Code examples were tested locally with dummy files and seem to work properly.


felixp8:
Does this look reasonable @CodyCBakerPhD?

docs/user_guide/expand_path.rst
```diff
@@ -71,6 +71,141 @@ If your session start time is present in your data path, you can indicate this f
...
+The Allen Institute's Visual Coding dataset contains, among other data, motion-corrected videos of each
+experimental session, with the directory structure shown below.
+
+.. code-block:: bash
+
+    allen-brain-observatory/
+    ¦   visual-coding-2p/
+    ¦   +-- ophys_movies/
+    ¦   ¦   +-- ophys_experiment_645691416.h5
+    ¦   ¦   +-- ophys_experiment_<session_id>.h5
```
CodyCBakerPhD, line 91: I have mixed feelings about including `<information that gets inserted inside angle braces>`

I used to do it all the time because it makes sense to developers or people familiar with programming

But this doc in particular is for people that may never have coded in their life, and I've personally encountered situations where they interpret the angle braces literally

I'd be fine if the code block here is just an abbreviated view of the file storage pattern using ~3 real file name examples; the f-string aspect becomes clear in the subsequent block

docs/user_guide/expand_path.rst
```diff
@@ -71,6 +71,141 @@ If your session start time is present in your data path, you can indicate this f
...
+    ¦   ¦   +-- ...
+
+To use :py:class:`~neuroconv.tools.path_expansion.LocalPathExpander` to find all ``ophys_movies`` files and
+extract their session IDs, you could use the following code.
+
+.. code-block:: python
+
+    >>> source_data_spec = {
```
CodyCBakerPhD, line 99: We only use the `>>>` directive in code blocks when pairing with doctest (such as the conversion gallery); not saying there shouldn't be tests for this on each of these example structures, but utilizing doctest for it would take a little bit of doing (generating the files under the hood, adding proper imports, setting up global paths, etc)

Probably best left to follow-up PRs if desired, so I'd remove the `>>>` for now

docs/user_guide/expand_path.rst
```diff
@@ -71,6 +71,141 @@ If your session start time is present in your data path, you can indicate this f
...
+
+The Buszaki Lab's SenzaiY dataset contains spiking and LFP data from mouse V1 with the directory structure
+shown below.
+
+.. code-block:: bash
+
+    SenzaiY/
+    ¦   YMV01/
+    ¦   +-- YMV01_170818/
+    ¦   ¦   +-- YMV01_170818.dat
```
CodyCBakerPhD, line 123:
```suggestion
    ¦   ¦   +-- YMV01_170818.eeg
```

I think it's fine to limit the scope of this path expander to the spiking and LFP; the `.dat` files are actually the raw ecephys data and the `.eeg` files are the LFP though

docs/user_guide/expand_path.rst
```diff
@@ -71,6 +71,141 @@ If your session start time is present in your data path, you can indicate this f
...
+the following commands on Linux to generate some dummy files mimicking the structure of the IBL data shown above.
+
+.. code-block:: bash
+
+    mkdir -p steinmetzlab/Subjects/NR_0017/2022-03-22/001/raw_video_data/
+    mkdir -p steinmetzlab/Subjects/NR_0019/2022-04-29/001/raw_video_data/
+    mkdir -p steinmetzlab/Subjects/NR_0019/2022-05-03/001/raw_video_data/
+    touch steinmetzlab/Subjects/NR_0017/2022-03-22/001/raw_video_data/_iblrig_leftCamera.raw.6252a2f0-c10f-4e49-b085-75749ba29c35.mp4
+    touch steinmetzlab/Subjects/NR_0019/2022-04-29/001/raw_video_data/_iblrig_leftCamera.raw.9041b63e-02e2-480e-aaa7-4f6b776a647f.mp4
+    touch steinmetzlab/Subjects/NR_0019/2022-05-03/001/raw_video_data/_iblrig_leftCamera.raw.05576d90-6fb7-4aba-99ae-7ba63cc50a9a.mp4
```
CodyCBakerPhD, line 208: I was thinking more along the lines of a one-line NeuroConv helper that can be called from any platform

Such as

```python
from neuroconv.utils import generate_path_expander_demo_ibl

generate_path_expander_demo_ibl(folder_path="path/to/generate/dummy/files")
```

CodyCBakerPhD:
Overall looks quite good!

Made some minor style-related suggestions for the first example that apply to the rest as well

felixp8:
Thanks! Will work through the suggestions today

docs/user_guide/expand_path.rst
felixp8: @CodyCBakerPhD Just making sure I understand this last bit - are you suggesting to replace the current demo using a `file_path` f-string to find `.dat` files with a demo using a `folder_path` f-string to find/parse the directories themselves instead? (since no additional info is in the file names)

docs/user_guide/expand_path.rst
felixp8: Largely unrelated to this specific suggestion, but looking closer at the IBL data, the hashes in the video file names aren't the same even within the same session (see [here](https://ibl.flatironinstitute.org/public/steinmetzlab/Subjects/NR_0017/2022-03-22/001/raw_video_data/), for example), so I assume they aren't actually session IDs? 

In this case, would it be proper `LocalPathExpander` usage to use empty braces instead? i.e. `/steinmetzlab/Subjects/{subject_id}/{session_start_time:%Y-%m-%d}/001/raw_video_data/_iblrig_leftCamera.raw.{}.mp4`

docs/user_guide/expand_path.rst
CodyCBakerPhD: The `NeuroScopeRecordingInterface` uses a `file_path` convention (to a `.lfp` or `.eeg` file) whereas the `NeuroScopeSortingInterface` uses a `folder_path` convention

The demo code below this stops after running the `PathExpander` to get the resulting source data, but if you wanted to use that source data to initialize interfaces it would follow that pattern

docs/user_guide/expand_path.rst
CodyCBakerPhD: > Largely unrelated to this specific suggestion, but looking closer at the IBL data, the hashes in the video file names aren't the same even within the same session (see [here](https://ibl.flatironinstitute.org/public/steinmetzlab/Subjects/NR_0017/2022-03-22/001/raw_video_data/), for example), so I assume they aren't actually session IDs?

Haha, wow good catch lol

You are absolutely 100% correct; technically speaking IBL has two caveats.

(a) each object has a separate ID, those IDs are then mapped to a session ID (called an `eid` over there) and in the context of code that would go after the `PathExpander` demo of this section you would use a reverse mapping from these object IDs to go back to the `eid`

(b) technically the preferred access method to the content is via `ONE` which only bothers with `eid` (and `pid` for ecephys data) and handles all this under the hood

But here's the thing, we're kind of strapped for nice but complicated public examples of this kind of thing because most of our actual cases are only for privately shared data so for all intents and purposes I'd say it's fine to demo it's usage portraying them as IDs; or if the blank braces actually works in the `PathExpander` code, that's fine too

felixp8:
All suggestions for the documentation should be addressed now!

Not sure the best way to implement the demo generator function. I opted not to scrape their data repository inside the function call since it's slow and I think it'd be better if the function didn't depend on an internet connection. But, I also didn't want to have like several thousand lines of hardcoded raw JSON in a python file, so I only included a small portion of the data - only video files for the first three subjects of `steinmetzlab` data. So, some questions:
- Does the function make sense where it is right now? i.e. in `neuroconv.tools.path_expansion`
- Is the partial hardcoded tree fine or would it be better to have a separate JSON file in the repo with the full IBL tree that the function can read from? (I already have a scraper that can generate that file)
- Is this something that should be noted in `CHANGELOG.md` and have tests added for it?

src/neuroconv/tools/path_expansion.py
```diff
@@ -89,3 +89,275 @@ def list_directory(self, base_directory: DirectoryPath) -> Iterable[FilePath]:
         base_directory = Path(base_directory)
         assert base_directory.is_dir(), f"The specified 'base_directory' ({base_directory}) is not a directory!"
         return (str(path.relative_to(base_directory)) for path in base_directory.rglob("*"))
+
+
+def generate_path_expander_demo_ibl(folder_path: Optional[str] = None) -> None:
```
bendichter, line 94: where is this being used? Is this the best place for this function?

CodyCBakerPhD:
> Is the partial hardcoded tree fine or would it be better to have a separate JSON file in the repo with the full IBL tree that the function can read from? (I already have a scraper that can generate that file)

The JSON approach is way more complicated than I was imagining

All you should need is a single flat list of the fake paths to generate

```python
video_file_paths = [
    demo_directory / "steinmetzlab" / "Subjects" / "NR_0017" / "2022-03-22" / "001" / "raw_video_data" / "_iblrig_bodyCamera.raw.2f88d70a-2172-4635-8ee9-88d9c5193aae.mp4"),
    ...
]

for video_file_path in video_file_paths:
    video_file_path.mkdir(exist_ok=True)  # if .touch() does not recursively create directory structures of a file
    video_file_path.touch()
```

there also shouldn't be a need to recursively use `mkdir`, the `parents` option should do that for you I believe

CodyCBakerPhD:
> Does the function make sense where it is right now? i.e. in neuroconv.tools.path_expansion

I don't mind it where it is right now; it is a path expander specific after all

If we had more demo's for NeuroConv in general, then it should probably be isolated to `neuroconv.demos`

Feel free to move it there if you wish to keep it removed from the rest of the source code

CodyCBakerPhD:
>  I think I'd rather separate the code the implements the actual functionality of path expansion from code that generates demo data. This might go better in src/neuroconv/tools/testing

`testing` could also be another place, but in that case we should probably just go ahead and add an actual test for it in the suite

felixp8:
> The JSON approach is way more complicated than I was imagining
> 
> All you should need is a single flat list of the fake paths to generate

Thanks, this is a lot cleaner! Do you have any thoughts on the extent of files I should be including? All of the Steinmetz data, or just the video files, etc.

CodyCBakerPhD:
> Is this something that should be noted in CHANGELOG.md and have tests added for it?

CHANGELOG modification is optional (the CI is just a reminder) for small changes, especially minor doc improvements.

This PR is on the threshold, but since you are adding some code for the demo and possibly a test for it, might as well update the CHANGELOG anyway

CodyCBakerPhD:
> Do you have any thoughts on the extent of files I should be including? All of the Steinmetz data, or just the video files, etc.

Only the minimal structure for showcasing the example from the docs, so just the video files in the nested directory structure

src/neuroconv/tools/testing/mock_files.py
```diff
@@ -0,0 +1,339 @@
...
+    directory tree for the video files of the Steinmetz Lab's data.
+
+    Parameters
+    ----------
+    folder_path : str, optional
+        Path to folder where the files are to be generated.
+        If None, the current working directory will be used.
+    """
+    if folder_path is None:
+        folder_path = os.getcwd()  # use current directory if no folder provided
```
CodyCBakerPhD, line 19: You can use `Path.cwd()` to keep everything here in `pathlib` land if you want

src/neuroconv/tools/testing/mock_files.py
```diff
@@ -0,0 +1,339 @@
...
+    folder_path = Path(folder_path)
```
CodyCBakerPhD, line 20: I ussually condense lines like tthis into

```python
folder_path = Path(folder_path or Path.cwd())
```

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -98,3 +99,62 @@ def test_expand_paths(tmpdir):
...
+        ),
+    )
+
+    # build expected output with glob and manual parsing
+    expected = []
+    for file_path in base_directory.glob("steinmetzlab/Subjects/*/*/*/raw_video_data/_iblrig_leftCamera.raw.*.mp4"):
+        subject_id = file_path.parts[-5]
+        session_start_time = datetime.strptime(file_path.parts[-4], "%Y-%m-%d")
+        session_id = file_path.parts[-1].split(".")[-2]
+        expected.append(
```
CodyCBakerPhD, line 131: We try to make our expected out come cases in tests as explicit as possible; what I would recommend is to take the returned dictionary from the `PathExpander` output and save it as a `.json`, then have this test load the dictionary back for testing

That way we can look at the full output structure outside of code to make sure it looks as truly expected (not that I have any doubts, just something we're in the habit of doing these days)

docs/user_guide/expand_path.rst
```diff
@@ -71,6 +71,221 @@ If your session start time is present in your data path, you can indicate this f
...
+
+If you would like to experiment locally with :py:class:`~neuroconv.tools.path_expansion.LocalPathExpander`, we
+provide a helper method in ``neuroconv.tools.path_expansion`` that partially replicates the directory structure of
+the IBL data with dummy files on your machine.
+
+.. code-block:: python
+
+    from neuroconv.tools.path_expansion import generate_path_expander_demo_ibl
+
+    generate_path_expander_demo_ibl(folder_path="path/to/generate/dummy/files")
```
CodyCBakerPhD, line 288: This will have to be updated

felixp8: ah good catch, completely forgot about that

felixp8:
Looks like the text file with video paths isn't included in the compiled package. Do I have to add it to `MANIFEST.in` @CodyCBakerPhD? (Also, feedback on the naming/locations of the text and json files is welcome)

CodyCBakerPhD:
> Do I have to add it to MANIFEST.in @CodyCBakerPhD?

Most likely, yes

src/neuroconv/tools/testing/ibl_video_files.txt
CodyCBakerPhD, line 1: This is great; I'd just change the name of the file to indicate they're for a demo of one of our features, not the files themselves to avoid any confusion in the future

Also perhaps not a bad idea to flag as a private entity using the leading '_' convention

That is; '_path_expander_demo_ibl_filepaths.txt'

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -98,3 +100,47 @@ def test_expand_paths(tmpdir):
...
+        ),
+    )
+
+    # build expected output from file
+    expected_file_path = Path(__file__).parent / "expand_paths_ibl_expected.json"
+    with open(expected_file_path, "r") as f:
+        expected = json.load(f)
+    for entry in expected:
+        if "NWBFile" in entry["metadata"]:  # `datetime` is not JSON serializable, so convert from string
+            if "session_start_time" in entry["metadata"]["NWBFile"]:
```
CodyCBakerPhD, line 132: You should be able to handle decoding and encoding of datetimes using the `neuroconv.utils.NWBMetadataEncoder` class instead of having to manually parse it here

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -98,3 +100,47 @@ def test_expand_paths(tmpdir):
...
+
+def test_expand_paths_ibl(tmpdir):
+    expander = LocalPathExpander()
+
+    # set up IBL Steinmetz video file structure
+    generate_path_expander_demo_ibl(tmpdir)
+    base_directory = Path(tmpdir)
+
+    # NOTE: empty brackets used b/c for some reason NR_0021/2022-06-30/ subdir is 002/ instead of 001/
+    out = expander.expand_paths(
```
CodyCBakerPhD, line 113:
```suggestion
    path_expansion_results = expander.expand_paths(
```

Recommend using more verbose/descriptive variable names, especially in tests

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -98,3 +100,47 @@ def test_expand_paths(tmpdir):
...
+                )
+        for source_data in entry["source_data"].values():  # update paths with base_directory
+            if "file_path" in source_data.keys():
+                source_data["file_path"] = str(base_directory / source_data["file_path"])
+            if "folder_path" in source_data.keys():
+                source_data["folder_path"] = str(base_directory / source_data["folder_path"])
+
+    # test results
+    for x in out:
+        assert x in expected
```
CodyCBakerPhD, line 145:
```suggestion
    for path_expansion_result in path_expansion_results:
        assert path_expansion_result  in expected
```

likewise from above

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -98,3 +100,47 @@ def test_expand_paths(tmpdir):
...
+    assert len(out) == len(expected)
```
CodyCBakerPhD, line 146: Although, in all honesty you ought to also be able to just do `assert path_expansion_results == expected_results` or otherwise harness `unittestt.TestCase.assertCountEqual` (which has the best error return message of any of the `unittest` variants)

tests/test_minimal/test_tools/test_expand_paths.py
felixp8: `neuroconv.utils.NWBMetadataEncoder` doesn't have decoding implemented, so I think I can either:
1) keep the manual datetime parsing code and compare the unmodified `expand_paths` output against the parsed JSON
2) encode the output of `expand_paths` using `NWBMetadataEncoder` (with `json.loads(json.dumps(..., cls=NWBMetaDataEncoder))`), so I can just load the target JSON file with no parsing and compare. Technically, this approach won't differentiate if `expand_paths` returns string-formatted session start times instead of `datetime.datetime` objects but that feels like a pretty negligible concern (?)

Thoughts between the two?

CodyCBakerPhD: Yes, option (2) is what we usually go with

CodyCBakerPhD:
@felixp8 Looking great, ready for final review?

bendichter:
@felixp8 I agree, this is really good! One minor thing: there are a few places where you have simply put code snippets where there should be links to the API docs. This is the case for the NeuroScopeSortingInterface class and for the neuroconv.tools.testing module. Could you fix these so they use the rst syntax to point to the appropriate place in the docs?

CodyCBakerPhD:
> Also, for the IBL example, "001" is the session_id

Depending on how accurate we want to be to the true IBL representation, that's the session number on the day

The true session ID would be the `eid`, which would be a hash that we are treating as a one-to-one correspondence with the one attributed to the assets (not exactly accurate either, but truer to the IBL form than the 3-digit number, which would be repeated for each session)

bendichter:
@CodyCBakerPhD I see. From a DANDI perspective, where the session IDs only need to be unique to a subject, it might make more sense to use the session number. That results in much more readable filepaths.

CodyCBakerPhD:
Yeah I suppose since this section is communicating best practices for users, we can use the 3-digit session number as the session ID towards the goal of making easier to read DANDI file paths

As long as that doesn't confuse anyone that goes and looks at the IBL dataset on DANDI and wonders why there is a difference

felixp8:
Thanks! will fix these last few things now

bendichter (approved):

bendichter (approved):

bendichter (approved):

bendichter (approved):

bendichter (approved):

bendichter (approved):

bendichter (approved):

PR #386, opened 2023-03-28
fix return character

bendichter:
fix #385

@simon-ball I am not able to test this, because I can't run this particular interface on an M1 mac and I don't have your file, but would be you able to let me know if this branch fixes your problem?

Issue #385, opened 2023-03-28
[Bug]: ScanImageImagingInterface fails to read any ScanImage2021 metadata


simon-ball:
### What happened?

Attempting to read metadata from a ScanImage generated Tiff fails to extract metadata such as the scanner sampling frequency. Stepping through [the code](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py#L11) for `neuroconv.datainterfaces.ophys.scanimage.scanimageimaginginterface.extract_extra_metadata` indicates that it can't, in fact, read _any_ of the `SI` flagged metadata tags that are written.

As far as I can see, the root cause is as follows:

`extract_extra_metadata` looks like so:
```python
description = ScanImageTiffReader.ScanImageTiffReader(str(file_path)).description(iframe=0)
extra_metadata = {x.split("=")[0]: x.split("=")[1] for x in description.split("\r") if "=" in x}
```

We can step through that with an example file (that I can provide, but it's both massive, and I would rather provide it direct to you than publicly)
```python
description
>>> 'frameNumbers = 1\nacquisitionNumbers = 1\nframeNumberAcquisition = 1\nframeTimestamps_sec = 0.000000000\nacqTriggerTimestamps_sec = -0.000088640\nnextFileMarkerTimestamps_sec = -1.000000000\nendOfAcquisition = 0\nendOfAcquisitionMode = 0\ndcOverVoltage = 0\nepoch = [2023  2  6 10 51 13.374]\nauxTrigger0 = []\nauxTrigger1 = []\nauxTrigger2 = []\nauxTrigger3 = []\nI2CData = {}\n'
```

The metadata extraction fails at the first hurdle, because the string isn't delimited by carriage return, but by newline (this might vary depending on the operating system ScanImage is running on, maybe?). So nearly all of the metadata is dropped, and what isn't dropped, is corrupted. 
```python
extra_metadata
>>> {'frameNumbers ': ' 1\nacquisitionNumbers '}
```

No metadata matching the pattern that the [ScanImageExtractorInterface](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py#L51) searches for  is present:
```python
        if "state.acq.frameRate" in self.image_metadata:
            sampling_frequency = float(self.image_metadata["state.acq.frameRate"])
```

[...]

PR #457, opened 2023-05-24
ScanImageImagingInterface to support modern versions of ScanImage

simon-ball:
Re issue #385 and superseding PR #386

This PR is intended to support Tif files produced by ScanImage which lack metadata keywords looking like an older example (ScanImage 3.8)

It adds the following aspects:
* a new method under `utils.dict` to support fetching a value from a dictionary matching any of several keywords, i.e. where the metadata structure has changed over time
* Changes to the metadata fields read from a tif
* Changes to support cross-platform new line indicators (re #386)


src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
```diff
@@ -65,9 +72,16 @@ def get_metadata(self) -> dict:
 
         metadata = super().get_metadata()
 
-        if "state.internal.triggerTimeString" in self.image_metadata:
-            extracted_session_start_time = dateparse(self.image_metadata["state.internal.triggerTimeString"])
```
CodyCBakerPhD, line 69: Can you re-add the `dateparse` wrapping around the return below? It's causing certain tests to fail in the full CI

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
```diff
@@ -48,9 +53,11 @@ def __init__(
         """
         self.image_metadata = extract_extra_metadata(file_path=file_path)
 
-        if "state.acq.frameRate" in self.image_metadata:
-            sampling_frequency = float(self.image_metadata["state.acq.frameRate"])
-        else:
+        try:
+            sampling_frequency = float(
+                get_metadata_value(self.image_metadata, "state.acq.frameRate", "SI.hRoiManager.scanFrameRate")
+            )
```
CodyCBakerPhD, line 59: Your `util` is certainly an interesting idea

However, I think it might be overkill for this simple of a task

Could we do

```python
if "state.acq.frameRate" in self.image_metadata:
    sampling_frequency = float(self.image_metadata.get("state.acq.frameRate"])
elif "SI.hRoiManager.scanFrameRate" in self.image_metadata:
    sampling_frequency = float(self.image_metadata["SI.hRoiManager.scanFrameRate"])
else:
```

instead? This would make the separate behavior across the cases more explicit when reading the code

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
```diff
@@ -65,9 +72,16 @@ def get_metadata(self) -> dict:
...
+        try:
+            extracted_session_start_time = get_metadata_value(
+                self.image_metadata,
+                "state.internal.triggerTimeString",
+                "epoch",
+            )
             metadata["NWBFile"].update(session_start_time=extracted_session_start_time)
+        except ValueError:
+            # If not present, ignore this aspect
+            pass
```
CodyCBakerPhD, line 84: Similar to above; we try to avoid using try/except clauses unless absolutely necessary

Would suggest something like

```python
if "state.internal.triggerTimeString" in self.image_metadata:
    extracted_session_start_time = dateparse(self.image_metadata["state.internal.triggerTimeString"])
    metadata["NWBFile"].update(session_start_time=extracted_session_start_time)
elif "epoch" in self.image_metadata:
    extracted_session_start_time = dateparse(self.image_metadata["epoch"])  # or whatever else is necessary to coerce/parse the value
    metadata["NWBFile"].update(session_start_time=extracted_session_start_time)
```

CodyCBakerPhD:
I think you mentioned in the issue that this works for you locally on the example files from the GIN dataset

Could you add it to the testing suite as well by adding extra test cases in this section?

https://github.com/catalystneuro/neuroconv/blob/main/tests/test_on_data/test_imaging_interfaces.py#L38-L48

Similar to what I've highlighted, hard-coding the expected metadata return values to show that we're extracting it all properly

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
simon-ball: Ah, fair point, and that does indeed make my utility function less useful

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
simon-ball: Fair enough. I dislike the duplication of the update line, but on the whole, my solution is both more verbose, and less in agreement with the style of the rest of the project

simon-ball: The format stored by ScanImage in more recent versions defeats `dateutils.parser.parse`, unfortunately, but `datetime.datetime.strptime` should do the job unless/until the format changes again

simon-ball:
Noting here: I contacted MBF on a semi-related note (broken link in the documentation), and the response included

> Unfortunately, we no longer support the Python Tiffreader,

I haven't found any official statement on this, and I have no idea how rapidly the Python library will become obsolete, but supporting future versions of ScanImage may require finding an alternative library.

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
```diff
@@ -68,6 +76,14 @@ def get_metadata(self) -> dict:
         if "state.internal.triggerTimeString" in self.image_metadata:
             extracted_session_start_time = dateparse(self.image_metadata["state.internal.triggerTimeString"])
             metadata["NWBFile"].update(session_start_time=extracted_session_start_time)
+        elif "epoch" in self.image_metadata:
+            # Versions of ScanImage at least as recent as 2020, and possibly earlier, store the start time under keyword
+            # `epoch`, as a string encoding of a Matlab array, example `'[2022  8  8 16 56 7.329]'`
+            # dateparse can't cope with this representation, so using strptime directly
```
CodyCBakerPhD, line 82: Thanks for leaving such a richly detailed comment!

CodyCBakerPhD:
> I see that you can express interface_kwargs as a list of dicts to run the tests against multiple examples - but check_extracted_metadata is defined for the whole class, so how to assign a set of hardcoded values, one for each instance, and compel correct choice?

Heh, yeah that's actually something we're going to change in the near future to reduce complexity

While there is a mechanism for how to deal with it in the current structure it's easier to just define a new class without using the list functionality; it should look something like this

Inserted below https://github.com/catalystneuro/neuroconv/blob/main/tests/test_on_data/test_imaging_interfaces.py#L37-L48

```python
@skipIf(platform.machine() == "arm64", "Interface not supported on arm64 architecture")
class TestScanImageImagingInterface(ImagingExtractorInterfaceTestMixin, TestCase):
    data_interface_cls = ScanImageImagingInterface
    interface_kwargs = dict(file_path=str(OPHYS_DATA_PATH / "imaging_datasets" / "path_to_new_test_file"))
    save_directory = OUTPUT_PATH

    def check_extracted_metadata(self, metadata: dict):
        assert metadata["NWBFile"]["session_start_time"] == datetime(2017, 10, 9, 16, 57, 7, 967000)  # update to new value
        assert (
            metadata["Ophys"]["TwoPhotonSeries"][0]["description"]
            == "the hard-coded string of all the grabbed & merged values"
        )
```

Also, since you're planning on using the interface I'll let you know the long-term plan for this is, instead of dumping a big messy description with all these random values, we want to make a proper NWB extension [similar to this one](https://github.com/catalystneuro/ndx-extract) for recording all of the fields as nicely described attributes

CodyCBakerPhD:
Last thing, could you update the CHANGELOG so this can get included in the next release notes?

Thanks a bunch for helping with this!

tests/test_on_data/test_imaging_interfaces.py
```diff
@@ -48,6 +48,24 @@ def check_extracted_metadata(self, metadata: dict):
...
+    # Second class for ScanImageImagingInterface as recommended easier than modifying the first to check metadata in
+    # each of several cases
+    data_interface_cls = ScanImageImagingInterface
+    interface_kwargs = dict(
+        file_path=str(OPHYS_DATA_PATH / "imaging_datasets" / "ScanImage" / "scanimage_20220801_volume.tif")
+    )
+    save_directory = OUTPUT_PATH
+
+    def check_extracted_metadata(self, metadata: dict):
+        assert metadata["NWBFile"]["session_start_time"] == datetime(2022, 8, 8, 16, 56, 7, 329000)
```
CodyCBakerPhD, line 62:
```suggestion
        assert metadata["NWBFile"]["session_start_time"] ==datetime.datetime(2022, 8, 8, 16, 39, 49, 190000)
```

according to the CI the timestamp was a bit off

tests/test_on_data/test_imaging_interfaces.py
```diff
@@ -48,6 +48,24 @@ def check_extracted_metadata(self, metadata: dict):
...
+    # Second class for ScanImageImagingInterface as recommended easier than modifying the first to check metadata in
+    # each of several cases
+    data_interface_cls = ScanImageImagingInterface
+    interface_kwargs = dict(
+        file_path=str(OPHYS_DATA_PATH / "imaging_datasets" / "ScanImage" / "scanimage_20220801_volume.tif")
+    )
+    save_directory = OUTPUT_PATH
+
+    def check_extracted_metadata(self, metadata: dict):
+        assert metadata["NWBFile"]["session_start_time"] ==datetime.datetime(2022, 8, 8, 16, 39, 49, 190000)
```
CodyCBakerPhD:
```suggestion
        assert metadata["NWBFile"]["session_start_time"] == datetime(2022, 8, 8, 16, 39, 49, 190000)
```
oops

according to the CI the timestamp was a bit off

tests/test_on_data/test_imaging_interfaces.py
```diff
@@ -48,6 +48,24 @@ def check_extracted_metadata(self, metadata: dict):
...
+    # Second class for ScanImageImagingInterface as recommended easier than modifying the first to check metadata in
+    # each of several cases
+    data_interface_cls = ScanImageImagingInterface
+    interface_kwargs = dict(
+        file_path=str(OPHYS_DATA_PATH / "imaging_datasets" / "ScanImage" / "scanimage_20220801_volume.tif")
+    )
+    save_directory = OUTPUT_PATH
+
+    def check_extracted_metadata(self, metadata: dict):
+        assert metadata["NWBFile"]["session_start_time"] == datetime.datetime(2022, 8, 8, 16, 39, 49, 190000)
```
CodyCBakerPhD, line 62:
```suggestion
        assert metadata["NWBFile"]["session_start_time"] == datetime(2022, 8, 8, 16, 39, 49, 190000)
```
oops

according to the CI the timestamp was a bit off

tests/test_on_data/test_imaging_interfaces.py
CodyCBakerPhD: @simon-ball Actually after fixing this it seems the extracted metadata also doesn't match expectation

Are you sure you copied/pasted the expected output corresponding to the file `scanimage_20220801_volume.tif`? Or was it one of the others?

simon-ball: That's possible. I have time to review further today

CodyCBakerPhD: Huh, how interesting 😆

Good catch!

CodyCBakerPhD (approved):

CodyCBakerPhD:
@simon-ball Thanks a bunch for all the work on this! LGTM

PR #459, opened 2023-05-26
[Sync VIIa] Improve temporal alignment testing

CodyCBakerPhD:
Here's those changes we talked about adding (still a bit of debugging to do)

The tests for ensuring the results are written correctly to the NWB file will be much easier when #455 is through; only the internal interface mechanisms are tested for now


src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -128,29 +176,64 @@ def check_align_starting_time_internal(self):
...
+
+        events_interface.align_by_interpolation(
+            unaligned_timestamps=unaligned_timestamps, aligned_timestamps=aligned_timestamps
+        )
+
+        aligned_behavior_event_times = events_interface.get_timestamps()
+        expected_aligned_behavior_event_times = [8.830632, 10.530796, 11.930964]
+        assert_array_almost_equal(x=aligned_behavior_event_times, y=expected_aligned_behavior_event_times)
+
+    def check_interface_temporal_alignment(self):
```
h-mayorquin, line 223: As you mentioned I do find this level indirection unecessary (better just to list all the things that you are going to test below) but I think you want to use this as organization for separating the properties that appended to the nwb from the ones that write to the nwb file.That seems fine but I still find the naming `internal` confusing with this idea. Maybe I am not getting it right.

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -128,29 +176,64 @@ def check_align_starting_time_internal(self):
         expected_timestamps = unaligned_timestamps + starting_time
         assert_array_equal(x=aligned_timestamps, y=expected_timestamps)
 
-    def check_align_timestamps_internal(self):
-        """Ensure that internal mechanisms for the timestamps getter/setter work as expected."""
-        unaligned_timestamps = self.interface.get_original_timestamps()
+    def check_interface_align_timestamps_mutation(self):
```
h-mayorquin, line 179: `check_interface_get_original_timestamps_inumability` seems better.

I see why you want to have this test (I guess you want to force people who implement this for a new interface in the future to recreate)  but I still find this strange to talk about the mutability of a the value returned by a function.

Maybe this hint that this should be a property that is read-only so this contract is enshrined in code?

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -116,10 +153,21 @@ def check_get_timestamps(self):
...
 
-    def check_align_starting_time_internal(self):
-        """Ensure that internal mechanisms for shifting timestamps given a starting time match expectation."""
+    def check_interface_align_timestamps(self):
+        """Ensure that internal mechanisms for the timestamps getter/setter work as expected."""
         fresh_interface = self.data_interface_cls(**self.test_kwargs)
-        unaligned_timestamps = fresh_interface.get_timestamps()
+        unaligned_timestamps = fresh_interface.get_original_timestamps()
+
+        aligned_timestamps = unaligned_timestamps + 1.23 + np.random.random(size=unaligned_timestamps.shape)
```
h-mayorquin, line 161: I think we should make all the uses of random deterministic in tests. We should use np.random.default_rng` which is the recommendation from numpy:

https://numpy.org/doc/stable/reference/random/generator.html

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -128,29 +176,64 @@ def check_align_starting_time_internal(self):
...
+        fresh_interface.align_timestamps(aligned_timestamps=aligned_timestamps)
 
-        retrieved_aligned_timestamps = self.interface.get_timestamps()
-        assert_array_equal(x=retrieved_aligned_timestamps, y=aligned_timestamps)
+        post_alignment_original_timestamps = fresh_interface.get_original_timestamps()
+        assert_array_equal(x=post_alignment_original_timestamps, y=pre_alignment_original_timestamps)
 
-    def check_align_starting_time_external(self):
-        pass  # TODO: generalize
+    def check_interface_interpolation(self):
```
h-mayorquin, line 190: I don't think this should be here. 

This right now is testing that the tested interface can be used as a source of timestamps for a third interface that you create here on the fly (i.e. `MockBehaviorEventInterface`).  Given that all the other methods used before are already tested somewhere else this seems redundant. 

The interface that you want to test `MockBehaviorEventInterface` is inicidental to this. This should be a test of its own I feel and does not need to be in the general mechanism just yet.

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -286,8 +341,29 @@ def check_align_timestamps_internal(self):
...
+        if fresh_interface._number_of_segments == 1:
+            aligned_timestamps = fresh_interface.get_timestamps()
+            expected_timestamps = all_unaligned_timestamps + starting_time
+            assert_array_equal(x=aligned_timestamps, y=expected_timestamps)
+        else:
+            all_aligned_timestamps = fresh_interface.get_timestamps()
+            all_expected_timestamps = [
+                unaligned_timestamps + starting_time for unaligned_timestamps in all_unaligned_timestamps
+            ]
+            [
```
h-mayorquin, line 360: How does having assertions inside list comprehensions plays with the error trace? 

I feel that this is a bad idea from a debugging standpoint as I would not be able to stop in the specific case that failed if that comes to happen.

h-mayorquin (commented):
I think this is a good improvement. I added some comments.

The more I think about this the more I feel that  `align` all over the place is not helpful for understanding. When I am looking at the tests and reasoning about what they do I need to remain myself that they are just setters and getters and are not doing anything else than to modify a -implicit?- property of the interfaces: the timestamps.  

I think align just gets into the way as nothing else other than getting and setting is taking place when using thsee methods. The name of this here is a good example I think:

`check_interface_align_starting_time_shifts_timestamps` -> `check_shift_timestamps_by_start_time` 

The original naming is hard to understand and convolutes and what it does but I think the second is quite clear.

What do you think?

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -96,17 +96,54 @@ def check_extracted_metadata(self, metadata: dict):
...
+            with self.subTest(str(num)):
+                self.case = num
+                self.test_kwargs = kwargs
+                self.interface = self.data_interface_cls(**self.test_kwargs)
+                self.check_metadata_schema_valid()
+                self.check_conversion_options_schema_valid()
+                self.check_metadata()
+                self.nwbfile_path = str(self.save_directory / f"{self.data_interface_cls.__name__}_{num}.nwb")
+                self.run_conversion(nwbfile_path=self.nwbfile_path)
+                self.check_read_nwb(nwbfile_path=self.nwbfile_path)
```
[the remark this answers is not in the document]
CodyCBakerPhD: The diff is right, it's just not very smart by GitHub's normal standards

The '-' part of this '+' got pushed down to roughly https://github.com/catalystneuro/neuroconv/pull/459/files#diff-4433d8d81063d5e6dd6a5cc301ce1e46f565bbf7f0dcf90b2953c9bab9b6dbc6L162-L167

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -96,17 +96,54 @@ def check_extracted_metadata(self, metadata: dict):
...
+    Generic class for testing temporal alignment methods.
+
+    This mixin must be paired with a DataInterfaceMixin and unittest.TestCase class.
+
+    Several of these tests are required to be run in a specific order. In this case,
+    there is a `test_conversion_as_lone_interface` that calls the `check` functions in
+    the appropriate order, after the `interface` has been created. Normally, you might
+    expect the `interface` to be simply created in the `setUp` method, but this class
+    allows you to specify multiple interface_kwargs.
+    """
```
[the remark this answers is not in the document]
CodyCBakerPhD: Oh, that's residual from the copy/paste of the main DataInterfaceMixin which I didn't alter in any way

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -96,17 +96,54 @@ def check_extracted_metadata(self, metadata: dict):
...
+
+    def check_interface_get_original_timestamps(self):
         """
         Just to ensure each interface can call .get_original_timestamps() without an error raising.
 
         Also, that it always returns non-empty.
         """
-        timestamps = self.interface.get_original_timestamps()
+        original_timestamps = self.interface.get_original_timestamps()
```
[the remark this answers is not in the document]
CodyCBakerPhD: If not available, then you're just generating `np.arange` with gaps `1/sampling_frequency`, which is not format specific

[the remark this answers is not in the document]
CodyCBakerPhD: Open to other names though, just not convinced that 'format' is any more accurate

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Sure thing, good call

Will adjust that in other tests too

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Where is `internal` used anymore? I thought I removed it

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Error gets thrown at same place and time as a normal for loop

Not sure what you mean about stopping; I assume that's some kind of IDE injection feature?

Still not sure why it was structured that way, no harm in making it a normal for loop

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Oh I see it in the docstring

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Oh I see; the point was for this to test the interpolation method for this interface, but will have to think about what that would mean more generally

CodyCBakerPhD: which is hard to do for arbitrary interfaces now that I think about it... especially higher frequency recordings, or format specific timestamps

Just removing it in that case

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: It is indeed supposed to be read-only, but how can you control that in a programmatic way for a newly created interface?

src/neuroconv/tools/testing/data_interface_mixins.py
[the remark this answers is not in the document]
CodyCBakerPhD: Fixed

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Fixed

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Fixed

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Removed

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Fixed

src/neuroconv/tools/testing/data_interface_mixins.py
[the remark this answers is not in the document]
h-mayorquin: In what cases is not available and why we would need to generate timestamps in those cases?

src/neuroconv/tools/testing/data_interface_mixins.py
h-mayorquin: I mean normal standard library debugger (pdb) tracepoint setting like:
```
    pdb.set_trace()  # This is where you set the debugger breakpoint
```

How do you debug when you need to jump to a certain line and inspect values?

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -241,7 +279,30 @@ def check_read_nwb(self, nwbfile_path: str):
                 if recording.has_scaled_traces() and nwb_recording.has_scaled_traces():
                     check_recordings_equal(RX1=recording, RX2=nwb_recording, return_scaled=True)
 
-    def check_align_starting_time_internal(self):
+    def check_interface_align_timestamps(self):
+        fresh_interface = self.data_interface_cls(**self.test_kwargs)
+        if fresh_interface._number_of_segments == 1:
+            unaligned_timestamps = self.interface.get_timestamps()
+            aligned_timestamps = unaligned_timestamps + 1.23 + np.random.random(size=unaligned_timestamps.shape)
```
h-mayorquin, line 286: non-deterministic, also same question, how comes this is not inherited from the mixing?

h-mayorquin (commented):
This is a big change, I need to check it once more.

src/neuroconv/tools/testing/data_interface_mixins.py
h-mayorquin: Yeah, I know how to do it. But I agree that it is better to leave it for another PR.

src/neuroconv/tools/testing/data_interface_mixins.py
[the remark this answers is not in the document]
h-mayorquin: When no available my argument is that `format_timestamps` should be None to make this clear. And any logic that requires them should be a combination of warnings, on-the-flygeneration of assertion depending on the context where they are required.

I think that's better than having an `original_timestamps` that has a different behavior depending on the format. Easier to reason about it. 

But maybe I am missing something?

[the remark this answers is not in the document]
h-mayorquin: Can you point out to the places where the fallback logic is implemented? I think this will be helpful for me to understand and think better.

src/neuroconv/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
```diff
@@ -127,6 +129,16 @@ def get_metadata(self) -> dict:
             metadata["NWBFile"]["session_start_time"] = session_start_time
         return metadata
 
+    def get_original_timestamps(self) -> np.ndarray:
```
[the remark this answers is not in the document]
CodyCBakerPhD: 3 debugs were necessary to interfaces that alias SI/neo from the NeuroConv `source_data` `__init__` signature

Specifically https://github.com/catalystneuro/neuroconv/pull/459/files#diff-5939435a8408da1ec7aba7f1b54b9eb49747778cc8376f1bac76e9c6fadc06b8R133 differs from the usual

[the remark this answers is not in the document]
CodyCBakerPhD: Is there a better approach to take overall? Probably but would take a lot more work to refactor every interface to utilize so just doing this spot fix for now

tests/test_on_data/test_behavior_interfaces.py
```diff
@@ -91,76 +84,88 @@ def check_read_nwb(self, nwbfile_path: str):
...
 
-    def check_align_starting_time_internal(self):
-        pass  # disabled since this interface follows a different strategy
+    def check_interface_align_timestamps(self):
+        all_unaligned_timestamps = self.interface.get_original_timestamps()
 
-    def check_align_starting_time(self):
-        fresh_interface = self.data_interface_cls(**self.test_kwargs)
+        aligned_timestamps = [
+            unaligned_timestamps + 1.23 + np.random.random(size=unaligned_timestamps.shape)
```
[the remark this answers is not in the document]
CodyCBakerPhD: Good catch, will fix

tests/test_on_data/test_recording_interfaces.py
```diff
@@ -93,23 +93,36 @@ class TestEDFRecordingInterface(RecordingExtractorInterfaceTestMixin, TestCase):
...
+    def test_interface_alignment(self):
+        interface_kwargs = self.interface_kwargs
+        if isinstance(interface_kwargs, dict):
+            interface_kwargs = [interface_kwargs]
+        for num, kwargs in enumerate(interface_kwargs):
+            with self.subTest(str(num)):
+                self.case = num
+                self.test_kwargs = kwargs
+
+                # TODO - debug hanging I/O from pyedflib
```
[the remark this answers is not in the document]
CodyCBakerPhD: Most likely

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Fixed

tests/test_on_data/test_behavior_interfaces.py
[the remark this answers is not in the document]
CodyCBakerPhD: Fixed

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -135,11 +135,19 @@ def align_timestamps(self, aligned_timestamps: Union[np.ndarray, List[np.ndarray
...
+        if self.sorting_extractor.has_recording():
+            if self._number_of_segments == 1:
+                self.align_timestamps(aligned_timestamps=self.get_timestamps() + starting_time)
             else:
-                sorting_segment._t_start += starting_time
+                self.align_timestamps(
+                    aligned_timestamps=[timestamps + starting_time for timestamps in self.get_timestamps()]
+                )
+        else:
+            for sorting_segment in self.sorting_extractor._sorting_segments:
```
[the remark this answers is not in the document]
CodyCBakerPhD: As discussed in meeting, will try to convince SI team to prevent the bad case of mixture of `None` segment start times in multi-segment sorting

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -116,10 +153,21 @@ def check_get_timestamps(self):
...
+
+        aligned_timestamps = unaligned_timestamps + 1.23 + np.random.random(size=unaligned_timestamps.shape)
+        fresh_interface.align_timestamps(aligned_timestamps=aligned_timestamps)
+
+        retrieved_aligned_timestamps = fresh_interface.get_timestamps()
+        assert_array_equal(x=retrieved_aligned_timestamps, y=aligned_timestamps)
+
+    def check_interface_align_starting_time_shifts_timestamps(self):
+        """Ensure that internal mechanisms for shifting timestamps by a starting time work as expected."""
+        fresh_interface = self.data_interface_cls(**self.test_kwargs)
```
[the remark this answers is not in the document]
h-mayorquin: Thanks. So it is going to be an attribute of the resulting composite class that is a list of dict. Can we add it as typed attribute to be overiden in the mixing classess as a documentation? The way that we add the attributes to be overiden in `BaseExtractorInterface`.

src/neuroconv/datainterfaces/ecephys/neuroscope/neuroscopedatainterface.py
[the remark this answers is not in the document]
h-mayorquin: I see. Fair enough. Exactly, I was wondering what exactly break and why did the general method logic did not work. In this case, do I understand well that is the variety of initialization signatures require specialized logic? 

If that's the case I think this makes sense. Let's have something that passess all the tests and then refactor to generalize later if possible. Can you add a small comment or docstring just mentioning this. Something like: "necessary to overide method because of different siganture for this extractor in spikeinterface` just so people know what we are overiden the base method here.

tests/test_on_data/test_behavior_interfaces.py
```diff
@@ -91,76 +84,89 @@ def check_read_nwb(self, nwbfile_path: str):
...
             assert f"Video: video_{video_type}" in nwbfile.acquisition
 
-    def check_align_starting_time_internal(self):
-        pass  # disabled since this interface follows a different strategy
+    def check_interface_align_timestamps(self):
+        all_unaligned_timestamps = self.interface.get_original_timestamps()
 
-    def check_align_starting_time(self):
-        fresh_interface = self.data_interface_cls(**self.test_kwargs)
+        generator = np.random.default_rng(seed=0)
```
h-mayorquin, line 90:
```suggestion
        rng = np.random.default_rng(seed=0)
```
This is the standard in numpy:
https://numpy.org/doc/stable/reference/random/generator.html

I would prefer the full random_number_generator but I def think it is better than generator.

tests/test_on_data/test_behavior_interfaces.py
```diff
@@ -91,76 +84,89 @@ def check_read_nwb(self, nwbfile_path: str):
...
+        aligned_timestamps = [
+            unaligned_timestamps + 1.23 + generator.random(size=unaligned_timestamps.shape)
```
h-mayorquin, line 92:
```suggestion
            unaligned_timestamps + 1.23 + rng.random(size=unaligned_timestamps.shape)
```

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -96,83 +96,121 @@ def check_extracted_metadata(self, metadata: dict):
...
-    def check_align_starting_time_internal(self):
-        """Ensure that internal mechanisms for shifting timestamps given a starting time match expectation."""
-        fresh_interface = self.data_interface_cls(**self.test_kwargs)
-        unaligned_timestamps = fresh_interface.get_timestamps()
+    def check_interface_align_timestamps(self):
+        """Ensure that internal mechanisms for the timestamps getter/setter work as expected."""
+        self.setUpFreshInterface()
+        unaligned_timestamps = self.interface.get_timestamps()
+
+        generator = np.random.default_rng(seed=0)
```
h-mayorquin, line 161:
```suggestion
        rng = np.random.default_rng(seed=0)
```

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -96,83 +96,121 @@ def check_extracted_metadata(self, metadata: dict):
...
+        aligned_timestamps = unaligned_timestamps + 1.23 + generator.random(size=unaligned_timestamps.shape)
```
h-mayorquin, line 162:
```suggestion
        aligned_timestamps = unaligned_timestamps + 1.23 + rng.random(size=unaligned_timestamps.shape)
```

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -241,7 +279,33 @@ def check_read_nwb(self, nwbfile_path: str):
                 if recording.has_scaled_traces() and nwb_recording.has_scaled_traces():
                     check_recordings_equal(RX1=recording, RX2=nwb_recording, return_scaled=True)
 
-    def check_align_starting_time_internal(self):
+    def check_interface_align_timestamps(self):
+        fresh_interface = self.data_interface_cls(**self.test_kwargs)
+        generator = np.random.default_rng(seed=0)
```
h-mayorquin, line 284:
```suggestion
        rng = np.random.default_rng(seed=0)
```

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -241,7 +279,33 @@ def check_read_nwb(self, nwbfile_path: str):
...
+        if fresh_interface._number_of_segments == 1:
+            unaligned_timestamps = self.interface.get_timestamps()
+
+            aligned_timestamps = unaligned_timestamps + 1.23 + generator.random(size=unaligned_timestamps.shape)
```
h-mayorquin, line 288:
```suggestion
            aligned_timestamps = unaligned_timestamps + 1.23 + rng.random(size=unaligned_timestamps.shape)
```

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -241,7 +279,33 @@ def check_read_nwb(self, nwbfile_path: str):
...
+            self.interface.align_timestamps(aligned_timestamps=aligned_timestamps)
+
+            retrieved_aligned_timestamps = self.interface.get_timestamps()
+            assert_array_equal(x=retrieved_aligned_timestamps, y=aligned_timestamps)
+        else:
+            all_unaligned_timestamps = self.interface.get_timestamps()
+
+            all_aligned_timestamps = [
+                unaligned_timestamps + 1.23 + generator.random(size=unaligned_timestamps.shape)
```
h-mayorquin, line 297:
```suggestion
                unaligned_timestamps + 1.23 + rng.random(size=unaligned_timestamps.shape)
```

h-mayorquin (approved):
This looks good to me. Some final comments but this is up to you. Just my opinions of what would be better:
1) I think it will be nice to add the interface_kwargs as typed attributes in the mixing classess so they work like documentantion as in the `BaseExtractor`.
2) Changed the randon number generator nam to the numpy convention in their documentation
3) Add clarifying notes as comments or docstrings when general methods have to be overridden so we know in the future why it was the case if we ever work on refactoring them.

tests/test_on_data/test_behavior_interfaces.py
CodyCBakerPhD: went with `random_number_generator`

CodyCBakerPhD:
1. Added typing
2. expanded all names to `random_number_generator`
3. added notes for SI aliasing

PR #460, opened 2023-05-26
improve docs for testing on data

bendichter:


h-mayorquin (approved):
This looks good to me. For spikeinterface I wrote more detailed instructions for how to install datalad in different OS:

https://spikeinterface.readthedocs.io/en/latest/development/development.html#installing-datalad

I wonder if you would like something like this here.

CodyCBakerPhD:
> I wonder if you would like something like this here.

@h-mayorquin That, or a link to there from here, assuming the DataLad website itself does not have such instructions (which would be the ideal link to point to)

PR #463, opened 2023-05-30
[Sync VIIb] Break up multi-segment alignment commands

CodyCBakerPhD:
@h-mayorquin My sketch for reducing magic for multi-segment alignment; thoughts?


h-mayorquin (commented):
I think this is good. Having a different method for the segmented case vs the single segment case seems like a good thing. The most important to me is that there is a method that works the same in the single segment interfaces and in the segmentless interfaces which I think is going to be the most common case. 

Another option that can be combined with this or in opposition to this is doing what spikeinterface does:

https://github.com/h-mayorquin/spikeinterface/blob/b9a3cca3fce268fbb01ad58ba1264d826677b430/src/spikeinterface/core/baserecording.py#L129

And make the method apply on the first segment for the single segment case when no segment_index is provided.

What do you think?

CodyCBakerPhD:
> The most important to me is that there is a method that works the same in the single segment interfaces and in the segmentless interfaces which I think is going to be the most common case.

AFAIK, via `get_num_segments`, an extractor always has at least 1 segment, or an integer strictly greater than that; unsure about 'segmentless', wouldn't that just be a number of segments equal to 1?


> And make the method apply on the first segment for the single segment case when no segment_index is provided.

To be clear, 

`align_timestamps` takes in just a single array and currently asserts that it can only be used in the case of a single segment, where it behaves like any other interface that doesn't have multiple segments

`align_segment_timestamps` takes a list of arrays and checks that the length matches the number of segments; nothing is stopping someone from having a list of a single element, but the goal of this PR is to avoid magic like being able to pass just a single array in case you tried using that method on a single segment case

Is including that kind of magic to `align_segment_timestamps` what you're suggesting?

h-mayorquin:
No, the only thing that I was suggestion is that maybe the method `align_timestamps` could take a segment_index like in spikeinterface. Or, even going further, implement that and drop the `align_segment_timestamps` as that can be donde by using the segment_index. But that was more of a suggestion just to  hear your thoughts.

My main comment is that I am very happy with this PR because the method `align_timestamps` will work the same for interfaces that don't have the notion of segments and for interfaces that have the notion but only have one segment. I think this is good as I think this is the common case and it is good to have a consistent API for that.

CodyCBakerPhD:
> No, the only thing that I was suggestion is that maybe the method align_timestamps could take a segment_index like in spikeinterface. Or, even going further, implement that and drop the align_segment_timestamps as that can be donde by using the segment_index. But that was more of a suggestion just to hear your thoughts.

Ahhh, I see now.

To avoid the problem of incomplete alignment across segments, I would not want to allow mix-and-match of segment indexing outside of specifying the full list for all segments (which acts as an explicit confirmation on behalf of the user)

It would also involve more complicated mechanisms on our side of the code for managing that through the alignment function(s)

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -174,18 +174,22 @@ def align_timestamps(self, aligned_timestamps: Union[np.ndarray, List[np.ndarray
...
-            ), f"The number of timestamp vectors ({len(aligned_timestamps)}) does not match the number of segments ({self._number_of_segments})!"
+        assert (
+            self._number_of_segments == 1
+        ), "This recording has multiple segments; please use 'align_segment_timestamps' instead."
 
-            for segment_index in range(self._number_of_segments):
-                self.recording_extractor.set_times(times=aligned_timestamps[segment_index], segment_index=segment_index)
+        self.recording_extractor.set_times(times=aligned_timestamps)
+
+    def align_segment_timestamps(self, aligned_timestamps: List[np.ndarray]):
```
CodyCBakerPhD, line 183: @h-mayorquin One last thing, what do you prefer between these two options

(a) keyword arguments are like shown here; `aligned_timestamps` for consistency with `align_timestamps` signature

(b) add 'segment' to the keyword name, as `aligned_segment_timestamps` for extra specificity

Same argument would be made about the `segment_starting_times` in the other function

CodyCBakerPhD:
@h-mayorquin OK I adjusted the name as well as improved the docstrings since the core methods can now just inherit those outright

Also extended these changes to the sorting interface, which works nearly identically due to the attachment of recording objects

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -6,6 +6,7 @@
 from typing import List, Optional, Type, Union
 
 import numpy as np
+from hdmf.testing import TestCase as HDMFTestCase
```
h-mayorquin, line 9:
```suggestion
```

You are not using ths I think.

h-mayorquin (commented):
This looks good to me. My biggest gripe before the release is still the use of the name align for the setter methods in our code.

Let me explain my concern by first reiterating your arguments to see that I am representing your view fairly: you claim that align for setters is a good name because the only way to align timestamps is to actually set them.

I agree with your claim that this might be the only way to set them, but I still think that this is a poor reason to name the method align. That is, I think the logic `align_timestamps(aligned_timestamps)`, is misleading. Let me illustrate how this looks in other contexts.

For example, in an auto repair class, you might have a method, `repair_motor()`, that installs a repaired motor into the car. Then you repair the motor and use the method `repair_motor(repaired_motor)`. This is misleading because the motor was repaired somewhere else; the function is just setting it.

In a restaurant, you could have a method named `cook_meal` that sets any meal on the table. The meal is cooked somewhere else and then you use the `method cook_meal(cooked_meal)` to implement this. This is misleading because the method is not cooking the meal; this was done somewhere else and the method is only setting it.

Now, do I care about this just because I really like logic? No, I think that in the future and for certain simple cases we can actually have a function that fully will deserve the name `align_timestamps` that gets the timestamps from the right place and use your setters to set the aligned timestamps. If we keep the current names then we overload the concept in advance for no good reason as far as I can see.

What do you think? Am I missing something?

h-mayorquin (approved):
I meant to approve this but clicked the wrong button.

Two minor comments about an unused method and something that I think will improve readability.

A larger comments that is probably out of the scope of this PR.

CodyCBakerPhD:
> A larger comments that is probably out of the scope of this PR.

Yeah, that's why I moved the discussion to the now re-opened https://github.com/catalystneuro/neuroconv/issues/301

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Ah, good catch

I had meant to use `assertRaisesWith` (the reason it's `hdmf` specific) as a class method like many `unittest.TestCase` do but for some reason it only operates as an instance method

I've added an `isinstance` assertion and added docstring note about it

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -134,20 +137,79 @@ def align_timestamps(self, aligned_timestamps: Union[np.ndarray, List[np.ndarray
...
+            else:
+                self.align_segment_timestamps(
+                    aligned_segment_timestamps=[
+                        segment_timestamps + segment_starting_time
+                        for segment_timestamps, segment_starting_time in zip(
+                            self.get_timestamps(), segment_starting_times
+                        )
+                    ]
+                )
+        else:
```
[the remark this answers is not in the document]
CodyCBakerPhD: Fair enough, fixed

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Yeah, I remember; I really prefer the HDMF implementation so that I don't have to think about how to cast everything into regex

Also makes the test case assertions easier to read

PR #464, opened 2023-05-30
[Improvement] Better video segment timing default

CodyCBakerPhD:
@h-mayorquin As per our discussion earlier, I think this should work for what you suggest

Will need some tests to be sure though


h-mayorquin (commented):
This makes sense to me but shouldn't the start_frames only be calculated when using an external file?

I think I don't understand that argument well now that I think about it.

h-mayorquin:
All right. I will put in my to-do, maybe I will finish some stuff before you do.

h-mayorquin:
Closing this, will come back to this after #1251 is merged.

PR #466, opened 2023-06-01
Align recording extractor properties between Neuroscope raw and LFP

h-mayorquin:
There are two functions that extract proeprties for the recording extractor on Neuroscope.

1. `subset_shank_channels` which takes a subset of the channels that are anatomically relevant according to the docstring.
2. `add_recording_extractor_properties` which uses the properties of the XML to add other properties such as shank electrode number and group memebership.

The problem right now is that they are in different order in `NeuroScopeLFPInterface` and in `NeuroScopeRecordingInterface` which leads to them having inconsistent properties when both of them are added in a conversion.

This PR fixes this by having them perform exactly the same operation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 * Added the `BaseTemporalAlignmentInterface` to serve as the new base class for all new temporal alignment methods. [PR #442](https://github.com/catalystneuro/neuroconv/pull/442)
 * Added direct imports for all base classes from the outer level; you may now call `from neuroconv import BaseDataInterface, BaseTemporalAlignmentInterface, BaseExtractorInterface`. [PR #442](https://github.com/catalystneuro/neuroconv/pull/442)
 * Added basic temporal alignment methods to the AudioInterface. `align_starting_time` is split into `align_starting_times` (list of times, one per audio file) and `align_global_starting_time` (shift all by a scalar amount). `align_by_interpolation` and other timestamp-based approaches is not yet implemented for this interface. [PR #402](https://github.com/catalystneuro/neuroconv/pull/402)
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD (approved):

CodyCBakerPhD:
@h-mayorquin I am noticing in the shared files that for NeuroScope the default chanels names are things like `ch20grp0`, which indicates the out-of-order channel naming from the amp order we would expect, but the addition of the `grp{shank_id}` seems unnecessary for the NWB electrode table since each `channel_name` also occurs with a `group_name`/`group` (ID)

Assuming this is just the default behavior for that property on the new SI extractor, right? What would you think about improving it on the NeuroConv interface side?

h-mayorquin:
@CodyCBakerPhD 
I don't understand this sentence:
```
 which indicates the out-of-order channel naming from the amp order we would expect, 
 ```
 
I think this will be quicker to discuss in our meeting today though.

h-mayorquin:
@CodyCBakerPhD 
> as I recall, these IDs are the order in which they are plugged into the amp,

This will be very useful information but I am not sure neo is aware of that:

https://github.com/NeuralEnsemble/python-neo/blob/4d4eb85e459c7e35cc1d43ce423fd58907fb3002/neo/rawio/neuroscoperawio.py#L82-L86

They are just reading the xml in a very similar way that we do here:
https://github.com/catalystneuro/neuroconv/blob/a212354ccb433112d5271fb3a25cbdf7c72e2033/src/neuroconv/datainterfaces/ecephys/neuroscope/neuroscope_utils.py#L62-L78

CodyCBakerPhD:
If they're reading the xml in the same way (following the iteration order from lines of the XML) then yes it's fine

I'm saying the `ch{channel_id}` part looks fine, I just don't want the `grp{group_d}` to be a part of the channel name

PR #476, opened 2023-06-14
remove deep_dict_update from make_nwbfile_from_metadata()

### LINKED ISSUE #483: [Bug]: Can't use make_or_load_nwbfile when metadata has array of objects
### What happened?

Using `make_or_load_nwbfile()` when metadata has an array of objects and these objects do not contain the key `name`, raises an error. This limitation is caused by the usage of `deep_dict_update()` down the line, within `make_nwbfile_from_metadata()`.

### Steps to Reproduce

```python
from neuroconv.tools.nwb_helpers import make_or_load_nwbfile
from datetime import datetime
from dateutil import tz

metadata = {
    "NWBFile": {
        "session_description": "session_description",
        "identifier": "identifier",
        "session_start_time": datetime(2018, 4, 25, 2, 30, 3, tzinfo=tz.gettz("US/Pacific")),
    },
    "FiberPhotometry": {
        "EcxitationsSourcesTable": {
            "description": "description",
            "items": [
                {"peak_wavelength": 475.0, "source_type": "LED"},
                {"peak_wavelength": 560.0, "source_type": "LED"},
            ]
        }
    }
}
    
with make_or_load_nwbfile(nwbfile_path="file.nwb", metadata=metadata, overwrite=True) as nwbfile:
    print(nwbfile)
```

### Traceback

```python
KeyError                                  Traceback (most recent call last)
/home/luiz/storage2/Github/shuler-lab-to-nwb/examples/fiber_photometry.ipynb Cell 17 in 2
      3 from dateutil import tz
      5 metadata = {
      6     "NWBFile": {
      7         "session_description": "session_description",
   (...)
     19     }
     20 }
---> 22 with make_or_load_nwbfile(nwbfile_path="file.nwb", metadata=metadata, overwrite=True) as nwbfile:
     23     print(nwbfile)

File /media/luiz/storage2/conda_environments/env_shuler/lib/python3.8/contextlib.py:113, in _GeneratorContextManager.__enter__(self)
    111 del self.args, self.kwds, self.func
    112 try:
--> 113     return next(self.gen)
    114 except StopIteration:
    115     raise RuntimeError("generator didn't yield") from None

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/tools/nwb_helpers.py:173, in make_or_load_nwbfile(nwbfile_path, nwbfile, metadata, overwrite, verbose)
    171 except Exception as e:
    172     success = False
--> 173     raise e
    174 finally:
    175     if nwbfile_path_in:

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/tools/nwb_helpers.py:169, in make_or_load_nwbfile(nwbfile_path, nwbfile, metadata, overwrite, verbose)
    167         nwbfile = io.read()
    168     elif nwbfile is None:
--> 169         nwbfile = make_nwbfile_from_metadata(metadata=metadata)
    170     yield nwbfile
    171 except Exception as e:

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/tools/nwb_helpers.py:50, in make_nwbfile_from_metadata(metadata)
     48 def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     49     """Make NWBFile from available metadata."""
---> 50     metadata = dict_deep_update(get_default_nwbfile_metadata(), metadata)
     51     nwbfile_kwargs = metadata["NWBFile"]
     52     if "Subject" in metadata:
     53         # convert ISO 8601 string to datetime

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/dict.py:189, in dict_deep_update(d, u, append_list, remove_repeats, copy, compare_key, list_dict_deep_update)
    187     sub_dict_to_update = dict_to_update.get(key_to_update, dict())
    188     sub_dict_with_update_values = update_values
--> 189     dict_to_update[key_to_update] = dict_deep_update(
    190         sub_dict_to_update, sub_dict_with_update_values, append_list=append_list, remove_repeats=remove_repeats
    191     )
    192 # Update with list calls the append_replace_dict_in_list function
    193 elif append_list and isinstance(update_values, list):

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/dict.py:189, in dict_deep_update(d, u, append_list, remove_repeats, copy, compare_key, list_dict_deep_update)
    187     sub_dict_to_update = dict_to_update.get(key_to_update, dict())
    188     sub_dict_with_update_values = update_values
--> 189     dict_to_update[key_to_update] = dict_deep_update(
    190         sub_dict_to_update, sub_dict_with_update_values, append_list=append_list, remove_repeats=remove_repeats
    191     )
    192 # Update with list calls the append_replace_dict_in_list function
    193 elif append_list and isinstance(update_values, list):

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/dict.py:196, in dict_deep_update(d, u, append_list, remove_repeats, copy, compare_key, list_dict_deep_update)
    194     for value in update_values:
    195         dict_or_list_of_dicts = dict_to_update.get(key_to_update, [])
--> 196         dict_to_update[key_to_update] = append_replace_dict_in_list(
    197             dict_or_list_of_dicts, value, compare_key, list_dict_deep_update, remove_repeats
    198         )
    199 # Update with something else
    200 else:
    201     dict_to_update[key_to_update] = update_values

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/dict.py:94, in append_replace_dict_in_list(ls, d, compare_key, list_dict_deep_update, remove_repeats)
     91     return d
     92 if isinstance(d, collections.abc.Mapping):
     93     indxs = np.where(
---> 94         [d.get(compare_key, None) == i[compare_key] for i in ls if isinstance(i, collections.abc.Mapping)]
     95     )[0]
     96     if len(indxs) > 0:
     97         for idx in indxs:

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/dict.py:94, in (.0)
     91     return d
     92 if isinstance(d, collections.abc.Mapping):
     93     indxs = np.where(
---> 94         [d.get(compare_key, None) == i[compare_key] for i in ls if isinstance(i, collections.abc.Mapping)]
     95     )[0]
     96     if len(indxs) > 0:
     97         for idx in indxs:

KeyError: 'name'
```

### Operating System

Linux

### Python Executable

Conda

### Python Version

3.9

### Package Versions

_No response_

### Code of Conduct

luiztauffer:
Fix #483

### DIFF
--- a/src/neuroconv/tools/nwb_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers.py
@@ -11,7 +11,7 @@ from pynwb import NWBHDF5IO, NWBFile
 from pynwb.file import Subject
 
 from ..utils import FilePathType, dict_deep_update
-from ..utils.dict import DeepDict
+from ..utils.dict import DeepDict, load_dict_from_file
 from ..utils.json_schema import validate_metadata
 
 
@@ -53,9 +53,7 @@ def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
 
     # Validate metadata
     schema_path = Path(__file__).resolve().parent.parent / "schemas/base_metadata_schema.json"
-    with open(schema_path) as schema_file:
-        base_metadata_schema = json.load(schema_file)
-
+    base_metadata_schema = load_dict_from_file(file_path=schema_path)
     validate_metadata(metadata=metadata, schema=base_metadata_schema)
 
     nwbfile_kwargs = metadata["NWBFile"]
--- a/src/neuroconv/tools/nwb_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers.py
@@ -59,7 +59,8 @@ def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
         # jsonschema.validate(metadata, base_metadata_schema)
         validate_metadata(metadata=metadata, schema=base_metadata_schema)
     except jsonschema.exceptions.ValidationError as e:
-        raise Exception(f"Base metadata is not valid:\n{e}")
+        print(f"Base metadata is not valid.\n")
+        raise e
 
     nwbfile_kwargs = metadata["NWBFile"]
     # convert ISO 8601 string to datetime
--- a/src/neuroconv/tools/nwb_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers.py
@@ -47,10 +47,11 @@ def get_default_nwbfile_metadata() -> DeepDict:
 
 def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     """Make NWBFile from available metadata."""
-    keys_to_keep = ["NWBFile", "Subject"]
-    metadata = {key: metadata[key] for key in keys_to_keep if key in metadata}
-    metadata = dict_deep_update(get_default_nwbfile_metadata(), metadata)
     nwbfile_kwargs = metadata["NWBFile"]
+    if "session_description" not in nwbfile_kwargs:
+        nwbfile_kwargs.update(session_description="No description.")
+    if "identifier" not in nwbfile_kwargs:
+        nwbfile_kwargs.update(identifier=uuid.uuid4())
     if "Subject" in metadata:
         # convert ISO 8601 string to datetime
         if "date_of_birth" in metadata["Subject"] and isinstance(metadata["Subject"]["date_of_birth"], str):
--- a/src/neuroconv/tools/nwb_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers.py
@@ -55,12 +55,8 @@ def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     schema_path = Path(__file__).resolve().parent.parent / "schemas/base_metadata_schema.json"
     with open(schema_path) as schema_file:
         base_metadata_schema = json.load(schema_file)
-    try:
-        # jsonschema.validate(metadata, base_metadata_schema)
-        validate_metadata(metadata=metadata, schema=base_metadata_schema)
-    except jsonschema.exceptions.ValidationError as e:
-        print(f"Base metadata is not valid.\n")
-        raise e
+
+    validate_metadata(metadata=metadata, schema=base_metadata_schema)
 
     nwbfile_kwargs = metadata["NWBFile"]
     # convert ISO 8601 string to datetime
--- a/src/neuroconv/tools/nwb_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers.py
@@ -45,9 +45,11 @@ def get_default_nwbfile_metadata() -> DeepDict:
     return metadata
 
 
-def make_nwbfile_from_metadata(metadata: dict, metadata_update_kwargs: Optional[dict] = None) -> NWBFile:
+def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     """Make NWBFile from available metadata."""
-    metadata = dict_deep_update(get_default_nwbfile_metadata(), metadata, **(metadata_update_kwargs or {}))
+    keys_to_keep = ["NWBFile", "Subject"]
+    metadata = {key: metadata[key] for key in keys_to_keep if key in metadata}
+    metadata = dict_deep_update(get_default_nwbfile_metadata(), metadata)
     nwbfile_kwargs = metadata["NWBFile"]
     if "Subject" in metadata:
         # convert ISO 8601 string to datetime
@@ -122,7 +124,6 @@ def make_or_load_nwbfile(
     metadata: Optional[dict] = None,
     overwrite: bool = False,
     verbose: bool = True,
-    metadata_update_kwargs: Optional[dict] = None,
 ):
     """
     Context for automatically handling decision of write vs. append for writing an NWBFile.
@@ -142,8 +143,6 @@ def make_or_load_nwbfile(
     verbose: bool, optional
         If 'nwbfile_path' is specified, informs user after a successful write operation.
         The default is True.
-    metadata_update_kwargs: dict, optional
-        Keyword arguments for the dict_deep_update() function. Only used if creating a new NWBFile.
     """
     nwbfile_path_in = Path(nwbfile_path) if nwbfile_path else None
     assert not (nwbfile_path is None and nwbfile is None and metadata is None), (
@@ -169,7 +168,7 @@ def make_or_load_nwbfile(
         if load_kwargs.get("mode", "") == "r+":
             nwbfile = io.read()
         elif nwbfile is None:
-            nwbfile = make_nwbfile_from_metadata(metadata=metadata, metadata_update_kwargs=metadata_update_kwargs)
+            nwbfile = make_nwbfile_from_metadata(metadata=metadata)
         yield nwbfile
     except Exception as e:
         success = False
--- a/src/neuroconv/tools/nwb_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers.py
@@ -1,12 +1,12 @@
+import json
 import uuid
 from contextlib import contextmanager
 from datetime import datetime
 from pathlib import Path
 from typing import Optional
 from warnings import warn
-import jsonschema
-import json
 
+import jsonschema
 from pynwb import NWBHDF5IO, NWBFile
 from pynwb.file import Subject
 
@@ -72,7 +72,9 @@ def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
         nwbfile_kwargs["subject"] = metadata["Subject"]
         # convert ISO 8601 string to datetime
         if "date_of_birth" in nwbfile_kwargs["subject"] and isinstance(nwbfile_kwargs["subject"]["date_of_birth"], str):
-            nwbfile_kwargs["subject"]["date_of_birth"] = datetime.fromisoformat(nwbfile_kwargs["subject"]["date_of_birth"])
+            nwbfile_kwargs["subject"]["date_of_birth"] = datetime.fromisoformat(
+                nwbfile_kwargs["subject"]["date_of_birth"]
+            )
         nwbfile_kwargs["subject"] = Subject(**nwbfile_kwargs["Subject"])
 
     return NWBFile(**nwbfile_kwargs)
--- a/src/neuroconv/tools/nwb_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers.py
@@ -4,6 +4,8 @@ from datetime import datetime
 from pathlib import Path
 from typing import Optional
 from warnings import warn
+import jsonschema
+import json
 
 from pynwb import NWBHDF5IO, NWBFile
 from pynwb.file import Subject
@@ -47,23 +49,32 @@ def get_default_nwbfile_metadata() -> DeepDict:
 
 def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     """Make NWBFile from available metadata."""
-    nwbfile_kwargs = metadata.get("NWBFile") or dict()
-    if "session_description" not in nwbfile_kwargs:
-        nwbfile_kwargs.update(session_description="No description.")
-    if "identifier" not in nwbfile_kwargs:
-        nwbfile_kwargs.update(identifier=uuid.uuid4())
-    if "Subject" in metadata:
-        # convert ISO 8601 string to datetime
-        if "date_of_birth" in metadata["Subject"] and isinstance(metadata["Subject"]["date_of_birth"], str):
-            metadata["Subject"]["date_of_birth"] = datetime.fromisoformat(metadata["Subject"]["date_of_birth"])
-        nwbfile_kwargs.update(subject=Subject(**metadata["Subject"]))
+
+    # Validate metadata
+    schema_path = Path(__file__).resolve().parent.parent / "schemas/base_metadata_schema.json"
+    with open(schema_path) as schema_file:
+        base_metadata_schema = json.load(schema_file)
+    try:
+        jsonschema.validate(metadata, base_metadata_schema)
+    except jsonschema.exceptions.ValidationError as e:
+        print("Base metadata is not valid.")
+        print(e)
+
+    nwbfile_kwargs = metadata["NWBFile"]
     # convert ISO 8601 string to datetime
-    assert "session_start_time" in nwbfile_kwargs, (
-        "'session_start_time' was not found in metadata['NWBFile']! Please add the correct start time of the "
-        "session in ISO8601 format (%Y-%m-%dT%H:%M:%S) to this key of the metadata."
-    )
-    if isinstance(nwbfile_kwargs.get("session_start_time", None), str):
-        nwbfile_kwargs["session_start_time"] = datetime.fromisoformat(metadata["NWBFile"]["session_start_time"])
+    if isinstance(nwbfile_kwargs.get("session_start_time"), str):
+        nwbfile_kwargs["session_start_time"] = datetime.fromisoformat(nwbfile_kwargs["session_start_time"])
+    if "session_description" not in nwbfile_kwargs:
+        nwbfile_kwargs["session_description"] = "No description."
+    if "identifier" not in nwbfile_kwargs:
+        nwbfile_kwargs["identifier"] = uuid.uuid4()
+    if "Subject" in metadata:
+        nwbfile_kwargs["subject"] = metadata["Subject"]
+        # convert ISO 8601 string to datetime
+        if "date_of_birth" in nwbfile_kwargs["subject"] and isinstance(nwbfile_kwargs["subject"]["date_of_birth"], str):
+            nwbfile_kwargs["subject"]["date_of_birth"] = datetime.fromisoformat(nwbfile_kwargs["subject"]["date_of_birth"])
+        nwbfile_kwargs["subject"] = Subject(**nwbfile_kwargs["Subject"])
+
     return NWBFile(**nwbfile_kwargs)
 
 
--- a/src/neuroconv/tools/nwb_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers.py
@@ -57,8 +57,7 @@ def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     try:
         jsonschema.validate(metadata, base_metadata_schema)
     except jsonschema.exceptions.ValidationError as e:
-        print("Base metadata is not valid.")
-        print(e)
+        raise(f"Base metadata is not valid:\n{e}")
 
     nwbfile_kwargs = metadata["NWBFile"]
     # convert ISO 8601 string to datetime
--- a/src/neuroconv/tools/nwb_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers.py
@@ -57,7 +57,7 @@ def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     try:
         jsonschema.validate(metadata, base_metadata_schema)
     except jsonschema.exceptions.ValidationError as e:
-        raise (f"Base metadata is not valid:\n{e}")
+        raise Exception(f"Base metadata is not valid:\n{e}")
 
     nwbfile_kwargs = metadata["NWBFile"]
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD:
Given that its usage in the `make_nwbfile_from_metadata` function really only concerns the outer, flat `NWBFile` level, what specific aspects of the `deep_dict_update` function do you need to make use of there? Most of them related to conflicts, array interactions, and deeper handling but I'd assume that's beyond what is required for that simple helper

Usually the encouraged workflow is to do more complicated metadata merging like this in between `get_metadata` and `run_conversion`, which also helps keep the functions themselves more modular

CodyCBakerPhD:
> which will in some cases cause errors like the one I found (metadata with lists of objects) because it will use deep_dict_update with this full metadata.

Can you elaborate? This is the part I'm interested in

As a matter of workflow, `run_conversion` ought only accept the full complete metadata `DeepDict` and should not modify it much beyond that (defaults ought to have been retrieved during the initial `get_metadata` call)

The `deep_dict_update` call inside `make_nwbfile_from_metadata` was probably done just for convenience so if it is the source of your issue, it can definitely be refactored in some way to make the operation safer on both ends (fewer modifications to in place structure, and safer guarantee of defaulted metadata making it into initial nwbfile object)

luiztauffer:
@CodyCBakerPhD 

> Can you elaborate? This is the part I'm interested in

more specifically, because the default value in `deep_dict_update()` is `"compare_key": "name"`, for array items without this key, it will break. I found it in the example metadata for fiber photometry for shuler’s lab: https://github.com/catalystneuro/shuler-lab-to-nwb/blob/modularize-data-interfaces/shuler_lab_to_nwb/fiberphotometry_datainterface/metadata.yaml


>  As a matter of workflow, run_conversion ought only accept the full complete metadata DeepDict and should not modify it much (if at all?) beyond that (defaults ought to have been retrieved during the initial get_metadata call)

I agree


> The deep_dict_update call inside make_nwbfile_from_metadata was probably done just for convenience

exactly, so probably that change on item n 2 over there could do the trick

luiztauffer:
@CodyCBakerPhD 

If the metadata passed at this stage down here is [this one](https://github.com/catalystneuro/shuler-lab-to-nwb/blob/modularize-data-interfaces/shuler_lab_to_nwb/fiberphotometry_datainterface/metadata.yaml) it breaks because it will try to find a key `name` in objects which do not have it
```python
def run_conversion(...):
        metadata = self.get_metadata()        
        with make_or_load_nwbfile(
            nwbfile_path=nwbfile_path, 
            nwbfile=nwbfile, 
            metadata=metadata, 
            overwrite=overwrite, 
            verbose=self.verbose,
        ) as nwbfile_out:
        # conversion code...
```

Error track:
```
...
File [/media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/dict.py:196](https://file+.vscode-resource.vscode-cdn.net/media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/dict.py:196), in dict_deep_update(d, u, append_list, remove_repeats, copy, compare_key, list_dict_deep_update)
    194     for value in update_values:
    195         dict_or_list_of_dicts = dict_to_update.get(key_to_update, [])
--> 196         dict_to_update[key_to_update] = append_replace_dict_in_list(
    197             dict_or_list_of_dicts, value, compare_key, list_dict_deep_update, remove_repeats
    198         )
    199 # Update with something else
    200 else:
    201     dict_to_update[key_to_update] = update_values

File [/media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/dict.py:94](https://file+.vscode-resource.vscode-cdn.net/media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/dict.py:94), in append_replace_dict_in_list(ls, d, compare_key, list_dict_deep_update, remove_repeats)
     91     return d
     92 if isinstance(d, collections.abc.Mapping):
     93     indxs = np.where(
---> 94         [d.get(compare_key, None) == i[compare_key] for i in ls if isinstance(i, collections.abc.Mapping)]
     95     )[0]
     96     if len(indxs) > 0:
     97         for idx in indxs:

File [/media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/dict.py:94](https://file+.vscode-resource.vscode-cdn.net/media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/dict.py:94), in (.0)
     91     return d
     92 if isinstance(d, collections.abc.Mapping):
     93     indxs = np.where(
---> 94         [d.get(compare_key, None) == i[compare_key] for i in ls if isinstance(i, collections.abc.Mapping)]
     95     )[0]
     96     if len(indxs) > 0:
     97         for idx in indxs:

KeyError: 'name'
```

You can reproduce it with [this](https://github.com/catalystneuro/shuler-lab-to-nwb/blob/modularize-data-interfaces/examples/fiber_photometry.ipynb), using the fiber photometry data shared by Shuler lab

luiztauffer:
The expected behavior is for `make_or_load_nwbfile()` to accept a full metadata and to not break if said metadata contains arrays of objects that do not contain the key `name`

luiztauffer:
@CodyCBakerPhD sure,
I already updated the changes here with the solution n 2
new args are not needed anymore, just a simple filter of which metadata keys should be handled by the `make_nwbfile_from_metadata()` function
This fixes the bug

luiztauffer:
>  and the other is that deep_dict_update doesn't work as expected on your structure without overriding that field

this is not really a bug, it works as expected. `deep_dict_update` already has the arg `compare_key` that can be used in that case

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -47,6 +47,8 @@ def get_default_nwbfile_metadata() -> DeepDict:
 
 def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     """Make NWBFile from available metadata."""
+    keys_to_keep = ["NWBFile", "Subject"]
+    metadata = {key: metadata[key] for key in keys_to_keep if key in metadata}
     metadata = dict_deep_update(get_default_nwbfile_metadata(), metadata)
     nwbfile_kwargs = metadata["NWBFile"]
```
CodyCBakerPhD, line 53:
```suggestion
    nwbfile_kwargs = metadata["NWBFile"]
    if "session_description" not in nwbfile_kwargs:
        nwbfile_kwargs.update(session_description="No description.")
    if "identifier" not in nwbfile_kwargs:
        nwbfile_kwargs.update(identifier=uuid.uuid4())
```

Though, y'know, looking deeper at it all it's really doing is adding the UUID and description (only two fields...); maybe we could just do direct operations on it like so, which would make the function more readable for the future and avoids in-place modifications to the `metadata` object (not a fun experience for most people to magically modify arguments passed into functions)

Either that or assert that those fields are in the metadata before creating the file - your choice - but I think we can avoid deep updating altogether towards the goal of simplifying what should have already been a very simple function

CodyCBakerPhD:
> this is not really a bug, it works as expected.

I disagree; that argument and block is only for comparing a list of objects where each object has the `name`+other fields definition structure that 'sometimes' but not always is present, and it should not be erroring if the schematic definition of that object fails to follow that pattern (really it shouldn't be triggering there at all in that case)

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -47,6 +47,8 @@ def get_default_nwbfile_metadata() -> DeepDict:
 
 def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
```
CodyCBakerPhD, line 48: Also can you add some tests illustrating whatever change we decide to go with?

CodyCBakerPhD (changes requested):

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -47,8 +47,11 @@ def get_default_nwbfile_metadata() -> DeepDict:
...
     """Make NWBFile from available metadata."""
-    metadata = dict_deep_update(get_default_nwbfile_metadata(), metadata)
     nwbfile_kwargs = metadata["NWBFile"]
+    if "session_description" not in nwbfile_kwargs:
+        nwbfile_kwargs.update(session_description="No description.")
+    if "identifier" not in nwbfile_kwargs:
+        nwbfile_kwargs.update(identifier=uuid.uuid4())
```
CodyCBakerPhD, line 54: Of course, the other exception here as raised in the tests is that not every received metadata structure is guaranteed to have the `NWBFile` outer key at all - can you add safe retrieval of that?

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -47,8 +47,11 @@ def get_default_nwbfile_metadata() -> DeepDict:
 
 def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     """Make NWBFile from available metadata."""
-    metadata = dict_deep_update(get_default_nwbfile_metadata(), metadata)
     nwbfile_kwargs = metadata["NWBFile"]
```
CodyCBakerPhD, line 50:
```suggestion
    nwbfile_kwargs = metadata.get("NWBFile") or dict()
```

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -47,8 +47,11 @@ def get_default_nwbfile_metadata() -> DeepDict:
...
 def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     """Make NWBFile from available metadata."""
-    metadata = dict_deep_update(get_default_nwbfile_metadata(), metadata)
-    nwbfile_kwargs = metadata["NWBFile"]
+    nwbfile_kwargs = metadata.get("NWBFile") or dict()
+    if "session_description" not in nwbfile_kwargs:
+        nwbfile_kwargs.update(session_description="No description.")
+    if "identifier" not in nwbfile_kwargs:
+        nwbfile_kwargs.update(identifier=uuid.uuid4())
     if "Subject" in metadata:
```
bendichter, line 55: This will fail because `session_start_time` is missing. I don't think we should accept a metadata that does not contain an "NWBFile" key. This is defined in the schema for the metadata. "NWBFile" is a required key and "session_start_time" is required to be defined within it. The alternative is to use a filler session_start_time, but I would rather require it. See https://github.com/catalystneuro/neuroconv/blob/3474d05ab066db309fdf47c226666d1becec8235/src/neuroconv/schemas/base_metadata_schema.json#L8-L21

CodyCBakerPhD: We agreed a while back to never to use default session start times (technically we used to use `datetime(1970,1,1)` and that was not good so we stopped doing it)

However, the change in question here is the same both before and after this update; `get_default_nwbfile_metadata()` **does not** yield a metadata structure with the `session_start_time` field - it MUST be contained in the passed dictionary

What is true is that if we hit the second condition of `"NWBFile"` not being in `metadata` at all, that is true... So I suppose we should make two assertions here

a) `assert "NWBFile" in metadata`

and

(b) `assert "session_start_time" in metadata["NWBFile"]`

before doing any further argument extraction/auto-population + file creation

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -47,8 +47,11 @@ def get_default_nwbfile_metadata() -> DeepDict:
 
 def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     """Make NWBFile from available metadata."""
-    metadata = dict_deep_update(get_default_nwbfile_metadata(), metadata)
-    nwbfile_kwargs = metadata["NWBFile"]
+    nwbfile_kwargs = metadata.get("NWBFile") or dict()
```
CodyCBakerPhD, line 50:
```suggestion
    assert "NWBFile" in metadata and "session_start_time" in metadata["NWBFile"], "The metadata dictionary must contain, at a minimum, a 'session_start_time' in metadata['NWBFile']."
    nwbfile_kwargs = metadata["NWBFile"]
```

CodyCBakerPhD: OR do the ultimate best thing, which would be full metadata JSON validation of either a (i) passed schema, or (ii) just the  `NWBFile` subset of the schema

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -47,20 +50,35 @@ def get_default_nwbfile_metadata() -> DeepDict:
...
+    # Validate metadata
+    schema_path = Path(__file__).resolve().parent.parent / "schemas/base_metadata_schema.json"
+    with open(schema_path) as schema_file:
+        base_metadata_schema = json.load(schema_file)
+    try:
+        # jsonschema.validate(metadata, base_metadata_schema)
+        validate_metadata(metadata=metadata, schema=base_metadata_schema)
+    except jsonschema.exceptions.ValidationError as e:
+        print(f"Base metadata is not valid.\n")
+        raise e
```
CodyCBakerPhD, line 63: Point 1: other people on the NeuroConv team feel strongly against default verbosity (which is now `False` BTW); need to either add control of that `print` statement or, probably better, just remove it and let the error do the talking

Point 2: Why have a try/except here at all? What is it avoiding? Especially since a non-ValidationError would get raised as normal and the only different is the print statement which should probably just be removed as per Point 1

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -47,20 +50,35 @@ def get_default_nwbfile_metadata() -> DeepDict:
 
 def make_nwbfile_from_metadata(metadata: dict) -> NWBFile:
     """Make NWBFile from available metadata."""
-    metadata = dict_deep_update(get_default_nwbfile_metadata(), metadata)
+
+    # Validate metadata
+    schema_path = Path(__file__).resolve().parent.parent / "schemas/base_metadata_schema.json"
+    with open(schema_path) as schema_file:
+        base_metadata_schema = json.load(schema_file)
```
CodyCBakerPhD, line 57: We have a `util` that would take less lines and I think it might use the deocder implicitly; I think it's named `load_dict_from_file`?

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -47,20 +50,35 @@ def get_default_nwbfile_metadata() -> DeepDict:
...
+        # jsonschema.validate(metadata, base_metadata_schema)
+        validate_metadata(metadata=metadata, schema=base_metadata_schema)
+    except jsonschema.exceptions.ValidationError as e:
+        print(f"Base metadata is not valid.\n")
+        raise e
+
     nwbfile_kwargs = metadata["NWBFile"]
+    # convert ISO 8601 string to datetime
+    if isinstance(nwbfile_kwargs.get("session_start_time"), str):
+        nwbfile_kwargs["session_start_time"] = datetime.fromisoformat(nwbfile_kwargs["session_start_time"])
```
CodyCBakerPhD, line 68: Remind me, PyNWB insists on it being a `datetime` object, right?

src/neuroconv/tools/nwb_helpers.py
```diff
@@ -47,20 +50,35 @@ def get_default_nwbfile_metadata() -> DeepDict:
...
+    if "session_description" not in nwbfile_kwargs:
+        nwbfile_kwargs["session_description"] = "No description."
```
CodyCBakerPhD, line 70: Now this one is interesting, especially insofar as it interacts with the base metadata schema (which is also now being used in the forms by the GUIDE)

Just to clarify...

a) Does the NWB Schema require this field at the top most level for all NWB files?

b) As I recall, the PyNWB `NWBFile` initializer does require it, regardless of (a), can you confirm?

c) But in NeuroConv, we view this field as optional, and so we fill it with a placeholder. Shouldn't we then include a `default` value in the NeuroConv metadata schema itself? Or, we don't want the GUIDE forms to auto-populate placeholders... Even though our forms then do not mark it as a true requirement even though it's needed to ultimately create the file in PyNWB...

tests/test_minimal/test_tools/test_nwb_helpers.py
```diff
@@ -25,12 +26,10 @@ def test_get_module(self):
...
-                "session in ISO8601 format (%Y-%m-%dT%H:%M:%S) to this key of the metadata."
-            ),
-        ):
+    def test_make_nwbfile_from_metadata_empty(self):
+        with self.assertRaises(ValidationError):
             make_nwbfile_from_metadata(metadata=dict())
+
+    def test_make_nwbfile_from_metadata_session_start_time(self):
+        with self.assertRaises(ValidationError):
+            make_nwbfile_from_metadata(metadata=dict(NWBFile=dict(session_description="Mouse exploring an open field")))
```
CodyCBakerPhD, line 35: To reflect key differences that are now being made in this PR, can you add a test that ensures the `metadata` dict being passed is not modified in-place (i.e., the `session_description` placeholder and `uuid` are not actually set back to the dictionary)? And add a docstring note to that test explaining the context and desired behavior?

tests/test_minimal/test_tools/test_nwb_helpers.py
```diff
@@ -25,12 +26,10 @@ def test_get_module(self):
...
-            exc_type=AssertionError,
-            exc_msg=(
-                "'session_start_time' was not found in metadata['NWBFile']! Please add the correct start time of the "
-                "session in ISO8601 format (%Y-%m-%dT%H:%M:%S) to this key of the metadata."
-            ),
-        ):
+    def test_make_nwbfile_from_metadata_empty(self):
+        with self.assertRaises(ValidationError):
             make_nwbfile_from_metadata(metadata=dict())
+
```
CodyCBakerPhD, line 32: Thanks, these tests are good at showing what kinds of basic `metadata` structures are now prohibited from being used to make a file

src/neuroconv/utils/json_schema.py
```diff
@@ -306,3 +307,14 @@ def get_metadata_schema_for_icephys():
...
 
     return schema
+
+
+def validate_metadata(metadata: Dict[str, dict], schema: Dict[str, dict], verbose: bool = False):
+    """Validate metadata against a schema."""
+    encoder = NWBMetaDataEncoder()
+    # The encoder produces a serialiazed object so we de serialized it for comparison
+    serialized_metadata = encoder.encode(metadata)
+    decoded_metadata = json.loads(serialized_metadata)
```
CodyCBakerPhD, line 317: Just out of curiosity, can you not just specify the class as one of the inputs to `json.loads`?

src/neuroconv/tools/nwb_helpers.py
luiztauffer: Out of the 3 NWB required fields (session_description, identifier and session_start_time) we only really enforce `sessions_start_time`. I see the inconsistency you’re pointing out. I agree with keep operating like this. 
Neuroconv (and the schema/guide) aims at facilitating conversion to NWB, and one important aspect of it is auto-filling as much as possible. We auto-fill session_start_time when it’s present on input files, but the other two fields are not possible to auto-extract like that. I guess one could argue for the importance of user-filled `session_description`, but I personally don’t see it for `identifier`, machines are better at doing unique identifiers.
In that sense, neuroconv schema could deviate from nwb schema in specific ways which would improve the user experience, imo.

Now, if we’re auto-filling required fields, we should do this explicitly. IMO, the schema should have a default value AND be required, and that should be reflected in the Forms: placeholder values and a red asterisk +  validation on form submission.

src/neuroconv/tools/nwb_helpers.py
luiztauffer: yes, it does
https://pynwb.readthedocs.io/en/stable/pynwb.file.html#pynwb.file.NWBFile

src/neuroconv/tools/nwb_helpers.py
luiztauffer: fair point, I removed it

src/neuroconv/tools/nwb_helpers.py
luiztauffer: done

CodyCBakerPhD (approved):

CodyCBakerPhD:
@luiztauffer Thanks for the hard work on this; we'll shift the discussion of `session_description` requirements/defaulting to a higher level and get this through as a critical refactor

PR #480, opened 2023-06-16
remove default placeholder for subject description

CodyCBakerPhD:
partial fix for https://github.com/NeurodataWithoutBorders/nwb-guide/issues/229

The GUIDE inherits that aspect all the way from the NeuroConv base schema

### DIFF
--- a/src/neuroconv/schemas/base_metadata_schema.json
+++ b/src/neuroconv/schemas/base_metadata_schema.json
@@ -142,7 +142,6 @@
             "U",
             "O"
           ],
-          "default": ""
         },
         "species": {
           "type": "string",
--- a/src/neuroconv/schemas/base_metadata_schema.json
+++ b/src/neuroconv/schemas/base_metadata_schema.json
@@ -142,7 +142,7 @@
             "U",
             "O"
           ],
-          "default": "U"
+          "default": ""
         },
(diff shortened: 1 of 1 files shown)


src/neuroconv/schemas/base_metadata_schema.json
```diff
@@ -143,7 +142,7 @@
             "U",
             "O"
           ],
-          "default": "U"
+          "default": ""
```
CodyCBakerPhD, line 145: Also including a fix for the default placeholder for `sex`, but I'm not entirely sure if the form will complain about a `default` value that is not in the `enum`

@garrettmflynn

garrettmflynn: At least for the form, this is equivalent to no default value. The final results object will have no "sex" key when an empty string is registered in the input.

Is there a reason you need a default value defined at all?

bendichter, line 145:
```suggestion
```

CodyCBakerPhD: No, removing it is the goal - I just wasn't sure what the forms expect

bendichter (approved):

PR #485, opened 2023-06-19
Add docstrings for filtering out non-neural channel functions in Neuroscope Interface

h-mayorquin:
I was discussing with @CodyCBakerPhD today about the logic of these functions as I wanted to have more explicit docstrings to indicate what they do. I have been working on a conversion that uses Neuroscope and thinking on how to best add metadata. I think this is useful to me and future developers that will need to read these functions to understand what they do.


CodyCBakerPhD (approved):

CodyCBakerPhD:
More docs and better naming is never a bad thing 👍

PR #486, opened 2023-06-19
pin upper bound on EDF lib numpy

CodyCBakerPhD:
CI has been failing on Python 3.11 over the weekend, seeing if this fixes it

Specifically it's an import problem through numpy caused by EDF lib


h-mayorquin (approved):
This is making the tests pass, seems good to me.

CodyCBakerPhD:
Issue already opened on pyedflib, with PR fix already in progress: https://github.com/holgern/pyedflib/issues/218

So this is a temporary thing here and their team has already been alerted

PR #488, opened 2023-06-20
Add `CellExplorerRecordingInterface` for adding electrode metadata

h-mayorquin:
I was discussing with @CodyCBakerPhD today on how to do this best. Our initial plan was to add this as metadata of the cell explorer interface but when I started working on this I realized that it will be more complicated than I thought. It seems easier to add this interface which we might expand in the future but right now can be used for adding the electrode metadata.

A detailed description of how this works in provided in the docstring.


h-mayorquin:
OK, turns out that adding the full interface was way less work than I expected:

 #492

PR #491, opened 2023-06-21
`CellExplorerSortingInterface` to get sampling frequency from new CellExplorer format

h-mayorquin:
As mentioned before, in the new CellExplorer format the sampling frequency is the in the same file where the spikes are. I made the changes in spikeinterface to take this into account https://github.com/SpikeInterface/spikeinterface/pull/1628 but we won't have a release in a while.

Meanwhile, we will need this temporary change to the interface so we can use it in the current conversions.

PR #492, opened 2023-06-21
Add `CellExplorerRecordingInterface` 

h-mayorquin:
OK, after working on https://github.com/catalystneuro/neuroconv/pull/488 I realized that we have already a Binary recorder that can be used directly to load the binary files. This, coupled with the information in the new `session.mat` of CellExplorer is all we need to build this extractor.

This should work for both adding the metadata and also the raw data and lfp for which we usually use the Neuroscope interface so we don't need to duplicate interfaces:

image

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
@@ -182,31 +182,48 @@ class CellExplorerRecordingInterface(BaseRecordingExtractorInterface):
 
     https://cellexplorer.org/
 
-    CellExplorer's new format contains a `basename.session.mat` file,
-    (where basename is isually a session identifier) containing the
-    following fields:
+    Parameters
+    ----------
+    folder_path : Path or str
+        The folder where the session data is located. It should contain a
+        `{folder.name}.session.mat` file and the binary files `{folder.name}.dat`
+        or `{folder.name}.lfp` for the LFP interface.
+    verbose : bool, default: True
+            Whether to output verbose text.
+    es_key : str, default: "ElectricalSeries" and "ElectricalSeriesLFP" for the LFP interface
+
+    Notes
+    -----
+    CellExplorer's new format contains a `basename.session.mat` file containing
+    rich metadata about the session. basename is the name of the session
+    folder / directory and works as a session identifier.
+
+    Link to the documentation detailing the file's structure:
+    https://cellexplorer.org/datastructure/data-structure-and-format/#session-metadata
+
+    Specifically, we can use the following fields from `basename.session.mat`
+    to create a recording extractor using `BinaryRecordingExtractor` from
+    spikeinterface:
 
     * Sampling frequency
     * Gains
     * Dtype
 
-    Link to the documentation detailing the file's structure:
-    https://cellexplorer.org/datastructure/data-structure-and-format/#session-metadata
+    Where the binary file is named `basename.dat` for the raw data and
+    `basename.lfp` for lfp data.
 
-    If the binary file is available (basename.dat or basename.lfp),
-    the interface will use the `BinaryRecordingExtractor` object from spikeinterface
-    to load the data.
-
-    The metadata for this electrode is also extracted from the `basename.session.mat`
+    The metadata for the channels is extracted from the `basename.session.mat`
     file. The logic of the extraction is described on the function:
 
     `add_channel_metadata_to_recorder_from_session_file`.
 
-    Note that, while all the new data produced by CellExplorer should have a `session.mat` file,  it is not clear
-    if all the channel metadata is always available.
+    Note that, while all the new data produced by CellExplorer should have a
+    `session.mat` file,  it is not clear if all the channel metadata is always
+    available.
 
-    Besides, the current implementation also supports extracting channel metadata from the `chanMap.mat` file used by
-    Kilosort. The logic of the extraction is described on the function:
+    Besides, the current implementation also supports extracting channel metadata
+    from the `chanMap.mat` file used by Kilosort. The logic of the extraction is
+    described on the function:
 
     `add_channel_metadata_to_recorder_from_channel_map_file`.
 
@@ -220,7 +237,7 @@ class CellExplorerRecordingInterface(BaseRecordingExtractorInterface):
     Detailed information can be found in the following link
     https://cellexplorer.org/datastructure/data-structure-and-format/#channels
 
-    Future versions of this interface will support the extraction of this metadata from these files.
+    Future versions of this interface will support the extraction of this metadata from these files as well
     """
 
     sampling_frequency_key = "sr"
--- a/src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
@@ -8,7 +8,7 @@ from pynwb import NWBFile
 from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
 from ..basesortingextractorinterface import BaseSortingExtractorInterface
 from ....tools import get_package
-from ....utils import FilePathType
+from ....utils import FilePathType, FolderPathType
 
 
 class CellExplorerRecordingInterface(BaseRecordingExtractorInterface):
@@ -42,7 +42,7 @@ class CellExplorerRecordingInterface(BaseRecordingExtractorInterface):
     sampling_frequency_key = "sr"
     binary_file_extension = "dat"
 
-    def __init__(self, folder_path, verbose=True, es_key: str = "ElectricalSeries"):
+    def __init__(self, folder_path: FolderPathType, verbose=True, es_key: str = "ElectricalSeries"):
         self.folder_path = Path(folder_path)
 
         # No super here, we need to do everything by hand
@@ -137,7 +137,7 @@ class CellExplorerLFPInterface(CellExplorerRecordingInterface):
     sampling_frequency_key = "srLfp"
     binary_file_extension = "lfp"
 
-    def __init__(self, folder_path, verbose=True, es_key: str = "ElectricalSeriesLFP"):
+    def __init__(self, folder_path: FolderPathType, verbose=True, es_key: str = "ElectricalSeriesLFP"):
         super().__init__(folder_path, verbose, es_key)
 
     def add_to_nwbfile(
--- a/src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
@@ -169,22 +169,23 @@ class CellExplorerRecordingInterface(BaseRecordingExtractorInterface):
 
     https://cellexplorer.org/
 
-    CellExplorer's new format contains a `session.mat` file, which has the following fields:
+    CellExplorer's new format contains a `basename.session.mat` file,
+    (where basename is isually a session identifier) containing the
+    following fields:
 
     * Sampling frequency
-    * Gains for both raw data (held in a file named session.dat) and lfp (located in session.lfp)
-    * Dtype for both raw data and lfp.
+    * Gains
+    * Dtype
 
     Link to the documentation detailing the file's structure:
     https://cellexplorer.org/datastructure/data-structure-and-format/#session-metadata
 
-    If the binary file is available (session.dat or session.lfp), the interface will use the
-    `BinaryRecordingExtractor` from spikeinterface to load the data. Otherwise, it will use the
-    `NumpyRecording` extractor to create a dummy extractor that only contains channel metadata.
-    This can be used to add the metadata to nwb using `write_electrical_series=False` as a conversion option.
+    If the binary file is available (basename.dat or basename.lfp),
+    the interface will use the `BinaryRecordingExtractor` object from spikeinterface
+    to load the data.
 
-    The metadata for this electrode is also extracted from the `session.mat` file. The logic of the extraction is
-    described on the function:
+    The metadata for this electrode is also extracted from the `basename.session.mat`
+    file. The logic of the extraction is described on the function:
 
     `add_channel_metadata_to_recorder_from_session_file`.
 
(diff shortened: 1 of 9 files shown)


src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
```diff
@@ -1,14 +1,172 @@
...
+
+    https://cellexplorer.org/datastructure/data-structure-and-format/#channels
+
+    As we accumulate more examples moving forward, we will be able to figure out what's better. This is a first
+    iteration.
+    """
+
+    sampling_frequency_key = "sr"
+
+    def __init__(self, folder_path, verbose=True, es_key: str = "ElectricalSeries"):
```
CodyCBakerPhD, line 46: Though I think I would actually prefer two separate file-path type arguments; one for the binary and one for the header. Folder mode is overkill for that I'd say

CodyCBakerPhD:
Just to confirm - the ecephys metadata extracted by both this and the CellExplorerSortingInterface are in agreement since they use the same header files?

Upon seeing it, the 'optional data file' approach seems pretty weird, whereas it seems appropriate to just have the sorting interface add ecephys metadata for its own sake

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
h-mayorquin: There is no such a clean distinction here. The relevant metadata is in a couple of files:

The metadata is in:
* `channelinfo.mat`
* `chanCoords`
* `chanMap` (if they used kilosort like in the current conversion)

The binary files are either:

* `{session}.dat`
* `session}.lfp`

h-mayorquin:
> Just to confirm - the ecephys metadata extracted by both this and the CellExplorerSortingInterface are in agreement since they use the same header files?

Which metadata specifically?

The current properties in `CellExplorerSortingInterface`  seem very sorting extractor related whereas the properties here are electrode related. 

> Upon seeing it, the 'optional data file' approach seems pretty weird, whereas it seems appropriate to just have the sorting interface add ecephys metadata for its own sake

What do you mean by the `optiona data file` approach?

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
CodyCBakerPhD: Oh I see, there's effectively more than one header file...

OK then

CodyCBakerPhD:
> Which metadata specifically?

All values in `metadata["Ecephys"]`, including a dummy recorder with identical property values


> What do you mean by the optional data file approach?

https://github.com/catalystneuro/neuroconv/pull/492/files#diff-15793905e5e0130bcd60f5603c6e9811d05ecfc65aca3e127c05ee61701059cfR72-R99

CodyCBakerPhD:
I'm saying a dummy recorder for setting metadata with the `write_ecephys_metadata` flag seems more OK than having an entire RecordingInterface that doesn't need any raw data to run (effectively equivalent to `write_electrical_series=False`, with the key exception being that we always have the sorting data, but don't always have raw data per session - hence more OK for the sorting interface to handle the metadata than to always have a RecordingInterface even if we don't have recording data)

h-mayorquin:
> > Which metadata specifically?
> 
> All values in `metadata["Ecephys"]`, including a dummy recorder with identical property values

You mean between this with a dummy recorder

> > What do you mean by the optional data file approach?
> 
> https://github.com/catalystneuro/neuroconv/pull/492/files#diff-15793905e5e0130bcd60f5603c6e9811d05ecfc65aca3e127c05ee61701059cfR72-R99

I am actually fine with removing the else branch of the loop. I had the hunch that you would want to run conversions where the binary files would not be present (as that the data that Gareth has available for example).

h-mayorquin:
> I'm saying a dummy recorder for setting metadata with the `write_ecephys_metadata` flag seems more OK than having an entire RecordingInterface that doesn't need any raw data to run (effectively equivalent to `write_electrical_series=False`, with the key exception being that we always have the sorting data, but don't always have raw data per session - hence more OK for the sorting interface to handle the metadata than to always have a RecordingInterface even if we don't have recording data)

I really think the two approaches as equally valid. More options is fine. 

Here, I am following the approach of buidling something as quick as possible that will generalize and save us labor in the future.  We can solve the current problem and if we do more conversions with Buzsaki we actually have a RecordingInterface for that.

CodyCBakerPhD:
> I am actually fine with removing the else branch of the loop. I had the hunch that you would want to run conversions where the binary files would not be present (as that the data that Gareth has available for example).

That is the goal we're working towards with all the CellExplorer PRs

We have sessions where we need to add ecephys metadata, but we don't have raw files for those sessions

But these chanMap and session.mat files are now a part of the newer CellExplorer format, right? So concievably the `CellExplorerSortingInterface.get_metadata` protocol could independently have responsibility for reading (and ultimately writing, with the conversion option) the ecephys metadata (Not just its own unit properties)

Just saying as a matter of practice it seems better to have an interface which actually has corresponding data be responsible for it instead of considering data files to be optional; OR be more concise and create a separate type of interface altogether (like `CellExplorerMetadataInterface` built specifically just to read the metadata files and write metadata to the NWB file

Overall I prefer the idea of the CellExplorerSortingInterface doing it since ecephys metadata is valuable context for the sorting data anyway

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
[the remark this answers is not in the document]
CodyCBakerPhD: Plan is to replace all such things with `pydantic.FilePathType` for added benefit of pydantic validation which goes beyond just 'is it a `Path` or a `str`'

h-mayorquin:
> Just saying as a matter of practice it seems better to have an interface which actually has corresponding data be responsible for it instead of considering data files to be optional; OR be more concise and create a separate type of interface altogether (like `CellExplorerMetadataInterface` built specifically just to read the metadata files and write metadata to the NWB file

But this might not be the case for other conversion projects. What if users use the CellExplorer format for the raw data but use kilosort only without transforming the output into the `{session}.spikes.cellinfo.mat`?  

I think having all the options is good:
A recording interface that can write the metadata independently of sorting. Conversion option:  `write_electrical_series`
A sorting interface that can write the mtadata independently of sorting. Conversion option: `write_ecephys_metadata`

The code for add he electrodes metadata can be shared between the two interfaces to avoid code repetition. That metadata is valuable context for the two of them. In the future for the probe properties such as the geometry we could have a ProbeInterface specifically designed for it.

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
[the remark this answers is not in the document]
h-mayorquin: Great.

CodyCBakerPhD:
> But this might not be the case for other conversion projects. What if users use the CellExplorer format for the raw data but use kilosort only without transforming the output into the {session}.spikes.cellinfo.mat?

Right, I'm totally on board with having a CellExplorerRecordingInterface eventually for that purpose (with the data file being required) but right now it's not the preferred solution as per 

> Here, I am following the approach of buidling something as quick as possible that will generalize and save us labor in the future.


I also agree that having all the options is good


>  In the future for the probe properties such as the geometry we could have a ProbeInterface specifically designed for it.

That is also the eventual hope, which I wanted you to kind of have in the back of your mind while playing around with this kind of stuff

h-mayorquin:
I am going to take a quick stab now to have the SortingInterface extract the electrode metadata as discussed. That way, we can use it in the current conversion. Yesterday I stopped in the tracks because I needed https://github.com/catalystneuro/neuroconv/pull/491 first but now that is merged and this is in a good shape maybe it will work out quick.

CodyCBakerPhD:
@h-mayorquin Is this ready for full review?

CodyCBakerPhD:
Approach seems fine otherwise - just requesting example file in https://github.com/catalystneuro/neuroconv/pull/494#issuecomment-1604351775

h-mayorquin:
@CodyCBakerPhD 
Added tests to the S3 which are being passed now. This is ready.

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
```diff
@@ -1,12 +1,328 @@
...
+
+        recording_extractor.set_property(key="channel_name", values=channel_name)
+        recording_extractor.set_property(key="group", ids=channel_ids, values=channel_groups)
+
+    return recording_extractor
+
+
+class CellExplorerRecordingInterface(BaseRecordingExtractorInterface):
+    """
+    Addds raw and lfp data from binary files with the new CellExplorer format:
```
CodyCBakerPhD, line 168:
```suggestion
    Adds raw and lfp data from binary files with the new CellExplorer format:
```

tests/test_on_data/test_recording_interfaces.py
```diff
@@ -84,6 +85,17 @@ class TestCEDRecordingInterface(RecordingExtractorInterfaceTestMixin, TestCase):
...
 
+class TestCellExplorerRecordingInterface(RecordingExtractorInterfaceTestMixin, TestCase):
+    data_interface_cls = CellExplorerRecordingInterface
+    interface_kwargs = [
+        dict(folder_path=str(DATA_PATH / "cellexplorer" / "dataset_4" / "Peter_MS22_180629_110319_concat_stubbed")),
+        dict(
+            folder_path=str(DATA_PATH / "cellexplorer" / "dataset_4" / "Peter_MS22_180629_110319_concat_stubbed_hdf5")
+        ),
+    ]
+    save_directory = OUTPUT_PATH
```
CodyCBakerPhD, line 96: Can you (i) break these into two separate testing classes so that (ii) you can add separate `check_metadata` tests to showcase all the metadata info they are automatically collecting?

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
```diff
@@ -1,12 +1,328 @@
...
+        stub_test: bool = False,
+        starting_time: Optional[float] = None,
+        write_as: Literal["raw", "lfp", "processed"] = "lfp",
+        write_electrical_series: bool = True,
+        compression: Optional[str] = None,
+        compression_opts: Optional[int] = None,
+        iterator_type: str = "v2",
+        iterator_opts: Optional[dict] = None,
+    ):
+        return super().add_to_nwbfile(
```
CodyCBakerPhD, line 314:
```suggestion
        super().add_to_nwbfile(
```

No need for `return` here since `add_to_nwbfile` returns `None`

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
```diff
@@ -1,12 +1,328 @@
...
+
+            traces_list = [np.empty(shape=(1, num_channels))]
+            dummy_recording = NumpyRecording(
+                traces_list=traces_list,
+                sampling_frequency=sampling_frequency,
+                channel_ids=channel_ids,
+            )
+
+            self.recording_extractor = dummy_recording
+            self.recording_extractor.set_channel_gains(channel_ids=channel_ids, gains=np.ones(num_channels) * gain)
```
CodyCBakerPhD, line 270: Thinking more on this, I really don't think we should have this `else` block

As is, there's nothing preventing someone from writing the dummy data to an NWB file by mistake because they forgot to disable `write_electrical_series` in the conversion options

Given the work in #494 the correct approach would be to use the sorting interface instead

And if we ever do run into the situation where there are (a) no raw data files (b) they used a different sorting format (such as kilosort) and (c) but they still have valuably cellexplorer metadata files, we should really just make either a specific `MetadataInterface` that adds no data to the `NWBFile` but only metadata, or a `MetadataMixin` that allows any Sorting or Recording Interface to incorporate different metadata protocols that might result from format mixing

tests/test_on_data/test_recording_interfaces.py
h-mayorquin: It is the same data in different formats. The metadata test if any should be the same.

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
h-mayorquin: If all the others are doing this that's fine but I feel it should return and nwbfile. I have not answered to your replies in the pipeiline interface because of lack of time.

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
h-mayorquin: I think this is a good point.

tests/test_on_data/test_recording_interfaces.py
h-mayorquin: Added the test for channel metadata extraction

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
h-mayorquin: I removed it.

h-mayorquin:
@CodyCBakerPhD 
`TestOpenEphysBinaryRecordingIn` are failin again and in #494

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -112,7 +112,7 @@ def test_conversion_as_lone_interface(self):
                 self.check_metadata_schema_valid()
                 self.check_conversion_options_schema_valid()
                 self.check_metadata()
-                self.nwbfile_path = str(self.save_directory / f"{self.data_interface_cls.__name__}_{num}.nwb")
+                self.nwbfile_path = str(self.save_directory / f"{self.__name__}_{num}.nwb")
```
CodyCBakerPhD, line 115:
```suggestion
                self.nwbfile_path = str(self.save_directory / f"{self.__classs__.__name__}_{num}.nwb")
```
derp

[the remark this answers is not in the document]
h-mayorquin: This seems to be working.

CodyCBakerPhD (approved):

PR #494, opened 2023-06-21
Add ecephys metadata to CellExplorerSorting

h-mayorquin:
OK, discussing with @CodyCBakerPhD on #492 he mentions that for the ongoing Buzsaki conversion he finds more convenient if the `CellExplorerSortingInterface` is the interface responsible for adding the metadata of the probe and recording device (i.e. electrode metadata). The argument, which makes sense, is that this is the data that will be available most of the time (as it is usually smaller).

The current PR extends #492 and gives `CellExplorerSortingInterface` the capability to write the recording device and probe metadata when using  `write_ecephys_metadata` as a conversion option. This is done in two steps:
 
* A recording is registered in the sorting extractor of the  `CellExplorerSortingInterface` and relevant metadata is attached as  properties to the recording. 
* When a recording is registered in a sorting interface and `write_ecephys_metadata` is called, the `BaseSortingExtractorInterface` adds those properties using our common methods (i.e. ` add_electrodes`).


src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -226,6 +226,38 @@ def subset_sorting(self):
         stub_sorting_extractor = self.sorting_extractor.frame_slice(start_frame=0, end_frame=end_frame)
         return stub_sorting_extractor
 
+    def add_recording_to_nwb(self, nwbfile: NWBFile, metadata: Optional[DeepDict] = None):
```
CodyCBakerPhD, line 229:
```suggestion
    def add_ecephys_metadata_to_nwbfile(self, nwbfile: NWBFile, metadata: Optional[DeepDict] = None):
```

Original method name not entirely accurate; if we had a full recording attached, it would not be writing the traces data

Proposing to rename to explicitly reflect what it adds while also making it more similar to the new base `add_to_nwbfile` method name

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -260,44 +292,31 @@ def add_to_nwbfile(
...
+        from ...tools.spikeinterface import add_sorting
 
-        from ...tools.spikeinterface import (
-            add_devices,
-            add_electrode_groups,
-            add_electrodes,
-            add_sorting,
-        )
+        if write_ecephys_metadata:
+            self.add_recording_to_nwb(nwbfile=nwbfile, metadata=metadata)
```
CodyCBakerPhD, line 298:
```suggestion
            self.add_ecephys_metadata_to_nwbfile(nwbfile=nwbfile, metadata=metadata)
```

CodyCBakerPhD:
One minor change regarding the method name, otherwise seems like the approach could work fine

@h-mayorquin Can you share or point me towards three example files for these different PRs?

One from running #492 without using the sorting interface from this branch

One from running the sorting interface on this branch without the recording from #492 

One from running both at the same time

h-mayorquin:
@CodyCBakerPhD 
Why 3?

You can use any of the files from the Huzsar or Valero conversion.

I am preparing a small example for the gin though that you can try. I will share the link on slack.

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
h-mayorquin: I agree that my choice is bad for the reasons that you describe. 
This is better.

I have a question though, isn't the units table also ecephys metadata and then this becomes overloaded?

Experimentally, what we are adding is the metadata of the recording device which is a combination of the recording device channels (that we add in the electrodes table), device and electrode groups. 

Can we do better or do you think is hard to disambiguate?

h-mayorquin: OK, after thinking for a while I think that the best name is `add_channel_metadata_to_nwb` it covers exactly the type of metadata that we are adding. It is not a probe, is it nos electrodes, it is not channels as those concepts are kind of intermixed in the way that we are handilng them right now.

What we are doing now is to add the channel metadata that is a combination of the things above. It is all not ecephys because it is possible that there is other ecephys data that is not channel related.

Added this in the last commit.

h-mayorquin:
> One minor change regarding the method name, otherwise seems like the approach could work fine
> 
> @h-mayorquin Can you share or point me towards three example files for these different PRs?
> 
> One from running #492 without using the sorting interface from this branch
> 
> One from running the sorting interface on this branch without the recording from #492
> 
> One from running both at the same time

OK
In the following zip you will find:
1) Both raw and LFP
2) Only sorting (writing write_ecephys_metadata True)
3) Combined
4) Both raw and LFP but with no write to electrical series
[files.zip](https://github.com/catalystneuro/neuroconv/files/11860165/files.zip)

tests/test_on_data/test_sorting_interfaces.py
```diff
@@ -44,6 +44,24 @@ class TestCellExplorerSortingInterface(SortingExtractorInterfaceTestMixin, TestC
...
         ),
+        dict(
+            file_path=str(
+                DATA_PATH
+                / "cellexplorer"
+                / "dataset_4"
+                / "Peter_MS22_180629_110319_concat_stubbed"
+                / "Peter_MS22_180629_110319_concat_stubbed.spikes.cellinfo.mat"
+            )
+        ),
```
CodyCBakerPhD, line 55: Similar to comment on other PR; can you break these into separate classes per test file and add the metadata check to showcase/assert what is being extracted?

CHANGELOG.md
```diff
@@ -18,6 +19,7 @@
 
 ### Testing
 * Added gin test for `CellExplorerRecordingInterface` [#488](https://github.com/catalystneuro/neuroconv/pull/488)
+* Added gin test for `CellExplorerSortinggInterface`  [PR #494](https://github.com/catalystneuro/neuroconv/pull/494)
```
CodyCBakerPhD, line 22: Can you add a note to these changelog lines that it's about the newer CellExplorer format? (with version info if there's any online since I know there aren't in the files)

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@
 * Added a common `.temporally_align_data_interfaces` method to the `NWBConverter` class to use as a specification of the protocol for temporally aligning the data interfaces of the converter. [PR #362](https://github.com/catalystneuro/neuroconv/pull/362)
 
 * Added `CellExplorerRecordingInterface` for adding data raw and lfp data from the CellExplorer format. [#488](https://github.com/catalystneuro/neuroconv/pull/488)
-* `CellExplorerSortingInterface` now supports the extacting sampling frequency from the new data format [PR #491](https://github.com/catalystneuro/neuroconv/pull/491)
+* `CellExplorerSortingInterface` now supports extracting sampling frequency from the new data format [PR #491](https://github.com/catalystneuro/neuroconv/pull/491)
+* `CellExplorerSortinggInterface` now supports adding channel metadata to the nwbfile with `write_ecephys_metadata=True` as a conversion option [PR #494](https://github.com/catalystneuro/neuroconv/pull/494)
```
CodyCBakerPhD, line 13:
```suggestion
* `CellExplorerSortingInterface` now supports adding channel metadata to the nwbfile with `write_ecephys_metadata=True` as a conversion option [PR #494](https://github.com/catalystneuro/neuroconv/pull/494)
```

CHANGELOG.md
CodyCBakerPhD, line 22:
```suggestion
* Added gin test for `CellExplorerSortingInterface`  [PR #494](https://github.com/catalystneuro/neuroconv/pull/494)
```

CodyCBakerPhD:
> You can generate the files yourself with the gin data (now in S3) and the following script:
https://gist.github.com/h-mayorquin/e9c2dd89622d5e6dd486debd929a7301

Thanks for the gist, as it explains a lot that I see in these files - however, what I meant was can you share some examples from your current stubbed conversion that use these interfaces as a part of the normal conversion pipeline (like a `ConverterPipe`)? What I really want to ensure is that merging of metadata through converter-level `get_metadata` calls is handled as expected

Especially pointing to the very specific handling of metadata-substructures at the interface-by-interface level of https://gist.github.com/h-mayorquin/e9c2dd89622d5e6dd486debd929a7301#file-examples_cellexplorer_recording-py-L51-L58; I wouldn't expect or ask users to be so delicate with their pipelines and instead they'd just initialize their interfaces (or pass their source data) and let the `NWBConverter` types handle it from there

CHANGELOG.md
h-mayorquin: Sure, I am not sure the changelog is a good place for this though. I will add in both in the docstring and in the changelog.

h-mayorquin: OK, added something. I think there is a confusion here. The format that contains `basename.session.mat` is referred as Cell Explorer. What I am referring as the "old format" here they refer as Buzcode (see one of the last emails that Petersen sent to us). 

So again, this is not a version thing, CellExplorer as they describe it should have a `session.mat` but because we were referring to their buzcode as CellExplorer I am making this distinction between old and new format.

What I added reflects this but feel free to add a diff that you think better reflects the situation.

CHANGELOG.md
h-mayorquin: Thanks!

tests/test_on_data/test_sorting_interfaces.py
h-mayorquin: They have the same data so I added a single test.

h-mayorquin:
@CodyCBakerPhD OK, here you go, new gist:

https://gist.github.com/h-mayorquin/f4c766b1eebc85e4312ae66c035345ef
[files_cody_cell_explorer_pipe.zip](https://github.com/catalystneuro/neuroconv/files/11866660/files_cody_cell_explorer_pipe.zip)

CodyCBakerPhD:
@h-mayorquin Seems to be getting an actual error related to the new tests now

h-mayorquin:
Yes, it is the same time memory error : /
This is actually no trivial for long recordings.

h-mayorquin:
There is a trivial solution with the new lazy extractors that I created but they are not in a spikeinterface release yet.

CodyCBakerPhD:
> Yes, it is the a memory error as well. The time vector is too large : /

Shouldn't the stubbed data have stubbed timestamps as well?

h-mayorquin:
> > Yes, it is the a memory error as well. The time vector is too large : /
> 
> Shouldn't the stubbed data have stubbed timestamps as well?

This is about the spikes.

It goes like this:
To generate the spikes as seconds spikeinterface spikeinterface uses two strategies:

https://github.com/SpikeInterface/spikeinterface/blob/5369cb0e799a52a1ccb1de6feacf05c41f65dd95/src/spikeinterface/core/basesorting.py#L115-L124

When you have a recording attached it uses the method `get_times` which generates a large timestamp array through this method:

https://github.com/SpikeInterface/spikeinterface/blob/5369cb0e799a52a1ccb1de6feacf05c41f65dd95/src/spikeinterface/core/baserecording.py#L647-L658

Now, the critical point, because we want to use a dummy extractor, we need the number of samples to be small, correct? (we don't want to create a large array of memory just as a dummy extractor) so the first strategy does not work for us. So I went for the second which is to create a time vector, but the time vector needs to be as large as the last frame of the spike. So I can't avoid creating this unecessary large time vector as this is better than creating and unecessary trace for the extractor (num_channels better!)

Now, there are two solutions:
1) Use a recording extractor that does not require large traces. This I have done but is not availalbe yet until the next release of spikeinterface.
2) Avoid using a registered recording so the spikes will be cast to seconds using the second strategy here:

https://github.com/SpikeInterface/spikeinterface/blob/5369cb0e799a52a1ccb1de6feacf05c41f65dd95/src/spikeinterface/core/basesorting.py#L115-L124

Which is what I have done in the other PR but needs a less standard path in neuroconv.

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
```diff
@@ -439,6 +433,41 @@ def __init__(self, file_path: FilePathType, verbose: bool = True):
...
+                channel_ids=channel_ids,
+            )
+
+            # Add the channel metadata
+            dummy_recording_extractor = add_channel_metadata_to_recoder(
+                recording_extractor=dummy_recording_extractor, folder_path=session_path
+            )
+
+            # Need a time vector for the recording extractor
+            last_spikes_frame = self.sorting_extractor.get_all_spike_trains()[0][0][-1]
```
h-mayorquin, line 466: @CodyCBakerPhD 
This is the trick that I mentioned. Creat a time vector that is as large as you need to cast the spikes (which spikeinterface stores as frames) to seconds later in `add_units_table`.

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -226,6 +226,69 @@ def subset_sorting(self):
...
+            if electrodes_metadata:
+                n_channels = max([len(x["data"]) for x in metadata["Ecephys"]["Electrodes"]])
+                from spikeinterface import NumpyRecording
+
+                recording = NumpyRecording(
+                    traces_list=[np.empty(shape=(1, n_channels))],
+                    sampling_frequency=self.sorting_extractor.get_sampling_frequency(),
+                )
+            else:
+                import warnings
```
CodyCBakerPhD, line 280: Can you move this to the top if you want to keep it?

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -226,6 +226,69 @@ def subset_sorting(self):
...
+            add_electrode_groups,
+            add_electrodes,
+        )
+
+        if hasattr(self, "generate_recording_with_channel_metadata"):
+            recording = self.generate_recording_with_channel_metadata()
+        else:
+            electrodes_metadata = metadata.get("Ecephys", dict()).get("Electrodes", list())
+            if electrodes_metadata:
+                n_channels = max([len(x["data"]) for x in metadata["Ecephys"]["Electrodes"]])
```
CodyCBakerPhD, line 272: "data" no longer exists in `metadata["Ecephys"]["Electrodes"]` - lines [270-279](https://github.com/catalystneuro/neuroconv/pull/494/files#diff-89e541faa3c286bbf53bd2387682d84bdde1235b6d752512f6869e7635b27a2dR270-R278) can be removed

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -226,6 +226,69 @@ def subset_sorting(self):
...
+                    sampling_frequency=self.sorting_extractor.get_sampling_frequency(),
+                )
+            else:
+                import warnings
+
+                warnings.warn(
+                    "No recording found in sorting extractor or electrodes_metadata in the metadata metadata"
+                    "Skipping addition of recording metadata to the NWBFile."
+                )
+                return None
```
CodyCBakerPhD, line 286:
```suggestion
                return
```

Any reason to include `None`? Kind of like annotating `-> None`, isn't it just the built-in default?

CodyCBakerPhD:
@h-mayorquin Final comments about removing legacy code - once that's cleaned up this should be good to go!

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
h-mayorquin: I think that all the else block can be removed, do you agree?

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
h-mayorquin: I feel that is more explicit and easier to infer that is an early termination of the function but I don't feel very strong about this. Do you prefer the empty return?

Anyway, as said above, this else block probably should go away anyway...

h-mayorquin: I feel that is more explicit and easier to infer that is an early termination of the function but I don't feel very strong about this. Do you prefer the empty return?

Anyway, as said above, this else block probably should go away anyway...

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
CodyCBakerPhD: Sure, removal of entire block is fine

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
CodyCBakerPhD, line 286:
```suggestion
```

CodyCBakerPhD: Pythonic explicit > implicit, I'm just wondering where the line is drawn 😆

Technically if we want to be super explicit we should fill every function call or initialization with all of their keywords, even the ones we  set to default; I see  `return None` as being at that same category, since the default of `return` is `return None` and the default of `def()` is `def() -> None`

Not a huge deal no matter

h-mayorquin: Yeah, it goes either way to me, same with annotation -> None for `__init__`. 

The else block was removed anyway.

CodyCBakerPhD (approved):

PR #495, opened 2023-06-23
Avoid redundant timestamp creation in `add_electrical_series` for recording objects without time vector

h-mayorquin:
I have a memory overflow error that is killing my full conversion for some sessions when working on this https://github.com/catalystneuro/buzsaki-lab-to-nwb/pull/66.

I am tracking down the issue and one of the things I noticed is that for very long-recordings extracting the timestamps in `add_electrical_series (to check if they are regular) create a large memory allocation. For example, for the 5 hours long recordings of some sessions the timestamps are ~ 2 GiB. We should avoid that if possible.

On the other hand, when the recording objects in spikeinterface do not have time vectors the call to `get_times` just creates artificial timestamps using the sampling rate which are regular by creation. As our interest is to check whether the timestamps are regular or not this is redundant and this is what this PR seek to avoid: when the recording segment has no time vector the check for regularity if skipped.

### DIFF
--- a/src/neuroconv/tools/spikeinterface/spikeinterface.py
+++ b/src/neuroconv/tools/spikeinterface/spikeinterface.py
@@ -656,18 +656,27 @@ def add_electrical_series(
         data=H5DataIO(data=ephys_data_iterator, compression=compression, compression_opts=compression_opts)
     )
 
-    # Timestamps vs rate
-    timestamps = recording.get_times(segment_index=segment_index)
-    rate = calculate_regular_series_rate(series=timestamps)  # Returns None if it is not regular
-    starting_time = starting_time if starting_time is not None else 0
+    # Now we decide whether to store the timestamps as a regular series or as an irregular series.
+    if recording.has_time_vector(segment_index=segment_index):
+        # First we check if the recording has a a time vector to avoid creating artificial timestamps
+        timestamps = recording.get_times(segment_index=segment_index)
+        rate = calculate_regular_series_rate(series=timestamps)  # Returns None if it is not regular
+        recording_t_start = timestamps[0]
+    else:
+        rate = recording.get_sampling_frequency()
+        recording_t_start = recording._recording_segments[segment_index].t_start
+        recording_t_start = 0 if recording_t_start is None else recording_t_start
 
+    starting_time = starting_time if starting_time is not None else 0
     if rate:
-        starting_time = starting_time + timestamps[0]
+        starting_time = float(starting_time + recording_t_start)
+        # Note that we call the sampling frequency again because the estimated rate might be different from the
+        # sampling frequency of the recording extractor by some epsilon.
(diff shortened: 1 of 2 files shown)


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -656,18 +656,27 @@ def add_electrical_series(
...
+    # Now we decide whether to store the timestamps as a regular series or as an irregular series.
+    if recording.has_time_vector(segment_index=segment_index):
+        # First we check if the recording has a a time vector to avoid creating artificial timestamps
+        timestamps = recording.get_times(segment_index=segment_index)
+        rate = calculate_regular_series_rate(series=timestamps)  # Returns None if it is not regular
+        recording_t_start = timestamps[0]
+    else:
+        rate = recording.get_sampling_frequency()
+        recording_t_start = recording._recording_segments[segment_index].t_start
+        recording_t_start = 0 if recording_t_start is None else recording_t_start
```
bendichter, line 668:
```suggestion
        recording_t_start = recording._recording_segments[segment_index].t_start or 0
```

if I understand correctly, this simpler version should work

bendichter:
Thank @h-mayorquin I agree, it's a good idea to skip the timestamp generation altogether if the timeseries is regular to begin with.

bendichter:
What would be the best approach to apply this logic to all timeseries?

src/neuroconv/tools/spikeinterface/spikeinterface.py
h-mayorquin: Yes, this is better!

h-mayorquin:
@bendichter 

I am not sure how we could generalize this to objects that are not from spikeinterface. I will think about it. In this case I know how the method that get the timestamps of the recording extractor works  (as in, I know that if it does not have an attribute called `time_vector` then it generates regular timestamps):

https://github.com/SpikeInterface/spikeinterface/blob/5369cb0e799a52a1ccb1de6feacf05c41f65dd95/src/spikeinterface/core/baserecording.py#L647-L658

This is the reason that I can avoid the call to `get_times`. 

Maybe something similar could be done for roiextractors but I am less sure.

CodyCBakerPhD:
> Maybe something similar could be done for roiextractors but I am less sure.

Yes, we have parallel `has_time_vector` method in [ROI Extractors](https://github.com/search?q=repo%3Acatalystneuro%2Froiextractors%20has_time_vector&type=code), though it's likely less of a problem over there due to the lower sampling frequencies

CodyCBakerPhD (approved):

CodyCBakerPhD:
LGTM - I'd ask for a test or two, but I know that would be quite tricky to assert significant different in memory usage, especially in small CI environments

h-mayorquin:
I will add better tests to this once we decouple the wrappers of compression and chunk iteration from the add_to_nwb methods.

PR #498, opened 2023-06-26
Add `MiniscopeConverter` 

weiglszonja:
Adds converter for `MiniscopeImagingInterface` and `MiniscopeBehaviorInterface`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
 -   repo: https://github.com/psf/black
-    rev: 23.3.0
+    rev: 23.7.0
     hooks:
     -   id: black
         exclude: ^docs/
--- a/src/neuroconv/datainterfaces/ecephys/requirements.txt
+++ b/src/neuroconv/datainterfaces/ecephys/requirements.txt
@@ -1,2 +1,2 @@
-spikeinterface>=0.97.1
+spikeinterface>=0.97,<0.98
 packaging<22.0


weiglszonja:
TODO:
maybe change it to Converter, no need for ConverterPipe when there are exactly these two data streams

src/neuroconv/tools/testing/converter_pipe_mixins.py
```diff
@@ -0,0 +1,92 @@
...
+from pathlib import Path
+from typing import List, Type, Union
+
+from jsonschema.validators import Draft7Validator, validate
+
+from neuroconv import ConverterPipe
+from neuroconv.utils import NWBMetaDataEncoder
+
+
+class ConverterPipeTestMixin:
```
weiglszonja, line 13: This should be done separately, first is sufficient to add the tests specific for this converter and can be revised after.

src/neuroconv/datainterfaces/ophys/miniscope/miniscopeconverter.py
```diff
@@ -0,0 +1,54 @@
...
+
+    data_interface_classes = dict(
+        Miniscope=MiniscopeImagingInterface,
+        BehavCam=MiniscopeBehaviorInterface,
+    )
+
+    @classmethod
+    def get_source_schema(cls):
+        source_schema = super().get_source_schema()
+        source_schema.update(additionalProperties=True)
```
CodyCBakerPhD, line 17: Remind me why it was necessary to allow additional properties here?

CodyCBakerPhD (approved):

CodyCBakerPhD:
@weiglszonja Thanks for the great work on this!

Just one last question about the source schema, but should be good to go with or without that

src/neuroconv/datainterfaces/ophys/miniscope/miniscopeconverter.py
weiglszonja: We can discuss this today but basically I get this error when trying to init the converter using `"folder_path"`:
`jsonschema.exceptions.ValidationError: Additional properties are not allowed ('folder_path' was unexpected)`
I thought it's better to use `"folder_path"` compared to `"source_data"` since both the imaging and the behavior data streams use the same folder.

CodyCBakerPhD: Oh, well sure - but then shouldn't we define the schema for the `folder_path` as a required property then? Via the same mechanism the interfaces use (`get_schema_from_method_signature`)?

I point this out because `additionalProperties=True` is a blanket shortcut used to bypass more rigorous specification, and so should be used only when truly necessary

src/neuroconv/datainterfaces/ophys/miniscope/miniscopeconverter.py
```diff
@@ -0,0 +1,54 @@
...
+        return source_schema
```
CodyCBakerPhD, line 18:
```suggestion
    @classmethod
    def get_source_schema(cls):
        return get_schema_from_method_signature(cls)
```

src/neuroconv/datainterfaces/ophys/miniscope/miniscopeconverter.py
```diff
@@ -0,0 +1,54 @@
+from ... import MiniscopeBehaviorInterface, MiniscopeImagingInterface
+from ....nwbconverter import NWBConverter
+from ....utils import FolderPathType
```
CodyCBakerPhD, line 3:
```suggestion
from ....utils import FolderPathType, get_schema_from_method_signature
```

CodyCBakerPhD (changes requested):

src/neuroconv/datainterfaces/ophys/miniscope/miniscopeconverter.py
weiglszonja: Ah sure, didn't think of `get_schema_from_method_signature`. Thank you!

tests/test_on_data/test_format_converters/test_miniscope_converter.py
```diff
@@ -0,0 +1,134 @@
...
+            self.behavcam_metadata,
+        )
+
+    def test_run_conversion(self):
+        nwbfile_path = str(self.test_dir / "test_miniscope_converter.nwb")
+        self.converter.run_conversion(nwbfile_path=nwbfile_path)
+
+        self.assertNWBFileStructure(nwbfile_path=nwbfile_path)
+
+    def test_run_conversion_updated_metadata(self):
```
CodyCBakerPhD, line 76: This test does a great job showcasing the metadata control of this converter; can you add some extra tests showcasing the `conversion_options` and its schema as well? (if there are any... showcase how it's empty otherwise?)

In particular, how it appears when nested inside another NWBConverter/ConverterPipe

src/neuroconv/nwbconverter.py
```diff
@@ -111,9 +111,16 @@ def _validate_source_data(self, source_data: Dict[str, dict], verbose: bool = Tr
...
+            if isinstance(data_interface, NWBConverter):
+                data_interface.add_to_nwbfile(
+                    nwbfile=nwbfile,
+                    metadata=metadata,
+                    conversion_options=conversion_options.get(interface_name, dict()),
+                )
+            else:
+                data_interface.add_to_nwbfile(
+                    nwbfile=nwbfile, metadata=metadata, **conversion_options.get(interface_name, dict())
+                )
```
weiglszonja, line 123: @CodyCBakerPhD 
the `conversion_options` for `MiniscopeConverter` would look like this when using it from inside an `NWBConverter`:
`conversion_options = dict(MiniscopeConverter=dict(Miniscope=dict(stub_test=True)))`
However
`**conversion_options.get(interface_name, dict())`
returns an error :

```
metadata = DeepDict: {'NWBFile': {'session_description': 'Auto-generated by neuroconv', 'identifier': '5f1bc608-b43b-4b22-a45c-20...2'}], 'ImageSeries': [{'name': 'BehavCamImageSeries', 'device': 'BehavCam2', 'dimension': [1280, 720], 'unit': 'px'}]}}
conversion_options = {'TestMiniscopeConverter': {'Miniscope': {'stub_frames': 2, 'stub_test': True}}}

    def add_to_nwbfile(self, nwbfile: NWBFile, metadata, conversion_options: Optional[dict] = None) -> None:
        for interface_name, data_interface in self.data_interface_objects.items():
>           data_interface.add_to_nwbfile(
                    nwbfile=nwbfile, metadata=metadata, **conversion_options.get(interface_name, dict())
            )
E           TypeError: add_to_nwbfile() got an unexpected keyword argument 'Miniscope'
../../../src/neuroconv/nwbconverter.py:114: TypeError
```
I'm not sure this is the best fix possible, what do you think?

CodyCBakerPhD: What conversion options does the `MiniscopeConverter` need to expose? Is it only `stub`?

If so the best fix will not be to modify the base NWBConverter, but do what we did with the source schema and go for a Converter class that behaves more like an Interface - define the conversion option schema explicitly as we would normally operate it on an interface, e.g., `get_schema_from_method_signature` applied to `add_to_nwbfile`

weiglszonja: Yes, the `MiniscopeBehaviorInterface` doesn't have any conversion options at the moment, so it's only what `MiniscopeImagingInterface` exposes in `add_to_nwbfile()` : `photon_series_type`, `stub_test`, `stub_frames`
See in https://github.com/catalystneuro/neuroconv/pull/498/commits/af2818b1e1559ec5f023be1680eb5e8f096d17af

CodyCBakerPhD: > photon_series_type, stub_test, stub_frames

For Minscope, shouldn't we be pinning `photon_series_type` to `OnePhotonSeries`?

src/neuroconv/datainterfaces/ophys/miniscope/miniscopeconverter.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @classmethod
+    def get_source_schema(cls):
+        return get_schema_from_method_signature(cls)
+
+    @classmethod
+    def get_conversion_options_schema(cls):
+        return get_schema_from_method_signature(
+            MiniscopeImagingInterface.add_to_nwbfile, exclude=["nwbfile", "metadata"]
+        )
```
CodyCBakerPhD, line 21: Might actually just be able to effectively alias this

```suggestion
        return MiniscopeImagingInterface.get_conversion_options_schema()
```

CodyCBakerPhD: Since the other interface doesn't have anything

weiglszonja: This returns: 
```
>       return MiniscopeImagingInterface.get_conversion_options_schema()
E       TypeError: get_conversion_options_schema() missing 1 required positional argument: 'self'
```
so maybe still `get_schema_from_method_signature` is needed?

src/neuroconv/datainterfaces/ophys/miniscope/miniscopeconverter.py
```diff
@@ -0,0 +1,73 @@
...
+class MiniscopeConverter(NWBConverter):
+    """Primary conversion class for handling Miniscope data streams."""
+
+    @classmethod
+    def get_source_schema(cls):
+        return get_schema_from_method_signature(cls)
+
+    @classmethod
+    def get_conversion_options_schema(cls):
+        return MiniscopeImagingInterface.get_conversion_options_schema(MiniscopeImagingInterface)
```
CodyCBakerPhD, line 19:
```suggestion
    def get_conversion_options_schema(cls):
        return self.data_interface_objects["MiniscopeImaging"].get_conversion_options_schema()
```

Bah, just remembered this isn't a class method at all anymore; may have to fix my suggestion on the key

src/neuroconv/datainterfaces/ophys/miniscope/miniscopeconverter.py
```diff
@@ -50,24 +46,27 @@ def __init__(self, folder_path: FolderPathType, verbose: bool = True):
...
+        return self.data_interface_objects["MiniscopeImaging"].get_conversion_options_schema()
+
     def add_to_nwbfile(
         self,
         nwbfile: NWBFile,
         metadata,
-        stub_test: bool = False,
-        stub_frames: int = 100,
+        conversion_options: Optional[dict] = None,
+        **kwargs,
```
weiglszonja, line 61: This works, but I know how this looks.. The issue when you're defining conversion_options for a standalone converter the `conversion_options` are passed as a dict, but when you're using it inside another converter the `conversion_options` are passed as `**conversion_options`. I will try to come up with a nicer solution but this is the best I have so far without touching `NWBConverter`.

https://github.com/catalystneuro/neuroconv/blob/cc15563a82cdc8dc6904d40372ee47123ad63737/src/neuroconv/nwbconverter.py#L111-L116

CodyCBakerPhD: Ahhh nice catch, I see the problem

That would probably be a fine minimal alteration of the core NWBConverter, instead of `**conversion_options.get(...)` -> `conversion_options=conversion_options.get(interface_name, dict())`

Then no need for `kwargs` in the Miniscope case

I think the new `add_to_nwbfile` syntax might make that possible where before it wasn't as easy to pass

weiglszonja: Thanks @CodyCBakerPhD, the tests are working for me locally. we'll see how it goes for the CI

src/neuroconv/nwbconverter.py
```diff
@@ -112,7 +112,9 @@ def add_to_nwbfile(self, nwbfile: NWBFile, metadata, conversion_options: Optiona
         conversion_options = conversion_options or dict()
         for interface_name, data_interface in self.data_interface_objects.items():
             data_interface.add_to_nwbfile(
-                nwbfile=nwbfile, metadata=metadata, **conversion_options.get(interface_name, dict())
+                nwbfile=nwbfile,
+                metadata=metadata,
+                conversion_options=conversion_options.get(interface_name, dict()),
```
CodyCBakerPhD, line 117:
```suggestion
                nwbfile=nwbfile, metadata=metadata, **conversion_options.get(interface_name, dict())
```

Never mind, this isn't going to word after all - `add_to_nwbfile` takes dynamic kwargs by design https://github.com/catalystneuro/neuroconv/blob/add_MiniscopeConverterPipe/src/neuroconv/basedatainterface.py#L55

like in https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py#L268-L275

src/neuroconv/datainterfaces/ophys/miniscope/miniscopeconverter.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+
+    def get_conversion_options_schema(self) -> dict:
+        return self.data_interface_objects["MiniscopeImaging"].get_conversion_options_schema()
+
+    def add_to_nwbfile(
+        self,
+        nwbfile: NWBFile,
+        metadata,
+        conversion_options: Optional[dict] = None,
```
CodyCBakerPhD, line 60: This seems like it might be the core of the problem - `add_to_nwbfile` is supposed to directly expose the conversion options as keyword arguments in the signature, not an argument called 'conversion_options`

This, again, is an oddity like with the source schema where this 'subconverter' behaves more like a DataInterface than a classic NWBConverter

So what you need to do is replace `conversion_options` here with the signature used in MiniscopeImaging; then you can probably go back to deriving the schema from the method signature if you want

Sorry for not catching that earlier

CodyCBakerPhD:
@weiglszonja This LGTM now, thanks for digging into that final issue

Merge when you're ready

PR #500, opened 2023-06-26
Add ecephys metadata to `CellExplorerSorting` through method

h-mayorquin:
This is another option to avoid the creation of an unecessary large time vector for the dummy sorting. 

This probably is telling that we need a lazy mechanism to handle timestamps in spikeinterface. I am thinking about how to do this.

Has #494 as a base.


CodyCBakerPhD:
OK as temporary solution 👍

Keep thinking on SI side of things - as we see more and more chronic recordings a lazy timestamper sounds like a good thing to have

PR #501, opened 2023-06-28
Cheaper test for series regularity (constant sampling_frequency)

h-mayorquin:
#495 got me thinking on how to avoid dangerous memory allocations when the series is not regular there is still the issue in how to minimize possible side effects when the series indeed has variable timestamps.

The current method that checks for regularity is a simple and logical implementation:

https://github.com/catalystneuro/neuroconv/blob/4601d2af7f9a60b476ff53a15081e94236ebd6da/src/neuroconv/utils/checks.py#L45-L55

But has the flaw that is creates another memory allocation (np.diff does) when used. If the timestamps are large this is problematic as it leads to a large memory spike that might kill the process.

This PR proposes a cheaper way of calculating series regularity which should avoid this problem.


CodyCBakerPhD:
From convo:

This method needs to be an algorithm not a heuristic, as such let's iterate over the entire array but efficiently - one proposal is to use a data chunk iterator (like the Sliceable one) to compute piecewise diffs and early exit on the first non-regular one of those

h-mayorquin:
For further reference. The problem with the current proposal is that it might happen by chance that the irregular differences in sampling cancel each other. This is a set of measure zero that becomes exponentially less likely as we increase the number of samples but is still an heuristic...

I think that an improvement of the test to make it exact is too complex and not worth it until the need comes. If that need to come to us, I would also probably prefer to work in improving spikeinterface which is where the most heavy timestmaps might come from (modality with the highest sampling rates).

Feel free to close this @CodyCBakerPhD .

CodyCBakerPhD:
> This is a set of measure zero

If we were talking about the temporal problem in reality (continuous space) that would certainly be true

Since this is an arbitrary discretization of that (even counting supposedly 'continuous' delays which have to be reduced down to a resolution of say, `float64`) our probability of being wrong will always be finite, and could even be high in situations where the timing rig is set up in a particular way

> I think that an improvement of the test to make it exact is too complex and not worth it until the need comes. 

The ideal way, similar to the zero-array detection in #527, would be simply to do the previous logic in an iterative fashion to yield both early return functionality, less memory usage per iteration, but still utilizing vectorization to a certain extent for speedups

Once we have the logic written down it should be easy to port across the related use cases


> If that need to come to us, I would also probably prefer to work in improving spikeinterface which is where the most heavy timestmaps might come from (modality with the highest sampling rates).

I mean, for SI we don't even request timestamps unless it `.has_times()` or whatever the command is; thus we know in `O(1)` time if we want to use regular or irregular approach - unless you're saying there are formats that generate and store regular timestamps on purpose?

PR #502, opened 2023-06-30
Hot fix for `CellExplorerSortingInterface`

h-mayorquin:
@garrettmflynn is running into some trouble when trying to load spikes. This makes the sampling frequency available even if the latest version of spikeinterface is not used. Following up on #491 for yet another configuration and missing parameters.


CodyCBakerPhD:
Bah, I should have caught this earlier, sorry.

Any way we can add a test for it? @garrettmflynn Is this branch working for you?

garrettmflynn:
Yep works great

h-mayorquin:
For the test you will need to stub some spikes.cellinfo files from the Valero conversion and their corresponding session.mat files.

CodyCBakerPhD (approved):

PR #505, opened 2023-07-03
Fix #504

### LINKED ISSUE #504: [Bug]: add_to_nwbfile fails if metadata is not explicitly passed
### What happened?

The new method `add_to_nwbfile()` which is supposed to be defined by new data interfaces now instead of `run_conversion()` fails if metadata is not explicitly passed by the user.

This is a bug, we're suppose to accept the metadata auto-generated by the `get_metadata()` method as a base minimal metadata object.

### Steps to Reproduce

```python
from pynwb.file import NWBFile
from neuroconv import BaseDataInterface


class MyInterface(BaseDataInterface):

    def __init__(self):
        super().__init__()
    
        return metadata_schema
    
    def get_metadata(self) -> dict:
        metadata["NWBFile"]["session_start_time"] = "2020-01-01T12:30:00-08:00"
        return metadata

    def add_to_nwbfile(self, nwbfile: NWBFile, metadata: dict, **conversion_options):
        # Conversion code comes here...
        print("converted")


di = MyInterface()
di.run_conversion(nwbfile_path="out.nwb")
```

### Traceback

```python
ValidationError                           Traceback (most recent call last)
/home/luiz/storage2/Github/shuler-lab-to-nwb/examples/fiber_photometry.ipynb Cell 26 in 2
     21         print("converted")
     24 di = MyInterface()
---> 25 di.run_conversion(nwbfile_path="out.nwb")
     26 # di.get_metadata()

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/basedatainterface.py:85, in BaseDataInterface.run_conversion(self, nwbfile_path, nwbfile, metadata, overwrite, **conversion_options)
     78 if nwbfile_path is None:
     79     warnings.warn(
     80         "Using DataInterface.run_conversion() without specifying nwbfile_path is deprecated. To create an "
     81         "NWBFile object in memory, use DataInterface.create_nwbfile(). To append to an existing NWBFile object,"
     82         " use DataInterface.add_to_nwbfile()."
     83     )
---> 85 with make_or_load_nwbfile(
     86     nwbfile_path=nwbfile_path,
     87     nwbfile=nwbfile,
     88     metadata=metadata,
     89     overwrite=overwrite,
     90     verbose=getattr(self, "verbose", False),
     91 ) as nwbfile_out:
     92     self.add_to_nwbfile(nwbfile_out, metadata=metadata, **conversion_options)

File /media/luiz/storage2/conda_environments/env_shuler/lib/python3.8/contextlib.py:113, in _GeneratorContextManager.__enter__(self)
    111 del self.args, self.kwds, self.func
    112 try:
--> 113     return next(self.gen)
    114 except StopIteration:
    115     raise RuntimeError("generator didn't yield") from None

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/tools/nwb_helpers.py:185, in make_or_load_nwbfile(nwbfile_path, nwbfile, metadata, overwrite, verbose)
    183 except Exception as e:
    184     success = False
--> 185     raise e
    186 finally:
    187     if nwbfile_path_in:

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/tools/nwb_helpers.py:181, in make_or_load_nwbfile(nwbfile_path, nwbfile, metadata, overwrite, verbose)
    179         nwbfile = io.read()
    180     elif nwbfile is None:
--> 181         nwbfile = make_nwbfile_from_metadata(metadata=metadata)
    182     yield nwbfile
    183 except Exception as e:

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/tools/nwb_helpers.py:57, in make_nwbfile_from_metadata(metadata)
     55 schema_path = Path(__file__).resolve().parent.parent / "schemas/base_metadata_schema.json"
     56 base_metadata_schema = load_dict_from_file(file_path=schema_path)
---> 57 validate_metadata(metadata=metadata, schema=base_metadata_schema)
     59 nwbfile_kwargs = metadata["NWBFile"]
     60 # convert ISO 8601 string to datetime

File /media/luiz/storage2/Github/neuroconv/src/neuroconv/utils/json_schema.py:318, in validate_metadata(metadata, schema, verbose)
    316 serialized_metadata = encoder.encode(metadata)
    317 decoded_metadata = json.loads(serialized_metadata)
--> 318 validate(instance=decoded_metadata, schema=schema)
    319 if verbose:
    320     print("Metadata is valid!")

File /media/luiz/storage2/conda_environments/env_shuler/lib/python3.8/site-packages/jsonschema/validators.py:1121, in validate(instance, schema, cls, *args, **kwargs)
   1119 error = exceptions.best_match(validator.iter_errors(instance))
   1120 if error is not None:
-> 1121     raise error

ValidationError: None is not of type 'object'
```

### Operating System

Linux

### Python Executable

Conda

### Python Version

3.9

### Package Versions

_No response_

### Code of Conduct

luiztauffer:
Fix #504

### DIFF
--- a/src/neuroconv/basedatainterface.py
+++ b/src/neuroconv/basedatainterface.py
@@ -86,11 +86,8 @@ class BaseDataInterface(ABC):
                 " use DataInterface.add_to_nwbfile()."
             )
 
-        base_metadata = self.get_metadata()
         if metadata is None:
-            metadata = base_metadata
-        else:
-            metadata = dict_deep_update(base_metadata, metadata)
+            metadata = self.get_metadata()
 
         with make_or_load_nwbfile(
             nwbfile_path=nwbfile_path,
--- a/src/neuroconv/basedatainterface.py
+++ b/src/neuroconv/basedatainterface.py
@@ -7,7 +7,7 @@ from typing import List, Optional
 from pynwb import NWBFile
 
 from .tools.nwb_helpers import make_nwbfile_from_metadata, make_or_load_nwbfile
-from .utils import get_schema_from_method_signature, load_dict_from_file
+from .utils import get_schema_from_method_signature, load_dict_from_file, dict_deep_update
 from .utils.dict import DeepDict
 
 
@@ -81,6 +81,11 @@ class BaseDataInterface(ABC):
                 "NWBFile object in memory, use DataInterface.create_nwbfile(). To append to an existing NWBFile object,"
                 " use DataInterface.add_to_nwbfile()."
             )
+        
+        base_metadata = self.get_metadata()
+        if metadata is None:
+            metadata = {}
+        metadata = dict_deep_update(base_metadata, metadata)
 
         with make_or_load_nwbfile(
             nwbfile_path=nwbfile_path,
--- a/src/neuroconv/basedatainterface.py
+++ b/src/neuroconv/basedatainterface.py
@@ -88,8 +88,9 @@ class BaseDataInterface(ABC):
 
         base_metadata = self.get_metadata()
         if metadata is None:
-            metadata = {}
-        metadata = dict_deep_update(base_metadata, metadata)
+            metadata = base_metadata
+        else:
+            metadata = dict_deep_update(base_metadata, metadata)
 
         with make_or_load_nwbfile(
             nwbfile_path=nwbfile_path,
--- a/src/neuroconv/basedatainterface.py
+++ b/src/neuroconv/basedatainterface.py
@@ -7,7 +7,11 @@ from typing import List, Optional
 from pynwb import NWBFile
 
 from .tools.nwb_helpers import make_nwbfile_from_metadata, make_or_load_nwbfile
-from .utils import get_schema_from_method_signature, load_dict_from_file, dict_deep_update
+from .utils import (
+    dict_deep_update,
+    get_schema_from_method_signature,
+    load_dict_from_file,
+)
 from .utils.dict import DeepDict
 
 
@@ -81,7 +85,7 @@ class BaseDataInterface(ABC):
                 "NWBFile object in memory, use DataInterface.create_nwbfile(). To append to an existing NWBFile object,"
                 " use DataInterface.add_to_nwbfile()."
             )
-        
+
         base_metadata = self.get_metadata()
         if metadata is None:
             metadata = {}


src/neuroconv/basedatainterface.py
```diff
@@ -82,6 +86,11 @@ def run_conversion(
                 " use DataInterface.add_to_nwbfile()."
             )
 
+        base_metadata = self.get_metadata()
+        if metadata is None:
+            metadata = {}
+        metadata = dict_deep_update(base_metadata, metadata)
```
bendichter, line 92: @luiztauffer thanks, I agree this is a bug. We should either make `metadata` required or properly handle when it is missing. Is there a situation where `self.get_metadata()` returns `None`? Otherwise, let's just simplify this as
```suggestion
        metadata = dict_deep_update(self.get_metadata(), metadata)
```

luiztauffer: @bendichter 
>  Is there a situation where self.get_metadata() returns None?

No, but `metadata` might be `None`, that's why we need lines 90-91

bendichter: oh, right I see

bendichter, line 92: how about this?

```suggestion
        base_metadata = self.get_metadata()
        if metadata is None:
            metadata = base_metadata
        else:
            metadata = dict_deep_update(base_metadata, metadata)
```

bendichter:
```pytb
AssertionError: The Audio metadata is incomplete (5 entry)! Expected 3 (one for each entry of 'file_paths').

During handling of the above exception, another exception occurred:

self = <test_audio_interface.TestAudioInterface testMethod=test_not_all_metadata_are_unique>

    def test_not_all_metadata_are_unique(self):
        metadata = deepcopy(self.metadata)
        metadata["Behavior"].update(
            Audio=[
                dict(name="Audio", description="Acoustic waveform series."),
                dict(name="Audio", description="Acoustic waveform series."),
                dict(name="Audio2", description="Acoustic waveform series."),
            ],
        )
        expected_error_message = "Some of the names for Audio metadata are not unique."
        with self.assertRaisesWith(exc_type=AssertionError, exc_msg=expected_error_message):
>           self.interface.run_conversion(nwbfile_path=self.nwbfile_path, metadata=metadata)
E           AssertionError: "^Some\ of\ the\ names\ for\ Audio\ metadata\ are\ not\ unique\.$" does not match "The Audio metadata is incomplete (5 entry)! Expected 3 (one for each entry of 'file_paths')."

tests/test_behavior/test_audio_interface.py:159: AssertionError
```
I'm surprised the deep update didn't handle the duplication. Any idea why this might be, @luiztauffer ?

luiztauffer:
With the current proposed change under `self.run_conversion()`, the list of audios from `self.get_metadata()` (3 items) will append the list of unique audios passed by the metadata argument (3 items, 2 with repeated `name`, so 2 items at the end). The final metadata before conversion becomes this:

```python
{
    'NWBFile': {
        'session_description': 'Auto-generated by neuroconv', 
        'identifier': '091babae-f732-43ab-9c22-62e9a758afd5', 
        'session_start_time': datetime.datetime(2023, 7, 4, 0, 13, 5, 499798, tzinfo=tzfile('/usr/share/zoneinfo/US/Pacific'))
    }, 
    'Behavior': {
        'Audio': [
            {'name': 'AcousticWaveformSeries', 'description': 'Acoustic waveform series.'}, 
            {'name': 'AcousticWaveformSeries1', 'description': 'Acoustic waveform series.'}, 
            {'name': 'AcousticWaveformSeries2', 'description': 'Acoustic waveform series.'}, 
            {'name': 'Audio', 'description': 'Acoustic waveform series.'}, 
            {'name': 'Audio2', 'description': 'Acoustic waveform series.'}
        ]
    }    
}
```

So the conversion is operating as expected, the correct error was raised (number of items in list different from number of audio files), it’s just not the error expected by the test. Btw, `dict_deep_update ()` will handle duplicates, so the error expected by the test can only happen with we don't pass the new metadata through `dict_deep_update ()`.

The question is, if `self.run_conversion()` receives `metadata!=None`, what do we want to do?
1. append passed metadata to base datainterface metadata, as it comes form `self.get_metadata()`. This is the current behavior.
2. accept the passed metadata as the final metadata, in which case we should change our code to this:
```python
if metadata is None:
    metadata = self.get_metadata()
```

The test seems to expect the result from number 2.
What do you think @bendichter ?

luiztauffer:
I think that, if the user/script is explicitly passing a metadata object to `run_conversion()`, then they should be responsible for passing a working version of it and, if no metadata is explicitly passed, the datainterface should use its default metadata from `get_metadata()`.

I changed the code to follow this behavior now, and the tests are passing again.

bendichter:
Yeah, that seems reasonable to me

bendichter (approved):

PR #506, opened 2023-07-03
add typehint outputs to abstract classes

bendichter:
I am adding typehint outputs for the methods in the abstract datainterface and nwbconverter classes. This should help ensure that downstream classes override these methods properly.


h-mayorquin (approved):
This looks good to me.

PR #507, opened 2023-07-05
Update `BrukerTiffImagingInterface` for volumetric imaging

weiglszonja:
Extractor PR: https://github.com/catalystneuro/roiextractors/pull/230


src/neuroconv/datainterfaces/ophys/brukertiff/brukertiffdatainterface.py
```diff
@@ -1,12 +1,127 @@
...
+
+        imaging_plane_metadata = metadata["Ophys"]["ImagingPlane"][0]
+        imaging_plane_metadata.update(
+            device=device_name,
+            imaging_rate=self.imaging_extractor.get_sampling_frequency(),
+        )
+        two_photon_series_metadata = metadata["Ophys"]["TwoPhotonSeries"][0]
+        two_photon_series_metadata.update(
+            description="The volumetric imaging data acquired from the Bruker Two-Photon Microscope.",
+            unit="px",
```
CodyCBakerPhD, line 90:
```suggestion
```

It's a fairly minor thing, but I'd say the units of 'pixels' are actually units of space; as previous in MICrONS they were used in reference to things like the estimated radius of pupil dilation w.r.t unknown physical units of space (say in 'cm') but rather in terms of known image space (number of pixels, 'px' which themselves would have unknown spatial equivalence 'n.a.')

Whereas the data of the two photon series is 'some amount of light' emitted from the sample and absorbed/detected by the camera over that pixels area. Also a very much unknown `n.a.` (the default of TwoPhotonSeries) until we start using Dmitri's measurement in terms of photon count per area over time

src/neuroconv/datainterfaces/ophys/brukertiff/brukertiffdatainterface.py
```diff
@@ -1,12 +1,127 @@
...
+            format="tiff",
```
CodyCBakerPhD, line 91:
```suggestion
```

I think this is for when using external mode, which is not Best Practice

CodyCBakerPhD:
Thanks! Sorry it took me so long to get to this...

The files look good for the most part; some comments about about minor metadata, and some minor remaining questions below

The file `bruker_dual_color_with_suite2p.nwb` has two devices, one properly named and one default, but the default doesn't seem to be linked to anything

Was this a previous (older) file or is it something that has since been fixed on the most recent version of this branch?

Also in `bruker_dual_plane_disjoint.nwb`, maybe I missed something, but I couldn't find any distinguishing information between the two disjoint planes (were they different depths?)

Also can you share or point towards an example of the multi-plane but contiguous?

weiglszonja:
Thank you @CodyCBakerPhD for checking the files! 
> 
> The file `bruker_dual_color_with_suite2p.nwb` has two devices, one properly named and one default, but the default doesn't seem to be linked to anything
> Was this a previous (older) file or is it something that has since been fixed on the most recent version of this branch?
> 
This was indeed an older version, I re-uploaded [stub_dual_color_imaging.nwb](https://s3.console.aws.amazon.com/s3/object/catalystneuro-processing?region=us-east-2&prefix=pinto-lab-to-nwb/stub_dual_color_imaging.nwb) which should have 1 properly named device linked to `ImagingPlaneCh1` and `ImagingPlaneCh2`
> Also in `bruker_dual_plane_disjoint.nwb`, maybe I missed something, but I couldn't find any distinguishing information between the two disjoint planes (were they different depths?)
> 
No, you're right. From the XML we know that the first plane "positionCurrent" parameter for the Z axis is `-130` and for the second plane it is `130`. But now I'm confused whether to interpret it as the first plane is at 130 microns depth and the second is 260.
> Also can you share or point towards an example of the multi-plane but contiguous?
Sure, [stub_dual_plane_imaging_contiguous.nwb](https://s3.console.aws.amazon.com/s3/object/catalystneuro-processing?region=us-east-2&prefix=pinto-lab-to-nwb/stub_dual_plane_imaging_contiguous.nwb) should have a default named `TwoPhotonSeries` with `(100, 512, 512, 2)` shape of data in it.

CodyCBakerPhD:
> No, you're right. From the XML we know that the first plane "positionCurrent" parameter for the Z axis is -130 and for the second plane it is 130. But now I'm confused whether to interpret it as the first plane is at 130 microns depth and the second is 260.

Based on the appearance of the nice looking contiguous file, I'd say make use of the `grid_spacing` and `field_of_view` arguments for the disjoint plane case; simply adjusted for the differences from being non-contiguous

Also can you add a CHANGELOG?

weiglszonja:
@CodyCBakerPhD files are in `bruker/`, look for `origin_coords` and `grid_spacing` in ImagingPlane, and `field_of_view` in `TwoPhotonSeries`

CodyCBakerPhD:
@weiglszonja Thanks so much! These files look amazing

Easily the best imaging files NeuroConv has produced

CodyCBakerPhD (approved):

PR #508, opened 2023-07-06
[Bug] fix test schema definition

CodyCBakerPhD:
Daily CI just caught an error caused by the latest release of `jsonschema` (7 hours ago) finding a mistake in a test case

Should be simple fix


CodyCBakerPhD:
Also looks like one of our external links, https://ibl.flatironinstitute.org/public is down, will keep an eye on if it's temporary or not

bendichter (approved):

PR #514, opened 2023-07-12
[Bug] Fix path format in source schema with Optional

CodyCBakerPhD:
@garrettmflynn found a minor issue with the source schema setter for certain interfaces - namely if the signature specifies a Union (which includes Optional) of one of our path types, then it's JSON 'format' won't get updated correctly

Has no effect on NeuroConv itself, but the GUIDE uses this JSON field to determine parameters of  drag-and-drop form component


garrettmflynn:
Works for me!

garrettmflynn (approved):
Works for me!

PR #517, opened 2023-07-24
Add FicTrac interface

h-mayorquin:
As in the title.

Moving from here https://github.com/catalystneuro/clandinin-lab-to-nwb/pull/1 as I managed to generate some examples file that allow to tests this independently of the conversion.

I added a gin data test but wainting on this:
https://gin.g-node.org/CatalystNeuro/behavior_testing_data/pulls/7

Meanwhile, I am sharing here so we can discuss two points here.

## The structure:

Everything in the interface is in radians, so everything is inside a `CompassDirection` multicontainer interface. For every grouping of variables, described here:
https://github.com/rjdmoore/fictrac/blob/master/doc/data_header.txt

I added a SpatialSeries in radians that then goes inside the `CompassDirection` multicontainer.

The result is something like this in the nwbfile:

image

##  Extraction framework / memory

Also, in the corresponding [conversion PR ](https://github.com/catalystneuro/clandinin-lab-to-nwb/pull/1) I was using polars because I was afraid the utf-8 conversion will be very memory heavy. After generating files myself I get a memory consumption of:

```
Memory per line  0.0002 Mib
 ```
Which leads to around:
```
 MiB 68.81103515625
```
Per hour.

This is with the type of fast camera [that is required](https://github.com/rjdmoore/fictrac/blob/9ac055e52d89f49f492a8eb4e1f7c5b8cd6df40a/README.md?plain=1#L37) for some insects (~100 fps).

While this is not small it might be a reasonable price to pay for using pandas which is a  more well known framework instead of taking the polars path. Let me know if you think it is better that I revert to polars.

In a further PR I will be adding the timing / aligning methods.

### CODE REVIEW

**Comment by h-mayorquin** [comment]
The files:
[fictrac_test.zip](https://github.com/catalystneuro/neuroconv/files/12175607/fictrac_test.zip)

**Comment by h-mayorquin** [comment]
And this is a figure should ease your understanding of the descriptions:
header_diagram1

**Comment by h-mayorquin** [comment]
I will add this to the gallery afte we merge this. As this is a big PR already.

File: src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py [hunk 1]
```diff
@@ -0,0 +1,523 @@
...
+
+        # Add integrated motion
+        integrated_motin_columns = [
+            "forward_motion_lab",
+            "side_motion_lab",
+        ]
+
+        description = (
+            "Integrated x/y position (radians) of the sphere in laboratory coordinates neglecting"
+            " heading. Equivalent to the output from two optical mice"
```
CodyCBakerPhD, line 365: Not entirely sure what is meant by the last part of this description, can you elaborate?

CodyCBakerPhD:
One minor comment about a description, everything else LGTM

I almost had a qualm about the speed field (usually a TimeSeries - which would then force us to use an extension to bundle all of these items together), but since it's in the same units and is simply normed amount in frame dimension, it should be fine with it's current representation

src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
h-mayorquin: Imagine the animal looking in a specific direction (that is the heading). Rotations about the z (yaw) axis of the animal can be integrated to obtain the animal’s heading direction in the current time frame. So you have the direction that the animal is moving towards.
Rotations in the two other two remaining axes can be give you the instataneous speed and if you integrate that in the direction of the heading get the x and y positions

image

(from the paper).

If you only integrate the rotations in the other remaining axis (wx and xy) you get the the position of the ball as it moves disregarding the heading. 

The reference to two optical mice comes from the documentation:

https://github.com/rjdmoore/fictrac/blob/9ac055e52d89f49f492a8eb4e1f7c5b8cd6df40a/doc/data_header.txt#L33-L36

And I think is added because capturing the original paper compares their estimation method to other methods that use two sensors from optical mice and they included that variable in the output for that comparison.

CodyCBakerPhD: Sorry I should have been more specific - the first sentence ('Integrated x/y position (radians) of the sphere in laboratory coordinates neglecting heading.') make sense

What I'm saying is the last statement 'Equivalent to the output from two optical mice' seems tacked on and and not defined what an 'optical mouse' is outside of the mental context of the original paper

Would you be OK with removing it?

CodyCBakerPhD: Or at the very least rephrasing to give more context (e.g., 'equivalent to approaches using dual head-mounted LEDs') or something

CodyCBakerPhD: Also this algorithm is applied to fly data, so is species generic, so specifically mentioning mouse subjects in the description is another thing about it that is a bit weird

h-mayorquin: I think it is fine to remove that addition. I also have the feeling that it confuses more than it clarifies.

h-mayorquin: Btw, it does not mean mice as species but as the stuff that you use to move your pointer in the computer. It is the optical sensor of the (computer) mice that the line is referring to.

CodyCBakerPhD: > Btw, it does not mean mice as species but as the stuff that you use to move your pointer in the computer.

Lol, ahhh.

> I also have the feeling that it confuses more than it clarifies.

Evidently 😆

CodyCBakerPhD (approved):

PR #520, opened 2023-07-25
[Feature] Add auxiliary interface for Intan

### LINKED ISSUE #519: [Feature]: IntanRecordingInterface - digital channels and aux channels

Currently the IntanRecordingInterface only reads in the analog channels for electrodes, and is not reading in the digital channels, aux channels or adc channels, which often have useful synchronization signals (similar to NIDQ interface). Adding in a feature to read in the additional channels would be useful.

_No response_

### Code of Conduct

CodyCBakerPhD:
fix #519 

Getting this started

Need to snip the shared example data and upload to the testing repo, then add tests here


h-mayorquin (commented):
Some general comments:
I think that the `BaseAuxiliaryExtractorInterface` would be better handled with composition and not inheritance. After adding the timing thing and the utility class to extract namings from spikeinterface (`BaseExtractorInterface`) our code is becoming harder to navigate.

This is a general recommendation of design as far as I am aware:
https://en.wikipedia.org/wiki/Composition_over_inheritance#:~:text=To%20favor%20composition%20over%20inheritance,and%20creating%20a%20family%20tree.

I also think that the overall general name is a code smell that this is not a good abstraction.

I also think that you should not use pyntan. I am not sure that @alejoe91 is mantining this anymore. When I asked him he mentioned that Sam did it and I find inconsistencies between the extraction there and the data extraction utilities that Intan Technologies provide:
https://intantech.com/downloads.html?tabSelect=Software

I think it would be better if we used those utilites instead.

CodyCBakerPhD:
> I think that the BaseAuxiliaryExtractorInterface would be better handled with composition and not inheritance.

From the article cited, the first and primary drawback from using a simpler Mixin is

> One common drawback of using composition instead of inheritance is that methods being provided by individual components may have to be implemented in the derived type

Since both methods in that class are specific to having already a loaded and set attribute `recording_extractor`, it only makes sense as a child class of the `BaseExtractorInterface` rather than re-implementing all of that in a separate class, which could enable possibility asynchrony between the two


> I also think that the overall general name ss BaseAuxiliaryExtractorInterface is a code smell that this is not a good abstraction.

What's your suggestion for a better name?


> I also think that we should not use pyntan. 

Yeah, I'd agree, but could be a separate PR - what is the pip installation name for the Intan supported tools? Would it be better if SI/neo just did it all for us (set a header attribute from the `read_rhd` which seems to be the only thing we need)

h-mayorquin:
You are right that in this case the proposed class needs the assumption about the recorder being already there. It was a premature reading of the situation by me. 

So If I understand well you are moving up the methods from `SpikeGLXNIDQInterface` because something similar will be needed for Intan. My first instinct is that is too early for creating another abstraction and I feel the unspecific name you came up with (Auxiliary) is an indication of this. I would just pay a little bit of code duplication and just have similar methods in intan. Duplication at the lower levels is cheaper in man hours and code complexity than having the wrong abstraction at the higher levels I think. 

That said, if there is something I am not seeing and you really want to go that way, I think that something like `EventRecordingInterface` or something along those lines would be better. At the end, it is just a wrapper to extract rising times from a specific channel (which will differ between extractors and maybe should be an attribute?) and could also contain the `get_rising_frames_from_ttl` as method that is only relevant in that context? Not very confident here.

As per the the pyntain issue, fair enough, there is no package fo that so you will just need to get the code and copy it to our code base as intan utilities. Against pyntan is the fact that it already confused  you by making you believe that the gain was part of the header format. We are making a tarde-off between convenience (for us) and being more likely to write wrong data to the nwb. I kind of think that the latter element should have more value but it is not a slam dunk as an argument.

CodyCBakerPhD:
> That said, if there is something I am not seeing and you really want to go that way, 

Yeah, we're potentially planning on having this for any format that could potentially have analog or digital channels


> EventRecordingInterface

Here's the justification/explanation for the naming

I interpret 'auxiliary' to mean 'either analog or digital, but definitely some continuous signal sampling at a resolution at or below the main neural acquisition traces'

'Events', such as those from blackrock, neuralynx, even some older NeuroScope, I interpret as being more of a collection of timestamps (list of individual timings)

Now, you can certainly parse events from TTL triggers or binary words (both examples of digital signals), but the idea is still that there is a full trace of the original information available and will be written as an ElectricalSeries (for provenance, especially important if complex parsing was involved), whereas an EventInterface (which may itself exist in NeuroConv one day for those formats) would map more onto a ndx-events structure I think, and would not usually have the source signal from which it was computed (if there was one at all)


> At the end, it is just a wrapper to extract rising times from a specific channel (which will differ between extractors and maybe should be an attribute?)

Exactly, a simple wrapper that only needs them to input the `channel_name` and calls the other trace access and tool function for them and returns a simple list of times that could theoretically be passed into the alignment methods

Though we should definitely expose a convenience wrapper for the `.get_channel_names()` (or `.get_property(key="channel_names")` often is more readable...) to make knowing which channel_name to use easier


> We are making a tarde-off between convenience (for us) and being more likely to write wrong data to the nwb. I kind of think that the latter element should have more value but it is not a slam dunk as an argument.

Can SpikeInterface or neo assume responsibility for guaranteeing this kind of accuracy so we don't have to worry about it (I don't care where they get the code as long as it's ground truth)? Or should we really just handle the header ourselves? What do you think? 😆

src/neuroconv/datainterfaces/ecephys/baseauxiliaryextractorinterface.py
```diff
@@ -0,0 +1,39 @@
...
+class BaseAuxiliaryExtractorInterface(BaseRecordingExtractorInterface):
+    """Parent class for all RecordingExtractorInterfaces."""
+
+    def get_channel_names(self) -> List[str]:
+        """Return a list of channel names as set in the recording extractor."""
+        return list(self.recording_extractor.get_channel_ids())
+
+    def get_event_times_from_ttl(self, channel_name: str) -> np.ndarray:
+        """
+        Return the start of event times from the rising part of TTL pulses on one of the NIDQ channels.
```
h-mayorquin, line 19: You probably want to change this docstring.

h-mayorquin:
Thanks for giving me more information about the situation. I see how Events is overloaded in the context of spikeinterface.

That said, I still have reservations about Auxiliar. Outside of the context of Intan is too general (It wasn't until I saw that you were moving you code from NIDQ that I got the code intent) and within the context of Intan it might not map well to its function (I am not sure all the auxiliary port in the controllers are used for this function). 

What about just `TTLExtractorInterface` ?

```python
class TTLExtractorInterface (BaseRecordingExtractorInterface):
    """Provides extra methods to locate and extract events from digial signals for extractors / formats that provide this feature."""
```

The main method is `get_event_times_from_ttl` anyway and I think it is easier to expand from something overly specific if something other than TLL is used to extract event timestamps in the future. `DigitalExtractorInterface` could also work.


> Though we should definitely expose a convenience wrapper for the .get_channel_names() (or .get_property(key="channel_names") often is more readable...) to make knowing which channel_name to use easier

Makes sense. I am hoping that we can constrain the output to only return the relevant channels in case the extractor combines both neural and the non-neural signals of interest.

> Can SpikeInterface or neo assume responsibility for guaranteeing this kind of accuracy so we don't have to worry about it (I don't care where they get the code as long as it's ground truth)? Or should we really just handle the header ourselves? What do you think? 😆

I actually don't know. Externalizing this is probably better for the community overall as we guarantee this are available for everyeone.  It also has the advantage that reduces our burden of maintenance which is not a small feat. But, as we have seen, other libraries have longer cycles of release and don't have provenance and metadata extraction as a high as a prioirty as we do. Plus, can we get them to commit? If it is us implementing that in their libraries I feel that we should build it here, get good tests and then push outwards once our needs are well specified. Not very confident here, just listing some trade-offs and ideas.

CodyCBakerPhD:
This PR will be stuck in draft until something happens with https://github.com/NeuralEnsemble/python-neo/pull/1249

> within the context of Intan it might not map well to its function (I am not sure all the auxiliary port in the controllers are used for this function).

This is true in both NIDQ and Intan from what I've seen; both can accept analog or digital signals

For the context of the `extract_ttl_times` function, only some of the digital signals count for strict on/off TTL parsing

AFAIK there is no way for the streams of neo/SI to know what type a given channel is, short of the name indicating something about that

Speculation: there might be something in the rich SpikeGLX metadata files about that setup, will look into it during that refactor

Actually, Intan might be even better about that since it has separate filenames for analogin, digitalin, etc... we shall see



> TTLExtractorInterface

The interface writes all the channels, not just the TTL ones. TTL parsing function may fail or be irrelevant for non-digital channels, it's just there as a convenience method when the user knows which channel they want to use

Even if we somehow were able to make a `DigitalExtractorInterface`, that also include streams with digital words for which the on/off TTL times aren't so meaningful as their actual (cleaned) voltage values


> I am hoping that we can constrain the output to only return the relevant channels in case the extractor combines both neural and the non-neural signals of interest.

If memory serves, I've not seen a case yet where neural signals are mixed with non-neural, short of the SY channel of AP/LF SpikeGLX (which we would parse out separately)

That's actually why NeuroScope extractor is designed to only return traces from the xml fields that indicate usage in spike detection, since the single binaries did occasionally include both



> If it is us implementing that in their libraries I feel is better to start here

As a matter of practice, I'd agree. Waiting on neo fixes anyway - given how close you are to both projects, perhaps you could help with that once you get back from leave?

zm711:
@CodyCBakerPhD 

Just saw this was linked to my Neo PR. Just wanted to drop my workaround here in case you want to move on while you wait for Neo to review my PR. You could just put this is as a side function in your class that can be called until Neo updates and then you could just delete this function and rely on the internal Neo functionality. 

Because the `memmap` requires the samples dedicated to the digital info you can actually just pull the digital info by hard coding the dict key as `"DIGITAL-IN"` as below:

```python
def _intan_neo_read_dig(self, reader: neo.rawio.IntanRawIO, time_slice=(None, None)) -> np.array:
        """
        Utility function that hacks the Neo memmap structure to be able to read
        digital events.

        Parameters
        ----------
        reader : neo.rawio.IntanRawIO
            The current file reader containing the memmap to the .rhd file.
        time_slice: tuple[start, stop]
            time slice of the data to analyze given in seconds with format (start, stop)
            None for start indicates start at 0, None for stop indicates go to end of
            recording

        Returns
        -------
        raw_digital_data : np.ndarray
            the raw digital data stored. Cannot be used. Must be processed first.

        """
        digital_memmap = reader._raw_data["DIGITAL-IN"]  # directly grab memory map from neo
        dig_size = digital_memmap.size
        dig_shape = digital_memmap.shape
        # below we have all the shaping information necessary
        if time_slice[0] is not None:
            i_start = int(np.rint(time_slice[0] * self.sample_frequency)) # wherever you get your sample frequency/rate from
        else:
            i_start = 0
        if time_slice[1] is not None:
            i_stop = int(np.rint(time_slice[1] * self.sample_frequency))
        else:
            i_stop = dig_size

        # this code is based on Neo's code for creating the array in their get_analogsignal_chunk function
        block_size = dig_shape[1]
        block_start = i_start // block_size
        block_stop = i_stop // block_size + 1

        sl0 = i_start % block_size
        sl1 = sl0 + (i_stop - i_start)

        raw_digital_data = np.squeeze(digital_memmap[block_start:block_stop].flatten()[sl0:sl1])
        # raw_digital_data is just one vector of all info
        return raw_digital_data
```

This is just one vector of all digital data. So in order to process the data you can either prompt the user for which digital channel they used or just iterate through all possible digital channels something like this (because Neo doesn't currently save the ids of the digital channels, if it did you could just iterate through the actual channels used rather than all of them, but this is another addition to Neo that I'm working on):

```python
for value in range(16): # 16 is standard number of possible Intan dig in channels
            final_digital_data[value, :] = np.not_equal(np.bitwise_and(raw_digital_data, (1 << value),),0,)
```

Whichever rows have data are the digital in channels which have data (ie if row 3 is all zeros then row 3 was not on during the recording, but if row 5 has some 1's in it then it was utilized during the recording. Then you can just do a rising-edge, falling edge detector on this matrix to get all your events. Of note, my current PR does not address the conversion from raw digital to this final digital (which would be a matrix of 1s and 0s).

h-mayorquin:
If we are going to have this I don't think that we can assume that the channels of the digital interface for intan should be written as a recording interface that is for ecephys. In the cases that I have seen those digital channels do not come from electrodes.

CodyCBakerPhD:
Are you talking about `ndx-extracellular-channels`? This old PR was developed under core models, so if it's in volts then it's an ElectricalSeries, and channels would still be channels, just in their own group annotated to not be neural

h-mayorquin:
Those signals are not in volts:
https://github.com/h-mayorquin/python-neo/blob/83a04080f390da593fa909e716edcf000cf17b27/neo/rawio/intanrawio.py#L1012-L1042

But good point, in the old schema the electrode table is actually a channel table and it works. Point taken.

CodyCBakerPhD:
What does unit 'TTL' mean? What does this data look like? TTL are usually truly in volts on the underlying system even if you've normalized them to just be on/off (such as 0 and 5 V)

CodyCBakerPhD:
An example of a non-ElectricalSeries version would be if the conversion factors for an analog signal (harder to imagine it for digital) already put it into some other non-electrical unit such as temperature, pressure, etc.

zm711:
We put in TTL to emphasize the units were meaningless. Intan encodes them as a single vector of values that require the demultiplexing. So not voltages, but just multiplexed 1s and 0s.  Here's the official documentation:

image

https://intantech.com/files/Intan_RHX_user_guide.pdf

zm711:
If you'd really prefer voltages (although I would recommend against it). The TTL logic for intan is 0-.8V for TTL 0 and 2.0-5.0 V for TTL 1.

CodyCBakerPhD:
> If you'd really prefer voltages (although I would recommend against it). The TTL logic for intan is 0-.8V for TTL 0 and 2.0-5.0 V for TTL 1.

Why do you recommend against that? In NWB that would simply correspond to

TTL 0: `conversion=0.8`
TTL 1: `conversion=3.0, offset=2.0`

zm711:
Because you can't actually say that. TTL low/0 can be any value between 0 and 0.8 volts and any value between 2 and 5 volts for TTL high/1. So you can go from volts to the TTL logic for Intan, but you can't get the "real" voltage based on the TTL. Intan only says the binary value. So any voltage you say for the TTL would be an arbitrary choice. For example you could say 3 V like your example, but maybe a lab uses a 4V pulse and so would be confused why your interface is saying 3V when they gave 4V.

CodyCBakerPhD:
I misunderstood what you were saying; the 'TTL 0' and 'TTL 1' modes are just ranges of allowed voltages, but no record of the original digital traces remains, only the version filtered to 0 or 1?

zm711:
Exactly! Only 1s and 0s remain. When you look at the recording trace on intan it also only shows up as 1s and 0s so I think their digital inputs convert it at the hardware level so there is no way to recover the info.

h-mayorquin:
I will take care of this in another PR.

PR #523, opened 2023-07-25
Allow additional metadata to be saved when using `expand_path`

tuanpham96:


src/neuroconv/tools/path_expansion.py
```diff
@@ -64,23 +64,31 @@ def expand_paths(self, source_data_spec: Dict[str, dict]) -> List[DeepDict]:
...
-                            source_data["base_directory"], path
-                        )  # return the absolute path
-                        if "session_id" in metadata:
-                            out[key]["metadata"]["NWBFile"]["session_id"] = metadata["session_id"]
-                        if "session_start_time" in metadata:
-                            out[key]["metadata"]["NWBFile"]["session_start_time"] = metadata["session_start_time"]
-                        if "subject_id" in metadata:
-                            out[key]["metadata"]["Subject"]["subject_id"] = metadata["subject_id"]
+                if path_type not in source_data:
+                    continue
```
CodyCBakerPhD, line 80: Thanks for reducing our cyclomatic complexity here 😉

CodyCBakerPhD (approved):

CodyCBakerPhD:
@tuanpham96 Thanks for the great PR! Especially thanks for the nice demonstrative tests

LGTM assuming the tests pass

CodyCBakerPhD:
pre-commit.ci run

CodyCBakerPhD:
Huh. pre-commit shows as having run fine on the users fork (not an external organization, which is known not to work...): https://github.com/catalystneuro/neuroconv/compare/main...tuanpham96:neuroconv:main

But still shows as failing here, though when I navigate to the log it says it says it reformatted things...

Although, the commit shown on the users fork does not show up in this PR... Especially odd... Did SI have to do anything special to get the bot to work on external forks @h-mayorquin?

src/neuroconv/tools/path_expansion.py
```diff
@@ -64,23 +64,31 @@ def expand_paths(self, source_data_spec: Dict[str, dict]) -> List[DeepDict]:
         ... )
         """
 
-        session_keys = {"session_start_time", "session_id", "subject_id"}
+        # session_keys = {"session_start_time", "session_id", "subject_id"}
```
h-mayorquin, line 67:
```suggestion

```
Do we need this?

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -104,6 +104,150 @@ def test_expand_paths(tmpdir):
     assert len(string_directory_out) == len(expected)
 
 
+def test_expand_paths_with_extras(tmpdir='tmp'):
```
h-mayorquin, line 107: We use pytest in our testing suite and we have it in the requirements as far as I remember. Would it not be more robust to use their dedicated fixtures instead of an argument?

https://docs.pytest.org/en/6.2.x/tmpdir.html

h-mayorquin:
@CodyCBakerPhD 
Not really. Not sure if the colabators should enable the bot? For example, I don't think that Sam activated this in its repo.

I can't recall.

src/neuroconv/tools/path_expansion.py
tuanpham96: I don't think so? I got rid of that in the recent commit.

tests/test_minimal/test_tools/test_expand_paths.py
tuanpham96: Ah, thank you! I didn't know about this feature. I go rid of the arguments.

CodyCBakerPhD:
@h-mayorquin It's magically working now 😂

Guess they just had to push another commit on top of the bot

PR #527, opened 2023-07-25
Need to change filtering all zero data from `add_fluorescence_traces`

weiglszonja:
Commenting out is not a solution, we'll need to do that part more efficiently.
https://github.com/catalystneuro/neuroconv/issues/528

### DIFF
--- a/tests/test_ophys/test_tools_roiextractors.py
+++ b/tests/test_ophys/test_tools_roiextractors.py
@@ -835,8 +835,8 @@ class TestAddFluorescenceTraces(unittest.TestCase):
         segmentation_extractor = generate_dummy_segmentation_extractor(
             num_rois=self.num_rois,
             num_frames=self.num_frames,
-            num_rows=self.num_rows,
-            num_columns=self.num_columns,
+            num_rows=300,
+            num_columns=400,
             has_raw_signal=True,
             has_deconvolved_signal=False,
             has_neuropil_signal=False,
@@ -924,8 +924,8 @@ class TestAddFluorescenceTraces(unittest.TestCase):
         ophys = get_module(self.nwbfile, "ophys")
         roi_response_series = ophys.get(self.fluorescence_name).roi_response_series
 
-        assert "Deconvolved" not in roi_response_series
-        self.assertEqual(len(roi_response_series), 2)
+        # assert "Deconvolved" not in roi_response_series
+        self.assertEqual(len(roi_response_series), 3)
 
     def test_no_traces_are_added(self):
         """Test that no traces are added to the nwbfile if they are all zeros or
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD:
@weiglszonja Looks like some minor testing behavior (related to assuming this feature works) broke, can you also comment those out with notes to re-enable when we think of a more efficient way to test it? Or we can discuss removing the feature entirely

For context, this was for a conversion of ~1G single-plane single-channel Suite2P data which took a surprisingly long time, hit processing snags at both the calculation of `roi_centroids` and `acceptance/rejection` lists, as well as writing the image mask...

We should try to get our hands on a large-sized session of Suite2P (not to be saved in the repo or CI, just for personal uses) to see if we can do anything to optimize our speeds, maybe as a part of the work @pauladkisson will get around to with ROIExtractors in coming months

weiglszonja:
Following @h-mayorquin suggestion:

```
# Filter empty and all zero data
    traces_to_add = {trace_name: trace for trace_name, trace in traces_to_add.items() if trace is not None and np.any(trace != 0)}
```

I tested with 4 dummy traces of 1.2GB in size (4000 rois x 40000 frames), this line executes in `26.5396130209` seconds

@CodyCBakerPhD are we happy with that or is there something else I should be looking at?

weiglszonja:
As discussed with @CodyCBakerPhD for now we are leaving this commented out and come back to this problem

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -924,8 +924,8 @@ def test_add_fluorescence_one_of_the_traces_is_all_zeros(self):
         ophys = get_module(self.nwbfile, "ophys")
         roi_response_series = ophys.get(self.fluorescence_name).roi_response_series
 
-        assert "Deconvolved" not in roi_response_series
-        self.assertEqual(len(roi_response_series), 2)
+        # assert "Deconvolved" not in roi_response_series
```
CodyCBakerPhD, line 927:
```suggestion
        # assert "Deconvolved" not in roi_response_series  # TODO: re-enable when/if we add smart zero-array skipping
```

CodyCBakerPhD (approved):

CodyCBakerPhD:
In summary

We want to skip writing of a RoiResponseSeries that is all zeros, most often occurring with the deconvolved responses

The original generator works best due to early return; it only loads a single value into memory at a time, only makes a single assertion at a time, and stops upon seeing the first non-zero value

For a sparse deconvolved series as we saw during the User days, this caused in incredible slow down on @seethakris's data

The numpy vectorized method of course works the quickest to make the assertion over the entire array, but is not lazy/iterated access and so loads the entire trace into memory to create the boolean array that `np.any` asserts on

I had also hoped that `np.any` had early-return functionality, but a quick test indicates no

```python
from timeit import timeit

num_trials = 10_000

timeit(setup="import numpy as np; arr = np.concatenate([np.zeros(shape=1000000), np.ones(shape=1)])", stmt="np.any(arr)", number=num_trials) / num_trials
> 0.00035983887500042327

timeit(setup="import numpy as np; arr = np.concatenate([np.zeros(shape=100), np.ones(shape=1)])", stmt="np.any(arr)", number=num_trials) / num_trials
> 1.5874583012191578e-06
```

PR #531, opened 2023-07-26
Set more efficient default for mask type

CodyCBakerPhD:
With `mask_type='image'` a relatively big segmentation dataset took forever and worse, stalled out in unstoppable ways (unless you completely kill the process from outside)

But swapping to `pixel` worked instantaneously. Proposing to set it as the default for both speed and reduced memory usage


bendichter:
I suppose this is fine. Let's add an entry to the README though since this will change behavior for some users.

CodyCBakerPhD:
Actually it could be a breaking change now that I think about it

`image` always work, but is the most inefficient for a large number of ROIs and/or large image resolution

`pixels` only works for 2D (non volumetric) imaging, so though it's more efficient specifying it in this way would break all volumetric pipelines that don't already override the option

Maybe a better distinction for the argument could be `full` or `sparse` (with `sparse` auto-detecting `pixel` or `voxel` appropriately based on image dimensions)

h-mayorquin:
Closing. It is a simple change, we would handle this if it becomes a priority again.

PR #532, opened 2023-07-26
Allow to add timestamps into DLC interface 

tuanpham96:
This allows users to inject a timestamps vector that users know to trust more than the inferred timestamps from videos, by DLC2NWB.

Rationale: For example, when 
(1) the videos cannot be found (e.g. files being transferred around and paths are incorrect), 
(2) the time information from the videos might be unreliable, 
(3) experiments were performed in discontinuous chunks but the videos do not have any metadata indication that,
(4) a more reliable timestamp recording is available

Note: there are currently no tests for this yet. I'm not sure where it should be


src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -49,6 +60,7 @@ def __init__(
         file_path: FilePathType,
         config_file_path: FilePathType,
         subject_name: str = "ind1",
+        timestamps: Optional[Union[List, np.ndarray]] = None,
         verbose: bool = True,
     ):
```
CodyCBakerPhD, line 65: Since the big `TemporalAlignmentInterface` revamp, we've been trying to move all `timestamps` references outside of `__init__` and into the new workflow, which is to set and get them using these methods: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/basetemporalalignmentinterface.py#L11-L56

What I would recommend here is to 

a) remove `timestamps` from `__init__`, BUT go ahead and set/reserve a `._timestamps` attribute that by default is `None` and uses the existing behavior (to use DLC timestamps) during write procedure

b) allow setting of `._timestamps` to a user defined vector using the alignment methods, and if `._timetamps` is not `None` during write procedure then use it instead of DLC timestamps (just as you have injected it into the `write_subject_to_nwb` function, which is perfectly fine)

@tuanpham96 What do you think?

CodyCBakerPhD: Yep, you got it!

CodyCBakerPhD:
Code looks good

Need

i) CHANGELOG entry

and

ii) tests

to get this finished

> Note: there are currently no tests for this yet. I'm not sure where it should be

I would say a simple test in this class: https://github.com/catalystneuro/neuroconv/blob/main/tests/test_behavior/test_video_interface.py#L82-L97

would be sufficient; set the timestamps to something specific using that method and a normal workflow, then read the file back to ensure what you specified was written successfully (and read back in) to/from disk

Let me know if any particular questions

CodyCBakerPhD:
@tuanpham96 Quick ping regarding those two last items

tuanpham96:
@CodyCBakerPhD sorry this slipped my mind. 

Regarding test, I was looking at `test_video_interface.py` and was confused how this would fit, since there wasn't any `DeepLabCutInterface` import in that test file. Wouldn't it be more appropriate to test in `test_behavior_interfaces.py`, particularly these lines https://github.com/catalystneuro/neuroconv/blob/main/tests/test_on_data/test_behavior_interfaces.py#L53-L76 ?

And is there a structure/format for updating changelog?

tuanpham96:
@CodyCBakerPhD the 2 tasks are done now, could you take a look? thanks!

CodyCBakerPhD (approved):

CodyCBakerPhD:
@tuanpham96 Fantastic, thanks so much!

PR #536, opened 2023-07-30
Pickle memory maps in SliceableDataChunkIterator

CodyCBakerPhD:
sister PR to https://github.com/hdmf-dev/hdmf/pull/924 and https://github.com/hdmf-dev/hdmf-zarr/pull/111

This acts as a first test implementation by allowing efficient pickling of a memory-map based iterator. Will add and test more conditions over time

Also, to be clear, a `np.memmap` (and Zarr arrays for that matter) are pickleable by default; however, they do it by pickling the data they contain, not a packaged set of instructions for re-instantiating that object after reloading (in our case, on a separate process)

So we want to override such methods to more efficiently allow re-initialization of the lazy mappings to the source data

Opening in draft for now


src/neuroconv/basedatainterface.py
```diff
@@ -79,7 +79,8 @@ def run_conversion(
             warnings.warn(
                 "Using DataInterface.run_conversion() without specifying nwbfile_path is deprecated. To create an "
                 "NWBFile object in memory, use DataInterface.create_nwbfile(). To append to an existing NWBFile object,"
-                " use DataInterface.add_to_nwbfile()."
+                " use DataInterface.add_to_nwbfile().",
+                stacklevel=2,
```
CodyCBakerPhD, line 83: Found this useful in a separate conversion project; kept getting the warning but was unsure where it was coming from (had several data interfaces in the conversion)

src/neuroconv/tools/hdmf.py
```diff
@@ -1,11 +1,14 @@
...
 import numpy as np
-from hdmf.data_utils import GenericDataChunkIterator as HDMFGenericDataChunkIterator
 
+# from hdmf.data_utils import GenericDataChunkIterator as HDMFGenericDataChunkIterator
+from hdmf.data_utils import GenericDataChunkIterator
 
-class GenericDataChunkIterator(HDMFGenericDataChunkIterator):
+
+# class GenericDataChunkIterator(HDMFGenericDataChunkIterator):
+class _GenericDataChunkIterator(GenericDataChunkIterator):
```
CodyCBakerPhD, line 11: This was a temporary reversion, I think I can restore it - just need to test it doesn't break the serialization

src/neuroconv/tools/hdmf.py
```diff
@@ -90,3 +95,42 @@ def _get_maxshape(self) -> tuple:
 
     def _get_data(self, selection: Tuple[slice]) -> np.ndarray:
         return self.data[selection]
+
+    def is_pickleable(self) -> bool:
+        if isinstance(self.data, (np.memmap, np.array)):
```
CodyCBakerPhD, line 100: Will need to remove `np.array` here; easy support is only offered for memory-map like lazy access to source data

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1143,8 +1143,8 @@ def add_units_table(
 
         extended_data[indexes_for_default_values] = default_value
         # Always store numpy objects as strings
-        if np.issubdtype(extended_data.dtype, np.object_):
-            extended_data = extended_data.astype("str", copy=False)
+        # if np.issubdtype(extended_data.dtype, np.object_):
+        #    extended_data = extended_data.astype("str", copy=False)
```
CodyCBakerPhD, line 1147: TODO: revert

h-mayorquin:
Closing this. I know how to do it. I will come back to be if it needs be.

PR #537, opened 2023-07-31
Use `ndx-probeinterface` to handle devices from SpikeInterface

alejoe91:
Instead of adding a default Device, the `add_devices` in the `spikeinterface` tools, will add `ndx_probeinterface.Probe` devices, which is specific to `probeinterface` objects.

Still WIP, needs some polish on the extension side.

### TODO 
- [ ] Add tests


h-mayorquin:
@alejoe91 are you OK with closing this now that we have:

https://github.com/catalystneuro/ndx-extracellular-channels

(now that we have integrated it already but we should probably use that)

PR #538, opened 2023-08-04
Abf electrode metadata fixes

CodyCBakerPhD:
@luiztauffer A user of the AbfInterface ran across some problems with custom metadata: https://github.com/NeurodataWithoutBorders/helpdesk/issues/53

I think I've fixed the issue; also added a test for it


luiztauffer (approved):
@CodyCBakerPhD looks good to me, I tried it in a couple of icephys conversions here and it worked well

PR #539, opened 2023-08-07
Pin ROIExtractors temporarily

CodyCBakerPhD:
Latest release of ROIExtractors split the BrukerTifExtractor into distinct categories (multi-channel, multi-plane, volumetric) that can't be automatically routed through a common class; so the current interface only works using the previous version of ROIExtractors

https://github.com/catalystneuro/neuroconv/pull/507 will loosen this pin as a part of its expansion


weiglszonja (approved):
LGTM, thanks for keeping an eye on it!

PR #546, opened 2023-08-17
Automatic pagination

CodyCBakerPhD:
Simple injection for automatic pagination through NeuroConv; came up in discussion with Paul earlier

Should this be exposed as an argument, i.e., optional behavior (even if enabling it is still set as the default)


CodyCBakerPhD:
Will work this into the backend refactor as configurable options for HDF5 backend

There appear to be fairly complicated conditions as for when pagination is ideal w.r.t. number and size of datasets in the files that we also have yet to discover as far as recommendations

PR #550, opened 2023-08-21
Fix tests on main

CodyCBakerPhD:
Latest HDMF release raised default chunk sizes which broke a couple of tests on `main`

Simple fixes, all


weiglszonja (approved):
LGTM

PR #551, opened 2023-08-22
Fix `subset_sorting` when sorting interface has recording registered

felixp8:
In `BaseSortingExtractorInterface.subset_sorting()`, if the sorting interface has recording registered, there's a chance that `1.1 * max_min_spike_time` is greater than the number of samples in the recording, which causes this [check](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/frameslicesorting.py#L51) to fail in SpikeInterface. Fixed by capping `end_frame` at the number of samples in the recording if one is registered.


h-mayorquin:
Why do we have the 1.1 at all?

CodyCBakerPhD:
> Why do we have the 1.1 at all?

I believe the reasoning was just to get a 'little bit' more spike timing data from all units. Clipping at the latest first spike time was otherwise usually too early to really see any spiking behavior

Out of all the subsetting methods this is the most heuristic, I suppose due to the irregular stochastic nature of the data type itself. Open to other ideas

CodyCBakerPhD:
I suppose the best interpreted setting of all stubbing anywhere would to report 'some small static amount (~10 MB per data stream?) of data that is likely some smaller percentage of the total data size'

So calculating how many spikes that is, equating that to how many frames gets you that many spikes, might be the best way to go

Might be something easier to do in the new backend refactor

CodyCBakerPhD:
@felixp8 Can you think of a way to make a test for this?

CodyCBakerPhD:
Great, thanks!

PR #555, opened 2023-08-28
[Backend Config II]: Add dataset configuration tools

CodyCBakerPhD:
Leaving integration with the interfaces/converter for separate follow

This will focus on introducing the basic tools and thoroughly testing them

Big focus on making this GUIDE-friendly in a schematic way to make that easy to integrate down the line


CodyCBakerPhD:
replaced by #569

PR #563, opened 2023-09-12
Improve sonpy installation for M1 Mac

CodyCBakerPhD:
As discovered by @rly on https://github.com/NeurodataWithoutBorders/nwb-guide/issues/354, despite the fact that neuroconv as a package works fine due to our previous lazy interface + exclusion principles at the interface level, it still attempts to install it (even seems to grab some files from the PyPI package) but fails at a point during the environment build process

This excludes it entirely from being bundled with NeuroConv in this situation

Also some docstring improvements

Unfortunately no way to test this in CI

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,5 @@
+import sys
+import platform
 from collections import defaultdict
 from pathlib import Path
 from shutil import copy
@@ -46,7 +48,7 @@ if not gin_config_file_local.exists():
     copy(src=gin_config_file_base, dst=gin_config_file_local)
 
 # Bug related to sonpy on M1 Mac being installed but not running properly
-if sys.platform == "darwin" and processor() == "arm":
+if sys.platform == "darwin" and platform.processor() == "arm":
     extras_require.pop("sonpy")
     extras_require["ecephys"].remove(
         next(requirement for requirement in extras_require["sonpy"] if "sonpy" in requirement)
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,12 @@ gin_config_file_local = Path("./tests/test_on_data/gin_test_config.json")
 if not gin_config_file_local.exists():
     copy(src=gin_config_file_base, dst=gin_config_file_local)
 
+# Bug related to sonpy on M1 Mac being installed but not running properly
+if sys.platform == "darwin" and processor() == "arm":
+    extras_require.pop("sonpy")
+    extras_require["ecephys"].remove(next(requirement for requirement in extras_require["sonpy"] if "sonpy" in requirement))
+    extras_require["full"].remove(next(requirement for requirement in extras_require["sonpy"] if "sonpy" in requirement))
+
 setup(
     name="neuroconv",
     version="0.4.3",
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,5 @@
-import sys
 import platform
+import sys
 from collections import defaultdict
 from pathlib import Path
 from shutil import copy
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -45,6 +47,15 @@
 if not gin_config_file_local.exists():
     copy(src=gin_config_file_base, dst=gin_config_file_local)
 
+# Bug related to sonpy on M1 Mac being installed but not running properly
+if sys.platform == "darwin" and platform.processor() == "arm":
```
h-mayorquin, line 51: For my own curiosity, why would the difference between `platform.processor()` and `plataform.machine()` would be?

src/neuroconv/datainterfaces/ecephys/spike2/spike2datainterface.py
```diff
@@ -10,7 +10,7 @@ def _test_sonpy_installation() -> None:
     get_package(
```
h-mayorquin, line 10: For my own curiosity, how did we ended up with this test here for `spike2`? This is exclusive for this datainterface?

h-mayorquin (approved):
As you mentioned -and I followed the linked threads- this was tested by Ryan and I guess will keep tested. 

This looks fine to me.

I am afraid that the complexity of our setup is increasing slowly but surely. I am the [pytoml standard dependency specification](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#example) would have helped us with this build problem.

Example:
https://github.com/Julian/dotfiles/blob/main/.local/share/virtualenvs/virtualenvs.toml#L27

Relevant links for the expressiveness of the grammar to see what kind of problems we might face (hopefully not!) in the future with build:
https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/
https://packaging.python.org/en/latest/specifications/dependency-specifiers/?highlight=platform_machine#environment-markers

That said, this is for the building / wheels  level. I am not sure we would get the nice error messages that we have had we gone this way?

Does the grammar adopted in the

setup.py
CodyCBakerPhD: On M1:

`platform.machine() = "arm64"`
`platform.processor() = "arm64"`

On my Windows, the `platform.processor()` is much more verbose about model number and specs from Intel, and machine is just `"AMD64"`

So if we ever need to extend, possibly `machine` might generalize cleaner but its similar to the older conversations about `sys.platform` syntax vs. all the other ways of getting string representations of platforms

I do prefer the .toml language however for handling all of that

src/neuroconv/datainterfaces/ecephys/spike2/spike2datainterface.py
CodyCBakerPhD: You mean the architecture subkey, or the `get_package` helper in general? Both have been around a while so they must have been for something else. Useful in general for any external import that has trouble with certain Python versions or platforms

CodyCBakerPhD:
Thanks, it's nice to have good .toml examples at that level of complexity

I'd also like to switch to .toml at some point as it will make our dependencies much more explicit

setup.py
h-mayorquin: Thanks, I just tested and they are the same in my system:

```python
In [2]: import platform

In [3]: platform.machine()
Out[3]: 'x86_64'

In [4]: platform.processor()
Out[4]: 'x86_64'
```

src/neuroconv/datainterfaces/ecephys/spike2/spike2datainterface.py
h-mayorquin: I realized I quote the wrong line. I am talking about the `_test_sonpy_installation()` function as a whole.

PR #557, opened 2023-09-01
[Backend Config III]: Configure DataIO for datasets of in-memory NWB

CodyCBakerPhD:
WIP for implementing the configuration of appropriate DataIO parameters given a backend specification

Needs a lot more tests

PR #549, opened 2023-08-21
[Backend Config I]: Add helper for finding candidate Datasets in-memory

CodyCBakerPhD:
Opening in Draft

First of many PRs breaking #475 into bite-size pieces with unit tests

This piece in particular differs in a couple small ways related to debugging on DynamicTable columns

Still needs some more tests on files opened via append mode as well as some 'combination' examples with `timestamps` and indexed columns. Also maybe columns added to a table explicitly as VectorData, etc... (should work in principle but hey that's what tests are for)

PR #568, opened 2023-09-18
[Backend Configuration Ib] Pydantic models for communicating backend-specific configuration information

CodyCBakerPhD:
Expands on #567

Harnesses the dataset models and narrows the specifications down to a minimal user-friendly structure for specifying the association between all of the datasets in the file-to-be and the backend of choice

If curious, the actual original workflow here was starting from #549 -> #555 -> #557 and then realizing that integrating the original models required so many modifications to the structure that it was worth it to rebase everything from its final debugged form


CodyCBakerPhD:
@h-mayorquin This is also the first PR in the chain that cannot be merged to `main` until the HDMF-Zarr compatibility fixes - but we can at least hash out any changes well ahead of that time so there's no delay

src/neuroconv/tools/nwb_helpers/_models/_hdf5_models.py
```diff
@@ -74,3 +75,18 @@ def get_data_io_keyword_arguments(self) -> Dict[str, Any]:
...
 
         return dict(chunks=self.chunk_shape, **compression_bundle)
+
+
+class HDF5BackendConfiguration(BackendConfiguration):
+    """A model for matching collections of DatasetConfigurations specific to the HDF5 backend."""
+
+    backend: Literal["hdf5"] = Field(
+        default="hdf5", description="The name of the backend used to configure the NWBFile."
+    )
```
bendichter, line 85: Wouldn't HDF5BackendConfiguration already imply that the backend is hdf5? It looks like you are making it possible to set to some other value

src/neuroconv/tools/nwb_helpers/_models/_hdf5_models.py
```diff
@@ -74,3 +75,18 @@ def get_data_io_keyword_arguments(self) -> Dict[str, Any]:
...
+    data_io_class: Type[H5DataIO] = Field(default=H5DataIO, description="The DataIO class that is specific to HDF5.")
```
bendichter, line 86: Similarly here, wouldn't HDF5 imply H5DataIO?

src/neuroconv/tools/nwb_helpers/_models/_hdf5_models.py
CodyCBakerPhD: I'm usings `Field` because that enhances the JSON schema output of the Pydantic model, which is what the GUIDE would use to make those little info icons

The `default` of a `Field` is equivalent syntax to `variance_name = "default value"`

Otherwise largely for consistency with the parent class where the choice is between string options "hdf5" and "zarr"

Though I will note that the workflow used later on would rarely have users instantiate these objects themselves - the convenience functions return it for them, so most of this is just for internal dataclass consistency

src/neuroconv/tools/nwb_helpers/_models/_hdf5_models.py
CodyCBakerPhD: Setting the `default` does set it as the value for that attribute; but again, users would never actually set this, it's mostly for internal convenience to have an object `<SomeBackend>BackendConfiguration` that reliably maps `.backend` as a short human readable string reference together with the actual `.data_io_class` which would be the actual thing you wrap the data data values in with `set_data_io`

Pretty much equivalent to having a dictionary `backend_to_data_io_class = dict(hdf5)` but a little more descriptive and extensible (primarily with respect to the list of contained DatasetConfigurations)

src/neuroconv/tools/nwb_helpers/_models/_hdf5_models.py
h-mayorquin: Datclasses have `init=False` to explichttps://docs.python.org/3/library/dataclasses.html#dataclasses.fieldtly indicate that something is not mean to be passed to the `__init__` method:

https://docs.python.org/3/library/dataclasses.html#dataclasses.field

Doesn't pydantic support something similar?

src/neuroconv/tools/nwb_helpers/_models/_zarr_models.py
```diff
@@ -120,3 +122,26 @@ def get_data_io_keyword_arguments(self) -> Dict[str, Any]:
...
+    )
+    dataset_configurations: Dict[str, ZarrDatasetConfiguration] = Field(
+        description=(
+            "A mapping from object locations to their ZarrDatasetConfiguration specification that contains all "
+            "information for writing the datasets to disk using the Zarr backend."
+        )
+    )
+    number_of_jobs: int = Field(
+        description="Number of jobs to use in parallel during write.",
+        ge=-psutil.cpu_count(),  # TODO: should we specify logical=False in cpu_count?
```
h-mayorquin, line 144: Mmm I think that the heuristic is that `logical=False` should be enabled if there is heavy processing but in this case is mostly IO for which you want all your logical cores.

src/neuroconv/tools/testing/_mock/_mock_dataset_models.py
```diff
@@ -55,3 +62,49 @@ def mock_ZarrDatasetConfiguration(
         filter_methods=filter_methods,
         filter_options=filter_options,
     )
+
+
+def mock_HDF5BackendConfiguration() -> HDF5BackendConfiguration:
+    """Mock instance of a HDF5BackendConfiguration with several NeuroPixel-like datasets."""
```
h-mayorquin, line 68:
```suggestion
    """Mock instance of a HDF5BackendConfiguration with two NeuroPixel-like datasets."""
```
?

src/neuroconv/tools/nwb_helpers/_models/_hdf5_models.py
CodyCBakerPhD: Easily, by simply overriding the `__init__` to exactly specify the signature. Would you prefer that?

The point is that these are more class attributes than they are initialized values of the dataclass, though once you have an instance it wouldn't matter either way

src/neuroconv/tools/nwb_helpers/_models/_hdf5_models.py
```diff
@@ -74,3 +75,18 @@ def get_data_io_keyword_arguments(self) -> Dict[str, Any]:
             compression_bundle = dict(compression=self.compression_method, compression_opts=self.compression_options)
 
         return dict(chunks=self.chunk_shape, **compression_bundle)
+
+
+class HDF5BackendConfiguration(BackendConfiguration):
```
h-mayorquin, line 80: Can you get away with just using instances of `BackendConfiguration`, why do you need a class for `HDF5BackendConfiguration` and another for `ZarrBackendConfiguration`?

src/neuroconv/tools/nwb_helpers/_models/_zarr_models.py
```diff
@@ -120,3 +122,26 @@ def get_data_io_keyword_arguments(self) -> Dict[str, Any]:
...
+        le=psutil.cpu_count(),
+        default=-2,  # -2 translates to 'all CPU except for one'
```
h-mayorquin, line 146: In the case someone has 4 logical cores this could be:

`-4, -3, -2, -1, 0, 1, 2, 3, 4` 

Questions:
What do negative numbers mean?
Why is -2 equivalent to 3? as per your comment in the default?
What does 0 mean?
 
is this a convention from somewhere?

h-mayorquin (commented):
Took the first look, left some comments.

src/neuroconv/tools/nwb_helpers/_models/_hdf5_models.py
CodyCBakerPhD: Discussed in person: summary, it would require us to use a root validator for the common parent, which is a bit hackier

This works cleaner at the price of some redundancy

src/neuroconv/tools/nwb_helpers/_models/_zarr_models.py
CodyCBakerPhD: TODO: need to document negative slicing better

src/neuroconv/tools/nwb_helpers/_models/_zarr_models.py
```diff
@@ -120,3 +122,26 @@ def get_data_io_keyword_arguments(self) -> Dict[str, Any]:
...
+        default=ZarrDataIO, description="The DataIO class that is specific to Zarr."
+    )
+    dataset_configurations: Dict[str, ZarrDatasetConfiguration] = Field(
+        description=(
+            "A mapping from object locations to their ZarrDatasetConfiguration specification that contains all "
+            "information for writing the datasets to disk using the Zarr backend."
+        )
+    )
+    number_of_jobs: int = Field(
+        description="Number of jobs to use in parallel during write.",
```
CodyCBakerPhD, line 143:
```suggestion
        description="Number of jobs to use in parallel during write. Negative slicing conforms with the pattern of indexing `list(range(total_number_of_cpu))[number_of_jobs]`; for example, `-1` uses all available CPU, `-2` uses all except one, etc.",
```

src/neuroconv/tools/nwb_helpers/_models/_zarr_models.py
```diff
@@ -134,3 +136,30 @@ def get_data_io_kwargs(self) -> Dict[str, Any]:
...
+    )
+    dataset_configurations: Dict[str, ZarrDatasetConfiguration] = Field(
+        description=(
+            "A mapping from object locations to their ZarrDatasetConfiguration specification that contains all "
+            "information for writing the datasets to disk using the Zarr backend."
+        )
+    )
+    number_of_jobs: int = Field(
+        description=(
+            "Number of jobs to use in parallel during write. Negative slicing conforms with the pattern of indexing "
```
h-mayorquin, line 158:
```suggestion
            "Number of jobs to use in parallel during write. Negative valures count from -1 being all the available cpus (including logical), -2 is all except one, etc. This is equivalent to the pattern of slicing ... "
```

src/neuroconv/tools/nwb_helpers/_models/_zarr_models.py
```diff
@@ -134,3 +136,30 @@ def get_data_io_kwargs(self) -> Dict[str, Any]:
...
+            " `list(range(total_number_of_cpu))[number_of_jobs]`; for example, `-1` uses all available CPU, `-2` "
+            "uses all except one, etc."
+        ),
+        ge=-psutil.cpu_count(),  # TODO: should we specify logical=False in cpu_count?
+        le=psutil.cpu_count(),
+        default=-2,  # -2 translates to 'all CPU except for one'
```
h-mayorquin, line 164:
```suggestion
        default=psutil.cpu_count() -1 
```
I would prefer this as the default.

h-mayorquin (approved):
OK, I actually had time to review it. 

This is fine for me some minor comments.

src/neuroconv/tools/nwb_helpers/_models/_zarr_models.py
```diff
@@ -134,3 +136,30 @@ def get_data_io_kwargs(self) -> Dict[str, Any]:
...
+    )
+    dataset_configurations: Dict[str, ZarrDatasetConfiguration] = Field(
+        description=(
+            "A mapping from object locations to their ZarrDatasetConfiguration specification that contains all "
+            "information for writing the datasets to disk using the Zarr backend."
+        )
+    )
+    number_of_jobs: int = Field(
+        description=(
+            "Number of jobs to use in parallel during write. Negative values, starting from -1, will use all the available CPUs (including logical), -2 is all except one, etc. This is equivalent to the pattern of indexing of "
```
CodyCBakerPhD, line 158:
```suggestion
            "Number of jobs to use in parallel during write. Negative values, starting from -1, "
            "will use all the available CPUs (including logical), -2 is all except one, etc. "
            "This is equivalent to the pattern of indexing of "
```

src/neuroconv/tools/nwb_helpers/_models/_base_models.py
```diff
@@ -180,3 +181,29 @@ def get_data_io_kwargs(self) -> Dict[str, Any]:
...
+
+class BackendConfiguration(BaseModel):
+    """A model for matching collections of DatasetConfigurations to a specific backend."""
+
+    backend: Literal["hdf5", "zarr"] = Field(description="The name of the backend used to configure the NWBFile.")
+    data_io_class: Type[DataIO] = Field(description="The DataIO class that is specific to this backend.")
+    dataset_configurations: Dict[str, DatasetConfiguration] = Field(
+        description=(
+            "A mapping from object locations (e.g. `
+acquisition/TestElectricalSeriesAP/data`)  to their DatasetConfiguration specification that contains all information "
```
CodyCBakerPhD, line 194:
```suggestion
            "A mapping from object locations (e.g. `acquisition/TestElectricalSeriesAP/data`) "
            "to their DatasetConfiguration specification that contains all information "
```

PR #570, opened 2023-09-18
[Backend Configuration IIb] Add backend collection tools

CodyCBakerPhD:
Expands on #569

This takes the datasets found from the previous dataset collection operation, which is then assembled into a single backend representation of the file contents, and will then be passed back to the user for confirmation before any final configuration of the datasets is performed


CodyCBakerPhD:
@h-mayorquin Ready for review now, should be an easy one

h-mayorquin (approved):
Sort PR, so mostly discussion. Now that we are getting to the end of this chain of PRs I have some usability questions but I think this will be better in person. I have some small requests but they are very small details and discussion points. Approving now, up to you if you want to change some of them.

Thanks for all the hard work.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_default_backend_configuration.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+processing/ecephys/NewProcessedTimeSeries/data
+----------------------------------------------
+  dtype : float64
+  full shape of source array : (4, 2)
+  full size of source array : 0.00 GB
+
+  buffer shape : (4, 2)
+  maximum RAM usage per iteration : 0.00 GB
```
h-mayorquin, line 255: One more think!
I think that maximum ram usage per iteration is a very strong promise. 
At least I know that promise IS NOT hold if you are working in spikeinterface. And I would not be suprised if that fails for other interfaces.

I would changes it to the lesser commitment of "expected RAM memory ussage" and also give the units in GiB as that is the standard for memory.

CodyCBakerPhD: 'expected' would be better, I agree

> At least I know that promise IS NOT hold if you are working in spikeinterface.

Can you give specific example here so I'm at least aware? Known cases where the `__init__` of an extractor reads entire array into memory? I was hoping every RecordingExtractor was fully lazy the whole way through

src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
```diff
@@ -206,3 +205,18 @@ def get_default_dataset_io_configurations(
                     continue  # skip
 
                 yield _get_dataset_metadata(neurodata_object=time_series, field_name=field_name, backend=backend)
+
+
+def get_default_backend_configuration(
+    nwbfile: NWBFile, backend: Literal["hdf5", "zarr"]
+) -> Union[HDF5BackendConfiguration, ZarrBackendConfiguration]:
+    """Fill a default backend configuration to serve as a starting point for further customization."""
+    BackendConfigurationClass = BACKEND_TO_CONFIGURATION[backend]
```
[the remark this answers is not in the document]
h-mayorquin: You could be right. I still feel like it is better to generalize and send it out once we need it and not generalizing this early. I am fine with having this disagreement though, I just wanted to point out a weak negative of the current configuration. Let's mark this as resolved and move on.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_default_backend_configuration.py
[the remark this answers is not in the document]
h-mayorquin: > Given that most of the purpose of this class, beyond convenience of forming the dictionary, is primarily focused on making user-friendly viewing the contents of said dictionary. Tests also showcase the clear difference (which can be educational to devs that are unaware) between the __repr__ and the print calls

All right, this explains it to me.

Completley agree with the point that there is very little cost to having such tests.

src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
```diff
@@ -206,3 +205,18 @@ def get_default_dataset_io_configurations(
                     continue  # skip
 
                 yield _get_dataset_metadata(neurodata_object=time_series, field_name=field_name, backend=backend)
+
+
+def get_default_backend_configuration(
```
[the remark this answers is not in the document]
h-mayorquin: I would like you to take this one.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_default_backend_configuration.py
h-mayorquin: It is not about loading the whole array, with that we are very careful I feel.

It is more than the wrong copy, casting or tranposition might trigger a memory allocation at some point of the extraction chain. This will generate memory allocations that within steps of the chain will produce ~2x3 times more ram consumption that they should based on chunk size alone and then there is this issue of **when** exactly that extra memory is liberated which is a topic that I don't understand fully an is OS dependent.

At some point I want to build general tests for the extractors to test that they don't have larger memory allocations but as usual, no time for that yet : (

See two relevant issues:
https://github.com/NeuralEnsemble/python-neo/pull/1265
https://github.com/SpikeInterface/spikeinterface/issues/2029

h-mayorquin:
@CodyCBakerPhD 
Thanks for engaging in discussion. Feel free to merge.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_default_backend_configuration.py
CodyCBakerPhD: Yes, OK I see...

That makes complete sense

PR #571, opened 2023-09-18
[Backend Configuration III] Configure all datasets in-memory to be written in specific way to disk

CodyCBakerPhD:
replaces #557 

This is the most useful piece of everything built up to this point; the application of the actual DataIO classes given a user-selected backend

These operations are still introduced as standalone tools for the time being; they can't be integrated into the main Interface/Converter workflow until the HDMF-zarr and HDMF package incompatibilities are resolved upstream


src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
```diff
@@ -209,4 +209,22 @@ def get_default_backend_configuration(
...
+def configure_backend(
+    nwbfile: NWBFile, backend_configuration: Union[HDF5BackendConfiguration, ZarrBackendConfiguration]
+) -> None:
+    """Configure all datasets specified in the `backend_configuration` with their appropriate DataIO and options."""
+    nwbfile_objects = nwbfile.objects.items()
+
+    data_io_class = backend_configuration.data_io_class
+    for dataset_configuration in backend_configuration.datset_configurations:
+        object_id = dataset_configuration.dataset_info.object_id
+        dataset_name = dataset_configuration.dataset_info.location.strip("/")[-1]
```
CodyCBakerPhD, line 225: TODO: in prior PR, add `dataset_name` to `DatasetInfo`

h-mayorquin:
@CodyCBakerPhD conflicts and some tests still failing.

src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py
```diff
@@ -75,6 +75,9 @@ def get_data_io_kwargs(self) -> Dict[str, Any]:
             elif self.compression_method is None:
                 compression_bundle = dict(compression=False)
         else:
-            compression_bundle = dict(compression=self.compression_method, compression_opts=self.compression_options)
+            compression_opts = None
+            if self.compression_options is not None:
```
h-mayorquin, line 79: I don't understand what are we doing here. Can you provide context?

src/neuroconv/tools/nwb_helpers/_configure_backend.py
h-mayorquin, line 1: My first instinct would be to merge `_backend_configuration.py`, `_configure_backend.py` and `dataset_configuration.py`  in the same file. Out of curiosity, why do you prefer this type of organization?

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -0,0 +1,24 @@
...
+
+    data_io_class = backend_configuration.data_io_class
+    for dataset_configuration in backend_configuration.dataset_configurations.values():
+        object_id = dataset_configuration.dataset_info.object_id
+        dataset_name = dataset_configuration.dataset_info.dataset_name
+        data_io_kwargs = dataset_configuration.get_data_io_kwargs()
+
+        # TODO: update buffer shape in iterator, if present
+
+        nwbfile_objects[object_id].set_data_io(dataset_name=dataset_name, data_io_class=data_io_class, **data_io_kwargs)
```
h-mayorquin, line 24: It would be nice if we could get the neurodata object by location.

I am also curios on why the `set_data_io` is at the neurodata object level and not at the data datset level. That's where the IO stuff ends up right?

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -0,0 +1,24 @@
+"""Collection of helper functions related to configuration of datasets dependent on backend."""
+from typing import Union
+
+from pynwb import NWBFile
+
+from ._configuration_models._hdf5_backend import HDF5BackendConfiguration
+from ._configuration_models._zarr_backend import ZarrBackendConfiguration
+
+
+def configure_backend(
```
h-mayorquin, line 10: I woud like a more specific name here. Consistency with our previos convention would get us `configure_io_backend` I think. We could also have `configure_write_backend` or `set_write_backend` as that is mainly what this is doing.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_overrides.py
```diff
@@ -0,0 +1,77 @@
...
+    backend_configuration = get_default_backend_configuration(nwbfile=nwbfile, backend=backend)
+    dataset_configuration = backend_configuration.dataset_configurations["acquisition/TestTimeSeries/data"]
+
+    smaller_chunk_shape = (30_000, 64)
+    smaller_buffer_shape = (60_000, 192)
+    higher_gzip_level = 5
+    dataset_configuration.chunk_shape = smaller_chunk_shape
+    dataset_configuration.buffer_shape = smaller_buffer_shape
+
+    higher_gzip_level = 5
```
h-mayorquin, line 52: You have repeated higher_gzip_level and I don't know if the if is necessary.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_overrides.py
```diff
@@ -0,0 +1,77 @@
...
+        dataset_configuration.compression_options = dict(level=higher_gzip_level)
+    elif backend == "zarr":
+        dataset_configuration.compression_options = dict(level=higher_gzip_level)
+
+    configure_backend(nwbfile=nwbfile, backend_configuration=backend_configuration)
+
+    if case_name != "unwrapped":  # TODO: eventually, even this case will be buffered automatically
+        assert nwbfile.acquisition["TestTimeSeries"].data
+
+    nwbfile_path = str(tmpdir / f"test_configure_{backend}_defaults_{case_name}_data.nwb.h5")
```
h-mayorquin, line 63: I don't think the string is necessary, right? Also, why the .h5 extension in the path name?

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_overrides.py
```diff
@@ -0,0 +1,77 @@
+"""Unit tests for `get_default_dataset_configurations`."""
+from pathlib import Path
+from typing import Literal
+
+import numcodecs
```
h-mayorquin, line 5: Question: should this be added to `requirements_testing`? how is it making it here?

h-mayorquin (commented):
Some questions and this should be ready to go.

src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py
```diff
@@ -75,6 +75,9 @@ def get_data_io_kwargs(self) -> Dict[str, Any]:
             elif self.compression_method is None:
                 compression_bundle = dict(compression=False)
         else:
-            compression_bundle = dict(compression=self.compression_method, compression_opts=self.compression_options)
+            compression_opts = None
```
CodyCBakerPhD, line 78:
```suggestion
            # Base filters only take particular form of a single input; single int for GZIP; 2-tuple for SZIP
            compression_opts = None
```

CodyCBakerPhD: Does this help?

src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py
CodyCBakerPhD: It's copied/pasted from https://github.com/catalystneuro/neuroconv/pull/571/files#diff-708845c5b3ce542157fc2da62133d4723159b4f250d1bc7fb83d7f2c1408feacL62-L64 (so a very small amount of duplicated code, doesn't seem worth to make a separate method though), but forgot the explaining comment above

The problem is that the `compression_opts` argument of the `H5DataIO` is very particular about how it likes things passed

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_overrides.py
CodyCBakerPhD: It is. `NWB{format}IO` require it to be a `str` or else it complains

The `.h5` is something being debated as a practice; including multiple file suffixes to allow external programs (and us...) to quickly tell what type of backend a given NWB file is without attempting to open it first

For example, now HDFView will always recognize such files without having to go into the settings and adjust it to recognize the `.nwb` suffix

The current opinion seems to be that for the next couple of years, `.nwb` will always be interpreted as being an HDF5 by default, and `.nwb.zarr` will be the identifier of files written using the new Zarr backend. DANDI in particular looks like it will use this suffix approach for deciding what required validations are appropriate (they have extra validations for Zarr assetts)

But if Zarr gains popularity, it might be nice to just get ahead of things and start doing `.nwb.h5` for symmetry along with that to add clarity at a glance

src/neuroconv/tools/nwb_helpers/_configure_backend.py
h-mayorquin: > Disagree in this case

You disagree that it would be good to have it?

It is good to have both. Most people are used to file systems so that's easy to use but unique identifiers are important for validation and computer use. What would be the downside?

 In hdf5 you can access stuff by its path within the file which I think makes sense from an usability perspective.

src/neuroconv/tools/nwb_helpers/_configure_backend.py
h-mayorquin: I am not sure about write it could be seomthing else but I guess the biggest thing is that I disagree with:

>  since Backend as a word makes it clearer it has to do with the storage structure on disk

Where are you getting this? Backend has multiple meanings and I am not sure that the one we are using here is the most common one. The most common use of Backend in SE is its use in the server-client model and most of the time is code for a database. There is also plotting backend. Point being that backend is not clear enough by itself, this is a backend for IO and at that we are only using it for writing (output) but I am fine with anything else. I am not that keen on write but I wanted to point out that Backend is the unprecise thing here.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_overrides.py
h-mayorquin: OK. This makes sense.

src/neuroconv/tools/nwb_helpers/_configure_backend.py
h-mayorquin: Makes sense. Thanks for taking the time to write.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_overrides.py
h-mayorquin: import h5py
import os
import random

def create_hdf5_file(filename):
    with h5py.File(filename, 'w') as f:
        dset = f.create_dataset("default", (100,), dtype='i')

def create_random_binary_file(filename):
    with open(filename, 'wb') as f:
        f.write(os.urandom(1024))  # Write 1024 random bytes

def create_nwb_file(filename):
    
    nwbfile = mock_NWBFile()
    # Save the file
    from pynwb import NWBHDF5IO
    with NWBHDF5IO(filename, 'w') as io:
        io.write(nwbfile)
    
    

def is_hdf5_file(filename):
    # Source for magic numbers https://www.loc.gov/preservation/digital/formats/fdd/fdd000229.shtml
    # We should find a better one though
    with open(filename, 'rb') as f:
        file_signature = f.read(8)
    return file_signature == b'\x89HDF\r\n\x1a\n'

# Create files
hdf5_filename = 'test.hdf5'
random_filename = 'random.bin'
nwb_filename = "nwbfile.nwb"

create_nwb_file(nwb_filename)
create_hdf5_file(hdf5_filename)
create_random_binary_file(random_filename)

# Check if files are HDF5
print(f"{hdf5_filename} is HDF5: {is_hdf5_file(hdf5_filename)}")
print(f"{nwb_filename} is HDF5: {is_hdf5_file(nwb_filename)}")
print(f"{random_filename} is HDF5: {is_hdf5_file(random_filename)}")

h-mayorquin: Test this. It works on my system.

src/neuroconv/tools/nwb_helpers/_configure_backend.py
CodyCBakerPhD: If you're asking/commenting in general that 'it would be nice to fetch an object in an NWB file knowing its path' then that already exists in the PyNWB API (it's that natural way of navigating group/dataset contents across the modules and submodules; but relies on closing each level of nesting)

But I think what you refer to is passing a single string that includes slashes to indicate the nested group structure (which you can also do in PyNWB; just pass it to the `nwbfile._file` attribute) but the location structure in the HDF5 file is not necessarily equivalent to the PyNWB module structure at all times. There are various aliases in the API that break that equivalency and even my 'location in file' isn't exactly either approach, nor is it intended to be

ALso this specific lookup is a piece of internal code masking what is really going on, which is reliance on object ID; what you seem to be asking for is a more generic features of PyNWB so I'd suggest just raise an issue over there to discuss with that team. I'm just saying it's not a feature I would ever use myself

Note that the 'location in file' used in the `BackendConfiguration` lookup dictionary here is only a convenient association to the object ID; is not 100% guaranteed to be perfectly identical to the PyNWB nesting structure at all times for all files, just 'generally close enough to distinguish references to contents'

That is, I've made this nice heuristic for mapping references of `location_in_file` -> `object_id` and then this line of internal operation, I go from `object_id` -> `neurodata_object`

src/neuroconv/tools/nwb_helpers/_configure_backend.py
CodyCBakerPhD: So you want

`BackendConfiguration` -> `BackendIOConfiguration`

and

`configure_backend` -> `configure_backend_io`

?

h-mayorquin: Yes, that's good.

I prefer `configure_io_backend` but `IOBackendConfiguration` maybe not so good? I think your native language ear might be good at this. Both are good to me.

src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py
CodyCBakerPhD: Well, technically the 78-81 lines are covered by the `test_minimal` run, it's just that the coverage report is only generated once by the full test run (unless you want the bot to spam twice from two uploads per push)

I guess I should I add a note and `pragma` for that... but rest assured that block is tested

As for 66-76 I think I ought to be able to add a test harnessing the [PyNWB tutorial for dynamic filters](https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/h5dataio.html#dynamically-loaded-filters), I guess I just didn't think of that. Might take a little bit

src/neuroconv/tools/nwb_helpers/_configure_backend.py
h-mayorquin: > But I think what you refer to is passing a single string that includes slashes to indicate the nested group structure 

Yes, exactly, that's what I was saying as a very later comment (not asking you) that it would be nice to have. For def I did not make it clear enough that is not related to this PR or even neuroconv. Sorry for making you write something larger than you would have had that context been clear.

> (which you can also do in PyNWB; just pass it to the nwbfile._file attribute) 
Really? I need something like this. This does not seem to work:
```
nwbfile._file["acquisition/ElectricalSeriesAp"]
AttributeError: 'NWBFile' object has no attribute '_file'
```

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_overrides.py
```diff
@@ -0,0 +1,76 @@
...
+    backend_configuration = get_default_backend_configuration(nwbfile=nwbfile, backend=backend)
+    dataset_configuration = backend_configuration.dataset_configurations["acquisition/TestTimeSeries/data"]
+
+    smaller_chunk_shape = (30_000, 64)
+    smaller_buffer_shape = (60_000, 192)
+    dataset_configuration.chunk_shape = smaller_chunk_shape
+    dataset_configuration.buffer_shape = smaller_buffer_shape
+
+    higher_gzip_level = 5
+    if backend == "hdf5":
```
h-mayorquin, line 52: What about these if? They seem to be the same?

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_defaults.py
```diff
@@ -0,0 +1,60 @@
...
+
+    nwbfile = mock_NWBFile()
+    time_series = mock_TimeSeries(name="TestTimeSeries", data=data)
+    nwbfile.add_acquisition(time_series)
+
+    backend_configuration = get_default_backend_configuration(nwbfile=nwbfile, backend=backend)
+    dataset_configuration = backend_configuration.dataset_configurations["acquisition/TestTimeSeries/data"]
+    configure_backend(nwbfile=nwbfile, backend_configuration=backend_configuration)
+
+    nwbfile_path = str(tmpdir / f"test_configure_{backend}_defaults_{case_name}_data.nwb.h5")
```
h-mayorquin, line 47: Ah, I think is the zarr one the one that complains about the string : /

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_overrides.py
CodyCBakerPhD: Sorry, can you clarify what the question is? What is the same?

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_defaults.py
CodyCBakerPhD: Oh, you mean the `NWBHDF5IO` allows a `pathlib.Path` nowadays?

Feel free to request `Path` support on HDMF-Zarr if that bugs you

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_overrides.py
CodyCBakerPhD: Ohhh lol OK I see it now

I mean, I _could_ just get rid of the `if/elif` I suppose, yes... less explicit but you're right it doesn't matter in this test case

CodyCBakerPhD:
OK then I say two follow-ups

- consider name change to make IO explicit
- increase coverage of dynamic filter cases for both HDF5 and Zarr

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_overrides.py
h-mayorquin: Yeah, no big deal, just in case they were actually meant to not be the same. Extra lines of code are fine.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_defaults.py
h-mayorquin: Yeah, I mean, it is midly more convenient. I was asking out of curiosity mostly in case you did not know that you could pass paths. But then I discovered that with the zarr one you can't not today when I was expereimenting with it.

src/neuroconv/tools/nwb_helpers/_configure_backend.py
h-mayorquin: And in the hdf5 case when streaming it i way cheaper to just know the specific location.

Issue #524, opened 2023-07-25
[Feature]: Adding video tracking data for neuralynx (.nvt file)


JohnStout:
### What would you like to see added to NeuroConv?

Current neuroconv NeuralynxRecordingInterface does not support .nvt file addition to the NWB file. 

The .nvt files represent video tracking data and contain at least X and Y position data (in pixels) as well as their associated timestamps (synchronized to the neural data)

### Do you have any interest in helping implement the feature?

Yes, but I would need guidance.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

CodyCBakerPhD:
Just a reminder if you could share some small-ish examples with us so we can add direct tests on the data

The format seems simple enough so the interface really shouldn't be too hard

CodyCBakerPhD:
@JohnStout ping

> Just a reminder if you could share some small-ish examples with us so we can add direct tests on the data
> The format seems simple enough so the interface really shouldn't be too hard

JohnStout:
Hey Cody,

Here is a link to a session recorded from a rat performing a t-maze task (
https://drive.google.com/drive/folders/1zLTovUMjh5S7RqPh1-JE8RlayMXsPYDJ?usp=share_link).
That file has the VT1.nvt and Events.nev variables. The Events variable
includes important information like timestamps based on infrared beam
breaks, while the VT1 file has X/Y/T video tracking data. It should also
have an angles variable for head direction, but I would ignore that because
it didn't work that well.

Here is another session that includes Tetrodes, CSC.ncs, Events.nev, and
VT1.nvt (
https://drive.google.com/drive/folders/1ifVId1cJF4oui2YJEGx7OT7fh2kcUsfB?usp=share_link
)

Thanks,
John

CodyCBakerPhD:
Thanks for sharing!

We'll dig into it and try to figure out a good approach for integration

bendichter:
It looks like neo covers NVT files: https://neo.readthedocs.io/en/latest/rawiolist.html#neo.rawio.NeuralynxRawIO

bendichter:
[...]

PR #580, opened 2023-09-28
Neuralynx NVT interface

bendichter:
fix https://github.com/catalystneuro/neuroconv/issues/524

- [x] draft of interface
- [x] upload test data to gin repo
- [x] write tests
- [x] write gallery example
- [x] write changelog entry


CodyCBakerPhD:
Can you share an example output NWB file?

bendichter:
[test.nwb.zip](https://github.com/catalystneuro/neuroconv/files/13168985/test.nwb.zip)

src/neuroconv/datainterfaces/behavior/neuralynx/neuralynx_nvt_interface.py
```diff
@@ -0,0 +1,111 @@
...
+            )
+
+            nwbfile.add_acquisition(Position([spatial_series], name="NvtPosition"))
+
+        if add_angle:
+            nwbfile.add_acquisition(
+                CompassDirection(
+                    SpatialSeries(
+                        name="NvtAngleSpatialSeries",
+                        data=data["Angle"],
```
CodyCBakerPhD, line 102: In the example data, this is an array of zeros

Is that a known anomaly of the testing data, or something that we think could happen in a real source file?

If the latter, we should add a pre-check to decide whether or not to automatically disable `add_angle` by default to avoid writing a dataset of effectively empty values

src/neuroconv/datainterfaces/behavior/neuralynx/neuralynx_nvt_interface.py
```diff
@@ -0,0 +1,111 @@
...
+                        reference_frame="unknown",
+                        unit="pixels",
```
CodyCBakerPhD, line 104: Also, an 'angle' series in units 'pixels' is rather confusing; shouldn't this be degrees or radians?

src/neuroconv/datainterfaces/behavior/neuralynx/neuralynx_nvt_interface.py
```diff
@@ -0,0 +1,111 @@
...
+        add_angle : bool, default=True
+        """
+
+        data = read_data(self.file_path)
+
+        if add_position:
+            # convert to float and change <= 0 (null) to NaN
+            xi = data["Xloc"]
+            x = xi.astype(float)
+            x[xi <= 0] = np.nan
```
CodyCBakerPhD, line 79: Out of curiosity, what causes the negative values in these series?

src/neuroconv/datainterfaces/behavior/neuralynx/nvt_utils.py
```diff
@@ -0,0 +1,147 @@
...
+
+    # Reading and parsing data
+    with open(filename, "rb") as file:
+        file.seek(HEADER_SIZE)
+        dtype = np.dtype(RECORD_FORMAT)
+        records = np.fromfile(file, dtype=dtype)
+        return {name: records[name].squeeze() for name, *_ in RECORD_FORMAT}
+
+
+def truncate_file(source_filename: FilePathType, dest_filename: str, n_records: int = 10) -> None:
```
CodyCBakerPhD, line 123: Technically, if this is code we want to support on this repo (as opposed to the GIN repo for example), we should add a specific test for it so our coverage doesn't drop too much

Or otherwise flag this with `# pragma: no cover`

src/neuroconv/datainterfaces/behavior/neuralynx/neuralynx_nvt_interface.py
bendichter: I don't really know but I can add a check for it

src/neuroconv/datainterfaces/behavior/neuralynx/neuralynx_nvt_interface.py
```diff
@@ -0,0 +1,112 @@
...
+
+            yi = data["Yloc"]
+            y = yi.astype(float)
+            y[yi <= 0] = np.nan
+
+            spatial_series = SpatialSeries(
+                name="NvtSpatialSeries",
+                data=np.c_[x, y],
+                reference_frame="unknown",
```
CodyCBakerPhD, line 89: I imagine reference frame is something that the person who set up the rig might know; should we expose the ability to specify this?

src/neuroconv/utils/json_schema.py
```diff
@@ -37,14 +37,21 @@ def default(self, obj):
...
+    base_schema = dict(
+        required=required or [],
+        properties=properties or {},
+        type="object",
+        additionalProperties=False,
+    )
     if tag is not None:
         base_schema.update(tag=tag)
-    if root:
+    if root is not None:
```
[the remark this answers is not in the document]
CodyCBakerPhD: And also deal with the case when `root = False`, in which case I doubt we want to update this level with the JSON schema version

src/neuroconv/datainterfaces/behavior/neuralynx/nvt_utils.py
CodyCBakerPhD, line 123:
```suggestion
def truncate_file(source_filename: FilePathType, dest_filename: str, n_records: int = 10) -> None:  # pragma: no cover
```

src/neuroconv/datainterfaces/behavior/neuralynx/neuralynx_nvt_interface.py
```diff
@@ -0,0 +1,142 @@
...
+
+            nwbfile.add_acquisition(Position([spatial_series], name="NvtPosition"))
+
+        if add_angle or (add_angle is None and not np.all(data["Angle"] == 0)):
+            nwbfile.add_acquisition(
+                CompassDirection(
+                    SpatialSeries(
+                        name="NvtAngleSpatialSeries",
+                        data=data["Angle"],
+                        reference_frame="unknown",
```
CodyCBakerPhD, line 134: Same reference frame would apply to the angular series as well

src/neuroconv/datainterfaces/behavior/neuralynx/neuralynx_nvt_interface.py
```diff
@@ -0,0 +1,142 @@
...
+            Path to the .nvt file
+        verbose : bool, default: True
+            controls verbosity.
+        """
+
+        self.file_path = file_path
+        self.verbose = verbose
+        self._timestamps = self.get_original_timestamps()
+        self.header = read_header(self.file_path)
+        self.nvt_filename = self.header["OriginalFileName"].split("\\")[-1]
```
CodyCBakerPhD, line 32: Definitely a good idea to use the filename below as the metadata reference key; one minor suggestion would be to strip the suffix and just use the stem

Also suggest using `Path` to accomplish this; the example data was clearly from a Windows system, but if it's possible for it to be from any other OS then the string splitting method might not have worked as expected

```suggestion
        self.nvt_filename = Path(self.header["OriginalFileName"]).stem
```

src/neuroconv/datainterfaces/behavior/neuralynx/neuralynx_nvt_interface.py
bendichter: missing values

bendichter: The data is originally stored as int, so NaN is not available to them

CodyCBakerPhD: Ahhh, gotcha

src/neuroconv/datainterfaces/behavior/neuralynx/nvt_utils.py
bendichter: I'd rather just not cover it. I added it since it might be nice to have in the future, but it's not something a user of NeuroConv would need for converting data, so I don't think it's necessary to write tests for it.

src/neuroconv/datainterfaces/behavior/neuralynx/neuralynx_nvt_interface.py
bendichter: It was trickier than I thought to parse filepaths from different operating systems, so I created a little utility function. I'd rather keep the suffix so it's clear to the user that this is from the filepath

src/neuroconv/datainterfaces/behavior/neuralynx/neuralynx_nvt_interface.py
CodyCBakerPhD: Or would it? For the angular, might need to additionally specify what convention the incidence is with respect to separately from the main x/y reference frame?

bendichter: It would need to be different. Reference frame here would correspond to where 0 degrees is, which I assume with either be up, going clockwise or to the right going counter clockwise.

bendichter:
@CodyCBakerPhD ok, ready for another round of review when you have time

CodyCBakerPhD (approved):

PR #588, opened 2023-10-06
[Bug] Split tables requirements across versions

CodyCBakerPhD:
Tests on `main` failed yesterday due to a new release of https://pypi.org/project/tables/#history which causes a dependency conflict with blosc2 on Python 3.8


weiglszonja (approved):
LGTM

PR #590, opened 2023-10-09
Fix tests on main 10/9/2023

CodyCBakerPhD:
The maxwell tests started failing today; seems to be due to the new release of `h5py` (maxwell requires a custom hdf5 plugin)

Pinning to previous `h5py` version for now


weiglszonja (approved):
LGTM

Issue #581, opened 2023-09-28
[Feature]: Add `OnePhotonSeries`/`TwoPhotonSeries` to processing


weiglszonja:
### What would you like to see added to NeuroConv?

Add `write_as` option to `add_photon_series` to support adding processed `OnePhotonSeries` data to processing.

For reference, the Tye lab has `OnePhotonSeries` data acquired from `Miniscope` and they want to store the processed and motion corrected data as well. Since we are soon over hours there, I only created a[ custom function](https://github.com/catalystneuro/tye-lab-to-nwb/blob/dee3239f959f1dc78281983a650f06cbce912f8d/src/tye_lab_to_nwb/ast_ophys/tools/add_photon_series_to_processing.py) at the repo for now, but eventually `add_photon_series` should have an option to write processed data.

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

CodyCBakerPhD:
Though I have had questions raised around the choice of `write_as` as the keyword name for this; @h-mayorquin @weiglszonja any ideas for a better more descriptive name of this role, now that it is needed in multiple places throughout our API?

I guess the most explicit would be `write_to_module` but even that makes it sound more like a boolean asking if we want to write or not?

CodyCBakerPhD:
Though I have had questions raised around the choice of `write_as` as the keyword name for this; @h-mayorquin @weiglszonja any ideas for a better more descriptive name of this role, now that it is needed in multiple places throughout our API?

I guess the most explicit would be `write_to_module` but even that makes it sound more like a boolean asking if we want to write or not?

weiglszonja:
@CodyCBakerPhD how about:
```
...
nwbfile_module_name: Literal["acquisition", "processing"] = "acquisition"
    The name of the module where the photon series data will be added to, default is `nwbfile.acquisition`
...
```

CodyCBakerPhD:
@weiglszonja Sure, that seems fine to me

h-mayorquin:
Using the fact that our appending functions follow this patttern:

```python
 add_something(object_to_add, nwbfile, **kwargs)`
```

[...]

Issue #269, opened 2022-12-26
[Feature]: change metadata lists to dictionaries


bendichter:
### What would you like to see added to NeuroConv?

This came from a conversation with @CodyCBakerPhD. Some of the metadata, e.g. Devices is added as a list. I'd like to change this to a dictionary. It would look something like this:

```
Devices:
  dev1:
    name: "device"
    description: "my device"
```

I would call `dev1` a "tag". It identifies the device but is not any of the device metadata. Then in `run_conversion` you should be able to specify the device tag you want to use `interface.run_conversion(..., device_tag="dev1")` This would allow multiple interfaces to easily share the same device or use different devices. If device_tag is not supplied and there is only one device, that device is used. If there is more than one device, it throws an error which informatively lists the possible values for `device_tag`

Thoughts? How hard would this be to implement? Are there down-sides other than implementation effort and backwards compatibility issues?

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

bendichter:
An alternative would be to make the names the keys of the dictionary. The advantages would be that the yaml file would be more compact and the uniqueness of the name would be enforced.

h-mayorquin:
This is great! I did not see this before.

CodyCBakerPhD:
I think this has mostly shifted over time

@h-mayorquin Maybe you want to do one last glance over the repo to see if you spot any cases and if so document them here? Or otherwise close

h-mayorquin:
Let me take a brief look, identify the ones that are missing and once we open a new issue we can close this.

h-mayorquin:
OK, so at least in ecephys. Both the devices and the electrode groups are passed as dictionaries. In Ophys, devices and imaging planes and even the series are passed as lists. 

So, I think it actually has shifted very little. Better to keep this open.

h-mayorquin:
The most glaring problem is the fact that ImageSeries are access by index instead of by key (like the ecephys do) but this is very hard to fix without breaking current working (and past) conversions. The ideal point is to change this when we integrate the microscopy extension (which hopefully should be .... sooon).

h-mayorquin:
[...]

PR #594, opened 2023-10-13
Change `add_imaging_plane` to add imaging plane by name instead of index

### LINKED ISSUE #575: [Question]: Assigning two `PhotonSeries` to the same `ImagingPlane`
The metadata for the `TwoPhotonSeries` and the `ImagingPlane` are lists (as we can have more than one).

Example from the current Clandinin conversion:
```
metadata["Ophys"]["TwoPhotonSeries"]
[{'name': 'FunctionalGreenTwoPhotonSeries', 'description': 'Imaging data acquire...ng readout', 'unit': 'px', 'imaging_plane': 'ImagingPlane', 'dimension': [...], 'format': '.nii', 'scan_line_rate': 15833.056254848874, 'field_of_view': [...]}, {'name': 'FunctionalRedTwoPhotonSeries', 'description': 'Imaging data acquire...istration.', 'unit': 'px', 'imaging_plane': 'ImagingPlane', 'dimension': [...], 'format': '.nii', 'scan_line_rate': 15833.056254848874, 'field_of_view': [...]}]
len(metadata["Ophys"]["TwoPhotonSeries"])
2
```

When writing a conversion with multiple interfaces we use the conversion options in `BaseImagingExtractorInterface` to specify which metadata belongs to a specific interface:

https://github.com/catalystneuro/neuroconv/blob/6164f98c8d271c5432cc45d4c5c745b15721060c/src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py#L101-L109

The argument is called `photon_series_index` and is used here:

https://github.com/catalystneuro/neuroconv/blob/6164f98c8d271c5432cc45d4c5c745b15721060c/src/neuroconv/tools/roiextractors/roiextractors.py#L355-L356

To my surprise, I realized that the the same index is used to determine which metadata is used to build the corresponding `image_plane_index`:

https://github.com/catalystneuro/neuroconv/blob/6164f98c8d271c5432cc45d4c5c745b15721060c/src/neuroconv/tools/roiextractors/roiextractors.py#L364-L365

Which, if I understand well, means that if I want to assign TwoPhotonSeries to the same imaging series I need to duplicated my metadata.

---

Maybe this work as intended and each `TwoPhotonSeries` should have ony `ImagingPlane` assigned? I think this is the decision that we did for Brucker.

I think this all comes from my confusion about the intent behind the `ImagingPlane`. When I first heard about volumetric data I was expecting that that would need multiple ImagingPlanes, one for every dimension of depth but this turned out to be false. On the other hand, I kind of expected that for formats with multiple channels (like Brucker) we could save one imagine plane with two `TwoPhotonSeries` referencing it and each `TwoPhotonSeries` somehow linked to their respective `OpticalChannel`. However, it seems that there is no way of doing this, as there is no way to directly link `OpticalChannel` back to the `TwoPhotonSeries`...

This is strange because `OpticalChannel` as an argument of `ImagingPlane` has a list type which I thought was meant to be used for ImagingPlanes with multiple channels but then how to disambiguate which series belongs to which channels? `TwoPhotonSeries` can only store one channel, either images or volumetric but not multiple channels.

weiglszonja:
- add `imaging_plane_name` keyword argument to `add_imaging_plane` function
- find which imaging plane to add from the metadata by name instead of index
- in `add_photon_series` and `add_plane_segmentation` the name of the imaging plane is identified from the metadata of photon series or plane segmentation
- add `imaging_plane` to default plane segmentation metadata
resolve #575


tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -246,6 +246,48 @@ def test_add_two_imaging_planes(self):
...
 
+    def test_add_imaging_plane_raises_when_name_not_found_in_metadata(self):
+        """Test adding an imaging plane raises an error when the name is not found in the metadata."""
+        imaging_plane_name = "test_imaging_plane"
+        with self.assertRaisesWith(
+            exc_type=ValueError, exc_msg=f"Imaging plane '{imaging_plane_name}' not found in metadata."
+        ):
+            add_imaging_plane(nwbfile=self.nwbfile, metadata=self.metadata, imaging_plane_name=imaging_plane_name)
+
+    def test_add_two_imaging_planes_from_metadata(self):
```
h-mayorquin, line 257: Nice.

We have:
* Test for adding a name when the name is not specified in the metadata
* Test for adding more than one ImagingPlane and distinguish them by name.

Maybe we can add one where the imaging plane is added by name but the name already exists on the nwbfile? Would that make sense?

h-mayorquin (commented):
This looks very good to me and makes sense. 

No major comments other than minor suggestions.

tests/test_ophys/test_tools_roiextractors.py
weiglszonja: Yes that could make sense, I'll add it and thanks for the comments!

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -1427,6 +1468,24 @@ def test_add_one_photon_series_roundtrip(self):
             expected_one_photon_series_shape = (self.num_frames, self.num_columns, self.num_rows)
             assert one_photon_series.shape == expected_one_photon_series_shape
 
+    def test_add_multiple_one_photon_series(self):
```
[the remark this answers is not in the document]
weiglszonja: I think I wanted to check here that multiple photon series can be added with the same imaging plane, but maybe this PR is not the right place to have this test. How about including this in #581 ? Where I could also check you could add photon series to processing and have your raw photon series in acquisition referencing the same imaging plane.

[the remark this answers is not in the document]
h-mayorquin: As you describe it, it does make sense here. I could just not infer intent from reading the code. Maybe changing the name of the test to:
`test_add_multiple_one_photon_series_with_same_imaging_plane` 
Or a comment under the current name could help.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -227,27 +233,43 @@ def add_imaging_plane(nwbfile: NWBFile, metadata: dict, imaging_plane_index: int
...
 
     # Set the defaults and required infrastructure
     metadata_copy = deepcopy(metadata)
     default_metadata = get_default_ophys_metadata()
     metadata_copy = dict_deep_update(default_metadata, metadata_copy, append_list=False)
     add_devices(nwbfile=nwbfile, metadata=metadata_copy)
 
-    imaging_plane_metadata = metadata_copy["Ophys"]["ImagingPlane"][imaging_plane_index]
-    imaging_plane_name = imaging_plane_metadata["name"]
+    default_imaging_plane_name = metadata_copy["Ophys"]["ImagingPlane"][0]["name"]
```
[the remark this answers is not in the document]
weiglszonja: @h-mayorquin Now that I'm looking at it, the logic that I wanted to use here is when you don't specify the name of the imaging plane, it will look up the "default" from the metadata. But it's not actually the default name "ImagingPlane" but after it has been updated in `metadata_copy`.
Does that make sense?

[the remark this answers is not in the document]
h-mayorquin: Let me think.
If we leave `metadata_copy` and the user passess some imaging plane metadata (but they don't specify`imaging_plane_name`) then this will choose the first name of the first element in their `metadata["Ophys"]["ImagingPlane"]` list. Correct?

Why would this be a good default and better than the value from `default_metadata`? How are you thinking about it?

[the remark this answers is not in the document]
weiglszonja: @h-mayorquin I'm not sure my reasoning was correct; but if we are comparing to the default_metadata then we would expose `imaging_plane_name` to `add_photon_series` and `add_segmentation`, because the user would have to define the name of the imaging plane they want to use in 1) get_metadata() and in 2) add_to_nwbfile(imaging_plane_name=""). Comparing to metadata_copy would silently choose it for them, but maybe that is not the good default.

tests/test_ophys/test_tools_roiextractors.py
weiglszonja: it's always good to check, and thanks! :)

src/neuroconv/tools/roiextractors/roiextractors.py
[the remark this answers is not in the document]
h-mayorquin: I am not sure I understand case 1.

Let me say something though, I think we don't need to expose the `imaging_plane_name` in `add_photon_series`. The metadata for the photon series already has the name of the plane on it. We should just swap these two lines:

https://github.com/catalystneuro/neuroconv/blob/0015f1a3b668a7dedef7ff391e4486bfe19e8bd1/src/neuroconv/tools/roiextractors/roiextractors.py#L364-L365

As for `add_segmentation` imaging_plane is an also a keyword argument of Segmentation:
https://pynwb.readthedocs.io/en/stable/pynwb.ophys.html#pynwb.ophys.PlaneSegmentation

So probably just having a default name in the metadaata for `get_default_segmentation_metadata` should work?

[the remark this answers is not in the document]
weiglszonja: Right, I think that could work! Thanks!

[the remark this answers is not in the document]
weiglszonja: Fixed the tests, ready for the final review @h-mayorquin

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -700,7 +725,8 @@ def add_plane_segmentation(
         imaging_plane_name = imaging_plane_metadata["name"]
         imaging_plane = nwbfile.imaging_planes[imaging_plane_name]
 
-        plane_segmentation_kwargs = dict(**plane_segmentation_metadata, imaging_plane=imaging_plane)
+        plane_segmentation_kwargs = deepcopy(plane_segmentation_metadata)
```
h-mayorquin, line 728: Same question here

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -200,7 +200,7 @@ def setUp(self):
         self.metadata["Ophys"].update(ImagingPlane=[self.imaging_plane_metadata])
 
     def test_add_imaging_plane(self):
-        add_imaging_plane(nwbfile=self.nwbfile, metadata=self.metadata)
+        add_imaging_plane(nwbfile=self.nwbfile, metadata=self.metadata, imaging_plane_name=self.imaging_plane_name)
```
h-mayorquin, line 203: Thanks form aking this explicit here and in the other ones.

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -1427,6 +1469,25 @@ def test_add_one_photon_series_roundtrip(self):
             expected_one_photon_series_shape = (self.num_frames, self.num_columns, self.num_rows)
             assert one_photon_series.shape == expected_one_photon_series_shape
 
+    def test_add_multiple_one_photon_series_with_same_imaging_plane(self):
+        """Test adding two OnePhotonSeries that use the same ImagingPlane."""
+        add_photon_series(
```
h-mayorquin, line 1474: What do you think of making this explicit by explictly setting the:

```python
first_photon_series_metadata = deepcopy(self.one_photon_series_metadata)
second_photon_series_metadata = deepcopy(self.one_photon_series_metadata)
first_photon_series_metadata["Ophys"]["OnePhotonSeries"][0]["imaging_plane"] = "same_imaging_plane_for_two_series"
second_photon_series_metadata["Ophys"]["OnePhotonSeries"][0]["imaging_plane"] = "same_imaging_plane_for_two_series"
```
Or we could use a single metadata dictionary for the two of them as well...

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -1427,6 +1469,25 @@ def test_add_one_photon_series_roundtrip(self):
...
+        second_photon_series_metadata["Ophys"]["OnePhotonSeries"][0]["name"] = "second_photon_series"
+        add_photon_series(
+            imaging=self.imaging_extractor,
+            nwbfile=self.nwbfile,
+            metadata=second_photon_series_metadata,
+            photon_series_type="OnePhotonSeries",
+        )
+
+        self.assertIn("second_photon_series", self.nwbfile.acquisition)
+
```
h-mayorquin, line 1490: Do you think it would be a good idea to check that we only have one imaging plane here written?

Something like 
```
assert len(self.nwbfile.imaging_planes) == 0
assert self.nwbfile.imaging_planes[0]["name"] == "the_default_name_or_the_name_we_chosen_in_this_test"
```

The second line is probably wrong but you get the idea.

h-mayorquin (commented):
This looks great. I am wondering if it would make sense to also get the photon_series properties by name instead of by index as well.

Just some final suggestions about the tests and a small question.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -346,18 +370,18 @@ def add_photon_series(
...
 
     # Add the image plane to nwb
-    # TODO: change imaging_plane_index to photon_series_key
-    nwbfile = add_imaging_plane(nwbfile=nwbfile, metadata=metadata_copy, imaging_plane_index=photon_series_index)
-    imaging_plane_name = photon_series_kwargs["imaging_plane"]
+    imaging_plane_name = photon_series_metadata["imaging_plane"]
+    add_imaging_plane(nwbfile=nwbfile, metadata=metadata_copy, imaging_plane_name=imaging_plane_name)
     imaging_plane = nwbfile.get_imaging_plane(name=imaging_plane_name)
+    photon_series_kwargs = deepcopy(photon_series_metadata)
     photon_series_kwargs.update(imaging_plane=imaging_plane)
```
weiglszonja, line 385: > My memory is that we needed to deepcopy the metadata so we don't modify the data that the user passes but the hpoton_series_metadata is already deepcopied so why copy it again?

You're right, the reason why I chose this is to avoid overwriting the "imaging_plane" in the metadata with the actual ImagingPlane object at this line.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -346,18 +370,18 @@ def add_photon_series(
...
         return nwbfile
 
     # Add the image plane to nwb
-    # TODO: change imaging_plane_index to photon_series_key
-    nwbfile = add_imaging_plane(nwbfile=nwbfile, metadata=metadata_copy, imaging_plane_index=photon_series_index)
-    imaging_plane_name = photon_series_kwargs["imaging_plane"]
+    imaging_plane_name = photon_series_metadata["imaging_plane"]
+    add_imaging_plane(nwbfile=nwbfile, metadata=metadata_copy, imaging_plane_name=imaging_plane_name)
     imaging_plane = nwbfile.get_imaging_plane(name=imaging_plane_name)
+    photon_series_kwargs = deepcopy(photon_series_metadata)
```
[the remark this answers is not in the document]
CodyCBakerPhD: To avoid mutating the other copy I assume 😅

tests/test_ophys/test_tools_roiextractors.py
weiglszonja: Yes, we should add a check for that! Thanks

tests/test_ophys/test_tools_roiextractors.py
weiglszonja: explicit is always better, but I'm not sure how you imagine the single metadata dictionary, something like this:
https://github.com/catalystneuro/neuroconv/blob/3c17f248da202be4320473d8ef71b4b4d021385c/tests/test_ophys/test_tools_roiextractors.py#L1474-L1480
https://github.com/catalystneuro/neuroconv/pull/594/commits/3c17f248da202be4320473d8ef71b4b4d021385c

CodyCBakerPhD:
First steps towards https://github.com/catalystneuro/neuroconv/issues/269

tests/test_ophys/test_tools_roiextractors.py
h-mayorquin: So,
`metadata["Ophys"]["OnePhotonSeries"]` will be a list with two elements (one per plane).

But if we are moving towards https://github.com/catalystneuro/neuroconv/issues/269 then maybe is not worth to do it like this.

src/neuroconv/tools/roiextractors/roiextractors.py
h-mayorquin: Ah, OK, probably not necessary anymore then.

tests/test_ophys/test_tools_roiextractors.py
h-mayorquin: No, not all, I am saying that if you don't want to have a single metadata dictionary that is totally fine. Because, when we eventually do #269 then we will need to change the list structure anyway.

So up to you.

src/neuroconv/tools/roiextractors/roiextractors.py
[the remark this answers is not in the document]
weiglszonja: Yes and it would definitely make sense to eventually also use photon_series_name as identifier instead of index!

h-mayorquin (approved):
I think this is done right?

weiglszonja:
Yes, I think it's finished.

h-mayorquin:
Yes, that's fine, I think the current one that is failing I fixed but another one seems to have appeared.

PR #595, opened 2023-10-14
[Fix] remove starting_time overwrite to 0

alessandratrapani:
starting_time value is overwritten to zero when adding rate in add_photon_series or add_fluorescence_traces when updating photon_series_kwargs or roi_response_series_kwargs.


h-mayorquin (approved):
This was updating the `starting_time`  in the metadata dict. The metadata dict should have the last word.

h-mayorquin:
The tests are failing because of a spikeinterface problem. Most likely we have come to the dreaded clash between annotations at runtime vs annotations at import that is implied by from __future__ import annotations.

h-mayorquin:
Yei for @alessandratrapani first bug fix in neuroconv : D

Issue #597, opened 2023-10-16
[Proposal] Automatically align `starting_time` or `timestamps` to `session_start_time` for interfaces that automatically extract the latter


h-mayorquin:
For some interfaces like SpikeGLX we automatically extract the `session_start_time`. If the latter is a dateime object we can automatically shift/align them if the user specifes the `session_start_time` in the metadata. 

The code would be something like this in their `add_object` method.

```python
    # Shift to the session start time
    metadata = metadata or self.get_metadata()
    interface_session_start_time = extract_session_start_time(self.file_path)   # This is the method of the interface
    conversion_session_start_time = metadata["NWBFile"]["session_start_time"]  # This is the converter metadata which might differ from the interface
    starting_time = interface_session_start_time.timestamp() - conversion_session_start_time.timestamp()
    # Shift timestamps if they are available
    timestamps_shifted = timestamps + starting_time
```

That way, if some interface / experiment is shifted (i.e. was started let's say 10 minutes after the beginning of the experiment) neuroconv will automatically align them.

What are some downsides on this?

CodyCBakerPhD:
There are some concerns I'd potentially have about feasibility but they all require speculation...

Can you provide a full real example of what this would look like now, and what an 'automated' sync method would look like as an alternative?

h-mayorquin:
The current state of affairs: 
Let's say a user as an interface with SpikeGLX and some ad-hoc behavior where the "correct" `session_start_time` is the one from behavior. For example, we started recording a video before we started the electrophysiology system. Let's say 10 minutes before. The user uses something like the code example from the conversion gallery:

https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/combinations/ecephys_pose_estimation.html

And correctly sets the `session_start_time` as indicated:

```
metadata["NWBFile"].update(session_start_time=behavior_start_time)
```

Then, if they run the code as in the gallery without doing anything else, then timestamps (or the starting_time) of the `ElectricalSeries` will be assumed to start from 0 and the resulting nwbfile wrong.

If they wanted to correct the time alignment, they would have to get discover that we have the time setter methods and then modify the timestamps of the SpikeGLX interface by shifting it ten minutes with the method `set_aligned_starting_time`.

Now, with the method that I am proposing this will be fixed automatically in some cases by calling this code before adding the object to the file.

```python
    # Shift to the session start time
    metadata = metadata or self.get_metadata()
[...]

PR #598, opened 2023-10-16
Add complete session start time to FicTrac

h-mayorquin:
When working on the Clandining conversion I realized that the first line of the data header has non-valid values save the timestamp. The timestamp is the unix time epoch in milliseconds which can be used to extract a more complete `session_start_time` as previously we were extracting only the date.

I did some other improvements to the code like adding `starting_time` when they are not zero which were not done before.

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
+++ b/src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
@@ -1,5 +1,5 @@
 import importlib.util
-from datetime import datetime
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import Optional
 
@@ -8,7 +8,7 @@ from pynwb.file import NWBFile
 
 # from ....basetemporalalignmentinterface import BaseTemporalAlignmentInterface TODO: Add timing methods
 from ....basedatainterface import BaseDataInterface
-from ....tools import get_module, get_package
+from ....tools import get_module
 from ....utils import FilePathType, calculate_regular_series_rate
 
 
@@ -105,18 +105,15 @@ class FicTracDataInterface(BaseDataInterface):
         timestamps_milliseconds = fictrac_data_df["timestamp"].values
         timestamps = timestamps_milliseconds / 1000.0
 
-        # Shift to the session start time
-        metadata = metadata or self.get_metadata()
-        interface_session_start_time = extract_session_start_time(self.file_path)
-        conversion_session_start_time = metadata["NWBFile"]["session_start_time"]
-        starting_time = interface_session_start_time.timestamp() - conversion_session_start_time.timestamp()
-        timestamps_shifted = timestamps + starting_time
-
         # Note: The last values of the timestamps look very irregular for the sample file in catalyst neuro gin repo
         # The reason, most likely, is that FicTrac is relying on OpenCV to get the timestamps from the video
         # In my experience, OpenCV is not very accurate with the timestamps at the end of the video.
-        rate = calculate_regular_series_rate(series=timestamps_shifted)
-        write_timestamps = False if rate else True  # calculate_regular_series_rate returns None if it is not regular
+        rate = calculate_regular_series_rate(series=timestamps)  # Returns None if the series is not regular
+        if rate:
+            write_timestamps = False
+            starting_time = timestamps[0]
+        else:
+            write_timestamps = True
 
         # All the units in FicTrac are in radians, the radius of the ball required to transform to
         # Distances is not specified in the format
@@ -427,15 +424,6 @@ class FicTracDataInterface(BaseDataInterface):
         processing_module.add_data_interface(compass_direction_container)
 
 
-def import_zone_info():
-    if importlib.util.find_spec("zoneinfo"):
-        return importlib.import_module("zoneinfo").ZoneInfo
-    elif importlib.util.find_spec("backports.zoneinfo"):
-        return importlib.import_module("backports.zoneinfo").ZoneInfo
-    else:
-        raise ImportError("Neither 'zoneinfo' nor 'backports.zoneinfo' is available.")
-
-
 def extract_session_start_time(file_path: FilePathType) -> datetime:
     """
     Lazily extract the session start datetime from a FicTrac data file.
@@ -444,7 +432,6 @@ def extract_session_start_time(file_path: FilePathType) -> datetime:
 
     The epoch in Linux is 1970-01-01 00:00:00 UTC.
     """
-    ZoneInfo = import_zone_info()
     with open(file_path, "r") as file:
         # Read the first data line
         first_line = file.readline()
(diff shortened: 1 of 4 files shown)


h-mayorquin:
Oh, I thought that `ZoneInfo` was introduced in python 3.8 but turns out is not available until python 3.9 : (

Related:
https://github.com/catalystneuro/neuroconv/issues/599

src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
```diff
@@ -72,17 +81,12 @@ def __init__(
...
-            date_string = self._config_file["build_date"]
-            date_object = datetime.strptime(date_string, "%b %d %Y")
-
-            metadata["NWBFile"].update(
-                session_start_time=date_object,
-            )
+        try:
+            session_start_time = extract_session_start_time(self.file_path)
+            metadata["NWBFile"].update(session_start_time=session_start_time)
+        except:
```
CodyCBakerPhD, line 88: I always try to avoid bare excepts; is there a specific error that is known to occur here? Is there no other way to determine the condition using `if/else`?

src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
```diff
@@ -72,17 +81,12 @@ def __init__(
...
+            warnings.warn("Unable to extract the session start time from the FicTrac data file. ")
```
CodyCBakerPhD, line 89: Do we warn about not finding session start time in any other interfaces? I had thought it was usually silent inclusion to the metadata if found, and nothing happens if not

src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
```diff
@@ -102,17 +106,25 @@ def add_to_nwbfile(
...
+        # Shift to the session start time
+        metadata = metadata or self.get_metadata()
+        interface_session_start_time = extract_session_start_time(self.file_path)
+        conversion_session_start_time = metadata["NWBFile"]["session_start_time"]
+        starting_time = interface_session_start_time.timestamp() - conversion_session_start_time.timestamp()
+        timestamps_shifted = timestamps + starting_time
+
+        # Note: The last values of the timestamps look very irregular for the sample file in catalyst neuro gin repo
+        # The reason, most likely, is that FicTrac is relying on OpenCV to get the timestamps from the video
+        # And we know that OpenCV is not very accurate with the timestamps, specially the last ones.
```
CodyCBakerPhD, line 125: Remind me why we know that? Past investigations to OpenCV, depending on codec, largely relate to the timestamps being calculated deterministically by `index / fps` equation

src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
h-mayorquin: Yes, this makes sense.

src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
h-mayorquin: You mentioned that before, where are you getting that?

It uses the `picture_pst` variable:
https://github.com/opencv/opencv/blob/2f63c583801cb9081cc696e9842fc9a43516904c/modules/videoio/src/cap_ffmpeg_impl.hpp#L1771-L1776

How is this obtained? Well, it tries to get the real presentation timestamp (or the package presentation time) depending on the version:

https://github.com/opencv/opencv/blob/2f63c583801cb9081cc696e9842fc9a43516904c/modules/videoio/src/cap_ffmpeg_impl.hpp#L1575-L1583

The problem is that the fallback when it does not find that is the decoded timestamps which for some reason fails at the end in my experience. This is what was happening on that old issue with Deep Lab Cut where the guy from datalad decided to interpolate the times.

h-mayorquin: Yeah, it is very annoying that this is soo unclear. I guess video people don't really care about the time to the degree that we do and then that explains it.

src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
```diff
@@ -102,21 +98,32 @@ def add_to_nwbfile(
...
-        if rate:
-            write_timestamps = False
 
-        processing_module = get_module(nwbfile=nwbfile, name="Behavior")
+        # Shift to the session start time
+        metadata = metadata or self.get_metadata()
+        interface_session_start_time = extract_session_start_time(self.file_path)
+        conversion_session_start_time = metadata["NWBFile"]["session_start_time"]
+        starting_time = interface_session_start_time.timestamp() - conversion_session_start_time.timestamp()
+        timestamps_shifted = timestamps + starting_time
```
CodyCBakerPhD, line 113: How is this supposed to interact with the alignment methods? I'd prefer to use those pathways, and for the `add_to_nwbfile` section to rely on retrieving whatever the user has modified with respect to that

(_e.g._, someone preparsed timestamps separate from FicTrac format and wants to set them explicitly; ditto for manually calculated start time distinct from w/e the header says)

In general we've also never encouraged relying on system times as cached by headers for calculating exact start times

Problem: what if SpikeGLX header (from a separate device) has a clock that is different from that of the FicTrac system (on yet another difference device); comparing the system timestamps is not trustworthy at all in this case, and even if the user wanted to do it they would be encouraged to do so via custom code outside of the interfaces (at the custom converter level or conversion script) to verify that they are OK with that approach - I don't think we should do it automatically in this case

h-mayorquin: This interface is not a `TemporalInterface` so those methods do not exists yet. That said, I think as it is, this implementation will clash with those methods when we upgrade this interface to be temporal. We need to think on this interaction better. I am removing this from this PR and we can keep this discussion here:

https://github.com/catalystneuro/neuroconv/issues/597

h-mayorquin: Removed. I also removed the whole zoneinfo thing. Turns out the utc can be imported from the standard library already so at least in this case no need to do extra fancy thing.

CodyCBakerPhD (approved):

tests/test_on_data/test_behavior_interfaces.py
```diff
@@ -37,6 +37,10 @@ class TestFicTracDataInterface(DataInterfaceTestMixin, unittest.TestCase):
 
     save_directory = OUTPUT_PATH
 
+    def check_extracted_metadata(self, metadata: dict):
```
CodyCBakerPhD, line 40: Actually, before merge, can we get a test for the new starting time functionality? Or is the example data irregular?

CodyCBakerPhD, line 40: Actually, before merge, can we get a test for the new starting time functionality? Or is the example data irregular?

CodyCBakerPhD (changes requested):

tests/test_on_data/test_behavior_interfaces.py
h-mayorquin: It is irregular in the example : (

CodyCBakerPhD (approved):

PR #612, opened 2023-10-24
Fix `MicroManagerTiffImagingInterface` to have `OnePhotonSeries` default metadata

weiglszonja:
This interface is used at https://github.com/catalystneuro/pinto-lab-to-nwb/ where they performed widefield imaging which should be of `OnePhotonSeries` instead of `TwoPhotonSeries`. 
This should fix the metadata to contain the default metadata for OnePhotonSeries and modifies the tests.

todo:
- [ ] Try this modification at https://github.com/catalystneuro/pinto-lab-to-nwb/

### DIFF
--- a/src/neuroconv/datainterfaces/ophys/micromanagertiff/micromanagertiffdatainterface.py
+++ b/src/neuroconv/datainterfaces/ophys/micromanagertiff/micromanagertiffdatainterface.py
@@ -1,7 +1,11 @@
+from typing import Optional, Literal
+
 from dateutil.parser import parse
+from pynwb import NWBFile
 
 from ..baseimagingextractorinterface import BaseImagingExtractorInterface
-from ....utils import FolderPathType
+from ....tools.roiextractors import get_nwb_imaging_metadata
+from ....utils import FolderPathType, dict_deep_update
 
 
 class MicroManagerTiffImagingInterface(BaseImagingExtractorInterface):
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
Well, the real question is if MicroManager is strictly widefield acquisition or if, as a device, it be used for any type of ophys acquisition and doesn't really have a default (in which case we fallback to the classic PyNWB default of 2P)

In the other previous case of Miniscope it was for sure 1P

Perhaps @alessandratrapani can shed light on the format/device and how it is most often used?

weiglszonja:
> Well, the real question is if MicroManager is strictly widefield acquisition or if, as a device, it be used for any type of ophys acquisition and doesn't really have a default (in which case we fallback to the classic PyNWB default of 2P)

Yeah this is a good point, also once `ndx-microscopy` takes over it will be much easier to define the metadata.

 If using .ome.tif in general can be used for both 2P (we've seen Bruker to use `.ome.tif`) and 1P, then maybe the custom conversion should override this metadata instead.

alessandratrapani:
So my first guess was that MicroManager would not be specific for any type of ophys acquisition since the differences between one or the others are more on the level of the laser source, the detectors and the techniques to capture the photons, but I checked the documentation, and they said:

> [µManager](https://micro-manager.org/) is an open-source microscopy software based on java and integrated with ImageJ to provide a suite of microscope control. This includes extensive support of various microscopy hardware such as stages and cameras. For a complete list of cameras supported by µManager, visit the [µManager website](https://micro-manager.org/wiki/Device_Support).

I also checked the list of hardware supported, and they can be used for different types of ophys acquisitions, depending on the configuration of the system.

PR #613, opened 2023-10-25
[Test] Error with future imports

h-mayorquin:
Testing.


src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
```diff
@@ -28,7 +28,7 @@ def get_streams(cls, folder_path: FolderPathType) -> List[str]:
     def __init__(
         self,
         folder_path: FolderPathType,
-        streams: Optional[List[str]] = None,
+        streams: Optional[list[str]] = None,
```
CodyCBakerPhD, line 31: I assume there's probably more unrelated issues that might pop up w.r.t. our schema generation tools for making the hard switch to this for the init signatures

the import was only a needed thing for latest SpikeInterface compatability right?

h-mayorquin: I was putting this in draft to discuss it with you later as I mentioned to you that using the future annotations sometimes breaks the schema. I wanted to show you this.


This is related to [PEP 563](https://peps.python.org/pep-0563/) and [PEP 649](https://peps.python.org/pep-0649/). As this stuff will not become part of the language until python 3.13 we have time to adapt. That said, I still wanted to make you guys aware.

Here is a good summary:

https://dev.to/derlin/python-type-hints-and-future-annotations-4kc5

CodyCBakerPhD:
@h-mayorquin Is this still needed, or replaced by #653?

h-mayorquin:
No, it should still produce errors with the schema generation tools.

PR #617, opened 2023-10-27
rmv redundant lines and add test for encoding numpy array

bendichter:


src/neuroconv/utils/json_schema.py
```diff
@@ -21,15 +21,10 @@ def default(self, obj):
...
-        # This should transforms numpy generic integers and floats to python floats
+        # Transform numpy generic integers and floats to python ints floats
         if isinstance(obj, np.generic):
             return obj.item()
 
-        # Numpy-versions of various numeric types
-        if isinstance(obj, np.integer):
-            return int(obj)
-        if isinstance(obj, np.floating):
-            return float(obj)
```
bendichter, line 32: The above line that handles generic types handles all float and integer types, so these lines are redundant

h-mayorquin (approved):
I did the same for spikeinterface at some point. LGTM.

PR #619, opened 2023-10-27
Add radius as conversion (scaling) instead of adding the scaled data in `FicTrac`

h-mayorquin:
This was a suggestion from @CodyCBakerPhD in our  last conversion which I think makes sense. We were adding the data scaled which increases the error if the measurement of the radius is not very precise. 

This method keeps both the radius, the right units and the original data without imprecisions. 

I modified the test to account for this.


CodyCBakerPhD:
Elegant separation of source data and utilization of existing mechanisms/attributes on the neurodata classes!

I will mention one more thing for a future improvement, dependent on resolution of https://github.com/NeurodataWithoutBorders/pynwb/pull/1766

For this kind of thing (any time there are radial + linearized interpretations of data) it is common to include two separate time series for each interpretation (each named clearly for easy identification), since even now the detail of the original radial data could be obscured by the conversion factor - the most common case of this is eye tracking, where we might include both the angle of the pupil relative to the target on the screen, and then the processed estimate of that x/y position given the angles - but this seems similar enough

This might seem like duplication, and it is, which is why I don't advocate for it right at the moment - but once the issue is resolved it should be possible to internally inter-link the `data` field of the two series so that the data is only written once

CodyCBakerPhD (approved):

Issue #608, opened 2023-10-23
[Discussion] semantics of `set_aligned_starting_time`


h-mayorquin:
This probably should have come up before but it escaped me back then. The current function for `set_aligned_starting_time` shifts the timestamps:
https://github.com/catalystneuro/neuroconv/blob/a239c27a07e49544acf0acea63906bd8abb0669c/src/neuroconv/basetemporalalignmentinterface.py#L58-L69

That is, if your timestamps are currently `[1, 2, 3, 4]` and the `starting_time` is `3` then the resulting timestamps of the interface will be `[4, 5, 6, 7]`.  We have shifted everything, in fact, that's how we call the test:

https://github.com/catalystneuro/neuroconv/blob/a239c27a07e49544acf0acea63906bd8abb0669c/src/neuroconv/tools/testing/data_interface_mixins.py#L175-L185

But the way that I am feeling now when working with this as a user is that after using `set_aligned_starting_time` the resulting timestamps should start from the set `starting_time`. That is, I would have expected the result above to be `[3, 4, 5, 6]`. That way, when the interface write the timestamps, the starting time will be in fact `3` and not `4` as it is currently.

It appears to me that the current method is shifting and not setting. What do you guys think?

h-mayorquin:
> It's setting a shift factor 
Yeah, that's the thing, I don't think on the `starting_time` as a shift factor. I want the semantics of the `starting_time` in neuroconv to be the be the `starting_time` that the corresponding nwb object will have. `set_aligned_starting_time` should fulfill that promise instead of shifting.

> if timestamps are unaligned or else why would you be using this method
Gross alignment vs fine alignment. The timestamps are fine I just want to set the correct start time because a video recording the behavior, for example, did not start until 20 seconds before a trial period.

Right now, I will need to extract the timestamps, modify them and use the `set_aligned_timestamps`  which adds complications. What's the gain of this?

CodyCBakerPhD:
> Gross alignment vs fine alignment.

There is no such separation in NWB

There is only the vector of timestamps


> Right now, I will need to extract the timestamps, modify them and use the set_aligned_timestamps which adds complications. What's the gain of this?

if you need to extract it anyway, just perform whatever operations you need to when you set the full timestamps - no need to interact with any other method.

The automated shifting is for interfaces that automatically fetch their vector, which may or may not be regular a priori, and determination of which convention (start time + rate or timestamps) occurs later on

h-mayorquin:
[...]

PR #621, opened 2023-10-28
Fix FicTrac timestamps. Coerce them to start at 0.

h-mayorquin:
Note that this has #619 as a base.
In the context of #608. 

Coercing the timestamps to start from 0 for FicTrac. In this case, the first row was actually a timestamp 0 that was modified by FicTrac logic see here:

https://github.com/rjdmoore/fictrac/issues/29

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
+++ b/src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
@@ -246,20 +246,34 @@ class FicTracDataInterface(BaseTemporalAlignmentInterface):
 
     def get_original_timestamps(self):
         """
-        In FicTrac version 2.1.1  the timestampts that are 0 are changed for system time in milliseconds. The
-        system time is set at the point in which the frame is processed. This means that the timestamps are not
-        monotonically increasing. This function checks if the timestamps are inconsistent.
+        Retrieve and correct the original timestamps from a FicTrac data file.
 
-        This happens for video sources whe OpenCV returns 0 at the beginning:
-        [0, t1, t2, t3, ...]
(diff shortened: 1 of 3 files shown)


src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
```diff
@@ -243,19 +247,67 @@ def add_to_nwbfile(
...
-        fictrac_data_df = pd.read_csv(self.file_path, sep=",", skiprows=1, header=None, usecols=[timestamp_index])
+        # Correct for the case when only the first timestamp was replaced by system time
+        first_difference = timestamps[1] - timestamps[0]
+        if first_difference < 0:
+            timestamps[0] = 0.0
 
-        return fictrac_data_df[timestamp_index].values / 1000.0
+        # Heuristic to test if timestamps are in Unix epoch
+        length_in_seconds_of_a_10_year_experiment = 10 * 365 * 24 * 60 * 60
+        all_timestamps_are_in_unix_epoch = np.all(timestamps > length_in_seconds_of_a_10_year_experiment)
```
CodyCBakerPhD, line 300: Aren't the `timestamps` variable already coerced to units of seconds above?

h-mayorquin: Which lines above you mean?

The timetamps are all in seconds from line 291 onwards.

CodyCBakerPhD: From discussion - underlying data is in unix time but relative to session start - makes sense now

CodyCBakerPhD (approved):

Issue #603, opened 2023-10-19
[Discussion]: Adding traces and segmentation images from a second channel (and/or plane)


weiglszonja:
### What would you like to see added to NeuroConv?

I am updating `Suite2pSegmentationInterface` (https://github.com/catalystneuro/neuroconv/pull/601) to support loading the fluorescence traces from a second channel. 

I need to have better control over the name of the traces, for instance:
https://github.com/catalystneuro/neuroconv/blob/ec748f124beefd56df3e66ebe408146e04702018/src/neuroconv/tools/roiextractors/roiextractors.py#L896

The same goes for the default naming of SegmentationImages:
https://github.com/catalystneuro/neuroconv/blob/ec748f124beefd56df3e66ebe408146e04702018/src/neuroconv/tools/roiextractors/roiextractors.py#L1051

The name of the traces and images should have a 1:1 link to the plane, for that using the name would be more informative.

Instead of `plane_index` and `plane_num` and `plane_segmentation_index` I would switch to `plane_segmentation_name` which would be propagated all the way to the interface `add_to_nwbfile()` level,  and that would control which plane segmentation is added. (starting to get in line with #269)

For adding traces from a second plane segmentation, I would create an additional `Fluorescence` or `DfOverF` container, where the name of the container would reflect the name of of the plane, e.g. "FluorescencePlaneSegmentationChan1Plane0"
In this way, each plane (and channel, if there are more than 1 channels) would have its own `Fluorescence` or `DfOverF` container where the name of the traces would be "RoiResponseSeries" etc. 
See #623 

As for adding summary images, I would add `SegmentationImages` to the metadata schema where name the of the Images container would reflect which plane the images belong to. 
See #622 

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

PR #601, opened 2023-10-17
Update `Suite2pSegmentationInterface` to support multiple channels and planes

weiglszonja:
- [x] Update Suite2pSegmentationExtractor to support multiple channels and planes https://github.com/catalystneuro/roiextractors/pull/242
- [x] Update all necessary metadata (Created #603 to discuss these questions)
- [x] Add test for stubbing, as mentioned in https://github.com/NeurodataWithoutBorders/nwb-guide/pull/460#issuecomment-1775175196
- [x] Update docs for usage example

- [x] Remove pin for roiextractors

PR #623, opened 2023-10-29
Change identifier of plane to `plane_segmentation_name`

weiglszonja:
Related to #603

- Add keyword argument `plane_segmentation_name` to identify which `PlaneSegmentation` to add from ophys metadata
- Expose `plane_segmentation_name` at `BaseSegmentationExtractorInterface.add_to_nwbfile()` to support adding segmentation output from multiple planes
Related to #601

### DIFF
--- a/src/neuroconv/tools/roiextractors/roiextractors.py
+++ b/src/neuroconv/tools/roiextractors/roiextractors.py
@@ -656,7 +656,8 @@ def add_plane_segmentation(
     segmentation_extractor: SegmentationExtractor,
     nwbfile: NWBFile,
     metadata: Optional[dict],
-    plane_segmentation_index: int = 0,
+    plane_segmentation_name: Optional[str] = None,
+    plane_segmentation_index: Optional[int] = None,  # TODO: to be removed
     include_roi_centroids: bool = True,
     include_roi_acceptance: bool = True,
     mask_type: Optional[str] = "image",  # Optional[Literal["image", "pixel"]]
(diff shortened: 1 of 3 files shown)


src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -1042,14 +1098,12 @@ def add_segmentation(
     # Add device:
     add_devices(nwbfile=nwbfile, metadata=metadata)
 
-    # Add imaging plane
-    add_imaging_plane(nwbfile=nwbfile, metadata=metadata)
-
-    # PlaneSegmentation:
```
weiglszonja, line 1048: The reason why this is removed is because `add_imaging_plane()` is also called from `add_plane_segmentation()`. If you feel like this change is not justified I can revert

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -1042,14 +1098,12 @@ def add_segmentation(
     # Add device:
     add_devices(nwbfile=nwbfile, metadata=metadata)
```
weiglszonja, line 1100:
```suggestion

# `add_imaging_plane()` is also called from `add_plane_segmentation()` no need to add it here
# add_imaging_plane(nwbfile=nwbfile, metadata=metadata)
```

weiglszonja: @CodyCBakerPhD how about this?

PR #624, opened 2023-10-31
Expose DANDI upload job parameters

CodyCBakerPhD:
@garrettmflynn 

Looks like this actually will need one small PR on DANDI to fully expose a certain argument

But there might be a way to work around it for now


garrettmflynn (approved):
Looks good to me! Let's see if we can get around this before the DANDI PR

PR #627, opened 2023-11-02
[Do not merge] Testing SI new release

h-mayorquin:
Should be testing the branc in this PR

https://github.com/SpikeInterface/spikeinterface/pull/2137


CodyCBakerPhD:
I mean, if the pre-release is equivalent to current SI main, it's covered in the 'dev branch' part of our CI on every run

h-mayorquin:
It is not exactly main as SI uses a specific branch to accrue changes and there might not be a an open PR in neuroconv at that specific point in time or the branch might not run together with the cron.

PR #630, opened 2023-11-06
Improve chunking pattern for ecephys recording

CodyCBakerPhD:
An improvement that is only applicable to latest releases of HDMF (might need to bump lower version here)


CodyCBakerPhD:
Will cut a new release after this to alleviate the issues seen by @magland

CodyCBakerPhD:
@bendichter Example result of this branch on a ~28 GB `.ap.bin` file

image

which is squarely at the 10 MB target

bendichter (approved):

PR #631, opened 2023-11-07
Document setting probe in intan.rst

magland:


CodyCBakerPhD:
I would not document this in the minimal gallery for a single format (or if so, do all formats that don't autoset probes)

Could you add a general section as a small tutorial instead? Perhaps with a real probe example as a use case?

magland:
> I would not document this in the minimal gallery for a single format (or if so, do all formats that don't autoset probes)
> 
> Could you add a general section as a small tutorial instead? Perhaps with a real probe example as a use case?

I was hoping to catch users when then do docs -> Extracellular electrophysiology -> Intan -> Copy Paste
because then I could just point labs to that page.

Since electrode locations is crucial information that should go along with every ephys data, I thought it would be appropriate to include it in the example gallery. If you think that's appropriate, I'd be happy to add it to all the formats that don't autoset probes. Of course it would be nice if that text could be maintained in a single location, but I don't know how best to do that. One possibility is to have a comment in the example script

```
# set the probe information
```

and then below the script have a link to a singe page with a description of how to set the probe information.

PR #636, opened 2023-11-09
Fix audio conversion in case of single-channel .wav

magland:
I tried converting a .wav file to .nwb based on this example: https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/behavior/audio.html

and got this error

```
Traceback (most recent call last):
  File "/mnt/ceph/users/magland/dandi_upload/prepare_ralph_av/prepare_ralph_av.py", line 18, in <module>
    interface.run_conversion(nwbfile_path=path_to_save_nwbfile, metadata=metadata)
  File "/mnt/home/magland/miniconda3/envs/dev/lib/python3.10/site-packages/neuroconv/basedatainterface.py", line 99, in run_conversion
    self.add_to_nwbfile(nwbfile_out, metadata=metadata, **conversion_options)
  File "/mnt/home/magland/miniconda3/envs/dev/lib/python3.10/site-packages/neuroconv/datainterfaces/behavior/audio/audiointerface.py", line 214, in add_to_nwbfile
    add_acoustic_waveform_series(
  File "/mnt/home/magland/miniconda3/envs/dev/lib/python3.10/site-packages/neuroconv/tools/audio/audio.py", line 71, in add_acoustic_waveform_series
    data=H5DataIO(SliceableDataChunkIterator(data=acoustic_series, **iterator_options), **compression_options),
  File "/mnt/home/magland/miniconda3/envs/dev/lib/python3.10/site-packages/neuroconv/tools/hdmf.py", line 83, in __init__
    super().__init__(**kwargs)
  File "/mnt/home/magland/miniconda3/envs/dev/lib/python3.10/site-packages/hdmf/utils.py", line 664, in func_call
    return func(args[0], **pargs)
  File "/mnt/home/magland/miniconda3/envs/dev/lib/python3.10/site-packages/hdmf/data_utils.py", line 220, in __init__
    self.buffer_shape = buffer_shape or self._get_default_buffer_shape(buffer_gb=buffer_gb)
  File "/mnt/home/magland/miniconda3/envs/dev/lib/python3.10/site-packages/neuroconv/tools/hdmf.py", line 28, in _get_default_buffer_shape
    smallest_chunk_axis, second_smallest_chunk_axis, *_ = np.argsort(self.chunk_shape)
ValueError: not enough values to unpack (expected at least 2, got 1)
```

Did some troubleshooting and found

```
Sampling rate: 192000 Hz
Acoustic series shape: (979200000,)
```

[...]

PR #638, opened 2023-11-10
Fix GenericDataChunkIterator for large arrays (affects .wav conversion)

magland:
This is a better alternative to https://github.com/catalystneuro/neuroconv/pull/636 as it fixes a more fundamental issue.

The previous version of GenericDataChunkIterator raises an exception in the case where
* The array has only one dimension and
* The array is large enough to not satisfy the early termination condition

This issue was not being picked up by the tests because the simulated .wav files are not nearly large enough to get passed the early termination condition.

The problematic line was

```
smallest_chunk_axis, second_smallest_chunk_axis, *_ = np.argsort(self.chunk_shape)
```

because in the case of a 1D array, there is no second smallest chunk axis. So with this PR, this case is handled separately.

In terms of verifying that this works properly, I temporarily removed the early termination condition and the tests passed (the sample .wav files are single channel). I'm not sure how to put in a proper test though because I don't think we want >1GB test files, and it wouldn't be easy to lower the buffer_gb threshold just for the test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 ### Fixes
 * Fixed GenericDataChunkIterator (in hdmf.py) in the case where the number of dimensions is 1 and the size in bytes is greater than the threshold of 1 GB. [PR #638](https://github.com/catalystneuro/neuroconv/pull/638)
-* changed np.floor/np.prod to math.floor/math.prod in various files
+* Changed `np.floor` and `np.prod` usage to `math.floor` and `math.prod` in various files. [PR #638](https://github.com/catalystneuro/neuroconv/pull/638)
 
 # v0.4.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ### Features
 * Added Pydantic data models of `BackendConfiguration` for both HDF5 and Zarr datasets (container/mapper of all the `DatasetConfiguration`s for a particular file). [PR #568](https://github.com/catalystneuro/neuroconv/pull/568)
 
-
+### Fixes
+* Fixed GenericDataChunkIterator (in hdmf.py) in the case where the number of dimensions is 1 and the size in bytes is greater than the threshold of 1 GB.
 
 # v0.4.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 * Added Pydantic data models of `BackendConfiguration` for both HDF5 and Zarr datasets (container/mapper of all the `DatasetConfiguration`s for a particular file). [PR #568](https://github.com/catalystneuro/neuroconv/pull/568)
 
 ### Fixes
-* Fixed GenericDataChunkIterator (in hdmf.py) in the case where the number of dimensions is 1 and the size in bytes is greater than the threshold of 1 GB.
+* Fixed GenericDataChunkIterator (in hdmf.py) in the case where the number of dimensions is 1 and the size in bytes is greater than the threshold of 1 GB. [PR #638](https://github.com/catalystneuro/neuroconv/pull/638)
 
 # v0.4.5
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### Fixes
 * Fixed GenericDataChunkIterator (in hdmf.py) in the case where the number of dimensions is 1 and the size in bytes is greater than the threshold of 1 GB. [PR #638](https://github.com/catalystneuro/neuroconv/pull/638)
+* changed np.floor/np.prod to math.floor/math.prod in various files
 
 # v0.4.5
 
--- a/src/neuroconv/tools/hdmf.py
+++ b/src/neuroconv/tools/hdmf.py
@@ -25,23 +25,25 @@ class GenericDataChunkIterator(HDMFGenericDataChunkIterator):
 
         buffer_bytes = chunk_bytes
         axis_sizes_bytes = maxshape * self.dtype.itemsize
+        target_buffer_bytes = buffer_gb * 1e9
         if num_axes > 1:
             smallest_chunk_axis, second_smallest_chunk_axis, *_ = np.argsort(self.chunk_shape)
+            # If the smallest full axis does not fit within the buffer size, form a square along the two smallest axes
+            sub_square_buffer_shape = np.array(self.chunk_shape)
+            if min(axis_sizes_bytes) > target_buffer_bytes:
+                k1 = np.floor((target_buffer_bytes / chunk_bytes) ** 0.5)
+                for axis in [smallest_chunk_axis, second_smallest_chunk_axis]:
+                    sub_square_buffer_shape[axis] = k1 * sub_square_buffer_shape[axis]
+                return tuple(sub_square_buffer_shape)
         elif num_axes == 1:
             smallest_chunk_axis = 0
             second_smallest_chunk_axis = -1
+            # Handle the case where the single axis is too large to fit in the buffer
+            if axis_sizes_bytes[0] > target_buffer_bytes:
+                k1 = np.floor(target_buffer_bytes / chunk_bytes)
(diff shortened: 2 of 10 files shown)


magland:
Made an adjustment. I realized you don't want to take the square root (forming a square) when there is only one axis.

src/neuroconv/tools/hdmf.py
```diff
@@ -25,16 +25,29 @@ def _get_default_buffer_shape(self, buffer_gb: float = 1.0) -> Tuple[int]:
...
+            # If the smallest full axis does not fit within the buffer size, form a square along the two smallest axes
+            sub_square_buffer_shape = np.array(self.chunk_shape)
+            if min(axis_sizes_bytes) > target_buffer_bytes:
+                k1 = np.floor((target_buffer_bytes / chunk_bytes) ** 0.5)
+                for axis in [smallest_chunk_axis, second_smallest_chunk_axis]:
+                    sub_square_buffer_shape[axis] = k1 * sub_square_buffer_shape[axis]
+                return tuple(sub_square_buffer_shape)
+        elif num_axes == 1:
+            smallest_chunk_axis = 0
+            second_smallest_chunk_axis = -1
```
CodyCBakerPhD, line 40: There doesn't look to be any reason to set this in this event

```suggestion
```

CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 ### Features
 * Added Pydantic data models of `BackendConfiguration` for both HDF5 and Zarr datasets (container/mapper of all the `DatasetConfiguration`s for a particular file). [PR #568](https://github.com/catalystneuro/neuroconv/pull/568)
 
-
+### Fixes
+* Fixed GenericDataChunkIterator (in hdmf.py) in the case where the number of dimensions is 1 and the size in bytes is greater than the threshold of 1 GB.
```
CodyCBakerPhD, line 7:
```suggestion
* Fixed GenericDataChunkIterator (in hdmf.py) in the case where the number of dimensions is 1 and the size in bytes is greater than the threshold of 1 GB. [PR #638](https://github.com/catalystneuro/neuroconv/pull/638)
```

CodyCBakerPhD:
Good catch

>  I'm not sure how to put in a proper test though because I don't think we want >1GB test files, and it wouldn't be easy to lower the buffer_gb threshold just for the test.

It would be fine to either extend the number of frames in https://github.com/catalystneuro/neuroconv/blob/01d23add0b59b664f88b972411898f42d03f9db5/tests/test_behavior/test_audio_interface.py#L45 to exceed 1 GB (in all honesty not much of a slowdown and they get cleaned up anyway)

But yes the better solution would be to reduce the `buffer_gb` via `iterator_options` to a value below the that size to cause multiple buffer conditions to trigger - could you add that to line https://github.com/catalystneuro/neuroconv/blob/01d23add0b59b664f88b972411898f42d03f9db5/tests/test_behavior/test_audio_interface.py#L202?

magland:
@CodyCBakerPhD I did what you suggested with that test line. It was a delicate balance for choosing buffer_gb high enough to avoid an error about buffer_gb must be greater than chunk size, and also low enough compared with the num. frames in the test (which I had to increase). I verified that in this test, it does cover past the early termination line.

Also I had to cast some `np.floor(...)` to `int` because I was getting a complaint somewhere mysterious within hdmf about np.int64 (or was it np.float64, I can't remember).

CodyCBakerPhD:
> Also I had to cast some np.floor(...) to int because I was getting a complaint somewhere mysterious within hdmf about np.int64 (or was it np.float64, I can't remember).

Oh, thank you for reminding me - if you're willing to make another set of adjustments, could you also swap all `np.prod` and `np.floor`-like operations to `math.prod` and `math.floor`? There's a critical overflow error that can rear its head otherwise, and I've been meaning to do it to this file

It should also fix that HDMF behavior you observe otherwise (I know `int` also has the same effect, but that's considered implicit knowledge of the cast, and zen of Python says to prefer explicit things, so `math.floor` is the recommendation)

magland:
I did as you requested with np.floor/np.prod, and I adjusted in other files as well (wasn't sure if that's what you were intending) so I made a note in the changelog, and you should take a look at it.

src/neuroconv/tools/hdmf.py
```diff
@@ -62,10 +75,10 @@ def _get_default_buffer_shape(self, buffer_gb: float = 1.0) -> Tuple[int]:
                 buffer_bytes *= chunks_on_axis
                 padded_buffer_shape[axis] = self.maxshape[axis]
             else:  # Found an axis that is too large to use with the rest of the buffer; calculate how much can be used
-                k3 = np.floor(target_buffer_bytes / buffer_bytes)
+                k3 = int(math.floor(target_buffer_bytes / buffer_bytes))
```
CodyCBakerPhD, line 78:
```suggestion
                k3 = math.floor(target_buffer_bytes / buffer_bytes)
```

`math.floor` returns a proper `int` type already

CHANGELOG.md
```diff
@@ -3,7 +3,9 @@
 ### Features
 * Added Pydantic data models of `BackendConfiguration` for both HDF5 and Zarr datasets (container/mapper of all the `DatasetConfiguration`s for a particular file). [PR #568](https://github.com/catalystneuro/neuroconv/pull/568)
 
-
+### Fixes
+* Fixed GenericDataChunkIterator (in hdmf.py) in the case where the number of dimensions is 1 and the size in bytes is greater than the threshold of 1 GB. [PR #638](https://github.com/catalystneuro/neuroconv/pull/638)
+* changed np.floor/np.prod to math.floor/math.prod in various files
```
CodyCBakerPhD, line 8:
```suggestion
* Changed `np.floor` and `np.prod` usage to `math.floor` and `math.prod` in various files. [PR #638](https://github.com/catalystneuro/neuroconv/pull/638)
```

minor changelog formatting

CodyCBakerPhD (approved):

CodyCBakerPhD:
> I did as you requested with np.floor/np.prod, and I adjusted in other files as well (wasn't sure if that's what you were intending) so I made a note in the changelog, and you should take a look at it.

Beautiful, thanks a bunch!

Honestly, it's not a bad idea to just use it as the go-to. The other places would have required some crazy size inputs to experience the overflow, but they still technically could have succumbed to it

src/neuroconv/tools/testing/mock_ttl_signals.py
```diff
@@ -84,7 +85,7 @@ def generate_mock_ttl_signal(
 
     if np.issubdtype(dtype, np.unsignedinteger):
         # If data type is an unsigned integer, increment the signed default values by the midpoint of the unsigned range
-        shift = np.floor(np.iinfo(dtype).max / 2).astype(int)
+        shift = math.floor(np.iinfo(dtype).max / 2).astype(int)
```
CodyCBakerPhD, line 88:
```suggestion
        shift = math.floor(np.iinfo(dtype).max / 2)
```

failing test

src/neuroconv/tools/testing/mock_ttl_signals.py
```diff
@@ -271,7 +272,7 @@ def regenerate_test_cases(folder_path: FolderPathType, regenerate_reference_imag
         if regenerate_reference_images:
             fig.add_trace(
                 go.Scatter(y=time_series_data, text=time_series_name),
-                row=np.floor(plot_index / num_cols).astype(int) + 1,
+                row=math.floor(plot_index / num_cols).astype(int) + 1,
```
CodyCBakerPhD, line 275:
```suggestion
                row=math.floor(plot_index / num_cols) + 1,
```
failing test

magland: oops

CodyCBakerPhD:
@magland Also, pro-tip when accepting several commented suggestions; the 'add to batch' command lets you commit all the suggestions together in one push (nice for avoiding multiple dispatches of the CI, which is technically wasted computation - not that we get billed for it thankfully)

Issue #634, opened 2023-11-08
[Bug]: Conversion from Intan ephys - electrode positions not present in NWB file


magland:
### What happened?

We used the [recommended script](https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/recording/intan.html) to convert Intan ephys data to NWB file. As part of the conversion we set the probe information using the following.

```
probe_path = 'C:\\path\\to\\probe.csv'
probe_df = pd.read_csv(probe_path); probe_df = probe_df.drop(columns = ['Unnamed: 0'])
probe = generate_multi_columns_probe(num_columns = 2)
probe = probe.from_dataframe(probe_df)
...
intan_interface.recording_extractor.set_probe(probe)
```

The resulting NWB does not have `rel_x` and `rel_y` columns in the the general/extracellular_ephys/electrodes table. There is a locations column, but it is filled with ['unknown', 'unknown', ...].

@oghenand Could you please provide the probe.csv file? @CodyCBakerPhD was requesting to see this.

### Steps to Reproduce

```python
# See above
```

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.7

### Package Versions

This screenshot shows that the probe_df has x and y columns.

![image](https://github.com/catalystneuro/neuroconv/assets/3679296/7023b938-dce8-4ab1-b55e-829f13b0ab31)

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

magland:
@alejoe91 is suggesting to try

```python
intan_interface.recording_extractor.set_probe(probe, in_place=True)
```

Note: `in_place=True`

alejoe91:
@CodyCBakerPhD @bendichter if this turns out to work (I'm 99.8% sure about it), we should add a page on Neuroconv docs to show how to properly add a `probeinterface.Probe`. Where would that page go?

magland:
Yes it seems to have worked... thanks @CodyCBakerPhD and @alejoe91 

I'm happy to help with the docs too once you decide where that should go.

magland:
[...]

PR #639, opened 2023-11-10
add set_probe method to BaseRecordingExtractorInterface

magland:
@alejoe91 @CodyCBakerPhD as we were discussing in #634


src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -225,6 +225,29 @@ def set_aligned_segment_starting_times(self, aligned_segment_starting_times: Lis
...
+        """
+        # Set the probe to the recording extractor
+        self.recording_extractor.set_probe(
+            probe,
+            in_place=True,
+            group_mode=group_mode,
+        )
+        # Spike interface sets the "group" property
+        # Buy neuroconv expects "group_name" to be set
+        self.recording_extractor.set_property("group_name", self.recording_extractor.get_property("group").astype(str))
```
CodyCBakerPhD, line 249: If you really want to keep the `group_name` stuff here, then I'd recommend at least

```python
if "group_name" in recording.get_property_keys():
    self.recording_extractor.set_property("group_name", self.recording_extractor.get_property("group").astype(str))
```

to avoid adding it unnecessarily when it hasn't been set during interface initialization

But I wonder, have you tried the deletion method yet? I think it would produce same group structure result but one less column of duplicated info on the table

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -225,6 +225,29 @@ def set_aligned_segment_starting_times(self, aligned_segment_starting_times: Lis
...
+            This is a required parameter to avoid the pitfall of using the wrong mode.
+        """
+        # Set the probe to the recording extractor
+        self.recording_extractor.set_probe(
+            probe,
+            in_place=True,
+            group_mode=group_mode,
+        )
+        # Spike interface sets the "group" property
+        # Buy neuroconv expects "group_name" to be set
```
CodyCBakerPhD, line 248:
```suggestion
        # But neuroconv allows "group_name" property to override spike interface "group" value
```

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
magland: The problem with the deletion method is that we need to also convert int to str

CodyCBakerPhD: Oh no, channel groups in SI are still allowed to be `int`? So `group_name` is still needed after all it seems... sounds fine then

CodyCBakerPhD:
Just need some unit tests for this; would you be willing to add a general tutorial in the user guide section as well in this PR or would you prefer that in a follow-up?

Basically, what a user would need is a clear way to 'how' to get that probe object. I know SI/PI has a couple of ways (including some predefined that can be loaded? But they can also be set entirely in-memory, as well as that CSV approach you used)

magland:
> Just need some unit tests for this; would you be willing to add a general tutorial in the user guide section as well in this PR or would you prefer that in a follow-up?
> 
> Basically, what a user would need is a clear way to 'how' to get that probe object. I know SI/PI has a couple of ways (including some predefined that can be loaded? But they can also be set entirely in-memory, as well as that CSV approach you used)

Yeah I think we can include the tutorial in this PR. Will probably need help from @alejoe91

I know Alessio also wanted to add a set_probe_group

magland:
Is there a way to get the test data on my local machine so I can run the tests that require access to that data?

CodyCBakerPhD:
> Is there a way to get the test data on my local machine so I can run the tests that require access to that data?

Yep, just give our [Developer Guide](https://neuroconv.readthedocs.io/en/main/developer_guide.html) a perusal

I would say though, that some cleaner unit tests might use the mock class (which uses randomly generated data via SI) https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/testing/mock_interfaces.py#L121 and belong in approximately this location of the testing suite: https://github.com/catalystneuro/neuroconv/blob/main/tests/test_ecephys/test_ecephys_interfaces.py

The tutorial could (probably should) use some of the real testing data however; but the only testing dataset I actually know the probe of would be https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/neuroscope/dataset_1 which became [dandiset 000003](https://dandiarchive.org/dandiset/000003/0.230629.1955) and has probe models [defined in the paper](https://www.cell.com/neuron/fulltext/S0896-6273(16)30950-3?_returnURL=https%3A%2F%2Flinkinghub.elsevier.com%2Fretrieve%2Fpii%2FS0896627316309503%3Fshowall%3Dtrue)

magland:
@CodyCBakerPhD 
This is a bit sloppy but it's the best I can do for now...

I added a call to `interface.set_probe()` in the tests

https://github.com/magland/neuroconv/blob/b8f793b5b7737d9c6789618bd80147a590b9f4d3/src/neuroconv/tools/testing/data_interface_mixins.py#L104-L130

This is only for the Intan interface, because when I did this for all interfaces, then I get into trouble with other tests that expect that the number of groups is 1.

And then I make sure the nwb recording matches with the properties rel_x, rel_y, rel_z, and group:

https://github.com/magland/neuroconv/blob/b8f793b5b7737d9c6789618bd80147a590b9f4d3/src/neuroconv/tools/testing/data_interface_mixins.py#L333-L342

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -225,6 +225,29 @@ def set_aligned_segment_starting_times(self, aligned_segment_starting_times: Lis
             ]
             self.set_aligned_segment_timestamps(aligned_segment_timestamps=aligned_segment_timestamps)
 
+    def set_probe(self, probe, *, group_mode: Union[Literal["by_shank"], Literal["by_probe"]]):
```
bendichter, line 228:
```suggestion
    def set_probe(self, probe, *, group_mode: Literal["by_shank", "by_probe"]):
```

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -120,6 +130,26 @@ def test_conversion_as_lone_interface(self):
                 self.run_custom_checks()
 
 
+def _create_mock_probe_for_recording(recording):
+    import probeinterface as pi
+    import spikeinterface as si
+
+    R: si.BaseRecording = recording
```
bendichter, line 137: What's on here? Why not just use a normal typehint?

```suggestion
def _create_mock_probe_for_recording(recording: si.BaseRecording):
    import probeinterface as pi
    import spikeinterface as si

    recording
```
Is it because you want passive loading of probeinterface and spikeinterface?

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -120,6 +130,26 @@ def test_conversion_as_lone_interface(self):
...
+    for i in range(n):
+        x = i // 4
+        y = i % 4
+        positions[i] = x, y
+    positions *= 20
+    probe = pi.Probe(ndim=2, si_units="um")
+    probe.set_contacts(positions=positions, shapes="circle", shape_params={"radius": 5})
+    probe.set_device_channel_indices(np.arange(n))
+    # set shank ids as 0, 1, 2, 0, 1, 2, ...
+    probe.set_shank_ids(np.arange(n) % 3)
```
bendichter, line 149: a bit odd to have electrodes that are near each other be assigned to different shanks, but I don't think this really poses a problem for now. We can always improve this to be more realistic later if needed

src/neuroconv/tools/testing/data_interface_mixins.py
magland: @bendichter yes... I didn't want to create a dependency on spikeinterface at the module level.

bendichter: It looks like the only attribute you are using from the recording in the number of electrodes. Would it be better to just accept that as an input? Or do you want to allow yourself the flexibility to use more attributes from the RecordingExtractor object in the future?

magland: you're right. I have adjusted this so it only takes in the num_channels parameter. This simplifies the typing as well.

src/neuroconv/tools/testing/data_interface_mixins.py
magland: With my latest commit I have adjusted this to be more realistic.

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -120,6 +131,33 @@ def test_conversion_as_lone_interface(self):
                 self.run_custom_checks()
 
 
+def _create_mock_probe(*, num_channels: int):
```
CodyCBakerPhD, line 134:
```suggestion
def _create_mock_probe(num_channels: int):
```

We don't generally allow arbitrary positional args to functions (the '*' in signature)

We also tend to follow the practice enforced these days by HDMF of always using keyword arguments

CodyCBakerPhD, line 134: Would it be possible to extend this function to handle `num_groups`? So it constructs a fake multi-shank probe for testing in conjunction with the other interfaces?

cc @samuelgarcia @alejoe91 @h-mayorquin

CodyCBakerPhD, line 134: Also, would you mind moving this to the [`_mock` submodule](https://github.com/catalystneuro/neuroconv/blob/9afd9a9efdf8a5fa4c6f97244b3ec698bdecfdd0/src/neuroconv/tools/testing/_mock/_mock_dataset_models.py) to be nearby other similar helpers?

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -27,6 +30,8 @@
 )
 from neuroconv.utils import NWBMetaDataEncoder
 
+interfaces_for_testing_probe = [IntanRecordingInterface]
```
CodyCBakerPhD, line 33: Since this PR is to add a general method to all `BaseRecordingExtractor`, we really should ensure to the absolute best of our ability that it works with all of the child classes with example data we claim to support, not just the one that started the discussion

I'd really love to see this applied to all the interfaces that rely on this mixin - happy to help with debugging the various cases that can occur throughout this process!

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -225,6 +225,29 @@ def set_aligned_segment_starting_times(self, aligned_segment_starting_times: Lis
             ]
             self.set_aligned_segment_timestamps(aligned_segment_timestamps=aligned_segment_timestamps)
 
+    def set_probe(self, probe, *, group_mode: Literal["by_shank", "by_probe"]):
```
CodyCBakerPhD, line 228:
```suggestion
    def set_probe(self, probe, group_mode: Literal["by_shank", "by_probe"]):
```

We don't tend to hard enforce keyword argument usage with input validation; if we started to, we should do it elsewhere but it would need to be discussed with the rest of the team

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD, line 134:
```suggestion
def _create_mock_probe(num_channels: int):
```

We don't tend to hard enforce keyword argument usage with input validation; if we started to, we should do it elsewhere but it would need to be discussed with the rest of the team first

CodyCBakerPhD:
@magland Thanks for working on this! The core function is nice and simple, but we should really make sure it can be used without error on as many formats as possible notwithstanding highly specific issues to be propagated up to SI/neo

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
h-mayorquin: I think enforcing this is a good idea. Let's discuss this on the next team meeting indeed.

src/neuroconv/tools/testing/data_interface_mixins.py
magland: I exposed the num_shanks parameter, and it defaults to 3.

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
magland: I think it is dangerous to not enforce it, for future-proofing backward compatibility. But for the sake of this PR I have removed the `*`.

src/neuroconv/tools/testing/data_interface_mixins.py
magland: As I said in the other comment, I think it is dangerous to not enforce it, for future-proofing backward compatibility. But for the sake of this PR I have removed the `*`.

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -225,6 +225,29 @@ def set_aligned_segment_starting_times(self, aligned_segment_starting_times: Lis
...
+        group_mode : {'by_shank', 'by_probe'}
+            How to group the channels. If 'by_shank', channels are grouped by the shank_id column.
+            If 'by_probe', channels are grouped by the probe_id column.
+            This is a required parameter to avoid the pitfall of using the wrong mode.
+        """
+        # Set the probe to the recording extractor
+        self.recording_extractor.set_probe(
+            probe,
+            in_place=True,
+            group_mode=group_mode,
```
alejoe91, line 245: just one note: if the probe has less channels than the recording (e.g., you want to only select the `A` channels from an Intan recording), the `set_probe(..., in_place=True)` will fail. 

We can't do anything about it, just something to keep in mind :)

magland:
> @magland Thanks for working on this! The core function is nice and simple, but we should really make sure it can be used without error on as many formats as possible notwithstanding highly specific issues to be propagated up to SI/neo

Thanks @CodyCBakerPhD 

I think I've addressed all of your comments, except possibly the location of the `generate_mock_probe()`. In particular, the test is now applied to all recording extractor interfaces.

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
CodyCBakerPhD: @alejoe91 What would this represent in a real world case?

Would it be indicative of a bad probe definition by the user?

Or are there cases where the user might actually specify a probe correctly, but extra channels show up (possibly from auxiliary sources or a mistake in SI/neo)?

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Oh uhh I guess those other ones ought to be in `_mock` too, will fix that in follow-up then

Thanks! This is fine for now

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -109,6 +111,13 @@ def test_conversion_as_lone_interface(self):
                 self.case = num
                 self.test_kwargs = kwargs
                 self.interface = self.data_interface_cls(**self.test_kwargs)
+                do_set_probe = isinstance(self.interface, BaseRecordingExtractorInterface)
+                if do_set_probe:
+                    assert isinstance(self.interface, BaseRecordingExtractorInterface)
+                    self.interface.set_probe(
+                        generate_mock_probe(num_channels=self.interface.recording_extractor.get_num_channels()),
+                        group_mode="by_shank",
+                    )
```
CodyCBakerPhD, line 120: I would move this section out of the base mixin (which covers things for all data modalities) down into the specific mixing for recording interfaces, then you don't have to have this `isinstance` call

https://github.com/catalystneuro/neuroconv/blob/e2298fa047bace71fc106b65ce4d197767998a0e/src/neuroconv/tools/testing/data_interface_mixins.py#L276

CodyCBakerPhD:
> I think I've addressed all of your comments

Thanks, looking a lot better!


> except possibly the location of the generate_mock_probe().

That place is fine, we can  move later if we want


> In particular, the test is now applied to all recording extractor interfaces.

Great!


One last comment about where the probe injection occurs in the Mixin class

src/neuroconv/tools/testing/data_interface_mixins.py
magland: But I think it belongs between

```
self.interface = self.data_interface_cls(**self.test_kwargs)
```

and

```
self.check_metadata_schema_valid()
```

so I'm not sure how to move it down into RecordingExtractorInterfaceTestMixin without adding another abstract method such as `self.mutate_the_interface_with_mock_data()`

LMK if I should go that route.

CodyCBakerPhD: You could just override the method entirely on the recording specific child mixin

I also have a suggestion at that level for a better trigger condition to use the mock instead of an existing probe (thinking of SpikeGLX in this case)

```python
class RecordingExtractorInterfaceTestMixin(DataInterfaceTestMixin, TemporalAlignmentMixin):
...

    def test_conversion_as_lone_interface(self):
        interface_kwargs = self.interface_kwargs
        if isinstance(interface_kwargs, dict):
            interface_kwargs = [interface_kwargs]
        for num, kwargs in enumerate(interface_kwargs):
            with self.subTest(str(num)):
                self.case = num
                self.test_kwargs = kwargs
                self.interface = self.data_interface_cls(**self.test_kwargs)

                if self.interface.recording_extractor.has_probe():  # to not override probes that may have been autoset per format
                    assert isinstance(self.interface, BaseRecordingExtractorInterface)
                    self.interface.set_probe(
                        generate_mock_probe(num_channels=self.interface.recording_extractor.get_num_channels()),
                        group_mode="by_shank",
                    )

                self.check_metadata_schema_valid()
                self.check_conversion_options_schema_valid()
                self.check_metadata()
                self.nwbfile_path = str(self.save_directory / f"{self.__class__.__name__}_{num}.nwb")
                self.run_conversion(nwbfile_path=self.nwbfile_path)
                self.check_read_nwb(nwbfile_path=self.nwbfile_path)

                # Any extra custom checks to run
                self.run_custom_checks()
```

magland: Do you mean we should override that method and duplicate all of that same code? It's okay with me, but I just wanted to check whether that's what you mean.

CodyCBakerPhD: >  It's okay with me, but I just wanted to check whether that's what you mean.

It is. go ahead

magland:
@CodyCBakerPhD I have moved the set_probe() test code to the RecordingExtractorInterfaceTestMixin class, and I have used the `not has_probe()` condition for setting the probe.

CodyCBakerPhD:
Looks great to me! Thanks for all the hard work on this

CodyCBakerPhD (approved):

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
alejoe91: yes I think the second option is what I had in mind. You have a recording with Ephys + Aux channels and you might want to select ephys channels with the probe slicing

PR #632, opened 2023-11-07
Change `Fluorescence`/`DFOverF` metadata schema for RoiResponseSeries from `array` type to `object`

weiglszonja:
Final piece for #601 
Resolve https://github.com/catalystneuro/neuroconv/issues/603

Changes the current metadata schema for `Fluorescence` and `DfOverF` 
- the name of the plane segmentation is used to identify which traces are added to the file. 
- the name of the traces can be changed in the inner schema for the response series metadata (see example below)

Example metadata structure:
```
dict(
    Ophys=dict(
        Fluorescence=dict(
            name="Fluorescence",
            PlaneSegmentationChan1Plane0=dict(
                # raw, deconvolved, neuropil as returned from extractor.get_traces_dict()
                # the "name" determines what the trace will be named in the nwb file in Fluorescence.roi_response_series
                raw=dict(name="RoiResponseSeriesChan1Plane0", description="raw fluorescence signal"),
                deconvolved=dict(name="DeconvolvedChan1Plane0", description="deconvolved fluorescence signal"),
                neuropil=dict(name="NeuropilChan1Plane0", description="neuropil fluorescence signal"),
            ),
            PlaneSegmentationChan1Plane1=dict(
                raw=dict(name="RoiResponseSeriesChan1Plane0", description="raw fluorescence signal"),
                deconvolved=dict(name="DeconvolvedChan1Plane0", description="deconvolved fluorescence signal"),
                neuropil=dict(name="NeuropilChan1Plane0", description="neuropil fluorescence signal"),
            ),
        ),
        DfOverF =dict(
            name="DfOverF"
            PlaneSegmentationChan1Plane0=dict(
                dff=dict(name="RoiResponseSeriesChan1Plane0", description="Array of df/F traces."),
            ),
            PlaneSegmentationChan1Plane1=dict(
                dff=dict(name="RoiResponseSeriesChan1Plane0", description="Array of df/F traces."),
            ),
        ),
    )
)

```

PR #640, opened 2023-11-12
Fix pattern properties

garrettmflynn:
This PR updates the `patternProperties` fields added in #632 to exist in the correct JSON Schema location for proper validation and use in the GUIDE (https://github.com/NeurodataWithoutBorders/nwb-guide/pull/505).


CodyCBakerPhD:
Enlisting @weiglszonja to help review since this was something @garrettmflynn noticed while we were at SfN and I failed to catch it in that last PR

weiglszonja:
@garrettmflynn thank you for adding a test for this, I'm not sure I understand why it is a problem that I included `patternProperties` under `properties`?  (I just run your unittest with the previous version of the schema that does this and it run without an error)

CodyCBakerPhD:
I think his reasoning was that according to some JSON examples, such as https://stackoverflow.com/questions/30769539/how-to-make-a-patternproperty-required-in-json-schema-ruby, show it acting at the same level as the additionalProperties level, saying effectively 'allow additional properties, but only if they follow this pattern'

Since there weren't any JSON validation error-raising tests in the original PR (my bad for not thinking to request those as proof the JSON feature works, or that we were using it correctly) it's hard to tell

But when we tried rendering basic forms using this schema adjustment, the pattern properties end up as items themselves (entries in the `Fluorescence` level)

garrettmflynn:
Thanks for the explanation, Cody! 

As a way to test that, you can set additionalProperties (set outside the properties field) to false on the old schema. Since patternProperties should not take effect there, you should get an error for any property besides "name"

garrettmflynn:
Another bit of evidence would be that simply translating the existing patternProperties regex to the outer level resulted in an error since it _included_ name. Apparently patternProperties take priority, so name was expected to be an object—although I also had to update the $ref to exist directly beneath the last patternProperties at that point too.

Overall, the simplest way to understand why patternProperties has to exist outside properties is simply to disambiguate this special property from any user-defined properties (which, using this system, could include something called patternProperties).

src/neuroconv/datainterfaces/ophys/basesegmentationextractorinterface.py
```diff
@@ -49,7 +49,7 @@ def get_metadata_schema(self) -> dict:
         roi_response_series_per_plane_schema = dict(
             type="object", patternProperties={"^[a-zA-Z0-9]+$": roi_response_series_schema}
         )
-        metadata_schema["properties"]["Ophys"]["properties"]["Fluorescence"]["properties"].update(
+        metadata_schema["properties"]["Ophys"]["properties"]["Fluorescence"].update(
             patternProperties={"^[a-zA-Z0-9]+$": roi_response_series_per_plane_schema}
```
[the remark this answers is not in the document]
garrettmflynn: Ah, thanks for trying here! I'd searched for the regex in the code but didn't find this location.

I'll see what I can do :)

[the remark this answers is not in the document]
garrettmflynn: @weiglszonja Can you point me to where the schema says that these properties should be an `array`? I haven't been able to see where that requirement is indicated.

CodyCBakerPhD:
@garrettmflynn Any updates/progress on this?

garrettmflynn:
No, I thought this was functional now.

Do we just need to update the changelog or are there blocking issues?

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,53 @@
...
+                    dff=dict(name="RoiResponseSeriesChan1Plane0", description="Array of df/F traces."),
+                ),
+                PlaneSegmentationChan1Plane1=dict(
+                    dff=dict(name="RoiResponseSeriesChan1Plane0", description="Array of df/F traces."),
+                ),
+            ),
+        ),
+    )
+
+    # jsonschema.validate(instance=metadata, schema=metadata_schema)
```
CodyCBakerPhD, line 52:
```suggestion
```

cleaning

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,53 @@
+import unittest
+from pathlib import Path
+
+import jsonschema
+
+from neuroconv.utils import load_dict_from_file
+from neuroconv.utils.json_schema import validate_metadata
+
+
+def test_metadata_schema():
```
CodyCBakerPhD, line 10: So, ideally, to really test that this schema catches violations properly, can you add a second test case that causes the error to be raised?

I'd use `pytest.raises` similar to [this example](https://github.com/catalystneuro/neuroconv/blob/554e07b96072eb6180eccfe3bdc4d049794a4f8d/tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_common_dataset_io_configuration_model.py#L63-L68)

garrettmflynn: I opted to use the `iter_errors` method of the `jsonschema.Draft7Validator` class to get a list of all the errors for a given instance. 

I had to update the schema to have `additionalProperties: false` to get the `patternProperties` regex to actually factor into the validation. 

Currently, the schema allows for a plane without any additional nested dictionaries:
```json
{
"DfOverF": "{
             "name":"DfOverF",
             "PlaneName": {}
       },
}
```

Is this intended?

CodyCBakerPhD: It's technically allowed. The `DfOverF` is actually a container of `RoiResponseSeries`. And it's possible to have an empty container

But TBH I'd disallow it if that's easy (`minItems=1` perhaps?)

garrettmflynn:
Added `minProperties: 2` to the top `patternProperties` so properties like `Fluorescence` must have a name and at least one plane.

garrettmflynn:
Actually, you were saying that we could directly import the metadata_schema.json file into the base segmentation interface? 

Going through the GUIDE, it does look like several properties (e.g. minProperties) are not included and have to be added to that class.

src/neuroconv/datainterfaces/ophys/basesegmentationextractorinterface.py
```diff
@@ -71,10 +76,10 @@ def get_metadata_schema(self) -> dict:
             properties=dict(
                 name=dict(type="string", default="SegmentationImages"),
                 description=dict(type="string"),
-                patternProperties={
-                    "^[a-zA-Z0-9]+$": summary_images_per_plane_schema,
-                },
             ),
+            patternProperties={
+                "^(?!(name|description)$)[a-zA-Z0-9]+$": summary_images_per_plane_schema,
```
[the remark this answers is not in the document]
garrettmflynn: You got it! It's a little weird, but otherwise the patternProperties schema will override the hand-coded one on "name" and other caught properties.

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,201 @@
...
+                "name": "DfOverF",
+                invalid_plane_name: dict(),  # Uncheked
+                create_valid_plane_name(invalid_plane_name): dict(),
+                create_valid_plane_name(invalid_plane_name, 1): dict(
+                    raw=dict(),
+                ),
+            },
+            SegmentationImages={
+                "name": "SegmentationImages",
+                invalid_plane_name: dict(),  # Uncheked
```
CodyCBakerPhD, line 105:
```suggestion
                invalid_plane_name: dict(),  # Unchecked
```

CodyCBakerPhD:
> Actually, you were saying that we could directly import the metadata_schema.json file into the base segmentation interface?

I would say let's do it in a follow-up; this PR as a fix to the new usage of patternProperties along with unit tests is sufficient for this PR

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,201 @@
...
+            session_start_time="2020-01-01T00:00:00",
+            session_description="Auto-generated by neuroconv",
+            identifier="1234",
+        ),
+        Ophys=dict(
+            Fluorescence={
+                "name": "Fluorescence",
+                invalid_plane_name: dict(),  # Unchecked
+                create_valid_plane_name(invalid_plane_name): dict(),
+                create_valid_plane_name(invalid_plane_name, 1): dict(
```
CodyCBakerPhD, line 91:
```suggestion
                "plane_segmentation_chan1_plane0": dict(),  # Unchecked
                "PlaneSegmentationChan1Plane0": dict(),
                "PlaneSegmentationChan1Plane1": dict(
```

I highly recommend making this aspect of the tests explicit so that a future developer can see at a glance what values are being asserted without having to scroll to the find the variable `invalid_plane_name` defined above, or having to understand how the `create_valid_plane_name` function sanitizes it

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,201 @@
...
+    metadata = dict(
+        NWBFile=dict(
+            session_start_time="2020-01-01T00:00:00",
+            session_description="Auto-generated by neuroconv",
+            identifier="1234",
+        ),
+        Ophys=dict(
+            Fluorescence={
+                "name": "Fluorescence",
+                invalid_plane_name: dict(),  # Unchecked
```
CodyCBakerPhD, line 89: Also, what does this comment mean; what is unchecked here?

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,201 @@
...
+                create_valid_plane_name(invalid_plane_name, 1): dict(
+                    raw=dict(),
+                ),
+            },
+        ),
+    )
+
+    validator = jsonschema.Draft7Validator(metadata_schema)
+
+    errors = list(map(lambda e: str(e).split("\n")[0], validator.iter_errors(metadata)))
```
CodyCBakerPhD, line 116:
```suggestion
    errors = [str(error).split("\n")[0] for error in validator.iter_errors(metadata)]
```

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,201 @@
...
+
+    iterable = iter(errors)
```
CodyCBakerPhD, line 118:
```suggestion
```

tests/test_minimal/test_metadata_schema.py
garrettmflynn: Updated. The value is unchecked because the key is incorrect.

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,201 @@
...
+                create_valid_plane_name(invalid_plane_name): dict(),
+                create_valid_plane_name(invalid_plane_name, 1): dict(
+                    raw=dict(),
+                ),
+            },
+            DFOverF={
+                "name": "DfOverF",
+                invalid_plane_name: dict(),  # Unchecked
+                create_valid_plane_name(invalid_plane_name): dict(),
+                create_valid_plane_name(invalid_plane_name, 1): dict(
```
CodyCBakerPhD, line 99: I would also give more unique valid and invalid name patterns so they show up more clearly in the assertions

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,201 @@
...
+    nExpectedErrors = 12
+    assert len(errors) == nExpectedErrors
+
+    assert device_error == next(iterable, None)
+
+    # Fluorescence
+    assert not_enough_properties_error in next(iterable, "")
+    assert name_error == next(iterable, None)
+    assert description_error == next(iterable, None)
+    assert plane_name_error in next(iterable, "")
```
CodyCBakerPhD, line 128: Something like the following is the go-to suggestion for making the assertion cleaner and less brittle

Though the repeated error headers across

```suggestion
    expected_errors = [
        "does not have enough properties",
        "does not have enough properties",
        "does not have enough properties",
        "'Devices' is a required property"
        "plane_segmentation_chan1_plane0"
        "'name' is a required property"
        "'description' is a required property"
        "'plane_segmentation_chan1_plane0' does not match any of the regexes"
    ]
    assert sort(errors) == sort(expected_errors)
```

or similarly if you only wanted to check that a handful of the really important expected ones were caught but don't care about the full values

```python
for expected_error in expected_errors:
    assert expected_error in errors
```

tests/test_minimal/test_metadata_schema.py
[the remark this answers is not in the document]
garrettmflynn: Turns out there's a quick way to grab identifying information like that from the error directly: `e.message` and `e.json_path`. I've adapted the checks to confirm the errors are coming from the right location in the metadata

tests/test_minimal/test_metadata_schema.py
garrettmflynn: Updated to match both the `message` and the `path` of each error. Might be too verbose, or there might be a better way to do this joint search in Python / for our testing suite—so just let me know

tests/test_minimal/test_metadata_schema.py
garrettmflynn: Moved to a method that searches the error list for matching errors instead of directly checking at a certain index.

tests/test_minimal/test_metadata_schema.py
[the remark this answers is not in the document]
CodyCBakerPhD: message and json_path both look better / more structured, yes

tests/test_minimal/test_metadata_schema.py
CodyCBakerPhD: matching more is better than matching less

tests/test_minimal/test_metadata_schema.py
CodyCBakerPhD: Names are indeed more appropriate now, thanks

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,285 @@
...
+    )
+
+    validate_metadata(metadata=metadata, schema=metadata_schema)
+
+
+invalid_plane_name = "plane_segmentation_chan1_plane0"
+name_error = "'name' is a required property"
+description_error = "'description' is a required property"
+plane_name_error_base = "does not match any of the regexes"
+not_enough_properties_error = "does not have enough properties"
```
CodyCBakerPhD, line 68: Thanks for making stuff below more explicit

Shouldn't need these defined globally anymore

```suggestion
```

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # One for each of Fluorescence, DFOverF, and SegmentationImages
+    assert (
+        count_matching_elements(
+            errors,
+            lambda o: f"'fluorescence_chan1_plane0' {plane_name_error_base}" in o["message"]
+            and o["path"] == "$.Ophys.Fluorescence",
+        )
+        == 1
+    )
```
CodyCBakerPhD, line 130: Still relying on a lot of implicit machinery here

I'd recommend the following pattern

```python
expected_errors = [
   {
        "message": "<insert full message, hard-coded>",
        "json_path": "<full, readable path for that message>"
    },
    ... # repeat for all messages you expect
]

for expected_error in expected_errors:
    assert expected_error in errors
```

That, combined with the assertion on the number of errors, is effectively testing equality of the lists but in a way more easily readable at a glance by anyone coming back to this months/years later

garrettmflynn: Alright pushing now!

tests/test_minimal/test_metadata_schema.py
```diff
@@ -0,0 +1,254 @@
...
+
+    validator = jsonschema.Draft7Validator(metadata_schema)
+
+    errors = [{"message": error.message, "path": error.json_path} for error in validator.iter_errors(metadata)]
+
+    def sorting_fn(o):
+        return o["message"]
+
+    errors.sort(key=sorting_fn)
+    expected_errors.sort(key=sorting_fn)
```
CodyCBakerPhD, line 159: These don't need to be sorted if we're using the strategy of iterating over elements asking if they are `in` the list, do we?

```suggestion
```

CodyCBakerPhD:
@garrettmflynn Tests are looking way better now, thank you!

One small leftover about not needing the sorting any more and this should be good to go

tests/test_minimal/test_metadata_schema.py
garrettmflynn: Ah duh. Thanks!

garrettmflynn:
This is somewhat relevent, though could be handled in a follow-up: https://github.com/NeurodataWithoutBorders/nwb-guide/pull/505#issuecomment-1834409627

PR #648, opened 2023-11-20
Fix number_of_jobs parsing

garrettmflynn:
Replaced `jobs` with the correct variable (`number_of_jobs`) when parsing the user's selection of job count from the GUIDE.


CodyCBakerPhD (approved):

CodyCBakerPhD:
Good catch; this wasn't caught before because apparently all my personal testing devices are single-threaded CPU's lol so the early return prevented reaching that line (pretty sure the GitHub action runners have same problem too)

PR #649, opened 2023-11-21
Modify the filtering of traces to also filter out traces with empty values

weiglszonja:
as seen with the Pinto lab, traces from a second channel can be "empty"

### DIFF
--- a/src/neuroconv/tools/roiextractors/roiextractors.py
+++ b/src/neuroconv/tools/roiextractors/roiextractors.py
@@ -903,7 +903,9 @@ def add_fluorescence_traces(
     traces_to_add = segmentation_extractor.get_traces_dict()
 
     # Filter empty data
-    traces_to_add = {trace_name: trace for trace_name, trace in traces_to_add.items() if trace is not None}
+    traces_to_add = {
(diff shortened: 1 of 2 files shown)


src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -903,7 +903,9 @@ def add_fluorescence_traces(
     traces_to_add = segmentation_extractor.get_traces_dict()
 
     # Filter empty data
-    traces_to_add = {trace_name: trace for trace_name, trace in traces_to_add.items() if trace is not None}
+    traces_to_add = {
+        trace_name: trace for trace_name, trace in traces_to_add.items() if trace is not None and trace.size
```
CodyCBakerPhD, line 907:
```suggestion
        trace_name: trace for trace_name, trace in traces_to_add.items() if trace is not None and trace.size != 0
```

Small suggestion for improved readability (more explicit)

CodyCBakerPhD:
Pretty interesting to see that this can happen - especially considering the work of Riche over on ROI extractors which assumed such files simply didn't exist in the Suite2P folder...

Simple cost effective fix however

LGTM

CodyCBakerPhD (approved):

PR #653, opened 2023-11-23
Improve testing name change

h-mayorquin:
As it is unclear when we will be able to support python 3.12 (as many of our dependencies might have upper limits on their own dependencies) I want to separate the re-formating of the tests that we did in that PR.

This PR does this and I also want to test if the patch to spikeinterface solve the problems that @CodyCBakerPhD pointed out to me.

Let's check.


h-mayorquin:
@CodyCBakerPhD 
So I fixed the first batch of erros. The second batch of errors happen because numba if available. One of the quality metrics that was recently added can only be imported if you have numba so somewhere here there is an error generated:

https://github.com/catalystneuro/neuroconv/blob/1d3d58d4b53570d20b5e81abde7ada7f9f02fa47/src/neuroconv/tools/spikeinterface/spikeinterface.py#L1319-L1330

We have some options here at neuroconv:
1) We drop that specific quality metric from here.
2) We add numba as a depedency to the context that test this. This is in line with @CodyCBakerPhD request  here https://github.com/SpikeInterface/spikeinterface/issues/1223.

CodyCBakerPhD:
@h-mayorquin Option 2 is what I would go with (just adding what needs to be included to the testing dependencies) - we generally want the SI and NeuroConv[ecephys] to be intercompatible with each other as much as possible

h-mayorquin:
All right, @CodyCBakerPhD . Fixed, and ready for review. I added the extras require to the spikeinterface dependency.

Now I am thinking though, this might be an overkill for installing ecephys. There are many quality metrics and only some of them require the added dependencies and I guess most users of neuroconv[ecephys] will not want write quality metrics.

Maybe there is a better way but I have no idea. 

@alejoe91 any suggestions?

CodyCBakerPhD:
> I guess most users of neuroconv[ecephys] will not want write quality metrics.

Well, users of `neuroconv[<some ecephys format>]` will want the easiest, fastest, most convenient way to convert `<some ecephys format>`

`neuroconv[ecephys]` is technically the base for that 'just because'

but generic users of `neuroconv[ecephys]` are people working with SI as an API and adding `neuroconv` on top of that at the end to output their data using our `tools.spikeinterface` submodule

Sorry for not being clearer above, I was proposing just modifying the testing requirements, not the base ecephys

requirements-testing.txt
```diff
@@ -2,6 +2,4 @@ pytest
 pytest-cov
 ndx-events>=0.2.0  # for special tests to ensure load_namespaces is set to allow NWBFile load at all timess
 parameterized>=0.8.1
-scikit-learn  # For SI Waveform tests
-numba; python_version <= '3.10'  # For SI Waveform tests
```
CodyCBakerPhD, line 6: I was proposing just moving `spikeinterface[qualitymetrics]>=0.99.1` here if that's possible

src/neuroconv/datainterfaces/ecephys/requirements.txt
```diff
@@ -1,2 +1,2 @@
-spikeinterface>=0.98.2
+spikeinterface[qualitymetrics]>=0.99.1
```
CodyCBakerPhD, line 1:
```suggestion
spikeinterface>=0.99.1
```

And then reducing this back to minimal for basic default usage (then the special quality metric is available for people using SI that know what they're doing, but is optionally imported and used by our tools)

h-mayorquin: Done

h-mayorquin:
> > I guess most users of neuroconv[ecephys] will not want write quality metrics.
> 
> Well, users of `neuroconv[<some ecephys format>]` will want the easiest, fastest, most convenient way to convert `<some ecephys format>`
> 
> `neuroconv[ecephys]` is technically the base for that 'just because'
> 
> but generic users of `neuroconv[ecephys]` are people working with SI as an API and adding `neuroconv` on top of that at the end to output their data using our `tools.spikeinterface` submodule
> 
> Sorry for not being clearer above, I was proposing just modifying the testing requirements, not the base ecephys

Yes, that makes sense. This is a way better solution, I added it to the test as suggested.
This is ready to go.

PR #655, opened 2023-11-25
Make DANDI optional

### LINKED ISSUE #651: Install neuroconv without dandi
Right now dandi is part of our basic dependencies and naturally it comes itself bundled with a bunch of libraries for handling downloading and uploading. 

For the use case of just transforming data within an analysis pipeline none of those dependencies are really needed. 

Maybe it would be worth to add an option to install neuroconv just as a t**ransformation tool** that other libraries can use without the dandi / web / tools-for-uploading-and-downloading data dependencies.

CodyCBakerPhD:
fix #651 

While I was poking around those helper functions, I reorganized distinct flavor of the data transfer tools into the same style as the backend/dataset config

Note that I'm thinking more and more how we might want to switch to using pyproject.toml soon to avoid implicit magic in our import structure

### DIFF
--- a/requirements-minimal.txt
+++ b/requirements-minimal.txt
@@ -6,8 +6,9 @@ h5py>=3.9.0
 hdmf>=3.11.0
 hdmf_zarr>=0.4.0
 pynwb>=2.3.2;python_version>='3.8'
+nwbinspector>=0.4.31
+pydantic>=1.10.13,<2.0.0
 psutil>=5.8.0
 tqdm>=4.60.0
-dandi>=0.57.0
 pandas
 parse>=1.20.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@
 
 ### Improvements
 * Change metadata extraction library from `fparse` to `parse`. [PR #654](https://github.com/catalystneuro/neuroconv/pull/654)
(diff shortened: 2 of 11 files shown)


requirements-minimal.txt
```diff
@@ -6,8 +6,9 @@ h5py>=3.9.0
 hdmf>=3.11.0
 hdmf_zarr>=0.4.0
 pynwb>=2.3.2;python_version>='3.8'
+nwbinspector>=0.4.31
```
h-mayorquin, line 9: So looking at the dependencies of the inspector this still does not solve the problem in #651. The inspector does have the following dependenies that mainly relate to download and uploading data.

https://github.com/NeurodataWithoutBorders/nwbinspector/blob/0e3640048f18838762af98ccbf071b07f0b39e48/requirements.txt#L2-L5

h-mayorquin (approved):
This is fine by me but see the comment about this not addressing the spirit #651.

We can merge that and then discuss how to realize #651 if that is possible. We could remove the inspector as a basic dependency or add an only inspect local files mode to the inspector (I mean, a mode as in an extras require setup.py option).

CodyCBakerPhD:
It's not really a change per-se (no back-compatibility broken), just a slow migration to the `sklearn` standard of managing how methods are exposed from their submodules (example: https://github.com/scikit-learn/scikit-learn/tree/main/sklearn/metrics)

Maybe we should just sit down and do it to get it over with, otherwise I've just been adding it along the way when something needs attention

PR #656, opened 2023-11-25
Try m1 action runner

CodyCBakerPhD:
We shall see...


CodyCBakerPhD:
Confirmed only available on paid organization-level plans (teams or enterprise)

h-mayorquin:
Good that you tested my clam  : ) (I mean it!)

PR #659, opened 2023-11-26
Refactor dataset configuration code as constructors 

h-mayorquin:
As discussed in https://github.com/catalystneuro/neuroconv/pull/569

https://github.com/catalystneuro/neuroconv/pull/569#issuecomment-1821388497

This should keep code that is related together and simplify code navigation.

### DIFF
--- a/src/neuroconv/tools/nwb_helpers/_models/_base_models.py
+++ b/src/neuroconv/tools/nwb_helpers/_models/_base_models.py
@@ -6,8 +6,50 @@ from typing import Any, Dict, Literal, Tuple, Type, Union
 import h5py
 import numcodecs
 import numpy as np
+import zarr
+from hdmf import Container
 from hdmf.container import DataIO
+from hdmf.data_utils import DataChunkIterator, DataIO, GenericDataChunkIterator
+from hdmf.utils import get_data_shape
 from pydantic import BaseModel, Field, root_validator
+from pynwb import NWBHDF5IO, NWBFile
+
+from ...hdmf import SliceableDataChunkIterator
+
+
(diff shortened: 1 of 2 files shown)


src/neuroconv/tools/nwb_helpers/_models/_base_models.py
```diff
@@ -61,6 +103,22 @@ def __init__(self, **values):
         values.update(dataset_name=dataset_name)
         super().__init__(**values)
 
+    @classmethod
+    def from_neurodata_object(cls, neurodata_object: Container, field_name: str) -> "DatasetInfo":
```
CodyCBakerPhD, line 107:
```suggestion
    def from_neurodata_object(cls, neurodata_object: Container, field_name: str) -> Self:
```

Shouldn't this use [`typing.Self`](https://peps.python.org/pep-0673/#use-in-classmethod-signatures)?

CodyCBakerPhD:
I think this pattern does indeed look much better, and is more standard/consistent with other class constructors in our ecosystem too

src/neuroconv/tools/nwb_helpers/_models/_base_models.py
h-mayorquin: Does that work? I thought that it was introduced in python 3.11 as the PEP states.

CodyCBakerPhD: Right you are; will raise this as an issue to not forget then

CodyCBakerPhD (approved):

CodyCBakerPhD:
Interesting - automerge did not trigger when a job was manually cancelled (having some kind of intermittent stall problem with mac runners...) even after all conditions passed

PR #663, opened 2023-11-27
Fix for latest DANDI release

CodyCBakerPhD:
Latest DANDI release (1 hour ago) broke tests due to an altered import path

This should fix it

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ with open(root / "requirements-testing.txt") as f:
     testing_suite_dependencies = f.readlines()
 
 extras_require = defaultdict(list)
-extras_require["dandi"].append("dandi>=0.58.0")
+extras_require["dandi"].append("dandi>=0.58.1")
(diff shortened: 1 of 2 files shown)


h-mayorquin (approved):

setup.py
```diff
@@ -18,7 +18,7 @@
     testing_suite_dependencies = f.readlines()
 
 extras_require = defaultdict(list)
-extras_require["dandi"].append("dandi>=0.58.0")
+extras_require["dandi"].append("dandi>=0.58.1")
```
CodyCBakerPhD, line 21:
```suggestion
extras_require["dandi"].append("dandi<0.58.1")  # TODO: fix >= 0.58.1
```

CodyCBakerPhD:
> Pin to be less than maybe?

Yeah, just to get the CI passing again to unblock other things

It looks like more changed than just that, investigation into why taking a little while

h-mayorquin (approved):

PR #664, opened 2023-11-27
Fix for latest DANDI release

CodyCBakerPhD:
Got the test working locally, not quite sure what CI is going on about yet


src/neuroconv/tools/yaml_conversion_specification/_yaml_conversion_specification.py
```diff
@@ -125,18 +125,27 @@ def run_conversion_from_yaml(
...
+            dandi_metadata = _get_pynwb_metadata(path=nwbfile_path_to_set)
+            if dandi_metadata["subject_id"] is not None:
+                dandi_metadata["subject_id"] = dandi_metadata["subject_id"].replace(" ", "_")
+            dandi_metadata.update(path=nwbfile_path_to_set)
             dandi_metadata_list.append(dandi_metadata)
-        named_dandi_metadata_list = create_unique_filenames_from_metadata(metadata=dandi_metadata_list)
+        dandi_metadata_with_set_paths = create_unique_filenames_from_metadata(metadata=dandi_metadata_list)
 
-        for named_dandi_metadata in named_dandi_metadata_list:
-            if "temp_nwbfile_name_" in named_dandi_metadata["path"].stem:
```
CodyCBakerPhD, line 137: Instead of doing filtering of paths that need to be updated here, doing it ahead of time at 128 (reduces cyclomatic complexity too)

CodyCBakerPhD:
OK there were a couple subtle changes to these functions since the last DANDI release; I think it's somewhat related to their Pydantic V2 migration, but I'm guessing they just didn't think it would alter any significant back-compatibility for downstream libraries (to be fair, we are probably the only ones that rely on subtle changes this sensitively)

- spaces in subject IDs are now changed to dashes automatically (this is great, 2 less lines for us to override)
- session IDs are not automatically used unless at least one pair of files in the passed metadata share the same subject (this is an annoyance I've brought up to them many times, but they won't budge on;  which is why our `automatic_dandi_upload` helper function forces it always)

As for why this caught us by surprise; as I recall in the past attempts to add DANDI CLI to our dev testing suite were not met with success because they require a very different setup in tests (I believe their suite relies on mixture of `tox` and `docker-compose`) and a simple `pip install -e` on a clone of the repo is not sufficient - I will however try again and see if anything has changed

bendichter (approved):

PR #665, opened 2023-11-27
[Testing] Add DANDI to dev tests

CodyCBakerPhD:
From https://github.com/catalystneuro/neuroconv/pull/664#issuecomment-1828409302


CodyCBakerPhD:
Well I'll be... seems to be working just fine

This should help us catch problems with DANDI a little bit earlier (hopefully)

.github/workflows/deploy-tests.yml
```diff
@@ -47,7 +47,7 @@ jobs:
   run-dev-tests:
     needs: assess-file-changes
     if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
-    uses: catalystneuro/neuroconv/.github/workflows/dev-testing.yml@main
+    uses: catalystneuro/neuroconv/.github/workflows/dev-testing.yml@try_to_add_dandi_to_dev
```
h-mayorquin, line 50: I checked and the test is there. let's revert this to main before we merge.
```suggestion
    uses: catalystneuro/neuroconv/.github/workflows/dev-testing.yml@main
```

.github/workflows/dev-testing.yml
```diff
@@ -48,6 +48,8 @@ jobs:
 
       - name: Dev gallery - ROIExtractors
         run: pip install git+https://github.com/CatalystNeuro/roiextractors@main
+      - name: Dev gallery - DANDI
+        run: pip install git+https://github.com/dandi/dandi-cli@master
```
h-mayorquin, line 52: Looking at the print out that you added this seems to be working. But you could specify the -no-cache-dir version to avoid nasty surprises.

```suggestion
  run: pip install --no-cache-dir git+https://github.com/dandi/dandi-cli@master
```

h-mayorquin (changes requested):

.github/workflows/dev-testing.yml
CodyCBakerPhD: added to all

h-mayorquin (approved):

PR #666, opened 2023-11-27
Fix api docs for YAML

CodyCBakerPhD:
Detected following merge of #655


docs/api/tools.yaml_to_nwb_conversion.rst
```diff
@@ -1,3 +1,3 @@
 YAML conversion specification
 -----------------------------
-.. automodule:: neuroconv.tools.yaml_conversion_specification.yaml_conversion_specification
+.. automodule:: neuroconv.tools.yaml_conversion_specification
```
CodyCBakerPhD, line 3: Apparently we can easily keep the API docs like this

src/neuroconv/tools/yaml_conversion_specification/__init__.py
```diff
@@ -1 +1,3 @@
 from ._yaml_conversion_specification import run_conversion_from_yaml
+
+__all__ = ["run_conversion_from_yaml"]
```
CodyCBakerPhD, line 3: All we have to do is specify the `__all__` (again, like sklearn does)

h-mayorquin: I thought that this only controls what `import * `

Does, can you expand on what this is solving?

h-mayorquin (approved):
OK,

src/neuroconv/tools/yaml_conversion_specification/__init__.py
CodyCBakerPhD: Automodule fails without pointing at the private submodule

Also not very nice to point to the private submodule since that makes it seem like a user ought to import from that private submodule

PR #667, opened 2023-11-28
Change `ImagingExtractorDataChunkIterator` `chunk_shape` to be a multiple of the image size and default `chunk_mb` to 10.0

weiglszonja:
Proposing to add a private method `_get_scaled_chunk_shape` to select `chunk_shape` less than the chunk_mb threshold while also making sure it is a multiple of the image size. Also changing default `chunk_mb` to 10.0.
@CodyCBakerPhD Let me know if these changes make sense to include here, or updating `chunk_shape` should be done separately for each conversion.

### DIFF
--- a/src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
+++ b/src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
@@ -62,12 +62,12 @@ class ImagingExtractorDataChunkIterator(GenericDataChunkIterator):
             assert chunk_mb * 1e6 <= buffer_gb * 1e9, "chunk_mb must be less than or equal to buffer_gb!"
 
         if chunk_mb is None and chunk_shape is None:
-            chunk_mb = 1.0
+            chunk_mb = 10.0
 
         self._maxshape = self._get_maxshape()
         self._dtype = self._get_dtype()
         if chunk_shape is None:
-            chunk_shape = super()._get_default_chunk_shape(chunk_mb=chunk_mb)
+            chunk_shape = self._get_scaled_chunk_shape(chunk_mb=chunk_mb)
 
         if buffer_gb is None and buffer_shape is None:
             buffer_gb = 1.0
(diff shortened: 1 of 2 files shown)


src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
```diff
@@ -82,6 +82,17 @@ def __init__(
             progress_bar_options=progress_bar_options,
         )
 
+    def _get_scaled_chunk_shape(self, chunk_mb: float) -> tuple:
```
CodyCBakerPhD, line 85: This already exists actually; a recent addition to HDMF that allows the strategy to be overridden

I would mimic the approach of the SI iterator here: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/spikeinterface/spikeinterfacerecordingdatachunkiterator.py#L74C18-L74C18

src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
```diff
@@ -82,6 +82,17 @@ def __init__(
...
+        """Select the chunk_shape less than the threshold of chunk_mb that is also a multiple of the image size."""
+        assert chunk_mb > 0, f"chunk_mb ({chunk_mb}) must be greater than zero!"
+
+        image_shape = self._maxshape[1:]
+        frame_size_bytes = np.prod(image_shape) * self._dtype.itemsize
```
CodyCBakerPhD, line 90: Fun fact from a nasty bug in HDMF a while back: always use `math.prod` instead of `np.prod`

Reason: buffer overflow

src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
weiglszonja: ah thank you, didn't know about that.

CodyCBakerPhD (approved):

src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
```diff
@@ -82,6 +82,18 @@ def __init__(
             progress_bar_options=progress_bar_options,
         )
 
+    def _get_default_chunk_shape(self, chunk_mb: float) -> tuple:
+        """Select the chunk_shape less than the threshold of chunk_mb while keeping the original image size."""
```
weiglszonja, line 86: @CodyCBakerPhD do you think this is clear enough? I want to make sure it's clear that we are keeping the image shape and fitting how many frames fit into `chunk_mb`?

CodyCBakerPhD: Yes, I think it makes perfect sense

CodyCBakerPhD: Maybe one day we will have to consider how to deal with super higher resolution images... but for now it always makes sense (both optimizing on disk and in cloud) to chunk by entire frame

PR #668, opened 2023-11-28
[Testing] Trying to stabilize mac runners

CodyCBakerPhD:
Since relenting DANDI to be optional, there is a difference in secondary dependency resolution that I suspect is causing some intermittent stall outs in the Mac runners

Testing to see if this stabilizes things in any way

That is, there a difference between installing `neuroconv[full]` in a fresh environment compared to `neuroconv; neuroconv[full]` incrementally (pip resolves dependencies in different ways when all requirements are presented at once rather than one at a time)

### DIFF
--- a/.github/workflows/deploy-tests.yml
+++ b/.github/workflows/deploy-tests.yml
@@ -30,7 +30,7 @@ jobs:
   run-tests:
     needs: assess-file-changes
     if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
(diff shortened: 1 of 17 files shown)


CodyCBakerPhD:
Example of sporadic error: https://github.com/catalystneuro/neuroconv/actions/runs/7025070771/job/19115202070?pr=668

CodyCBakerPhD:
From that, I suspect it might just be the latest release of `tables` on 11/27/2023 (when I started seeing this)

I'll try downpinning

CodyCBakerPhD:
Re-triggered just to be sure, but seems to have a positive effect on the runners

PR #669, opened 2023-11-29
Parse param.annotation as type in case of `__future__.annotations`

alejoe91:
Small fix for a very hard-to-debug problem.

If someone has a `from __future__ import annotations` in a package and implements a `DataInterface` in it, the `inspect.signature(...).parameters` return annotations as `str` instead of actual `type`.

## To reproduce

Instantiate a simple interface that does nothing in an `interface.py`:

```
from neuroconv import BaseDataInterface
from neuroconv.utils import FolderPathType

class MockInterface(BaseDataInterface):
    def __init__(self, folder_path: FolderPathType):
        self.folder_path = folder_path

    def add_to_nwbfile(
        self,
        nwbfile,
        metadata,
    ):
        pass
```

Then run:
```
from interface import MockInterface

MockInterface.get_source_schema() 
# works fine
```

Now add `from __future__ import annotations` at the top of the `interface.py` and run in a new session:
```
from interface import MockInterface

MockInterface.get_source_schema() 
# fails
```
The latter fails with this error without this PR:
```
AttributeError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 MockInterface.get_source_schema()

File ~/Documents/codes/catalystneuro/neuroconv/src/neuroconv/basedatainterface.py:26, in BaseDataInterface.get_source_schema(cls)
     23 @classmethod
     24 def get_source_schema(cls):
     25     """Infer the JSON schema for the source_data from the method signature (annotation typing)."""
---> 26     return get_schema_from_method_signature(cls, exclude=["source_data"])

File ~/Documents/codes/catalystneuro/neuroconv/src/neuroconv/utils/json_schema.py:134, in get_schema_from_method_signature(method, exclude)
    130 arg = param.annotation
    131 # in case __future__.annotations is imported, the annotation is a string
    132 # if isinstance(arg, str):
    133 #     arg = type(arg)
--> 134 if arg.__name__ in annotation_json_type_map:
    135     args_spec[param_name]["type"] = annotation_json_type_map[arg.__name__]
    136 else:

AttributeError: 'str' object has no attribute '__name__'
```

### DIFF
--- a/src/neuroconv/utils/json_schema.py
+++ b/src/neuroconv/utils/json_schema.py
@@ -128,6 +128,9 @@ def get_schema_from_method_signature(method: Callable, exclude: list = None) -> 
                     input_schema["properties"].update({param_name: dict(format="directory")})
             else:
                 arg = param.annotation
+                # in case __future__.annotations is imported, the annotation is a string
+                if isinstance(arg, str):
+                    arg = type(arg)
                 if arg.__name__ in annotation_json_type_map:
                     args_spec[param_name]["type"] = annotation_json_type_map[arg.__name__]
                 else:
--- a/tests/test_minimal/test_utils/test_json_schema_utils.py
+++ b/tests/test_minimal/test_utils/test_json_schema_utils.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import json
 import os
 from copy import deepcopy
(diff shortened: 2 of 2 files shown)


h-mayorquin:
This is the same error as described in:
https://github.com/catalystneuro/neuroconv/pull/613

You should try adding lines there to see if this indeed solves the problem.

alejoe91:
Sorry I didn't see that open PR!

h-mayorquin:
Ah no worries @alejoe91 . I just open the PR to point the problem out to @CodyCBakerPhD . I did not have time to push a solution, this is great!

h-mayorquin:
If this fixes the problem that's great!

I was telling Cody that in python 3.10 there is a function that automatically does this for us:
`https://docs.python.org/3/library/inspect.html#inspect.get_annotations`

And there is some functions that could be used. See this section "In Python 3.9 and older" here:
:
https://dev.to/derlin/python-type-hints-and-future-annotations-4kc5

But we can fix a problem and then build a more general patch later if the needs to come to be : )

h-mayorquin:
Let's tag @samuelgarcia as he is very excited about typing and I like to keep him on the loop for the latest developments on the field.

CodyCBakerPhD:
Well, apparently all the new types under this behavior become `str` and so everything gets mapped to `string` in the JSON type map...

https://github.com/catalystneuro/neuroconv/actions/runs/7048911302/job/19186337897?pr=669

CodyCBakerPhD:
See conversation on #671 for how this doesn't entirely resolve the issue

PR #670, opened 2023-11-30
stub options

luiztauffer:
small modification which allows for changing the sample size of the sub recording conversion


CodyCBakerPhD:
Can you describe what you need this for?

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -300,6 +303,7 @@ def add_to_nwbfile(
         nwbfile: NWBFile,
         metadata: Optional[dict] = None,
         stub_test: bool = False,
+        stub_test_size: int = 100,
```
CodyCBakerPhD, line 306: I would also recommend calling this `stub_frames` or anything else to indicate it is not stubbing channels along the way (I know it's in the docstring, but if it's in the variable name that's even clearer)

CodyCBakerPhD:
> hey @CodyCBakerPhD these are some changes I made locally to be able to create a converted file with a meaningful subset of the spikeglx recordings, I thought it could be useful in general

Were you attempting to make the preview files longer or shorter than the previous default stubbing?

Not against adding control to this in general, would just prefer it be done in a smooth way

In general the hope was actually for stubbing to be a part of the general backend configuration mechanism but that particular aspect proves a little tricky ATM so a bit delayed on that front

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
h-mayorquin: We use this already, don't we?

https://github.com/catalystneuro/neuroconv/blob/74689fdfce22ad70d01e273dc5830354820b3f4e/src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py#L109-L110

PR #671, opened 2023-11-30
[Test] Test Alessio's fix to schema generation

CodyCBakerPhD:


src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
```diff
@@ -23,7 +23,7 @@ class VideoInterface(BaseDataInterface):
     """Data interface for writing videos as ImageSeries."""
 
     def __init__(
-        self, file_paths: list[str | Path], verbose: bool = False
+        self, file_paths: list, verbose: bool = False
```
CodyCBakerPhD, line 26: Supporting the ability for people to call `from __future__ import annotations` in the library is one thing; adding full support for schema inference from data interface signatures of the updated kind is another, and would in preferance be done with Pydantic

Removing for now

src/neuroconv/utils/json_schema.py
```diff
@@ -130,6 +130,9 @@ def get_schema_from_method_signature(method: Callable, exclude: list = None) ->
                     input_schema["properties"].update({param_name: dict(format="directory")})
             else:
                 arg = param.annotation
+                # in case __future__.annotations is imported, the annotation is a string
+                if isinstance(arg, str):
+                    arg = type(arg)
```
CodyCBakerPhD, line 135: Well, apparently all the new types under this behavior become `str` and so everything gets mapped to `string` in the JSON type map...

CodyCBakerPhD: Let's try it!

CodyCBakerPhD: Despite the input argument `eval_str` saying it handles exactly this case, it doesn't look like it actually does what is says anywhere, and ends up at https://github.com/shawwn/get-annotations/blob/main/get_annotations/__init__.py#L86 regardless

src/neuroconv/utils/json_schema.py
```diff
@@ -130,6 +131,9 @@ def get_schema_from_method_signature(method: Callable, exclude: list = None) ->
                     input_schema["properties"].update({param_name: dict(format="directory")})
             else:
                 arg = param.annotation
+                # in case __future__.annotations is imported, the annotation is a string
+                if isinstance(arg, str):
+                    arg = get_annotations(arg, eval_str=True)
```
CodyCBakerPhD, line 136:
```suggestion
                    arg = eval(arg)
```

CodyCBakerPhD: Trying this instead

CodyCBakerPhD:
Yeah IDK my efforts here can resolve the issues with common source data patterns, but the same logic is applied to conversion options and there are some cases that cause more and other issues after that

I think we might just want to fix this by offloading all the effort required to Pydantic as in re-implementing #264

CodyCBakerPhD:
replaced by #1016

PR #676, opened 2023-12-04
Add check for series regularity when passing timestamps in `VideoInterface`

h-mayorquin:
I am using the method `set_aligned_timestamps` in the VideoInterface but then there is no check for series regularity for this case:

https://github.com/catalystneuro/neuroconv/blob/5687c07ca8517a3d258309c077c7378b43938215/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py#L340-L341

I just added that and removing one inlined definition that was getting in the way of my debugging.

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
+++ b/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
@@ -14,7 +14,11 @@ from .video_utils import VideoCaptureContext
 from ....basedatainterface import BaseDataInterface
 from ....tools import get_package
 from ....tools.nwb_helpers import get_module
-from ....utils import get_base_schema, get_schema_from_hdmf_class
+from ....utils import (
+    calculate_regular_series_rate,
+    get_base_schema,
+    get_schema_from_hdmf_class,
+)
 
 
 class VideoInterface(BaseDataInterface):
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
The intention is that if you are bothering to explicitly set `timestamps` in this way, you already know they are irregular

Real question is, if you're fetching and setting a vector of timestamps in this case, why do you want to check if they are regular after the fact?

CodyCBakerPhD:
> Outside of that, we do the checks everywhere to never write regular timestamps.

Specific to video data, this does not hold true

Both DLC and SLEAP appear to use timestamps from their original videos: https://github.com/talmolab/sleap/blob/2d24296c2149402c7ee07601e4d0272fdd308441/sleap/io/format/ndx_pose.py#L247 + https://github.com/DeepLabCut/DLC2NWB/blob/main/dlc2nwb/utils.py#L216C30-L216C30

And see this PR from an external user exposing that control on our side: https://github.com/catalystneuro/neuroconv/pull/532


> People might get the correct timestamps from somewhere else and they set them but they don't know they are regular.

My point is that if a user did that, they might be confused why their output NWB file did not contain any timestamps (and was magically transformed to a `rate` + `starting_time`)

Hence why I've been pushing for the following structure...

-> If user specifies nothing, let format automatically determine. Could be irregular timestamps, a constant rate, or a constant rate inferred from regular timestamps (but point is user is unaware of any of it)

-> If user specifies timestamps via the alignment methods, then use the timestamps they give (maybe we could check for regularity and warn?)

-> If user wants to override `rate` or `starting_time`, they should be able to (though exposing all of this in a globally consistent way across NeuroConv is still WIP; some `starting_time` values are still considered `conversion_options`, some rates are required in the `__init__` signature...)


> No cost to do it here as well.

Minor RAM/CPU cost for very long videos. If user knows what they want to do ahead of time, no sense wasting the operation

PR #677, opened 2023-12-05
Add depth to ROI iterator

CodyCBakerPhD:
Detected during review of Clandinin example files; the chunks are way too large since they include all depths (which have a huge number of planes)

Needs tests

### DIFF
--- a/src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
+++ b/src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
@@ -89,19 +89,14 @@ class ImagingExtractorDataChunkIterator(GenericDataChunkIterator):
         num_frames = self._maxshape[0]
         width = self._maxshape[1]
         height = self._maxshape[2]
-        image_shape = self._maxshape[1:]
 
-        depth = None
-        if len(self._maxshape) == 4:
-            depth = self._maxshape[3]
+        frame_size_bytes = width * height * self._dtype.itemsize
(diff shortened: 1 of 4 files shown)


src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
```diff
@@ -87,11 +87,18 @@ def _get_default_chunk_shape(self, chunk_mb: float) -> tuple:
...
+        width = self._maxshape[1]
+        height = self._maxshape[2]
+
+        frame_size_bytes = width * height * self._dtype.itemsize
+        chunk_size_bytes = chunk_mb * 1e6
+        num_frames_per_chunk = int(chunk_size_bytes / frame_size_bytes)
+
+        if len(self._maxshape) == 3:
+            chunk_shape = (max(min(num_frames_per_chunk, num_frames), 1), width, height)
+        elif len(self._maxshape) == 4::
```
CodyCBakerPhD, line 99:
```suggestion
        elif len(self._maxshape) == 4:
```

h-mayorquin (approved):
This looks good to me. 
There are two concerns I have but they are minor and kind of outside of the scope of this PR:
* I think backend is not enough as a word to describe `configure_backend` (which backend?)
* It seems (maybe I am wrong) that there is no coverage for the filter compression options with hdf5.

h-mayorquin:
Oh, auto-merge.

PR #678, opened 2023-12-05
Add compression to `FicTracDataInterface`

h-mayorquin:
As discussed in today's meeting.

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
+++ b/src/neuroconv/datainterfaces/behavior/fictrac/fictracdatainterface.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Optional, Union
 
 import numpy as np
+from hdmf.backends.hdf5.h5_utils import H5DataIO
 from pynwb.behavior import Position, SpatialSeries
 from pynwb.file import NWBFile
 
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
I agree that it is mildly annoying to add this only for it to be immediately deprecated after the backend refactor (only 2 PRs away I think now...)

But deadlines are deadlines and this data really should be compressed

h-mayorquin:
Not a lot of work but is fine. To be honest I did not concerned mysefl because it is 30 MB of data. Do we have any threshold for when compression makes sense?

CodyCBakerPhD:
> Do we have any threshold for when compression makes sense?

The word from Satra and Yarik on the archive is that every byte that can be compressed, should be compressed.

If anyone was actually paying $$$ for the bucket it would be an even bigger deal (hard enforced through validation), because why would you pay for storage you don't need [but this is then an indirect implication with the AWS Open Data team because someone is paying for it somehow; a hidden opportunity cost perhaps]

Also bandwidth cost. When the data is ready hundreds, thousands of times over the years by users, each unnecessary byte that gets sent is wasted time and theoretically also transfer cost if using cloud resources

So yeah, even 30 MB of data, if it gets streamed 1,000+ times, adds up. So if there was something that could have been done to easily reduce that amount, we should try

h-mayorquin:
Thanks for explaining the rationale, makes sense.

PR #679, opened 2023-12-06
Fix path expander II

### LINKED ISSUE #674: `LocalPathExpander` should match only `folder_paths` or `file_paths` depending on the spec.

h-mayorquin:
After #675 @bendichter [mentioned ](https://github.com/catalystneuro/neuroconv/pull/675#issuecomment-1840847452) that we are still missing the case of nested over matching:

> "a_simple_pattern_{session_id}" applied to "tmp/base_directory/a_simple_pattern_2/nested/" will still return session_id="2/nested/" when it should not match

and proposed a solution of excluding metadata entries that have file separators (slashes in posix or double front slashes on windows) from the matches. This PR implements this solution and amends the tests to reflect that.

This should fix #674

### DIFF
--- a/tests/test_minimal/test_tools/test_expand_paths.py
+++ b/tests/test_minimal/test_tools/test_expand_paths.py
@@ -13,7 +13,7 @@ from neuroconv.utils import NWBMetaDataEncoder
 def test_only_folder_match(tmpdir):
     base_directory = Path(tmpdir)
 
-    sub_directory1 = base_directory /  "subject1" / "a_simple_pattern_1"
+    sub_directory1 = base_directory / "subject1" / "a_simple_pattern_1"
     sub_directory2 = base_directory / "subject2" / "a_simple_pattern_2"
 
     sub_directory1.mkdir(exist_ok=True, parents=True)
@@ -41,13 +41,12 @@ def test_only_folder_match(tmpdir):
         }
     }
 
-    # Instantiate LocalPathExpander
-
     path_expander = LocalPathExpander()
     metadata_list = path_expander.expand_paths(source_data_spec)
     folder_paths = [metadata_match["source_data"]["a_source"]["folder_path"] for metadata_match in metadata_list]
 
-    expected = {str(sub_directory1), str(sub_directory2), str(sub_directory3)}
+    # Note that sub_directory3 is not included because it does not conform to the pattern
+    expected = {str(sub_directory1), str(sub_directory2)}
 
     assert set(folder_paths) == expected
 
@@ -55,11 +54,11 @@ def test_only_folder_match(tmpdir):
 def test_only_file_match(tmpdir):
     base_directory = Path(tmpdir)
 
-    sub_directory1 = base_directory / "a_simple_pattern_1"
-    sub_directory2 = base_directory / "a_simple_pattern_2"
+    sub_directory1 = base_directory / "subject1" / "a_simple_pattern_1"
+    sub_directory2 = base_directory / "subject2" / "a_simple_pattern_2"
 
-    sub_directory1.mkdir(exist_ok=True)
-    sub_directory2.mkdir(exist_ok=True)
+    sub_directory1.mkdir(exist_ok=True, parents=True)
+    sub_directory2.mkdir(exist_ok=True, parents=True)
 
     # Add files with the same name to both folders
     file1 = sub_directory1 / "a_simple_pattern_1.bin"
@@ -70,8 +69,8 @@ def test_only_file_match(tmpdir):
     file2.touch()
 
     # Add another sub-nested folder with a folder
-    sub_directory3 = sub_directory1 / "a_simple_pattern_3"
-    sub_directory3.mkdir(exist_ok=True)
+    sub_directory3 = sub_directory1 / "intermediate_nested" / "a_simple_pattern_3"
+    sub_directory3.mkdir(exist_ok=True, parents=True)
     file3 = sub_directory3 / "a_simple_pattern_3.bin"
     file3.touch()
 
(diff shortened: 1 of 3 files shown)


bendichter:
@h-mayorquin thanks for working on this. Let me see if it's possible with regex

bendichter:
oh, that's not easy with with parse library. hmm.

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -13,11 +13,11 @@
...
     base_directory = Path(tmpdir)
 
-    sub_directory1 = base_directory / "a_simple_pattern_1"
-    sub_directory2 = base_directory / "a_simple_pattern_2"
+    sub_directory1 = base_directory / "subject1" / "a_simple_pattern_1"
+    sub_directory2 = base_directory / "subject2" / "a_simple_pattern_2"
 
-    sub_directory1.mkdir(exist_ok=True)
-    sub_directory2.mkdir(exist_ok=True)
+    sub_directory1.mkdir(exist_ok=True, parents=True)
```
bendichter, line 19: oh cool I didn't know about the parents arg

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -28,38 +28,37 @@ def test_only_folder_match(tmpdir):
...
 
-    # Instantiate LocalPathExpander
-
     path_expander = LocalPathExpander()
     metadata_list = path_expander.expand_paths(source_data_spec)
     folder_paths = [metadata_match["source_data"]["a_source"]["folder_path"] for metadata_match in metadata_list]
 
-    expected = {str(sub_directory1), str(sub_directory2), str(sub_directory3)}
+    # Note that sub_directory3 is not included because it does not conform to the pattern
+    expected = {str(sub_directory1), str(sub_directory2)}
```
bendichter, line 49: can you also test that the metadata was extracted correctly?

h-mayorquin: Done. I also added a docstring to `extract_metadata`

bendichter:
@h-mayorquin what do you think of this? I factored out the file creation logic so it can be used my multiple tests and I added a test case to ensure that the parser is not combining multiple nested directories into a single metadata field

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -94,10 +105,13 @@ def test_only_file_match(tmpdir):
 
     path_expander = LocalPathExpander()
     matches_list = path_expander.expand_paths(source_data_spec)
-    file_paths = [match["source_data"]["a_source"]["file_path"] for match in matches_list]
+    file_paths = set(match["source_data"]["a_source"]["file_path"] for match in matches_list)
```
h-mayorquin, line 108: Minor point but the `file_paths` are transformed to set before the test.

h-mayorquin (commented):
@bendichter This is OK to me. The code duplication across the two tests did bother me and I was thinking to use a module scoped pytest fixture to get rid of it. I was on the edge though because of the trade-off between your test being very explicit which is good and some code duplication which is bad. 

But if you think this makes more sense and I was already on the edge it seems tha the average favors a bit of abstraction. 

If you are Ok with this we should mege!

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -2,96 +2,137 @@
...
+    matches_list = path_expander.expand_paths(source_data_spec)
+    file_paths = set(match["source_data"]["a_source"]["file_path"] for match in matches_list)
 
-    expected = {str(file1), str(file2), str(file3)}
+    # Note that file3 is not included because it does not conform to the pattern
+    expected = {
+        str(base_directory / "subject1" / "a_simple_pattern_1" / "a_simple_pattern_1.bin"),
+        str(base_directory / "subject2" / "a_simple_pattern_2" / "a_simple_pattern_2.bin"),
+    }
     assert set(file_paths) == expected
```
bendichter, line 115:
```suggestion
    assert file_paths == expected
```

bendichter (approved):

PR #680, opened 2023-12-06
[Feature] Create template fstring path from real path

bendichter:
address https://github.com/NeurodataWithoutBorders/nwb-guide/issues/528

### DIFF
--- a/src/neuroconv/tools/path_expansion.py
+++ b/src/neuroconv/tools/path_expansion.py
@@ -159,13 +159,13 @@ def construct_path_template(path: str, *, subject_id: str, session_id: str, **me
     if subject_id not in path or session_id not in path:
         raise ValueError("Subject ID and Session ID must be present in the path")
 
-    path = path.replace(subject_id, "{subject_id}").replace(session_id, "{session_id}")
+    format_string = path.replace(subject_id, "{subject_id}").replace(session_id, "{session_id}")
 
     for key, val in metadata_kwargs.items():
         if val == "":
             raise ValueError(f"Value for '{key}' cannot be an empty string")
-        if val not in path:
+        if val not in format_string:
             raise ValueError(f"Value for '{key}' not found in the path")
-        path = path.replace(val, f"{{{key}}}")
+        format_string = format_string.replace(val, f"{{{key}}}")
 
-    return path
+    return format_string
--- a/src/neuroconv/tools/path_expansion.py
+++ b/src/neuroconv/tools/path_expansion.py
@@ -159,7 +159,7 @@ def construct_path_template(path: str, *, subject_id: str, session_id: str, **me
     if subject_id not in path or session_id not in path:
         raise ValueError("Subject ID and Session ID must be present in the path")
 
-    path = path.replace(subject_id, '{subject_id}').replace(session_id, '{session_id}')
+    path = path.replace(subject_id, "{subject_id}").replace(session_id, "{session_id}")
 
     for key, val in metadata_kwargs.items():
         if val == "":
--- a/src/neuroconv/tools/path_expansion.py
+++ b/src/neuroconv/tools/path_expansion.py
@@ -103,3 +103,69 @@ class LocalPathExpander(AbstractPathExpander):
         base_directory = Path(base_directory)
         assert base_directory.is_dir(), f"The specified 'base_directory' ({base_directory}) is not a directory!"
         return (str(path.relative_to(base_directory)) for path in base_directory.rglob("*"))
+
+
+def construct_path_template(path: str, *, subject_id: str, session_id: str, **metadata_kwargs) -> str:
+    """
+    Construct a path template by replacing specific parts of a given path with placeholders.
+
+    This function takes a real path example and replaces the occurrences of subject ID, session ID,
+    and any additional metadata values with their respective placeholders.
+
+    Parameters
+    ----------
+    path : str
+        The path string containing actual data that needs to be templated.
+    subject_id : str
+        The subject ID in the path that will be replaced with the '{subject_id}' placeholder.
+    session_id : str
+        The session ID in the path that will be replaced with the '{session_id}' placeholder.
+    **metadata_kwargs : dict
+        Additional key-value pairs where the key is the placeholder name and the value is the actual data
+        in the path that should be replaced by the placeholder.
+
+    Returns
+    -------
+    str
+        The path string with specified parts replaced by placeholders.
+
+    Raises
+    ------
+    ValueError
+        If `subject_id`, `session_id`, or any value in `metadata_kwargs` is an empty string, or if `subject_id`
+        or `session_id` placeholders are not found in the path.
+
+    Examples
+    --------
+    >>> construct_path_template(
+    >>>     "/data/subject456/session123/file.txt",
+    >>>     subject_id="subject456",
+    >>>     session_id="session123"
+    >>> )
+    '/data/{subject_id}/{session_id}/file.txt'
+
+    >>> construct_path_template(
+    >>>     "/data/subject789/session456/image.txt",
+    >>>     subject_id="subject789",
+    >>>     session_id="session456",
+    >>>     file_type="image"
+    >>> )
+    '/data/{subject_id}/{session_id}/{file_type}.txt'
+    """
+
+    if subject_id == "" or session_id == "":
+        raise ValueError("Subject ID and Session ID cannot be empty strings")
+
+    if subject_id not in path or session_id not in path:
+        raise ValueError("Subject ID and Session ID must be present in the path")
+
+    path = path.replace(subject_id, '{subject_id}').replace(session_id, '{session_id}')
+
+    for key, val in metadata_kwargs.items():
+        if val == "":
+            raise ValueError(f"Value for '{key}' cannot be an empty string")
+        if val not in path:
+            raise ValueError(f"Value for '{key}' not found in the path")
+        path = path.replace(val, f"{{{key}}}")
+
+    return path
--- a/tests/test_minimal/test_tools/test_expand_paths.py
+++ b/tests/test_minimal/test_tools/test_expand_paths.py
@@ -383,37 +383,33 @@ def test_expand_paths_ibl(tmpdir):
     tc = unittest.TestCase()
     tc.assertCountEqual(path_expansion_results, expected)
 
+
 # contruct_path_template tests
 
 
 test_cases = [
     dict(
         test_input=dict(
-            path = "/data/subject456/session123/file.txt",
-            metadata = dict(subject_id="subject456", session_id="session123")
+            path="/data/subject456/session123/file.txt", metadata=dict(subject_id="subject456", session_id="session123")
         ),
-        expected = "/data/{subject_id}/{session_id}/file.txt"
+        expected="/data/{subject_id}/{session_id}/file.txt",
     ),
     dict(
         test_input=dict(
             path="/data/subject789/session456/image.txt",
-            metadata=dict(subject_id="subject789", session_id="session456", file_type="image")
+            metadata=dict(subject_id="subject789", session_id="session456", file_type="image"),
         ),
-        expected="/data/{subject_id}/{session_id}/{file_type}.txt"
+        expected="/data/{subject_id}/{session_id}/{file_type}.txt",
     ),
     dict(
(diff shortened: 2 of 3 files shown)


bendichter:
@CodyCBakerPhD thoughts on this?

src/neuroconv/tools/path_expansion.py
```diff
@@ -103,3 +103,69 @@ def list_directory(self, base_directory: DirectoryPath) -> Iterable[FilePath]:
...
+    path = path.replace(subject_id, "{subject_id}").replace(session_id, "{session_id}")
+
+    for key, val in metadata_kwargs.items():
+        if val == "":
+            raise ValueError(f"Value for '{key}' cannot be an empty string")
+        if val not in path:
+            raise ValueError(f"Value for '{key}' not found in the path")
+        path = path.replace(val, f"{{{key}}}")
+
+    return path
```
CodyCBakerPhD, line 171: It would likely make debugging/logging easier going forward to name the output of the method chaining to something else so the original `path` input stays in the namespace for internal side-by-side comparison (granted you can always compare it to the input argument it received, but that's a namespace above, just mildly inconvenient)

```suggestion
    format_string = path.replace(subject_id, "{subject_id}").replace(session_id, "{session_id}")

    for key, val in metadata_kwargs.items():
        if val == "":
            raise ValueError(f"Value for '{key}' cannot be an empty string")
        if val not in format_string:
            raise ValueError(f"Value for '{key}' not found in the path")
        format_string = format_string.replace(val, f"{{{key}}}")

    return format_string
```

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -380,3 +381,49 @@ def test_expand_paths_ibl(tmpdir):
 
     tc = unittest.TestCase()
     tc.assertCountEqual(path_expansion_results, expected)
+
+
+class TestConstructPathTemplate(unittest.TestCase):
+    def setUp(self):
+        self.path = "/data/subject456/session123/file.txt"
```
CodyCBakerPhD, line 388: Would love to see more test cases of this in practice, especially on real paths used in our documentation

At a minimum, all the ones from the main sections of https://neuroconv.readthedocs.io/en/main/user_guide/expand_path.html#example-usage

Peeled out:

```
"file_path": "/allen-brain-observatory/visual-coding-2p/ophys_movies/ophys_experiment_496908818.h5"

"file_path": "/SenzaiY/YMV01/YMV01_170818/YMV01_170818.eeg"

"folder_path": "/SenzaiY/YMV01/YMV01_170818/"

"file_path": "/steinmetzlab/Subjects/NR_0017/2022-03-22/001/raw_video_data/_iblrig_leftCamera.raw.6252a2f0-c10f-4e49-b085-75749ba29c35.mp4"
```

Each of those has some interesting edge cases that would be useful to verify; the Allen one doesn't have a subject ID (doesn't need it either apparently, path expansion just needs the session ID?), the Buzsaki ones include subject twice and/or as a part of the session ID (bit of a user choice there), the IBL one has two extra metadata that are useful to test (session start time, which path expansion parses as datetime - and the lab name is technically in there too if we wanted to try to extract it)

I'd also toss one of our most nested+structured file formats in there to test rigorously this way, SpikeGLX

```
"file_path": "<base_folder_that_probably_has_subject_info>/Session1/Session1_g0/Session1_g0_t0.ap.bin"
```

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -380,3 +381,49 @@ def test_expand_paths_ibl(tmpdir):
...
+
+    def test_normal_operation(self):
+        result = construct_path_template(self.path, subject_id="subject456", session_id="session123")
+        self.assertEqual(result, "/data/{subject_id}/{session_id}/file.txt")
+
+    def test_with_additional_metadata(self):
+        result = construct_path_template(
+            "/data/subject789/session456/image.txt", subject_id="subject789", session_id="session456", file_type="image"
```
CodyCBakerPhD, line 396: I see what you're checking here, but just my 2 cents: if sticking to `unittest` structure, then would be more consistent to peel this case out into a separate `TestCase`, since the current one mixes application on some on-the-fly cases with the `self.path` example. Just makes it harder to catch if quickly skimming through the testing suite

I'd also say though that `pytest` has been the growing preference preference for new additions to the library, not that that is official policy or anything. Just that it would make this all flatter (all your example paths could be fixtures and each test would be much more obviously unit-like; `pytest.excinfo` is also a bit more convenient to write thorough tests around) and more readable IMO

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -380,3 +381,49 @@ def test_expand_paths_ibl(tmpdir):
...
+        )
+        self.assertEqual(result, "/data/{subject_id}/{session_id}/{file_type}.txt")
+
+    def test_empty_subject_id(self):
+        with self.assertRaises(ValueError):
```
CodyCBakerPhD, line 401: To ensure the right `ValueError` is raised in the right condition, can you also assert the message expected in each case? (the code itself has 4 separate places where that type is raised, each with different messages)

src/neuroconv/tools/path_expansion.py
```diff
@@ -103,3 +103,69 @@ def list_directory(self, base_directory: DirectoryPath) -> Iterable[FilePath]:
...
+        raise ValueError("Subject ID and Session ID must be present in the path")
+
+    path = path.replace(subject_id, "{subject_id}").replace(session_id, "{session_id}")
+
+    for key, val in metadata_kwargs.items():
+        if val == "":
+            raise ValueError(f"Value for '{key}' cannot be an empty string")
+        if val not in path:
+            raise ValueError(f"Value for '{key}' not found in the path")
+        path = path.replace(val, f"{{{key}}}")
```
CodyCBakerPhD, line 169: It's certainly a style choice so no strong opinion either way; but I'd say

```suggestion
        path = path.replace(val, f"{ {key} }")
```

is easier to read (less confusing why so many curly braces)

CodyCBakerPhD: Though I see in the case of string inputs the behavior is different, nevermind

CodyCBakerPhD:
@bendichter Sorry it took me a while to get around to this; the Simons slam followed by taking a bit of time to decompress afterward

Left some comments on things, mostly requesting more test cases

What would you think about incorporating some round-trip tests to ensure that all the cases we test this helper function on can and do also feed back into a `LocalPathExpander` to reproduce the expected metadata extraction and iteration?

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -380,3 +382,96 @@ def test_expand_paths_ibl(tmpdir):
...
+        ),
+        expected="/SenzaiY/{subject_id}/{subject_id}_{session_id}/{subject_id}_{session_id}.eeg",
+    ),
+    dict(
+        test_input=dict(path="/SenzaiY/YMV01/YMV01_170818/", metadata=dict(session_id="170818", subject_id="YMV01")),
+        expected="/SenzaiY/{subject_id}/{subject_id}_{session_id}/",
+    ),
+    dict(
+        test_input=dict(
+            path="/steinmetzlab/Subjects/NR_0017/2022-03-22/001/raw_video_data/_iblrig_leftCamera.raw.6252a2f0-c10f-4e49-b085-75749ba29c35.mp4",
```
bendichter, line 416: @CodyCBakerPhD this is a tricky one. What would you consider the session_id? "2022-03-22/001"?

tests/test_minimal/test_tools/test_expand_paths.py
bendichter: The date isn't going to be possible for the feature. You could put in `session_start_time=date(2022, 3, 22)`, but then you'd also need to pass a format string. Users will have to do this manually, by editing the template string with `"{session_start_time:%Y-%M-%D}"`. I'd rather not generalize this feature to enable this- I think it's easier to edit manually.

tests/test_minimal/test_tools/test_expand_paths.py
CodyCBakerPhD: It could be `2022-03-22`, or `2022-03-22/001`, or the UUID before the MP4 file `6252a2f0-c10f-4e49-b085-75749ba29c35`

tests/test_minimal/test_tools/test_expand_paths.py
```diff
@@ -421,3 +425,96 @@ def test_expand_paths_ibl(tmpdir):
...
+            metadata=dict(
+                session_id="001",
+                subject_id="NR_0017",
+                camera_id="leftCamera.raw.6252a2f0-c10f-4e49-b085-75749ba29c35",
+                session_date="2022-03-22",
+            ),
+        ),
+        expected="/steinmetzlab/Subjects/{subject_id}/{session_date}/{session_id}/raw_video_data/_iblrig_{camera_id}.mp4",
+    ),
```
CodyCBakerPhD, line 468:
```suggestion
    dict(
        test_input=dict(
            path="/steinmetzlab/Subjects/NR_0017/2022-03-22/001/raw_video_data/_iblrig_leftCamera.raw.6252a2f0-c10f-4e49-b085-75749ba29c35.mp4",
            metadata=dict(
                session_id="001",
                subject_id="NR_0017",
                camera_id="leftCamera.raw.6252a2f0-c10f-4e49-b085-75749ba29c35",
                session_date="2022-03-22",
            ),
        ),
        expected="/steinmetzlab/Subjects/{subject_id}/{session_date}/{session_id}/raw_video_data/_iblrig_{camera_id}.mp4",
    ),
    dict(
        test_input=dict(
            path="/steinmetzlab/Subjects/NR_0017/2022-03-22/001/raw_video_data/_iblrig_leftCamera.raw.6252a2f0-c10f-4e49-b085-75749ba29c35.mp4",
            metadata=dict(
                subject_id="NR_0017",
                session_date="2022-03-22",
                camera_number="001",
                video_id="6252a2f0-c10f-4e49-b085-75749ba29c35",
            ),
        ),
        expected="/steinmetzlab/Subjects/{subject_id}/{session_date}/{camera_number}/raw_video_data/_iblrig_{video_id}.mp4",
    ),
```

At least one more little test case here; the intention being that programmatically there exists a map from from the `video_id` hash to the actual `session_id` (just the way they did things in the IBL dataset), though the actual `eid` (as they refer to it) is not a part of the video filename

src/neuroconv/tools/path_expansion.py
```diff
@@ -132,3 +132,69 @@ def list_directory(self, base_directory: DirectoryPath) -> Iterable[FilePath]:
         base_directory = Path(base_directory)
         assert base_directory.is_dir(), f"The specified 'base_directory' ({base_directory}) is not a directory!"
         return (str(path.relative_to(base_directory)) for path in base_directory.rglob("*"))
+
+
+def construct_path_template(path: str, *, subject_id: str, session_id: str, **metadata_kwargs) -> str:
```
CodyCBakerPhD, line 137: Actually, the IBL case gives an example where `session_id` could be optional because there is another `metadata_kwarg` value in the filename that has a one-to-one correspondence with the true `session_id` value

I suppose the same point could actually be made with respect to the subject ID as well...

Perhaps the requirement should be that there is at least one keyword specified out of the important reserved words `subject_id` or `session_id`, or otherwise at least one key in `metadata_kwargs`?

CodyCBakerPhD:
One last requested test case then we can get this through

@garrettmflynn If you have any free time from the other priorities, you can use this branch to start https://github.com/NeurodataWithoutBorders/nwb-guide/issues/528; the function of relelvance is `neuroconv.tools.path_expansion.construct_path_template` with arguments `path: str, subject_id: str, session_id: str` ` and optional listing of key/value pairs `metadata_kwargs` (you may have to expose that one in a custom manner similar to 'add properties'

src/neuroconv/tools/path_expansion.py
CodyCBakerPhD: Though really this just highlights different usage between programmatic side and and GUIDE exposure

GUIDE wants this function for use in the session manager, which means it needs session and subject IDs explicitly specified in the paths

CodyCBakerPhD (approved):

PR #681, opened 2023-12-06
general cleanup

bendichter:
None of these changes change the functionality of the code. They are all just tidying up a bit.

* remove unused imports
* fix typos
* attributes like `data_interface_cls: BaseSortingExtractorInterface` should be `Type[BaseSortingExtractorInterface]`, indicating that it takes a class, not an object
* fix some typehint annotations

### DIFF
--- a/src/neuroconv/tools/testing/data_interface_mixins.py
+++ b/src/neuroconv/tools/testing/data_interface_mixins.py
@@ -277,7 +277,6 @@ class RecordingExtractorInterfaceTestMixin(DataInterfaceTestMixin, TemporalAlign
 
     data_interface_cls: Type[BaseRecordingExtractorInterface]
 
-
     def check_read_nwb(self, nwbfile_path: str):
         from spikeinterface.extractors import NwbRecordingExtractor
 
--- a/src/neuroconv/tools/testing/data_interface_mixins.py
+++ b/src/neuroconv/tools/testing/data_interface_mixins.py
@@ -277,9 +277,6 @@ class RecordingExtractorInterfaceTestMixin(DataInterfaceTestMixin, TemporalAlign
 
     data_interface_cls: Type[BaseRecordingExtractorInterface]
 
-    def __init__(self):
-        super().__init__()
-        self.nwb_recording = None
 
     def check_read_nwb(self, nwbfile_path: str):
         from spikeinterface.extractors import NwbRecordingExtractor
--- a/src/neuroconv/tools/testing/data_interface_mixins.py
+++ b/src/neuroconv/tools/testing/data_interface_mixins.py
@@ -60,6 +60,13 @@ class DataInterfaceTestMixin:
     conversion_options: dict = dict()
     maxDiff = None
 
+    def __init__(self):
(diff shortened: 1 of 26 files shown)


tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -35,24 +33,20 @@
...
-    @classmethod
-    def setUpClass(cls):
-        """Use common recording objects and values."""
-        cls.num_channels = 3
-        cls.sampling_frequency = 1.0
-        cls.durations = [3.0]  # 3 samples in the recorder
-        cls.test_recording_extractor = generate_recording(
-            sampling_frequency=cls.sampling_frequency, num_channels=cls.num_channels, durations=cls.durations
-        )
-
```
CodyCBakerPhD, line 47: I'm pretty sure we initialize `cls.test_recording_extractor` at the class level to avoid having to regenerate it again for every test, which should decrease the overall amount of time it takes to run this particular set of tests without altering their behavior (I don't think they modify the `test_recording_extractor` in place)

tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -52,7 +50,7 @@ def setUp(self):
         )
 
     def test_default_values(self):
-        add_electrical_series(recording=self.test_recording_extractor, nwbfile=self.nwbfile, iterator_type=None)
+        add_electrical_series(recording=self.test_recording_extractor, nwbfile=self.nwbfile)
```
CodyCBakerPhD, line 53: This type of change does actually alter behavior of the tests

`iterator_type = None` means 'disable iteration completely', as in the `data` is not wrapped in an iterator at all

This is one of those behaviors that will be changed soon in the backend integration but for now, as a non-altering cleanup PR, I'd say revert that too

bendichter: ah, ok I didn't notice that. the typehints of `the iterator_type` arg of `add_electrical_series` indicate that the arg must be a string, so I thought this was a mistake. I'll just update those typehints instead.

CodyCBakerPhD: Yeah the annotations in the signature are not correct, those could be updated here

CodyCBakerPhD:
CI is angry about something... Perhaps the `__init__` changes in the mixin? Error log isn't very clear

bendichter:
yeah, those changes are overriding the init from TestCase

bendichter:
I'll revert those as well

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -277,6 +277,10 @@ class RecordingExtractorInterfaceTestMixin(DataInterfaceTestMixin, TemporalAlign
 
     data_interface_cls: Type[BaseRecordingExtractorInterface]
 
+    def __init__(self):
+        super().__init__()
+        self.nwb_recording = None
```
bendichter, line 282:
```suggestion
```

PR #683, opened 2023-12-09
Add router to alternate SI class for pl2

CodyCBakerPhD:
I missed this when I heard `pl2` support had been finalized in SI; I figured they added the proper routing to the main Plexon recording extractor, but apparently it was added as a different class


CodyCBakerPhD:
Didn't fix the issue the user as experiencing; something was simply off with their pl2 files - directed to issues on neo

PR #695, opened 2023-12-19
Exposed block_index to OpenEphys interfaces

### LINKED ISSUE #692: [Feature]: block index for OpenEphysBinaryRecordingInterface

Please add support for loading multi-block Open Ephys datasets, either one block at a time or all at once.

Right now OpenEphysBinaryRecordingInterface throws an error if you try to load a multi-block Open Ephys dataset (i.e. one with more than one experiment folder under the top-level folder). Right now as far as I can tell the only workaround would be to split out each experiment into separate folders, but that's fiddly for users with lots of multi-block recordings. With a two-line change to pass the block_index through to the underlying SpikeInterface extractor it seems to load just fine. Not sure how feasible or necessary it would be to load all blocks with a single OpenEphysBinaryRecordingInterface instance, I think the single-block solution is fine for my purposes.

I have a multi-experiment OpenEphys recording and I want to load it with OpenEphysBinaryRecordingInterface.

No.

### Code of Conduct

CodyCBakerPhD:
fix #692

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/openephys/openephysbinarydatainterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/openephys/openephysbinarydatainterface.py
@@ -20,8 +20,11 @@ def _open_with_pyopenephys(folder_path: FolderPathType):
 
 
 class OpenEphysBinaryRecordingInterface(BaseRecordingExtractorInterface):
-    """Primary data interface for converting binary OpenEphys data (.dat files). Uses
-    :py:class:`~spikeinterface.extractors.OpenEphysBinaryRecordingExtractor`."""
+    """
+    Primary data interface for converting binary OpenEphys data (.dat files).
+
+    Uses :py:class:`~spikeinterface.extractors.OpenEphysBinaryRecordingExtractor`.
+    """
 
     ExtractorName = "OpenEphysBinaryRecordingExtractor"
 
@@ -47,6 +50,7 @@ class OpenEphysBinaryRecordingInterface(BaseRecordingExtractorInterface):
         self,
         folder_path: FolderPathType,
         stream_name: Optional[str] = None,
+        block_index: Optional[int] = None,
         stub_test: bool = False,
         verbose: bool = True,
         es_key: str = "ElectricalSeries",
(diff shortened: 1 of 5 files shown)


src/neuroconv/datainterfaces/ecephys/openephys/openephysbinarydatainterface.py
```diff
@@ -86,15 +92,19 @@ def __init__(
         available_streams = self.get_stream_names(folder_path=folder_path)
         if len(available_streams) > 1 and stream_name is None:
             raise ValueError(
-                "More than one stream is detected! Please specify which stream you wish to load with the `stream_name` argument. "
-                "To see what streams are available, call `OpenEphysRecordingInterface.get_stream_names(folder_path=...)`."
+                "More than one stream is detected! "
+                "Please specify which stream you wish to load with the `stream_name` argument. "
+                "To see what streams are available, call "
```
h-mayorquin, line 97: I think you should print the streams for users here given that you already calculated them.

tests/test_on_data/test_recording_interfaces.py
```diff
@@ -409,6 +413,21 @@ def check_extracted_metadata(self, metadata: dict):
         assert metadata["NWBFile"]["session_start_time"] == datetime(2020, 11, 24, 15, 46, 56)
 
 
+class TestOpenEphysBinaryRecordingInterface_version_0_5_3_block_1_stream_1(
```
h-mayorquin, line 416: I would prefer to name this test by purpose `TestOpenEphysBinaryWithBlocks` and add the provenance as a comment. But this is just my preference I see the argument for adding the provenance.

h-mayorquin (commented):
This looks good to me. Thanks for sorting this out so quick. I would approve it but I am afraid of auto-merge. Let's see if the user tests this today or tomorrow and if we don't we merge. I can also take a look at the data that he shared with us.

tests/test_on_data/test_recording_interfaces.py
```diff
@@ -409,6 +413,21 @@ def check_extracted_metadata(self, metadata: dict):
...
+    RecordingExtractorInterfaceTestMixin, TestCase
+):
```
CodyCBakerPhD, line 418:
```suggestion
class TestOpenEphysBinaryRecordingInterfaceWithBlocks_version_0_6_block_1_stream_1(
    RecordingExtractorInterfaceTestMixin, TestCase
):
    """From Issue #695, exposed `block_index` argument and added tests on data that include multiple blocks."""
```

@h-mayorquin

src/neuroconv/datainterfaces/ecephys/openephys/openephysbinarydatainterface.py
CodyCBakerPhD: Feel free to add this is a separate PR, I probably won't in this one (I was just breaking up a line length violation in this change)

h-mayorquin: Makes sense.

CodyCBakerPhD:
@h-mayorquin User OK'd the changes

h-mayorquin (approved):

PR #578, opened 2023-09-25
[Draft] Ecephys integration of backend configuration

CodyCBakerPhD:
Opening this because as I work on it I realize, doing this straight up as in the reference PR has the effect that this is a hard, immediate deprecation that breaks backcompatability

So this PR would specifically be a major version bump on the package; and the only way I can think to soften it is maybe keep the keyword arguments around for a while but throw DeprecationWarnings that they no longer do anything, and instructions on the new approach to setting the compression+iteration options

What do you think @h-mayorquin?

PR #699, opened 2024-01-02
Disable H5DataIO compression if compression is None

alejoe91:
Disables hardcoded `H5DataIO` wrapping in `tools.spikeinterface.write_recording` if `compression=None`.

This change enables one to use the `configure_backend` functionality (added a note in the docstring)


alejoe91:
Closing in favor of #578

CodyCBakerPhD:
#578 still needs a little work to debug some stuff

I'd be fine with this one for now; also no need to add a test since those options are to be deprecated in the follow-up anyway

alejoe91:
Then I'll reopen! Couldn't test the full thing for the issue on `set_data_io` for tables, but electrical series worked fine when configured with a Zarr compressor!

PR #701, opened 2024-01-09
add description for NP probe type NHP

luiztauffer:


CodyCBakerPhD:
I'd greatly appreciate having a test for this; do you think you could manually stub the example file they shared down to < 10 MB (or as small as you want really) and send it to the GIN folks + add to the S3 bucket?

Also changelog

CodyCBakerPhD (approved):

PR #703, opened 2024-01-11
[Backend Configuration IIIc]: Improved testing and debugs

CodyCBakerPhD:
Bug caught by @alejoe91 resulting from some odd mutation of the dataset resulting from it being wrapped unnecessarily in a `DataChunkIterator` (@h-mayorquin had good intuition that this might cause a problem and should be avoided; I've reverted to more basic dtype retrieval for that step until we figure out a better path forward; will work fine for source data that are numpy arrays or anything similar)


alejoe91:
Working as a charm on my side with custom Zarr compressors!!!

alejoe91 (approved):

h-mayorquin:
Side effects are dangerous! Thanks for catching this @alejoe91

Issue #540, opened 2023-08-08
[Bug]: Example GIN dataset for TdT results in conversion_factor of `1.0`


CodyCBakerPhD:
### What happened?

Not sure if it's a problem with the dataset being an 'example' (and so not correctly depicting a true example of the data) or if it is just the way the format is (like NeuroScope)

Will investigate using some alternative TdT data

If the second case, then need to expose the conversion factor as an `__init__` argument to the interface

### Steps to Reproduce

```python
Generate TdT file from test suite
```

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.7

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

PR #704, opened 2024-01-12
Add gain to `TdtRecordingInterface`

weiglszonja:
Expose gain as an `__init__` argument for `TdtRecordingInterface`
resolve https://github.com/catalystneuro/neuroconv/issues/540

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/tdt/tdtdatainterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/tdt/tdtdatainterface.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
 from ....utils.types import FolderPathType
 
@@ -6,7 +8,12 @@ class TdtRecordingInterface(BaseRecordingExtractorInterface):
     """Primary data interface class for converting Tucker-Davis Technologies (TDT) data."""
 
     def __init__(
-        self, folder_path: FolderPathType, stream_id: str = "0", verbose: bool = True, es_key: str = "ElectricalSeries"
+        self,
+        folder_path: FolderPathType,
+        stream_id: str = "0",
+        gain: Optional[float] = None,
+        verbose: bool = True,
+        es_key: str = "ElectricalSeries",
     ):
         """
         Initialize reading of a TDT recording.
@@ -17,6 +24,9 @@ class TdtRecordingInterface(BaseRecordingExtractorInterface):
             Path to the directory with the corresponding files (TSQ, TBK, TEV, SEV)
         stream_id : str, "0" by default
             Select from multiple streams.
+        gain : float, optional
+            The conversion factor from int16 to Volts.
+            The default is None.
         verbose : bool, default: True
             Allows verbose.
         es_key : str, optional
@@ -37,3 +47,6 @@ class TdtRecordingInterface(BaseRecordingExtractorInterface):
         channel_names = self.recording_extractor.get_property("channel_name")
         channel_names = [name.replace("'", "")[1:] for name in channel_names]
         self.recording_extractor.set_property(key="channel_name", values=channel_names)
+
+        if gain is not None:
+            self.recording_extractor.set_channel_gains(gain)
--- a/tests/test_on_data/test_recording_interfaces.py
+++ b/tests/test_on_data/test_recording_interfaces.py
@@ -539,6 +539,28 @@ class TestTdtRecordingInterface(RecordingExtractorInterfaceTestMixin, TestCase):
     save_directory = OUTPUT_PATH
 
 
+class TestTdtRecordingInterfaceWithGain(RecordingExtractorInterfaceTestMixin, TestCase):
+    data_interface_cls = TdtRecordingInterface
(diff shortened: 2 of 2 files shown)


src/neuroconv/datainterfaces/ecephys/tdt/tdtdatainterface.py
```diff
@@ -6,7 +8,12 @@ class TdtRecordingInterface(BaseRecordingExtractorInterface):
     """Primary data interface class for converting Tucker-Davis Technologies (TDT) data."""
 
     def __init__(
-        self, folder_path: FolderPathType, stream_id: str = "0", verbose: bool = True, es_key: str = "ElectricalSeries"
+        self,
+        folder_path: FolderPathType,
+        stream_id: str = "0",
+        gain: Optional[float] = None,
```
CodyCBakerPhD, line 14:
```suggestion
        gain: float,
```

I would be in favor of making this required, since the 'default' is not to be trusted

What do you think?

tests/test_on_data/test_recording_interfaces.py
```diff
@@ -539,6 +540,29 @@ class TestTdtRecordingInterface(RecordingExtractorInterfaceTestMixin, TestCase):
...
+class TestTdtRecordingInterfaceWithGain(RecordingExtractorInterfaceTestMixin, TestCase):
+    data_interface_cls = TdtRecordingInterface
+    interface_kwargs = dict(folder_path=str(DATA_PATH / "tdt" / "aep_05"), gain=1.0)
+    save_directory = OUTPUT_PATH
+
+    def run_custom_checks(self):
+        # Check that the gain is applied
+        recording_extractor = self.interface.recording_extractor
+        gains = recording_extractor.get_channel_gains()
+        expected_channel_gains = [1.0] * recording_extractor.get_num_channels()
```
CodyCBakerPhD, line 552: I would use something other than `1.0` to make the test a little clearer, since `1.0` is often a default of other things

weiglszonja: applied that in: [b95cfaf](https://github.com/catalystneuro/neuroconv/pull/704/commits/b95cfaff1f61ad7d4c9bd2c35ae54b18ff604b95)

src/neuroconv/datainterfaces/ecephys/tdt/tdtdatainterface.py
weiglszonja: I think until https://github.com/NeuralEnsemble/python-neo/issues/1369 is resolved it is probably the best we can do, the question is whether the gain should be `1.0` in the docs ?

src/neuroconv/datainterfaces/ecephys/tdt/tdtdatainterface.py
```diff
@@ -17,6 +24,8 @@ def __init__(
             Path to the directory with the corresponding files (TSQ, TBK, TEV, SEV)
         stream_id : str, "0" by default
             Select from multiple streams.
+        gain : float
+            The conversion factor from int16 to Volts.
```
CodyCBakerPhD, line 28:
```suggestion
            The conversion factor from int16 to microvolts.
```

The `gain` specified here sets the `gain_to_uV` value on the SI recording object, right? So the `gain` we want is `1.0`, but that is to map the `int16` to `uV`, the current problem of which is that it's actually applying to Volts, resulting in the incorrect value of `1e6`

src/neuroconv/datainterfaces/ecephys/tdt/tdtdatainterface.py
CodyCBakerPhD: Yeah, I'd say the conversion gallery using that value is a good idea for documentations

src/neuroconv/datainterfaces/ecephys/tdt/tdtdatainterface.py
weiglszonja: Yes, you're right!

CodyCBakerPhD (approved):

docs/conversion_examples_gallery/recording/tdt.rst
```diff
@@ -19,7 +19,7 @@ Convert TDT data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.tdt.t
     >>> # For this data interface we need to pass the folder_path with the location of the data
     >>> folder_path = f"{ECEPHY_DATA_PATH}/tdt/aep_05"
     >>> # Change the folder_path to the location of the data in your system
-    >>> interface = TdtRecordingInterface(folder_path=folder_path, verbose=False)
+    >>> interface = TdtRecordingInterface(folder_path=folder_path, gain=1.0, verbose=False)
```
h-mayorquin, line 22: @CodyCBakerPhD maybe we should add a note to users on where or how to get this? 

It could also be in the docstring of the class.

CodyCBakerPhD: Aside from justifying our choice of `1.0` (which I thought Szonja did) we have no instructions for how to find - assumption is user knows where/who they bought it from and would be able to consult their docs or get in touch with manufacturer

PR #705, opened 2024-01-12
Fix sleap python range

CodyCBakerPhD:
Recent release of sleap doesn't support Python 3.9 and causes error in CI; not one of our dev test suite packages so didn't catch it until today


weiglszonja (approved):
LGTM, thanks @CodyCBakerPhD!

PR #709, opened 2024-01-19
cleaning

bendichter:
* rename TestSLEAPInterface since there is another class by that name
* rmv duplicate test_non_overwriting_unit_names_sorting_property
* rmv unnecessary parens
* rmv unused imports
* fix typo

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -72,7 +72,7 @@ add_module_names = False
 def _correct_signatures(app, what, name, obj, options, signature, return_annotation):
     if what == "class":
         signature = str(inspect.signature(obj.__init__)).replace("self, ", "")
-    return (signature, return_annotation)
+    return signature, return_annotation
 
 
 def setup(app):  # This makes the data-interfaces signatures display on the docs/api, they don't otherwise
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -72,7 +72,7 @@ add_module_names = False
 def _correct_signatures(app, what, name, obj, options, signature, return_annotation):
     if what == "class":
         signature = str(inspect.signature(obj.__init__)).replace("self, ", "")
-    return signature, return_annotation
+    return (signature, return_annotation)
 
 
 def setup(app):  # This makes the data-interfaces signatures display on the docs/api, they don't otherwise
--- a/tests/test_ecephys/test_tools_spikeinterface.py
+++ b/tests/test_ecephys/test_tools_spikeinterface.py
@@ -909,7 +909,7 @@ class TestAddUnitsTable(TestCase):
         self.assertListEqual(unit_names_in_units_table, expected_unit_names_in_units_table)
 
     def test_non_overwriting_unit_names_sorting_property(self):
-        "add_units_table function should not ovewrtie the sorting object unit_name property"
+        "add_units_table function should not overwrite the sorting object unit_name property"
         unit_names = ["name a", "name b", "name c", "name d"]
         self.sorting_1.set_property(key="unit_name", values=unit_names)
         add_units_table(sorting=self.sorting_1, nwbfile=self.nwbfile)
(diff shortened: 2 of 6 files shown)


tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -942,16 +942,6 @@ def test_string_unit_names_overwrite(self):
         unit_names_in_units_table = list(self.nwbfile.units["unit_name"].data)
         self.assertListEqual(unit_names_in_units_table, expected_unit_names_in_units_table)
 
-    def test_non_overwriting_unit_names_sorting_property(self):
```
bendichter, line 945: This is a duplicate

tests/test_on_data/test_behavior_interfaces.py
```diff
@@ -480,7 +480,7 @@ def check_metadata(self):
         assert metadata["NWBFile"]["session_start_time"] == datetime(2023, 5, 15, 10, 35, 29)
 
 
-class TestSLEAPInterface(TestCase):
+class CustomTestSLEAPInterface(TestCase):
```
bendichter, line 483: there is another class in this file named TestSLEAPInterface

docs/conf.py
```diff
@@ -72,7 +72,7 @@
 def _correct_signatures(app, what, name, obj, options, signature, return_annotation):
     if what == "class":
         signature = str(inspect.signature(obj.__init__)).replace("self, ", "")
-    return (signature, return_annotation)
+    return signature, return_annotation
```
CodyCBakerPhD, line 75: I'd say that casting as a tuple at line of return is more explicit than knowing that any comma-separated returns do so automatically

tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -909,7 +909,7 @@ def test_string_unit_names(self):
         self.assertListEqual(unit_names_in_units_table, expected_unit_names_in_units_table)
 
     def test_non_overwriting_unit_names_sorting_property(self):
-        "add_units_table function should not ovewrtie the sorting object unit_name property"
+        "add_units_table function should not overwrite the sorting object unit_name property"
```
CodyCBakerPhD, line 912: Surprised codespell didn't catch this one

tests/test_minimal/test_metadata_schema.py
```diff
@@ -1,9 +1,6 @@
-import traceback
```
CodyCBakerPhD, line 1: Odd that isort did not catch this; maybe it's only running on `src` files?

tests/test_on_data/test_behavior_interfaces.py
CodyCBakerPhD: This one looks to be the older style before the new mixin was added (and when mixin was added, this one wasn't removed)

I think the goal should be to remove this one in a follow up and absorb any differences with its test function into the new mixin (if there are any differences that is) - probably in a follow-up

docs/conf.py
bendichter: I'm used to the simpler non-parens syntax, and this is suggested by PyCharm's auto checker. I don't feel that strongly about it though. We can change it back if you want.

tests/test_on_data/test_behavior_interfaces.py
bendichter: Yeah, I thought about deleting it but it looked like there were some differences and I didn't know exactly how to merge them. I agree, there is probably a lot of redundancy here.

tests/test_minimal/test_metadata_schema.py
bendichter: yeah, maybe

docs/conf.py
CodyCBakerPhD: I would probably change it back. One of those style things that yes, saves characters in a line, but avoids any potential confusion especially in cases of returning nested tuples or tuple-likes such as arraays (which you'd then have to know are nested at a particular level; classic example of this is `np.where`)

CodyCBakerPhD, line 75:
```suggestion
    return (signature, return_annotation)
```

CodyCBakerPhD (approved):

PR #710, opened 2024-01-19
Try to add test folder to config of isort

CodyCBakerPhD:
Testing to see if this does anything


CodyCBakerPhD:
Oh, I guess isort doesn't actually remove them. It's been a requested feature for a while but doesn't look like it was integrated yet

PR #713, opened 2024-01-22
Allow control over timestamps write behavior for recording extractors

### LINKED ISSUE #712: [Feature]: allow user to force storing timestamps (as opposed to start time + rate) for BaseRecordingExtractorInterface subclasses

When writing ephys data to NWB using neuroconv, I would like to be able to explicitly control whether the timestamps are sorted as a vector, or a simple rate & start time.

Right now, `BaseRecordingExtractorInterface.add_to_nwbfile` via `neuroconv.tools.spikeinterface.add_electrical_series` decides whether to store timings for a recording as an explicit timestamp vector or a starting time & rate based on the regularity of the timestamps (i.e. are they identical to within 6 decimal places). This can introduce timing drift in certain scenarios. Unless I'm missing something, the user has no control over this behaviour.

As far as I can tell by inspecting the code, there is no trivial way to override this behaviour by subclassing whichever data interface you're working with. It's not possible to add timestamps to the ElectricalSeries once it's added to the NWB file. The only option would be to reimplement `BaseRecordingExtractorInterface.add_to_nwbfile`, which would be a huge amount of code duplication.

This is related to https://github.com/catalystneuro/neuroconv/issues/302 and https://github.com/catalystneuro/neuroconv/issues/711. I have two systems (one Neuropixels, one Intan) with similar but not identical sampling rate. I can align them using a TTL pulse train sent from one system to the other, but the aforementioned regularity check means that only the aligned starting time and original sampling rate are saved to the NWB file. The precise, per-sample alignment is lost. There is non-trivial drift between the aligned timestamps and those inferred from the sampling rate: about 0.5 ms/min. For my current purposes that's not catastrophic, but for longer recordings (tens of minutes) it's not hard to imagine experiments where that kind of timing drift could meaningfully affect the results. 

### Code of Conduct

CodyCBakerPhD:
fix #712


h-mayorquin (commented):
If our default is going to be follow best practices if apperars to me that `use_timestamps` should probably be named `force_timestamps` instead, what do you think?

Save this, this looks all right to me. As this is a future request maybe we can add it as it is. We can then after think about adding this into our mixing tests and propagate into the other modalities. What do you think?

Btw, this would be the third example of parameter (together with `write_in_processing_module` and `stub_test`) as per our discussion of the configuration classess, right?

h-mayorquin:
OK, I can finish this here:
https://github.com/catalystneuro/neuroconv/pull/1091

Can we close this @CodyCBakerPhD ?

CodyCBakerPhD:
@h-mayorquin As long as you ensure the needs of #712 are satifisfied, sure

PR #715, opened 2024-01-24
change docs to pydata theme

### LINKED ISSUE #347: [Documentation]: use pydata theme

Lots of leading numerical computing Python libraries are using [the PyData Theme for readthedocs](https://pydata-sphinx-theme.readthedocs.io/en/stable/index.html), and I think it's a really nice style. Let's look into adopting this style. I tried just naively applying it and it looks like this will take a bit of reorganization of the docs to make it look good.

* API documentation will need to re reorganized so that API is on the top level and all API documentation are beneath it
* The catalog should be moved somewhere else. Maybe at the end of the user guide?

Maybe

### Code of Conduct

bendichter:
fix #347

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,0 +1,11 @@
+API
+===
+
+.. toctree::
+  :maxdepth: 2
+
+  api/basedatainterface
+  api/interfaces
+  api/nwbconverter
+  api/tools
+  api/utils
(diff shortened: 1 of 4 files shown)


bendichter:
I made a few changes:
* Made API its own top-level section, which works much better now that it is in the header
* Added a link to the GitHub page
* Added an "Edit this page" link, which allows any reader to easily suggest changes and file a PR.

for some reason, the favicon is now broken

CodyCBakerPhD:
Looks great!

CodyCBakerPhD (approved):

PR #720, opened 2024-02-06
add display_name and help for each interface to be used by GUIDE

bendichter:
display_name, when present, should be used instead of the class name help, when presesnt, should be used for the info hover instead of the docstring of the class


bendichter:
@CodyCBakerPhD and @garrettmflynn , thoughts on this?

CodyCBakerPhD (approved):

CodyCBakerPhD:
LGTM - should only require a minimal injection as mentioned in https://github.com/NeurodataWithoutBorders/nwb-guide/pull/589

PR #724, opened 2024-02-07
Update plexon example to only update timezone

CodyCBakerPhD:
@bendichter This part should no longer be necessary then, I assume


bendichter:
Yes perfect thank you!

bendichter (approved):

PR #726, opened 2024-02-07
try new structure for conversion examples

bendichter:
The problem is that currently with the new pydata skin there are 3 different tables that basically do the same thing:

image

With this new navigation, it may be more convenient to put all the conversion examples on a single page, making it easier to CTRL+F. You can easily navigate to the desired conversion using the sidebar.

What do you guys think of this as an alternative structure/layout?

image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -116,4 +116,3 @@ extlinks = {
 html_sidebars = {
     "conversion_example_gallery": ["page-toc", "edit-this-page", "sourcelink"]
 }
-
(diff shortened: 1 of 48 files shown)


bendichter:
Here's another alternative:

image

h-mayorquin:
I think this is a good idea. We discussed something like this with the core-team If I remember well. With the navigation tables I see no lost as the user can still go to the code they want directly, the links to share a specific section seem to be working as well:

https://neuroconv--726.org.readthedocs.build/en/726/conversion_example_gallery.html#alphaomega

I prefer the current version with all the navigation collapsed on the left. The version of this with two side bars (like Pandas) [here](https://pandas.pydata.org/docs/user_guide/indexing.html#different-choices-for-indexing
) works only when I think there is a deep structure which we don't have here. 

I checked in mobile and it also seems to look OK.

PR #727, opened 2024-02-07
Remove roundabout JSON interaction with electrode table

CodyCBakerPhD:
As per meeting; moving this from a generic interface method that has no programmatic use to GUIDE specific interactions


garrettmflynn (approved):
Perfect. Good call

PR #730, opened 2024-02-08
Add an introductory section on "how to use the documentation" to the read the docs landing page

h-mayorquin:
I feel that there is no clear directions on how to navigate our documentation from the landing page. This PR adds a section describing how the documentation is structured and gives users a short description on what they can find on each of the sections:

image

Let me know what you think.

I also added a couple of links that I think could be useful at the end of the landing page such as a link to the guide project and to the nwb overview repo.


bendichter:
Third doesn't hurt but in general people don't read exposition, they just
click and rapidly until they find what they want so it's very hard to fix a
UX problem with text

Ben Dichter, PhD
Data Science Consultant
personal website <http://bendichter.com/>

h-mayorquin:
@bendichter 

That's a good ponint. This will benefit the most the minority of users who do read. Good that it does not hurt though. 
From that's perspective, what do you think of switching Conversion Gallery and User Guide order so they click first the most introductory section with simple code examples?

weiglszonja:
> From that's perspective, what do you think of switching Conversion Gallery and User Guide order so they click first the most introductory section with simple code examples?

I think [this](https://neuroconv.readthedocs.io/en/main/) is already how we have it, to me it makes sense.

Maybe it is not related to this PR, but do you think it would be useful to have a "How to get started (with your conversion)?" page? I'm thinking something similar to [this](https://spikeinterface.readthedocs.io/en/latest/how_to/get_started.html), that would very briefly explain what are data interfaces, converters, how to install and update the metadata etc. it would also be great to show where we are adding the ophys/ephys data in NWB, if that makes sense.

pauladkisson:
If we are going to add more detail to each section header, I vote that we
1. Remove the block of links under "Contents" since it becomes redundant and takes up a lot of space
2. Adopt the graphical grid structure used by [NumPy](https://numpy.org/doc/stable/index.html), [SciPy](https://docs.scipy.org/doc/scipy/), and [pandas](https://pandas.pydata.org/docs/)
  Note: pandas exposes its source so you can just grab the formatting from there.

pauladkisson (changes requested):
If we are going to add more detail to each section header, I vote that we
1. Remove the block of links under "Contents" since it becomes redundant and takes up a lot of space
2. Adopt the graphical grid structure used by [NumPy](https://numpy.org/doc/stable/index.html), [SciPy](https://docs.scipy.org/doc/scipy/), and [pandas](https://pandas.pydata.org/docs/)
  Note: pandas exposes its source so you can just grab the formatting from there.

h-mayorquin:
> > From that's perspective, what do you think of switching Conversion Gallery and User Guide order so they click first the most introductory section with simple code examples?
> 
> I think [this](https://neuroconv.readthedocs.io/en/main/) is already how we have it, to me it makes sense.

I meant that the Conversion Gallery should be first (more to the left) than the User Guide.

> Maybe it is not related to this PR, but do you think it would be useful to have a "How to get started (with your conversion)?" page? I'm thinking something similar to [this](https://spikeinterface.readthedocs.io/en/latest/how_to/get_started.html), that would very briefly explain what are data interfaces, converters, how to install and update the metadata etc. it would also be great to show where we are adding the ophys/ephys data in NWB, if that makes sense.

Do you expect this would end up looking very different from the current User Guide? 

@pauladkisson 
> If we are going to add more detail to each section header, I vote that we

That was not my intention, I just wanted to add a "what is what" words about how the documentation is structured somewhere.

The grid proposal seems like a good idea -because of familiarity- but probably better discussed on a separate issue.

pauladkisson:
> The grid proposal seems like a good idea -because of familiarity- but probably better discussed on a separate issue.

Ok, but I still think we should ditch the Contents section.

h-mayorquin:
@pauladkisson 
What is the thing that you want to remove? Was it added on this PR? Can you add a screenshot?

docs/index.rst
```diff
@@ -19,9 +19,42 @@ Features:
...
+* **API Reference**
+
+  Detailed documentation of the NeuroConv API can be found in the `API <api>`_ section.
+
+
+Do you find that some information is missing or some section lacking or unclear? Reach out with an issue or pull request on our `GitHub repository <https://github.com/catalystneuro/neuroconv>`_.
+We are happy to help and appreciate your feedback.
 
 .. toctree::
   :maxdepth: 2
```
pauladkisson, line 60: I was suggesting we delete this section bc it's redundant

pauladkisson (changes requested):

docs/index.rst
h-mayorquin: Thanks.
Let's move this to another PR:
https://github.com/catalystneuro/neuroconv/pull/796

pauladkisson (commented):
Something is failing with the external links. Otherwise LGTM

pauladkisson (approved):

PR #731, opened 2024-02-08
Reference ROIExtractors for ScanImage metadata

pauladkisson:
Fixes #729.

PR #733, opened 2024-02-10
[Bug] Fix non-lazy ROIExtractors import

CodyCBakerPhD:
Failed to notice this with PR #731, my bad. Imports like this should be lazy and done within the `__init__` of the class so the outer imports don't fail under a minimal installation

### DIFF
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -4,4 +4,3 @@ ndx-events>=0.2.0  # for special tests to ensure load_namespaces is set to allow
 parameterized>=0.8.1
 ndx-miniscope
 spikeinterface[qualitymetrics]>=0.99.1
-roiextractors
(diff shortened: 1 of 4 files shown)


requirements-testing.txt
```diff
@@ -4,4 +4,3 @@ ndx-events>=0.2.0  # for special tests to ensure load_namespaces is set to allow
 parameterized>=0.8.1
 ndx-miniscope
 spikeinterface[qualitymetrics]>=0.99.1
```
h-mayorquin, line 6: why do we still have spikeinterface though?

h-mayorquin (approved):
Seems good. You are checking that the you can import interfaces but that they fail at run time, right?

requirements-testing.txt
CodyCBakerPhD: Not SpikeInterface in particular, but the deps it needs for the `[qualitymetrics]` extra are used to test an advance feature of SI (sparisty I believe? Or something like that)

CodyCBakerPhD:
> You are checking that the you can import interfaces but that they fail at run time, right?

In the particular case of minimal install in fresh environment, yess

PR #734, opened 2024-02-11
Propose suffix association

CodyCBakerPhD:
replaces #344 

Instead of using this in validation as a rule (which might be too strict in some cases), this is just to indicate common associations

Like the `help` and `display_name`, the GUIDE could both display these on interface selection and leverage them in the source data  or example fstring inference popup file selector as a 'default file suffix filter' to make it easy to choose the most likely appropriate file out of a folder that might otherwise have a lot of contents


CodyCBakerPhD:
cc @h-mayorquin @bendichter @garrettmflynn

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
```diff
@@ -35,11 +35,12 @@ def add_recording_extractor_properties(recording_extractor) -> None:
 
 
 class SpikeGLXRecordingInterface(BaseRecordingExtractorInterface):
-    help = "Interface for SpikeGLX recording data."
     display_name = "SpikeGLX Recording"
     keywords = BaseRecordingExtractorInterface.keywords + [
         "Neuropixels",
     ]
+    common_suffixes = ["^.imec\d+$", ".ap", ".lf", ".bin"]
```
CodyCBakerPhD, line 42: These could either be regex patterns in the case of suffix patterns like `.imec0`, `.imec1`, etc. here - or just static strings

bendichter:
I like this. For neo interfaces we should be able to snag this from the reader object.

src/neuroconv/basedatainterface.py
```diff
@@ -20,7 +20,10 @@
 class BaseDataInterface(ABC):
     """Abstract class defining the structure of all DataInterfaces."""
 
+    display_name: Union[str, None] = None
     keywords: List[str] = []
+    common_suffixes: List[str] = []
+    help: Union[str, None] = None
```
bendichter, line 26: Any reason you aren't using Optional here?

CodyCBakerPhD: @h-mayorquin and I have gone back and forth on this philosophy - knowing to use `Optional` because it is equivalent is an 'implicit' behavior (PEP8 says explicit > implicit) and therefore not always obvious at first glance or to newcomers, or to even seasoned devs who forget in a moment of debugging

So we find it a bit easier to read to see at a glance the `Union` of all allowed types for the value

Also in (3.11 I think?) this syntax becomes `str | None` so it's even cleaner than `Optional` wrapping at that point too

CodyCBakerPhD:
> For neo interfaces we should be able to snag this from the reader object.

From

```python
from neuroconv.datainterfaces import SpikeGLXRecordingInterface

interface = SpikeGLXRecordingInterface(file_path="F:/GIN/ephy_testing_data/spikeglx/Noise4Sam_g0/Noise4Sam_g0_imec0/Noise4Sam_g0_t0.imec0.ap.bin")

neo_reader = interface.recording_extractor.neo_reader

neo_reader.extensions
> ['meta', 'bin']
```

so missing the `imec{x}` and `ap` and `lf` characteristics

We could change the type of `common_suffixes` to a `set` and take union of hardcoded ones we know of plus these `extensions`?

CodyCBakerPhD:
> We could change the type of common_suffixes to a set and take union of hardcoded ones we know of plus these extensions?

Actually, this isn't going to work either; we only have the `recording_extractor`, and by way its `neo_reader` after initialization on source data -  whereas the point to these attributes is they are available directly from the classes we've imported

(the GUIDE listing functionality is extremely minimal; it only iterates through the imported interfaces and grabs a handful of attributes from each one)

Might be best then to just offload complexity of checking consistency with neo in simple test in the recording mixin

h-mayorquin:
+1 on this. I think this is helpful and I see no other cost  than the initial work of adding it.

I would vote against regex in the strings though.

src/neuroconv/basedatainterface.py
```diff
@@ -20,7 +20,10 @@
 class BaseDataInterface(ABC):
     """Abstract class defining the structure of all DataInterfaces."""
 
+    display_name: Union[str, None] = None
     keywords: List[str] = []
+    common_suffixes: List[str] = []
```
CodyCBakerPhD, line 25:
```suggestion
    associated_suffixes: List[str] = []
```

CodyCBakerPhD: Changing to 'associated_suffixes' since some of these might refer to 'nearby' files, such as headers (example: `.meta` for SpikeGLX, `.xml` for NeuroScope, etc.)

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py
CodyCBakerPhD, line 42:
```suggestion
    associated_suffixes= [".imec", ".ap", ".lf", ".meta", ".bin"]
```

CodyCBakerPhD:
> I would vote against regex in the strings though.

How about semi-format strings of the kind `.imec{probe_number}`? It's actually fairly important for certain formats like NeuroScope Sorting (last suffix is numeric of group number) or Blackrock (stream type)

CodyCBakerPhD:
I'm also proposing to change from `help` to `info` since `help` is a reserved built-in phrase for Python.

bendichter:
It seems to me there are two different use cases here. The first would be to inform users in their usage of neuroconv and guide. Here, it seems to me suffixes would work. Another use case is to use these programmatically, either to constrain the default settings of a file selector dialog box, or to inspect input paths, etc. For that use case, full regex seems better to me

garrettmflynn:
Just to make sure we aren’t over-engineering for use on browsers, keep in
mind that filtering by file type is fairly limited and doesn’t support
regex:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept

If we wanted more complex behaviors, we’d have to build our own file
selection pop-up.

CodyCBakerPhD:
Yeah so I guess this is mostly for human readability then, in which case I assume regex is not super friendly

CodyCBakerPhD:
Yeah so I guess the fancier references are mostly for human readability then, in which case I assume full regex is not super friendly

However, it seems it does support the wildcard, and most people understand that at least

So how about `.imec*`, `.ns*` so it can be used in both readability and file filtering?

garrettmflynn:
From what I read, wildcards are only acceptable when specifying a mimetype (e.g. audio, video, image, etc.) that usually needs a subspecification.

If you want to confirm whether a certain pattern will work, I'd create a minimal HTML file with a single file element inside it and start putting different values into the accept attribute.

I'm AFK so can't test/send anything from my end at the moment.

tests/imports.py
```diff
@@ -23,69 +23,456 @@ def _strip_magic_module_attributes(ls: list) -> list:
     return list(filter(lambda key: key not in exclude_keys, ls))
 
 
-class TestImportStructure(TestCase):
```
CodyCBakerPhD, line 26: Not sure why I had this in a TestCase; seems unnecessary now

CodyCBakerPhD: Though it does amplify the +/- appearance simply because I'm unnesting it 🤷

tests/imports.py
```diff
@@ -23,69 +23,456 @@ def _strip_magic_module_attributes(ls: list) -> list:
...
+def test_guide_attributes():
+    from neuroconv.datainterfaces import interface_list
+
+    guide_attribute_names = ["display_name", "keywords", "associated_suffixes", "info"]
+    guide_attributes_per_interface = {
+        interface.__name__: {attribute: getattr(interface, attribute) for attribute in guide_attribute_names}
+        for interface in interface_list
+    }
+
+    assert guide_attributes_per_interface == {
```
CodyCBakerPhD, line 104: It's a bit structure to have in the test; I can maybe offload it to an external JSON file

But it sure it nice to be able to see all the results aggregated together

Some comments below about standardizing text

tests/imports.py
```diff
@@ -23,69 +23,456 @@ def _strip_magic_module_attributes(ls: list) -> list:
...
+        "NeuralynxRecordingInterface": {
+            "display_name": "Neuralynx Recording",
+            "keywords": ("extracellular electrophysiology", "voltage", "recording"),
+            "associated_suffixes": (),
+            "info": "Interface for Neuralynx recording data.",
```
CodyCBakerPhD, line 109: Some of these just say

"Interface for {format name} {modality data type} data."

It's something to show up on a hover icon, but is there anything more valuable or informative we could put here? As is I think many users could figure out what the selection in the form would be based on the display name

Others here say

"Interface for converting {format name} {modality data type} data." and may include the "to NWB." bit on the end

That certainly adds something

I got in the habit of swapping 'converting' to 'handling', I plan to swap back to converting if that's the one we want to go with.

tests/imports.py
```diff
@@ -23,69 +23,456 @@ def _strip_magic_module_attributes(ls: list) -> list:
...
+        },
+        "NeuralynxSortingInterface": {
+            "display_name": None,
```
CodyCBakerPhD, line 112: Still some TO-DO's

bendichter:
I just put these in as placeholders. We can edit them as necessary based on what we think would help users most.

bendichter:
I'd by fine with an auto-generated help for default

h-mayorquin:
> > I would vote against regex in the strings though.
> 
> How about semi-format strings of the kind `.imec{probe_number}`? It's actually fairly important for certain formats like NeuroScope Sorting (last suffix is numeric of group number) or Blackrock (stream type)

I think this would be better. To be more specific, I just thought that showing regex to end users is a bad idea.

garrettmflynn:
@CodyCBakerPhD Just to confirm the behavior on browsers, it seems to be even more constrained than expected.

For matching `.imec0.lf.bin`, for instance, the only `accept` value that will match correctly is `.bin`. So it seems that browsers take a very naive approach to the match, simply breaking off the canonical extension (i.e. the last period and text combination) and comparing it to what you provide.

So in many cases, we won't be able to do too much to help users programmatically until _after_ they've selected a file.

CodyCBakerPhD:
> In that case, exposing a regex to validate the selected file would be quite nice—maybe in a pattern attribute? 

The point is actually that we're _not_ validating because that might be too strict (not uncommon for people to modify things in a certain way, including suffixes, when it in reality can still be read exactly the same way)

Everything here is just trying to amplify convenience; is it possible for the file selector to match display of these two files `.imec0.lf.bin` + `.imec0.lf.meta` using the filter criteria `[".bin", ".meta"]`?

Otherwise the extra suffix patterns are less for the file selector and more for the user reading/seeing it on the interface selection screen and noting that their own files have similar values

garrettmflynn:
Ah thanks for the clarification.

> Everything here is just trying to amplify convenience; is it possible for the file selector to match display of these two files .imec0.lf.bin + .imec0.lf.meta using the filter criteria [".bin", ".meta"]?

Yes, though this leaves room for false positives, which isn't ideal but approaches the desired behavior.
Screenshot 2024-02-12 at 10 46 42 AM

Screenshot 2024-02-12 at 10 44 54 AM

garrettmflynn:
Also, note that users will still be able to drag-and-drop any file that they'd like. This is impossible to block without some post-input validation.

CodyCBakerPhD:
Oh, IDK if I would entirely call those false positives. It's correctly highlighting SpikeGLX format data and suppressing non-SpikeGLX format. For the SpikeGLX converter it's great to show all of those like that since you're selecting the top level folder path as input

I guess there might be a potential exception for a single-stream SpikeGLXRecording, since this filter shows the NIDQ files the user might be inclined to select, BUT I think that particular can (or is/should) raise some other error during `__init__` of the respective classes and otherwise combatted via documentation that the 'Recording' applies to the AP/LF streams only

> Also, note that users will still be able to drag-and-drop any file that they'd like.

Exactly; the user could also change the selection mode to '*' inside the selector popup itself, just like normal window behavior for that. We're just providing a suggested default to help make it clearer/easier to find the files of interest

tests/imports.py
CodyCBakerPhD: JSON doesn't preserve the typing in this case

CodyCBakerPhD: Could go with `pickle` I suppose but that defeats the purpose of human readability

tests/imports.py
CodyCBakerPhD: Moved to #740

tests/imports.py
CodyCBakerPhD: Not actually necessary any more. Might try it again in follow-up

CodyCBakerPhD:
Wound up having to remove the neo tests in the end; neo extensions do not disambiguate between recording, sorting, or behavior data streams, and we wouldn't want the recording interface to associate its suffixes with the spiking data of that format (if they are in different files that is)

Could have still included them, but for every single interface you'd then have to specify which neo extensions to filter out of the test, which is effectively just as much work as just manually asserting them anyway

CodyCBakerPhD:
@bendichter @h-mayorquin @weiglszonja @garrettmflynn 

This is ready for final review now

Sister branch on GUIDE https://github.com/NeurodataWithoutBorders/nwb-guide/pull/589 needs to be updated to latest state of this to get a preview of appearance

garrettmflynn:
The GUIDE is now updated for these updates :)

tests/test_minimal/test_tools/test_importing.py
```diff
@@ -0,0 +1,536 @@
...
+
+    # Blocking assertion when adding new interfaces
+    for name, summary in summaries.items():
+        for key, value in summary.items():
+            assert value is not None, f"{name} is missing GUIDE related attribute {key}."
+            if isinstance(value, tuple):
+                assert len(value) > 0, f"{name} is missing entries in GUIDE related attribute {key}."
+
+    # For easier reference to global commonalities
+    assert summaries == {
```
bendichter, line 18: I think I'd rather check a few keys here rather than the entire dict. As it stands, we'd have to add each new data interface to this test

bendichter:
Looks good, thanks for putting this together, Cody!

weiglszonja:
This is great! looks good to me.

tests/test_minimal/test_tools/test_importing.py
CodyCBakerPhD: @bendichter Would you be OK with at least having a static reference dumped to a JSON stored somewhere here (but not required for tests)?

It makes it a lot faster to reference at a glance whenever we have a question like 'why does this thing on the GUIDE look this way', instead of having to call code to generate it

CodyCBakerPhD:
Need approval to move forward on the GUIDE merge

weiglszonja (approved):

PR #732, opened 2024-02-10
[Bug] Fix configure_backend to account for data or timestamp links

CodyCBakerPhD:
It's possible for TimeSeries to link their `data` or `timestamps` fields to those of any other `TimeSeries`, thereby forming an internal link between the values to avoid having to duplicate the data

This was forgotten as a test case in the original PR for backend configuration.

- [x] TODO: add tests

PR #735, opened 2024-02-11
[Bug] Fix special case of buffer shape estimation

CodyCBakerPhD:
The new `configure_backend` tool is making more widespread usage of the `GenericDataChunkIterator` shape estimation methods than has been done previously

Caught this case on an ImageSeries with `maxshape=(3600, 304, 608)` and `buffer_gb = 0.5`

- [x] Need to add tests


CodyCBakerPhD:
Also upgraded buffer overflow warnings to errors (good idea in general) and casted some arrays as tuples prior to taking products, which was triggering the warnings to begin with

CodyCBakerPhD:
@h-mayorquin Added a test; this would be merged after its parent #732 (both were needed for debugging a conversion)

src/neuroconv/tools/hdmf.py
```diff
@@ -42,56 +48,55 @@ def estimate_default_chunk_shape(chunk_mb: float, maxshape: Tuple[int, ...], dty
...
                 # If the smallest full axis does not fit within the buffer size, form a square along the smallest axes
                 sub_square_buffer_shape = np.array(chunk_shape)
                 if min(axis_sizes_bytes) > target_buffer_bytes:
                     k1 = math.floor((target_buffer_bytes / chunk_bytes) ** 0.5)
                     for axis in [smallest_chunk_axis, second_smallest_chunk_axis]:
                         sub_square_buffer_shape[axis] = k1 * sub_square_buffer_shape[axis]
                     return tuple(sub_square_buffer_shape)
-            elif num_axes == 1:
-                smallest_chunk_axis = 0
+            elif num_axes == 1 and axis_sizes_bytes[0] > target_buffer_bytes:
```
CodyCBakerPhD, line 88:
```suggestion
            elif num_axes == 1:
```

Covered by line 65

src/neuroconv/tools/hdmf.py
```diff
@@ -1,11 +1,17 @@
 """Collection of modifications of HDMF functions that are to be tested/used on this repo until propagation upstream."""
 
 import math
+import warnings
 from typing import Tuple
 
 import numpy as np
 from hdmf.data_utils import GenericDataChunkIterator as HDMFGenericDataChunkIterator
 
+# Elevate any overflow warnings to trigger error, hopefully get full traceback too...
```
h-mayorquin, line 10: Shouldn't this be inside the class at it is `__init__`. I am afraid this will modify a very general behavior from a very implicit / hidden place.

Can we constrain this to work only in the context of the changes here somehow?

src/neuroconv/tools/hdmf.py
```diff
@@ -114,13 +119,14 @@ def estimate_default_buffer_shape(
                 k3 = math.floor(target_buffer_bytes / buffer_bytes)
                 padded_buffer_shape[axis] *= k3
                 break
+        padded_buffer_shape = tuple(int(value) for value in padded_buffer_shape)  # To avoid overflow from math.prod
```
h-mayorquin, line 122: Does this comment corresopnding to casting the elements to int?

src/neuroconv/tools/hdmf.py
CodyCBakerPhD: The `__init__` might not be the right place

Moved it inside the relevant local scopes though

src/neuroconv/tools/hdmf.py
CodyCBakerPhD: It corresponds to converting the numpy array of numpy ints to a tuple of python ints. Otherwise it's another potential source of the overflow

h-mayorquin: Got you I guess we could do it from line 109 already?

CodyCBakerPhD: Needs to be an array for line 125

Doesn't need to be an array on exiting the conditions after 126

h-mayorquin: Got you, I looked for numpy operations and did not see that implicit one : )

PR #736, opened 2024-02-12
Update spikeinterface to the last version

h-mayorquin:
Update the comparison function to take into account the latest changes in the NWBRecordingExtractor in Spikeinterface.

### DIFF
--- a/src/neuroconv/tools/testing/data_interface_mixins.py
+++ b/src/neuroconv/tools/testing/data_interface_mixins.py
@@ -286,26 +286,35 @@ class RecordingExtractorInterfaceTestMixin(DataInterfaceTestMixin, TemporalAlign
         electrical_series_name = self.interface.get_metadata()["Ecephys"][self.interface.es_key]["name"]
 
         if recording.get_num_segments() == 1:
+
             # Spikeinterface behavior is to load the electrode table channel_name property as a channel_id
             self.nwb_recording = NwbRecordingExtractor(
                 file_path=nwbfile_path, electrical_series_name=electrical_series_name
             )
-            if "channel_name" in recording.get_property_keys():
-                renamed_channel_ids = recording.get_property("channel_name")
+
+            ## Set channel_ids right for comparison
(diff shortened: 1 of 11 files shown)


CodyCBakerPhD:
Looks like this cleans up all the SI related issues

Can you add a CHANGELOG describing what to expect for the new behavior? I get they're all casted as strings now, but what was the previous behavior? (I thought they would always be strings in NWB file in the end)

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -314,6 +323,20 @@ def check_read_nwb(self, nwbfile_path: str):
                 if recording.has_scaled_traces() and self.nwb_recording.has_scaled_traces():
                     check_recordings_equal(RX1=recording, RX2=self.nwb_recording, return_scaled=True)
 
+            ## Compare channel groups
```
h-mayorquin, line 326: @CodyCBakerPhD and added here to

h-mayorquin:
> Looks like this cleans up all the SI related issues
> 
> Can you add a CHANGELOG describing what to expect for the new behavior? I get they're all casted as strings now, but what was the previous behavior? (I thought they would always be strings in NWB file in the end)

Added. `NWBRecordingExtractor` always is using `group_name` to set the group property. Previously, we were comparing `group` (in the original recording) to `group` (in the `NWBRecordingExtractor`). I added logic to compare `group name` to group when this exists.

Basically,  I am reflecitng the writing logic here:
https://github.com/catalystneuro/neuroconv/blob/374ad9d89f15a1be1407134975e85e3ecb10b8dc/src/neuroconv/tools/spikeinterface/spikeinterface.py#L279-L287

On the comparison testing here:
https://github.com/catalystneuro/neuroconv/blob/374ad9d89f15a1be1407134975e85e3ecb10b8dc/src/neuroconv/tools/testing/data_interface_mixins.py#L325-L335

CodyCBakerPhD (approved):

PR #747, opened 2024-02-15
fix call signatures that are broken

bendichter:


src/neuroconv/datainterfaces/behavior/miniscope/miniscopedatainterface.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, folder_path: FolderPathType):
         miniscope_config_files = natsort.natsorted(list(folder_path.glob(f"*/BehavCam*/{configuration_file_name}")))
         assert (
             miniscope_config_files
-        ), f"The configuration files ({configuration_file_name} files) are missing from '{self.folder_path}'."
+        ), f"The configuration files ({configuration_file_name} files) are missing from '{folder_path}'."
```
CodyCBakerPhD, line 47: `folder_path` is defined at [line 37](https://github.com/catalystneuro/neuroconv/pull/747/files#diff-8e6b9e66669cdffffa2cfebeef3f6c9df638af7fb60de11b84d5c8920d9cdf73L37), right?

bendichter: yes, `folder_path` is defined on line 37. `self.folder_path` is never defined, though.

CodyCBakerPhD: Oh I see now

CodyCBakerPhD (approved):

PR #743, opened 2024-02-14
Updated make_environment file to match tested python versions

pauladkisson:
Fixes #742 by specifying only the python versions that are actually tested (3.8-3.11).  Before, make_environment would use python 3.12 which led to a bug.

PR #753, opened 2024-02-19
Remove conda env file and improve README

### LINKED ISSUE #742: [Bug]: make_environment.yml leads to installation error
### What happened?

I created the `neuroconv_environment` via the `make_environment.yml` file and then installed neuroconv (full and test) and got an installation error.

Note the python version is 3.12.1 NOT 3.7 (there was no option on the dropdown)

### Steps to Reproduce

```python
conda env create -f make_environment.yml
conda activate neuroconv_environment
pip install -e .'[full,test]'
```

### Traceback

```python
× Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [33 lines of output]
        File "/opt/anaconda3/envs/neuroconv_environment/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/opt/anaconda3/envs/neuroconv_environment/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/opt/anaconda3/envs/neuroconv_environment/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 112, in get_requires_for_build_wheel
          backend = _build_backend()
                    ^^^^^^^^^^^^^^^^
        File "/opt/anaconda3/envs/neuroconv_environment/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 77, in _build_backend
          obj = import_module(mod_path)
                ^^^^^^^^^^^^^^^^^^^^^^^
        File "/opt/anaconda3/envs/neuroconv_environment/lib/python3.12/importlib/__init__.py", line 90, in import_module
          return _bootstrap._gcd_import(name[level:], package, level)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
        File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
        File "<frozen importlib._bootstrap>", line 1310, in _find_and_load_unlocked
        File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
        File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
        File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
        File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
        File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
        File "<frozen importlib._bootstrap_external>", line 994, in exec_module
        File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
        File "/private/var/folders/s3/qb42pmpn5jd0xb6fm2gt65qm0000gn/T/pip-build-env-tus00p4_/overlay/lib/python3.12/site-packages/setuptools/__init__.py", line 16, in <module>
          import setuptools.version
        File "/private/var/folders/s3/qb42pmpn5jd0xb6fm2gt65qm0000gn/T/pip-build-env-tus00p4_/overlay/lib/python3.12/site-packages/setuptools/version.py", line 1, in <module>
          import pkg_resources
        File "/private/var/folders/s3/qb42pmpn5jd0xb6fm2gt65qm0000gn/T/pip-build-env-tus00p4_/overlay/lib/python3.12/site-packages/pkg_resources/__init__.py", line 2172, in <module>
          register_finder(pkgutil.ImpImporter, find_on_path)
                          ^^^^^^^^^^^^^^^^^^^
      AttributeError: module 'pkgutil' has no attribute 'ImpImporter'. Did you mean: 'zipimporter'?
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.7

### Package Versions

_No response_

### Code of Conduct

CodyCBakerPhD:
Closes #743
Fixes #742


pauladkisson (changes requested):
Do you think we should mention here which versions we support? That way a new user wouldn't have to guess which version to select for their conda environment.

CodyCBakerPhD:
> Do you think we should mention here which versions we support?

If you can think of a way to do that in text without needing to update it over time (which we're inevitably bound to forget if its decentralized) I'm all for it

Otherwise, the attempt to `pip install neuroconv` (without specific version) will always route them to the latest/last release for each Python version they might realistically select (or warn that it is not available otherwise)

README.md
```diff
@@ -36,26 +36,44 @@ Features:
...
+```
+
+To install the latest stable release of **neuroconv** though PyPI, run:
+
 ```shell
 pip install neuroconv
 ```
 
-For more flexibility we recommend installing the latest version directly from GitHub. The following commands create an environment with all the required dependencies and the latest updates:
+To install the current unreleased `main` branch (requires `git` to be installed in your environment, such was via `conda install git`), run:
```
h-mayorquin, line 52: I think it would be better to point them to the official instalation instructions of the git project:
https://git-scm.com/downloads

CodyCBakerPhD: IDK, that tends to do different things than you would expect depending on the platform. Suggested added both just in case

On windows for instance, it's a standalone interpreter with it's own console (doesn't play well with standard command prompt)

You can technically add it to the system registry but behaves best with conda when installed as a separate package

README.md
```diff
@@ -36,26 +36,44 @@ Features:
...
+
+```shell
+pip install git+https://github.com/catalystneuro/neuroconv.git@main
+```
+
+To install the package in [editable mode](https://pip.pypa.io/en/stable/cli/pip_install/#editable-installs), run:
```
h-mayorquin, line 58: Just for your information.

For another PR as you are just refactoring but I think we probably should remove mentions of editable installs on general readme documents and send them over to the developer sections. 

With the migration away from setup tools editable installs are less and less consistent I jumped into this rabbit hole last week. Just two point out two common problems:

* Poetry which is one of the most pouplar npm-like experiences for python does not suppport editable installs.
* static analysis tools would not be able to use the enabled editable installs because they use import hooks:

https://github.com/microsoft/pylance-release/issues/3473

I can take care of this later.

README.md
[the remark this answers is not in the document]
h-mayorquin: I rather just ommit mentioning conda here but this works.

h-mayorquin:
This is OK to me. If no one else is using these files I will trust @CodyCBakerPhD gut that the mainentance is not worth the convenience.

README.md
h-mayorquin: Ah I see. Ok, I did not see this before writing my comment below. Ok, so in your experience installing conda with git is the easier for windows users and hence you are making that suggestion. 

The current two-fold solution seems fine but we could also explicitly mention that we suggest git on windows. As you feel is better.

README.md
CodyCBakerPhD: Removed editable install instructions; I believe the idea of this README duplicating our user guide in the docs is to reduce redirection for people just finding the package on PyPI or main GitHub page. So just quickest and dirtiest instructions to get going. Otherwise I'm in greater favor of going the GUIDE README approach and offloading everything of detail to the main docs

h-mayorquin: Good. Not sure if they should be removed from the cookie cutter for the conversion projects as well. Here we expect users, there we expect people to change stuff? leaning to remove it as well.

CodyCBakerPhD: Conversion projects depend on the client. Some want/need to tamper with stuff, so yes to editable there

Others want static PyPI release with CLI tooling; what we do most with them with respect to bug fixes is do the `git` based `pip` install of a particular branch or commit

pauladkisson (approved):

CodyCBakerPhD:
> Some people (ex. [python-gitlab](https://github.com/python-gitlab/python-gitlab)) look like they have pypi keep track of it and then link a badge, but tbh it seems like a bit more work than its worth in this case.

Oooo that is a nice looking actually

PR #754, opened 2024-02-21
Try new badges

CodyCBakerPhD:
From Paul in #753


pauladkisson:
Badge just says python: missing.

maybe adjust some setting in pypi?

CodyCBakerPhD:
There's not many things customizable directly on PyPI - I believe it needs the classifiers in the setup when built, so this will show missing until next release, which I'll look into cutting tomorrow now that we have those bug fixes to configuration and buffering through

h-mayorquin:
Cool!

pauladkisson (approved):

h-mayorquin:
I guess we are only missing python 3.12 : )

PR #755, opened 2024-02-21
Use relative paths instead of absolute for CI actions

h-mayorquin:
From the discussion in #626. 

Pushed here to streamline other CI changes that might be necessary for #626.

For future reference:
https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,4 +86,12 @@ setup(
         ],
     },
     license="BSD-3-Clause",
+    classifiers=[
+        "Framework :: NeuroConv",
+        "Intended Audience :: Science/Research",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3.11",
+    ],
 )


.github/workflows/dailies.yml
```diff
@@ -31,4 +31,4 @@ jobs:
         subject: NeuroConv Daily Failure
         to: ${{ secrets.DAILY_FAILURE_EMAIL_LIST }}  # add more with commas, no separation
         from: NeuroConv
-        body: "The daily workflow for the NWB Inspector failed: please check status at https://github.com/catalystneuro/neuroconv/actions/workflows/dailies.yml"
+        body: "The daily workflow for the NWB Inspector failed: please check status at https://github.com/./actions/workflows/dailies.yml"
```
CodyCBakerPhD, line 34: ? Won't this break my link

h-mayorquin: Yes, I was too aggressive with refex. Let me correct this.

CodyCBakerPhD (approved):

PR #758, opened 2024-02-21
Add ignoring linkcheck 

h-mayorquin:
Buzsaki lab webpage has an SSL certificate expired and therefore the action is failing:

https://buzsakilab.com/wp/


CodyCBakerPhD:
Stackoverflow links have also been breaking everywhere lately

Was going to just keep an eye on it for a while and hope it resolves

The point of the checks is so that we don't have broken links in our docs; setting it to ignore will make the check mark green but now we have a broken link in the docs

So instead of ignoring, I'd say either remove the link from the docs if it bothers you that much or give it some time and hope for the best (external links check is never blocking)

h-mayorquin:
The link is not broken. It just does not have the SSL certificate. Anyway, I also realized that this action is failing at many other levels. 

That said, I think that if we get accustomed to see the action in red and just ignore it, then what purpose is the action failing? It should not give a lot of false positives otherwise is just gona end up like the kid who cried wolf.

If you trust your pulse to monitor it often so the latter problem does not affect you, maybe it should just be a cron job where you can periodically can check for them and correct them.

CodyCBakerPhD:
> The link is not broken. It just does not have the SSL certificate.

You can't get to it without jumping through hoops on standard browsers, so is there really a practical difference? Someone clicks on that through our docs its gonna leave a bad impression

> That said, I think that if we get accustomed to see the action in red and just ignore it

I don't ignore it, I keep an eye on it for some number of days until deciding how to proceed. We're at that point with the stackoverflow links, should we just remove them? Those ones are weird because I can actually follow the link just fine, it's something about how the sphinx tester performs it (I think when I looked into it might be possible that the GitHub action IP was banned from making requests to their site), but if we set it to ignore and then they break in the future we wouldn't know

Otherwise grateful if you'd be able to figure that one out

> If you trust your pulse to monitor it often so the latter problem does not affect you, maybe it should just be a cron job where you can periodically can check for them and correct them.

Which it also is, but people also add links on new PRs, so shouldn't we check those links are valid on the PRs? It's not an expensive test either way

h-mayorquin:
> > The link is not broken. It just does not have the SSL certificate.
> 
> You can't get to it without jumping through hoops on standard browsers, so is there really a practical difference? Someone clicks on that through our docs its gonna leave a bad impression

I don't see in that way. I don't get a bad impression on us at allf rom this. It is just that Buzsaki has not updated their web page on a while. That's it.

> > That said, I think that if we get accustomed to see the action in red and just ignore it
> 
> I don't ignore it, I keep an eye on it for some number of days until deciding how to proceed. We're at that point with the stackoverflow links, should we just remove them? Those ones are weird because I can actually follow the link just fine, it's something about how the sphinx tester performs it (I think when I looked into it might be possible that the GitHub action IP was banned from making requests to their site), but if we set it to ignore and then they break in the future we wouldn't know
> 
> Otherwise grateful if you'd be able to figure that one out
> 
> > If you trust your pulse to monitor it often so the latter problem does not affect you, maybe it should just be a cron job where you can periodically can check for them and correct them.
> 
> Which it also is, but people also add links on new PRs, so shouldn't we check those links are valid on the PRs? It's not an expensive test either way

Yeah, I am arguing that the workflow that you follow is better suited for a cronjob. We can use a cronjob where the ignorelinks are... well, ignored. And then, in the main job here we will ignore the jobs and only check for the new ones that are added.

Something like (pseudocode):

```
if running_cronjob:
            sphinx-build -b linkcheck -D linkcheck_ignore=[] ./docs ./docs/_build/
else:
                 sphinx-build -b linkcheck ./docs ./docs/_build/

```

CodyCBakerPhD:
Sure, two separate workflows would be fine, one with ignore on the CLI invocation and another without

Two separate email notification headers would be cool too

h-mayorquin:
> sphinx-build -b linkcheck -D linkcheck_ignore=[] ./docs ./docs/_build/

All right, let's move forward this way.

h-mayorquin:
@CodyCBakerPhD 
OK, this should be working and I saw that the cronjob is failing (which is the correct and expected behavior). I think this is ready to go on my side.

.github/workflows/test-external-links-cron-job.yml
```diff
@@ -0,0 +1,23 @@
...
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - run: git fetch --prune --unshallow --tags
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+      - name: Install package for API docs
+        run: pip install -e .[docs]
```
h-mayorquin, line 21:
```suggestion
        run: pip install .[docs]
```

.github/workflows/test-external-links-cron-job.yml
```diff
@@ -0,0 +1,25 @@
+name: Testing External Links Cron Job
+
+on:
```
CodyCBakerPhD, line 3:
```suggestion
on:
  workflow_dispatch:
```

CodyCBakerPhD: It's nice to have a manual trigger IMO

CodyCBakerPhD (approved):

.github/workflows/test-external-links-cron-job.yml
h-mayorquin: Agree!

.github/workflows/test-external-links-cron-job.yml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+      - name: Install package for API docs
+        run: pip install .[docs]
+      - name: Test External Links
+      # This command runs Sphinx's linkcheck with a configuration override to ensure no links are ignored during the check.
+      # The -D option here explicitly sets linkcheck_ignore to an empty list, overriding any ignore patterns set in conf.py.
+        run: sphinx-build -b linkcheck -D linkcheck_ignore=[] ./docs ./docs/_build/
```
CodyCBakerPhD, line 26: Wait, actually, could you add a final step here to do the email notification? And give it a email header about neuroconv external links?

CodyCBakerPhD (changes requested):

.github/workflows/test-external-links-cron-job.yml
h-mayorquin: I have no idea how to do that.

h-mayorquin: I actually would like to not receive emails myself.

h-mayorquin: I want the action to be useful for avoiding adding new broken links. That function is less likely to be performed if the action is failing all the time as it is now.  

At the same time, you want to keep the action that is failing all the time to see if at some point it will stop failing. 

I am having two actions for these two different purposes.

CodyCBakerPhD: > I want the action to be useful for avoiding adding new broken links.

PR reviewers should also ensure links work as expected, so it's not like the CI should exclusively be relied on for that. It's just a nicety when it works

> I am having two actions for these two different purposes.

Except the burden of maintaining the 'ignore' list to achieve that 'passing' status is yet another thing I need to deal with now, instead of just resolving the source problematic links. I'd only be willing to add this layer if I can get help maintaining it

h-mayorquin: > PR reviewers should also ensure links work as expected, so it's not like the CI should exclusively be relied on for that. It's just a nicety when it works

If we can make a better classifier to help us I don't see why we shouldn't. I classifier with less false positives is better.

>Except the burden of maintaining the 'ignore' list to achieve that 'passing' status is yet another thing I need to deal with now, instead of just resolving the source problematic links. I'd only be willing to add this layer if I can get help maintaining it

I am fine adding links that start failing to the list myself.

>  That happens a lot more often than you think lol

Could be. I don't think the stackoverflow ones can or will be solved. But in any case, having a link that is working on the list does not have any downside.

h-mayorquin: But do we agree that this notification will be added to the action that runs in the PRs right?  The one that uses the ignore list.

If so, that makes sense to me and I will add it.

CodyCBakerPhD: The other one, the daily (or coarser CRON) that checks current links on `main`. We would not want notifications from pushes to PRs lol

h-mayorquin: No, then we disagree. I think that that signal is not useful. And is not related to what I volunteer for: adding links to the list that ar failing.

I think getting a notification that the links are failing everyday is not a useful signal. Why do you want that?

CodyCBakerPhD: The entire origin of these CI checks is to ensure that links on `main` (in our public docs) do not go out of date or experience other disruptions (and yes, generally, SSL certificate issues would fall under that category, since it could indicate any number of problems or reasons why we wouldn't want to send users to that website through our own)

Links are a live service; their health must be checked periodically.

A PR-based workflow helps inform users that a link being added to a new PR needs attention; no one needs to be notified about that besides the people writing and reviewing the PR in question.

A daily-based workflow ensure the links stays good after merging, that's the maintenance cost to that

Your suggestion to make the PR-based workflow more effective by adding a specific ignore list adds a cost of maintenance (for items like the Buzsaki link; the alternative on stackoverflow is also to just remove the link entirely to make it a non-issue) since it should be periodically checked to see if any can be taken off of ignore, or likewise on regular daily failures if something needs to be either fixed or added to ignore if no other options. That's what I'm saying I would want help with in that case

h-mayorquin: To satisfy your critieria here:

> A daily-based workflow ensure the links stays good after merging, that's the maintenance cost to that

I see two options to move forward:
* We make the action with the ignore list also a cron job and we add the notification to that.
* We just remove the Buzsaki links and the stackoverflow ones and we leave the current action as it is.

The things that I don't find value on having are:
1) An action that is always failing in the PR. Makes the work of non-adding broken links harder.
2) A daily action that is always failing and give us notification on that.

I though that you were interested on having the action without ignore list because you want to check if at some point the links start working but maybe I am wrong.

What do you think?

h-mayorquin: OK, I prefer number 1. 

I think it will be more flexible long term. 

Buzsaki might be fixed, I don't think Stackoverflow will be fixed as I don't think they want bots accessing their links after the AI developments of the last year.

Do we have something now that is doing this?
> A daily-based workflow ensure the links stays good after merging, that's the maintenance cost to that

h-mayorquin: Ah, OK, the current configuration is exactly like:

https://github.com/catalystneuro/neuroconv/blob/a4f9b7581eb58301f7f868dc1c101b454e0c878f/.github/workflows/dailies.yml#L1-L37

That is, the dallies will call the current check for the links and test it as I proposed above. I also added the codecov token in the dallies which is causing a failure in the CI as of  now.

h-mayorquin:
OK, @CodyCBakerPhD, this is the current state of affairs:
1) The current `test-external-links.yml` has the ignore list. This means that in the PRs, the action  does not report failures for links that we already know are broken (e.g. stackoverflow stuff). Making easier to detect if a newly added link is broken.
2) Through the dallies, the current `test-external-links.yml` is called daily and will inform us if one of the already existing links is broken.
3) The nofication for the broken link is separate from the notification of the tests failures. You can add my name on the notification from the broken links and I will take care of that.

Does this addresses all of your concerns?

.github/workflows/dailies.yml
```diff
@@ -13,22 +13,39 @@ jobs:
...
+    if: ${{ needs.run-daily-doc-link-checks.result == 'failure' }}
+    steps:
+    - uses: dawidd6/action-send-mail@v3
+      with:
+        server_address: smtp.gmail.com
+        server_port: 465
+        username: ${{ secrets.MAIL_USERNAME }}
+        password: ${{ secrets.MAIL_PASSWORD }}
+        subject: NeuroConv Daily Doc Link Check Failure
+        to: ${{ secrets.DAILY_FAILURE_EMAIL_LIST }}
```
h-mayorquin, line 49: You can add a new secret with my email there.

.github/workflows/dailies.yml
```diff
@@ -13,22 +13,39 @@ jobs:
...
-        server_port: 465  # TSL
+        server_port: 465
         username: ${{ secrets.MAIL_USERNAME }}
         password: ${{ secrets.MAIL_PASSWORD }}
-        subject: NeuroConv Daily Failure
-        to: ${{ secrets.DAILY_FAILURE_EMAIL_LIST }}  # add more with commas, no separation
+        subject: NeuroConv Daily Test Failure
+        to: ${{ secrets.DAILY_FAILURE_EMAIL_LIST }}
         from: NeuroConv
-        body: "The daily workflow for the NWB Inspector failed: please check status at https://github.com/catalystneuro/neuroconv/actions/workflows/dailies.yml"
```
CodyCBakerPhD, line 34: Hah

.github/workflows/dailies.yml
CodyCBakerPhD: Would you like to be on both or just the docs one?

.github/workflows/dailies.yml
```diff
@@ -13,22 +13,39 @@ jobs:
...
   run-daily-doc-link-checks:
     uses: ./.github/workflows/test-external-links.yml
 
-  notify:
+  notify-test-failure:
     runs-on: ubuntu-latest
-    needs: [run-daily-tests, run-daily-doc-link-checks]
-    if: ${{ always() && contains(needs.*.result, 'failure') }}
+    needs: [run-daily-tests]
+    if: ${{ needs.run-daily-tests.result == 'failure' }}
```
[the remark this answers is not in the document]
h-mayorquin: We should leave it then. Probably not worth diggin for clarification if this has worked so far.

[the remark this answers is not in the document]
h-mayorquin: Restored : )

h-mayorquin:
> > Does this addresses all of your concerns?
> 
> It does! Thanks for joining me to share the load

All out of my love for you and my OCD hate for red buttons : D

CodyCBakerPhD (approved):

CodyCBakerPhD:
Might also need the `contains(needs.*.result, 'failure') `, I forget why. We'll find out

.github/workflows/dailies.yml
h-mayorquin: Add me on both, that's fine. But I will prioritze the documentation ones as that is what I promised. 

The other one will probably be useful for spikeinterface failures though.

h-mayorquin:
> Might also need the `contains(needs.*.result, 'failure') `, I forget why. We'll find out

I thought that was only for checking that any of the two jobs that were in need fail. As in, to use the regular expression with the wildcard `*` to avoid writing two instructions instead?

CodyCBakerPhD:
> I thought that was only for checking that any of the two jobs that were in need fail. As in, to use the regular expression with the wildcard * to avoid writing two instructions instead?

Hmm OK I see, the `contains` is with respect to the expected list output, but is still an equality check in that regard

PR #759, opened 2024-02-21
Correct path for `codecov.yml`

h-mayorquin:
The current argument in the CI is non-existent:
https://github.com/codecov/codecov-action?tab=readme-ov-file#arguments

I am redirecting the file to the correct path and add verbosity ofr failure debuggin.


CodyCBakerPhD:
Guessing that happened last time versions were bumped?

.github/workflows/testing.yml
```diff
@@ -139,4 +139,5 @@ jobs:
           file: ./codecov.xml
           flags: unittests
           name: codecov-umbrella
-          yml: ./codecov.yml
+          codecov_yml_path: ./codecov.yml
+          verbose: true
```
CodyCBakerPhD, line 143:
```suggestion
          fail_ci_if_error: true
          verbose: true
```

I'd also recommend this one

h-mayorquin: I wanted to do it but did not want to have that discussion right. Great!

h-mayorquin: Note that this will probably make the CI fail as it seems that tokens are not available on branches":

https://github.com/catalystneuro/neuroconv/actions/runs/7994541423/job/21832860179?pr=756

But I think we should solve this.

h-mayorquin: Right this is on-going:

https://github.com/catalystneuro/neuroconv/actions/runs/7994978795/job/21834447378?pr=759

CodyCBakerPhD: It's not about it being a separate branch, it's about it not being passed from deployment

Though the real problem is that without that CI fail flag codecov never complains about such a major issue

CodyCBakerPhD (approved):

.github/workflows/testing.yml
h-mayorquin: What do you mean not being passed from deployment? 
It seems that is working on the CI actions that run from main.

CodyCBakerPhD: > It seems that is working on the CI actions that run from main.

What makes you think that

.github/workflows/deploy-tests.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       S3_GIN_BUCKET: ${{ secrets.S3_GIN_BUCKET }}
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
h-mayorquin, line 38: @CodyCBakerPhD ah, got you.

.github/workflows/testing.yml
h-mayorquin: I looked a the actions from main in the list and I can see that is being uploaded there:

https://github.com/catalystneuro/neuroconv/actions/runs/7976494254/job/21777198212

CodyCBakerPhD: Mmmmm, yes

Because that was the daily cron trigger. Thankfully that never broke

Because it triggers from https://github.com/catalystneuro/neuroconv/blob/main/.github/workflows/testing.yml#L3 it doesn't need the CODECOV token passed

So I suppose in a roundabout way it is an issue affecting branches because they rely exclusively on the workflow call

But if I had, for instance, triggered a manual dispatch from `main` the same problem might have occured

h-mayorquin: Ah, got you, it is because it is called from from the `deploy-test.yml` and that does not propagate secrets.

PR #760, opened 2024-02-21
Remove conda for most of the ci

h-mayorquin:
The only place were we required it now is to upload the data to S3 when that fails. I am confining the use of the action there as the CI is giving deprecations warnings for this actions already.

Let's see if my assumption of this not being used is correct.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -63,6 +63,9 @@ body:
         - 3.7
         - 3.8
         - 3.9
+        - 3.10
+        - 3.11
+        - 3.12
     validations:
       required: true
   - type: textarea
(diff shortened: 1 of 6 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -63,6 +63,9 @@ body:
         - 3.7
         - 3.8
         - 3.9
+        - 3.10
+        - 3.11
+        - 3.12
```
CodyCBakerPhD, line 68:
```suggestion
        - 3.8
        - 3.9
        - 3.10
        - 3.11
```

Then 3.12 in the respective branch

We could also change this to an optional write-in box so we don't have to keep the options up to date

h-mayorquin: That would be better. Let's do that in another PR. though.

CodyCBakerPhD (approved):

PR #767, opened 2024-03-02
Upgrade to Pydantic 2

### LINKED ISSUE #662: [Reminder]: When Python >= 3.11, using typing.Self

Starting from https://github.com/catalystneuro/neuroconv/pull/659, any time we use annotation return types that would use the modern [`typing.Self`](https://peps.python.org/pep-0673/) available in Python 3.11, we currently use `-> "NameOfClass"` instead

This is just a reminder to do the proper thing once our lower bound reaches that point in a couple of years

_No response_

Yes.

### Code of Conduct

CodyCBakerPhD:
DANDI just finished the upgrade to Pydantic 2, so we need to keep up for our live service automatic upload + GUIDE bundling

Also fixes #662 by using `typing_extensions.Self`


requirements-minimal.txt
```diff
@@ -4,10 +4,11 @@ jsonschema>=3.2.0
 PyYAML>=5.4
 scipy>=1.4.1
 h5py>=3.9.0
-hdmf>=3.12.1
+hdmf>=3.12.2
```
CodyCBakerPhD, line 7: Required to avoid the deprecation warning on passing `**kwargs` instead of `data_io_kwargs`

src/neuroconv/tools/nwb_helpers/__init__.py
```diff
@@ -37,21 +37,20 @@
     "BACKEND_CONFIGURATIONS",
     "DATASET_IO_CONFIGURATIONS",
     "BACKEND_NWB_IO",
+    "BackendConfiguration",
```
CodyCBakerPhD, line 40: Reordered to look slightly better on certain autocompleters

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -13,9 +14,11 @@
 class BackendConfiguration(BaseModel):
     """A model for matching collections of DatasetConfigurations to a specific backend."""
 
-    backend: ClassVar[Literal["hdf5", "zarr"]] = Field(
-        description="The name of the backend used to configure the NWBFile."
-    )
+    backend: ClassVar[Literal["hdf5", "zarr"]]
```
CodyCBakerPhD, line 17: So, we had talked about whether or not this should be a property of the model instead of a class attribute

However, especially after the `from_*` class constructors, our treatment of these fields become solidified as class attributes, which feels right to me

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -31,11 +39,20 @@ def _find_location_in_memory_nwbfile(current_location: str, neurodata_object: Co
             if isinstance(parent_field_value, dict) and neurodata_object.name in parent_field_value:
                 return parent_field_name + "/" + neurodata_object.name + "/" + current_location
         return neurodata_object.name + "/" + current_location
-    return _find_location_in_memory_nwbfile(
+    return _recursively_find_location_in_memory_nwbfile(
         current_location=neurodata_object.name + "/" + current_location, neurodata_object=parent
     )
 
 
+def _find_location_in_memory_nwbfile(neurodata_object: Container, field_name: str) -> str:
```
CodyCBakerPhD, line 47: Just a minor readability change to the private methods; otherwise below there was a line that read as `_find_location_in_memory_nwbfile(current_location=field_name)` which looked weird until you followed the redirection and implicit logic enough to realize how the location is constructed recursively

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -67,105 +84,55 @@ def _infer_dtype(dataset: Union[h5py.Dataset, zarr.Array]) -> np.dtype:
     return data_type
 
 
-class DatasetInfo(BaseModel):
```
CodyCBakerPhD, line 70: This is the biggest change as per our previous discussions on the matter

Prior to Pydantic 2 the entire model was either mutable or immutable; post Pydantic 2 individual fields can be frozen

So instead of having an entire nested class inside the other models (the `DatasetIOConfiguration`s) these values are now unpacked and attributed directly to the dataset models

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -67,105 +84,55 @@ def _infer_dtype(dataset: Union[h5py.Dataset, zarr.Array]) -> np.dtype:
...
+    compression_method: Union[
+        str, InstanceOf[h5py._hl.filters.FilterRefBase], InstanceOf[numcodecs.abc.Codec], None
+    ] = Field(
         description="The specified compression method to apply to this dataset. Set to `None` to disable compression.",
     )
     compression_options: Union[Dict[str, Any], None] = Field(
         default=None, description="The optional parameters to use for the specified compression method."
     )
 
+    @abstractmethod
```
CodyCBakerPhD, line 127: Minor style on ordering of methods; I feel like an abstract method ought to be as close as possible to the top of the class definition since it will need to be defined on any child classes

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -204,12 +171,18 @@ def __str__(self) -> str:
 
         return string
 
-    @root_validator
+    @model_validator(mode="before")
```
CodyCBakerPhD, line 174: `model_validator(mode="before")` is equivalent to the old `root_validator` in that it checks the key/value pairs passed to `__init__`, but technically prior to actually calling the `__init__` on those values

src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
```diff
@@ -114,7 +114,7 @@ def get_default_dataset_io_configurations(
                     continue
 
                 dataset_io_configuration = DatasetIOConfigurationClass.from_neurodata_object(
-                    neurodata_object=column, field_name="data"
+                    neurodata_object=column, dataset_name="data"
```
CodyCBakerPhD, line 117: Minor style change, mentioned in the CHANGELOG

Since it doesn't affect the recommended, or common way of using these brand-new features I'm bypassing the need for a deprecation alert

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -67,105 +84,55 @@ def _infer_dtype(dataset: Union[h5py.Dataset, zarr.Array]) -> np.dtype:
...
         description=(
             "The specified shape to use when iteratively loading data into memory while writing the dataset. "
             "For optimized writing speeds and minimal RAM usage, a total size of around 1 GB is recommended."
-        )
+        ),
     )
-    # TODO: When using Pydantic v2, wrap h5py._hl.filters.FilterRefBase and numcodecs.abc.Codec with InstanceOf
-    compression_method: Union[str, h5py._hl.filters.FilterRefBase, numcodecs.abc.Codec, None] = Field(
-        default="gzip",
+    compression_method: Union[
```
h-mayorquin, line 118: I wish we could use | instead of Union : P

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
h-mayorquin: Thanks for writting an explanation.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
h-mayorquin: OK.

src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
h-mayorquin: I think it makes sense.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -13,9 +14,11 @@
...
+    data_io_class: ClassVar[Type[DataIO]]
+
+    model_config = ConfigDict(validate_assignment=True)  # Re-validate model on mutation
+
     data_io_class: Type[DataIO] = Field(description="The DataIO class that is specific to this backend.")
```
h-mayorquin, line 22: Isn't `data_io_class:` repeated here (lines 22 and 18)?

h-mayorquin (commented):
This seems OK to me. 

It is flattening the structure that we use to write the backend plus some minor changes that I agree with: field for dataset_name, creating another function for finding the location which has better signature semantics and making some properties class variables.

The most difficult thing is to read carefully through the test as the flattening of the structure implied that they have to change.

I also think that the diff in the tests resulted a bit larger than it had to be because we are testing string representations. 

I will re-read the tests tomorrow just as a sanity check for my eyes and approve after as I think this is ready to go.

requirements-minimal.txt
h-mayorquin: OK

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
CodyCBakerPhD: I think an NWBFile can have something that's a `Container` but not an `NWBContainer`? Maybe, maybe not - the typing and logic should apply to the most generic case however

requirements-minimal.txt
```diff
@@ -4,10 +4,11 @@ jsonschema>=3.2.0
...
 hdmf_zarr>=0.4.0
 pynwb>=2.3.2;python_version>='3.8'
-pydantic>=1.10.13,<2.0.0
+pydantic>=2.0.0
+typing_extensions>=4.1.0
```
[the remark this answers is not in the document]
CodyCBakerPhD: I actually don't know - would be good to try in #671 or whatever other follow-up truly fixes those typing issues

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
CodyCBakerPhD: Soon

Soon...

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
CodyCBakerPhD: Whoops, good catch

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_common_dataset_io_configuration_model.py
```diff
@@ -9,183 +9,131 @@
...
 ):
     """Any divisibility is allowed when the buffer shape is capped at the full length of an axis."""
-    dataset_configuration_class(
-        dataset_info=mock_DatasetInfo(),
-        chunk_shape=(78_125, 7),
-        buffer_shape=(1_250_000, 384),
-    )
-
-
-@pytest.mark.parametrize(
```
h-mayorquin, line 172: I guess that this is done by pydantic now.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_default_dataset_io_configurations.py
```diff
@@ -39,10 +39,10 @@ def test_configuration_on_time_series(iterator: callable, backend: Literal["hdf5
 
     dataset_configuration = dataset_configurations[0]
     assert isinstance(dataset_configuration, DATASET_IO_CONFIGURATIONS[backend])
-    assert dataset_configuration.dataset_info.object_id == time_series.object_id
-    assert dataset_configuration.dataset_info.location_in_file == "acquisition/TestTimeSeries/data"
-    assert dataset_configuration.dataset_info.full_shape == array.shape
-    assert dataset_configuration.dataset_info.dtype == array.dtype
+    assert dataset_configuration.object_id == time_series.object_id
```
h-mayorquin, line 42: less nesting

src/neuroconv/tools/testing/_mock/_mock_dataset_models.py
```diff
@@ -7,46 +7,48 @@
...
-    DatasetInfo,
     HDF5BackendConfiguration,
     HDF5DatasetIOConfiguration,
     ZarrBackendConfiguration,
     ZarrDatasetIOConfiguration,
 )
 
 
-def mock_DatasetInfo(
+def mock_HDF5DatasetIOConfiguration(
```
h-mayorquin, line 17: mock at the beginning of the name. It could also be mock_DatasetIOConfigurarion(backend: ["hdf5", "zarr"])) to save some code duplcation. 

But I am not sure is better. This also works given that we have only two objects.

h-mayorquin (approved):
The tests look OK to me. This is good to go.

PR #768, opened 2024-03-05
Fix group names in intan

h-mayorquin:
As requested per @CodyCBakerPhD 

This should fix some on-going errors.

Context for @CodyCBakerPhD 

In https://github.com/NeuralEnsemble/python-neo/pull/1415 we chaned Intan to use `native_channel_name` as `channel_ids` and now we use `custom_channel_name` as channel names. 

The thing is that here in neuroconv the group name was extracted from the channel names (which is now the ID). So this makes the changes in neuroconv to keep the old behavior. 

I commented out WaveformExtractor because I think it has been deprecated in the dev of Spikeinterface.


CodyCBakerPhD:
> The remaining error is because Spikeinterfacd dropped the WaveformExtractor I think.

Cool - should we let @alejoe91 decide how to fix that then?

CodyCBakerPhD:
> The remaining error is because Spikeinterfacd dropped the WaveformExtractor I think.

Could you set those tests to skip if a version of SI is imported that doesn't have that class (i.e., dev branch for next few months?)

alejoe91:
> > The remaining error is because Spikeinterfacd dropped the WaveformExtractor I think.
> 
> Could you set those tests to skip if a version of SI is imported that doesn't have that class (i.e., dev branch for next few months?)

should be skip if `SI version>=0.101`

h-mayorquin:
Why not drop them all together if SI is not supporting that anymore?

h-mayorquin:
ok, let's discuss 

> > The remaining error is because Spikeinterfacd dropped the WaveformExtractor I think.
> 
> Could you set those tests to skip if a version of SI is imported that doesn't have that class (i.e., dev branch for next few months?)

OK, this is done in another PR:
https://github.com/catalystneuro/neuroconv/pull/777

CodyCBakerPhD (approved):

h-mayorquin:
Oh, failing because of code coverage failure : /

PR #771, opened 2024-03-11
get json schema descriptions from docstrings

bendichter:
In GUIDE, the we need data entry help text for fields of the interface constructors. See for example:

image

Without this help field, it is sometimes quite unclear which specific file or folder the user is meant to select. This help text is taken from the "description" field in the source json schema. This particular field for the `SpikeGLXRecordingInterface` was added manually to the source schema:

https://github.com/catalystneuro/neuroconv/blob/f80899b00a3749e5c8c9adf812ca0b69b45e7dd3/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxdatainterface.py#L27-L31

While it is possible to do this with every DataInterface, that would be redundant with the docstring of the init and would add a lot of lines of code.

This PR solves this problem in general by instead taking this description directly from the docstring of the `__init__`. It adds a new dependency, docstring-parser, to the minimal installation, which is a downside. docstring-parser Is not a very well-established package and may break in the future.

Note that with this PR, all source schemas are already being generated and validated against json schema with these lines:

https://github.com/catalystneuro/neuroconv/blob/f80899b00a3749e5c8c9adf812ca0b69b45e7dd3/src/neuroconv/tools/testing/data_interface_mixins.py#L64-L66

One additional test simply uses an arbitrary DataInterface to confirm that the descriptions are indeed being added to the schema.

I can see two alternative solutions.

1. manually add these descriptions to the json schema as it is currently done with the `SpikeGLXRecordingInterface`. This is explicit, but also a bit redundant.
2. Change all DataInterfaces to extend from pydantic BaseModels, and replace our custom json schema generation with the corresponding functionality within pydantic.

The current solution achieves the same goal with minimal changes to the existing codebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * Absorbed the `DatasetInfo` model into the `DatasetIOConfiguration` model. [PR #767](https://github.com/catalystneuro/neuroconv/pull/767)
 * Keyword argument `field_name` of the `DatasetIOConfiguration.from_neurodata_object` method has been renamed to `dataset_name` to be more consistent with its usage. This only affects direct initialization of the model; usage via the `BackendConfiguration` constructor and its associated helper functions in `neuroconv.tools.nwb_helpers` is unaffected. [PR #767](https://github.com/catalystneuro/neuroconv/pull/767)
 * Manual construction of a `DatasetIOConfiguration` now requires the field `dataset_name`, and will be validated to match the final path of `location_in_file`. Usage via the automated constructors is unchanged. [PR #767](https://github.com/catalystneuro/neuroconv/pull/767)
-
+* Enhance `get_schema_from_method_signature` to extract description from the method docval. [PR #771](https://github.com/catalystneuro/neuroconv/pull/771)
 
 
 # v0.4.7 (February 21, 2024)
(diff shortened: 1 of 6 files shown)


CodyCBakerPhD:
LGTM, given that it's silent and optional (will not error if docstring is not numpy-compliant)

> 2. Change all DataInterfaces to extend from pydantic BaseModels, and replace our custom json schema generation with the corresponding functionality within pydantic.

Yes, this is already the plan for the future (and we can actually work on it soon since we live in Pydantic v2 now!) as the best way to resolve the slew of upcoming annotation changes coming via `from __futures__ import annotations`: #669, #671

But the `get_schema_from_method_signature` function would only have its internal workings changed in that event, likely just being reduced to

```python
model = make_model_from_signature(signature)
schema = model.json_schema()
return schema
```

and these lines related to docstrings would then go in the `make_model_from_signature` method pretty much exactly as they are, only they would map to the `description` of a `pydantic.Field` (which then maps to the JSON description as expected)

CodyCBakerPhD:
Oh, I did just observe that the `docstring-parser` tests are failing for MacOS: https://github.com/rr-/docstring_parser/actions/runs/7276222152

Should we be concerned about adding it in this case, or is it working fine as far as you can tell?

CodyCBakerPhD:
Also, to get full benefit from this, we should revisit pydocstyle CI checks to ensure numpy style compliance: https://github.com/catalystneuro/neuroconv/pull/276

@pauladkisson has gained a lot of expertise in this area

CodyCBakerPhD (approved):

PR #773, opened 2024-03-11
Return organized files from DANDI upload

CodyCBakerPhD:
Requested by https://github.com/NeurodataWithoutBorders/nwb-guide/issues/658#issuecomment-1989243735

Give this a try and let me know

### DIFF
--- a/src/neuroconv/tools/data_transfers/_dandi.py
+++ b/src/neuroconv/tools/data_transfers/_dandi.py
@@ -97,14 +97,16 @@ def automatic_dandi_upload(
             dandi_stem_split.insert(1, f"ses-{session_id}")
             corrected_name = "_".join(dandi_stem_split) + ".nwb"
             organized_nwbfile.rename(organized_nwbfile.parent / corrected_name)
-    organized_nwbfiles = dandiset_path.rglob("*.nwb")
+
+    organized_nwbfiles = [str(x) for x in dandiset_path.rglob("*.nwb")]
     # The above block can be removed once they add the feature
 
     assert len(list(dandiset_path.iterdir())) > 1, "DANDI organize failed!"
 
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
Ah, right.

For GUIDE purposes I'd just strip everything before the second to last path. It should always have the pattern of a file `sub-.../sub-....ses-....nwb`

CodyCBakerPhD:
Which is really what you're trying to display/communicate back the user I think

garrettmflynn:
Hmm I brought this up because where these sessions originally came for is _not_ actually encoded anywhere in the leftover path. 

These were actually the outputs of two conversion pipelines using different data, which I'd selected as two folders in `~/NWB_GUIDE/conversions`

CodyCBakerPhD:
> Hmm I brought this up because where these sessions originally came for is not actually encoded anywhere in the leftover path.

I don't understand what you mean - for reporting what was uploaded as a result of DANDI upload, do you not want to report the file paths as they will appear on the Dandiset? (not their original names chosen through the GUIDE, let alone the conversion output folder)

CodyCBakerPhD:
> At least in standalone upload mode, I'd imagine that users would like to see a mapping between the discovered files and their DANDI equivalent, right?

I agree, but it's a different feature ('file organization preview') with its own challenges (direct mapping isn't returned by constituent functions; I know how to make it but would take little work)

> If not, then we should probably rename this on the GUIDE to New Files on DANDI.

Sure, that's a fine intermediate

garrettmflynn:
Works on https://github.com/NeurodataWithoutBorders/nwb-guide/pull/665

garrettmflynn (approved):
Works on https://github.com/NeurodataWithoutBorders/nwb-guide/pull/665

PR #774, opened 2024-03-11
fix phy source schema

bendichter:


CodyCBakerPhD:
Only it's not truly a list of strings, it's an enum of the unique group names (if any are even labeled)

So sure, a user could specify the names in the GUIDE but do we want them to be able to do that even when it will have no effect?

bendichter:
I think of this as an advanced option where a user has a particular named cluster in mind. Yes, users could enter in incorrect information, and yes, it would be even better if we validated this against the possible responses, but this is an improvement over the current situation which does not enable users to exclude clusters at all

CodyCBakerPhD (approved):

PR #777, opened 2024-03-13
Drop tests of Spikeinterface when `WaveformExtractor` is not available

h-mayorquin:
This will get dev tests working properly where the latest version of Spikeinterface is used.

We will discuss more slowly how we want to move forward in supporting the new way of extracting waveforms in Spikeinterface.


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1254,7 +1254,7 @@ def write_sorting(
 
 
 def add_waveforms(
-    waveform_extractor: WaveformExtractor,
+    waveform_extractor,
```
CodyCBakerPhD, line 1257:
```suggestion
    waveform_extractor: "WaveformExtractor",
```
Should we at least do this?

h-mayorquin: Yeah, I think that's better. Let me push this.

CodyCBakerPhD: Or else nest entirely in a try/except scope

try:
    from spikeinterface import WaveformExtractor

    def add_waveforms(waveform_extractor: WaveformExtractor, ...)
    ...
except Exception as exception:
    assert type(exception) is <type of exception - I think an ImportError?)
    assert "Some string part of the message unique to the failured import of waveforms extractor" in str(exception)

CodyCBakerPhD: Or, or,

LIke you do in tests, nest entirely in `if SI_VERSION > threshold` block?

h-mayorquin: I rather just add the docstring. I don't think is worth it with an object that will be dropped anyway.

CodyCBakerPhD (approved):

Issue #291, opened 2023-01-30
[New Format]: Mini2p


bendichter:
### What format would you like to see added to NeuroConv?

Mini2p

### Does the format have any documentation?

https://github.com/kavli-ntnu/MINI2P_toolbox

### Existing APIs for format

I could only find MATLAB code

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

simon-ball:
The Zong2022 dataset noted here includes multiple evolutions of the way tracking data was stored. I can potentially share some more recent datasets representing the current state of the project, although I will need to confirm access with the relevant authors first.

Issue #530, opened 2023-07-26
[Bug]: Cannot disable mask_type from Suite2p Convertor


seethakris:
### What happened?

Stub_test in suite2p fails at the masking step. When mask_type is disabled, conversion validator fails because it says it expects a string and does not accept None. None as a string also raises an error although the assertion error says it does accept None to disable mask_type. 

### Steps to Reproduce

```python
suite2pinterface = Suite2pSegmentationInterface(folder_path=self.suite2pfile, verbose=True)
converter = ConverterPipe(data_interfaces=[suite2pinterface], verbose=True)
converter.run_conversion(
            nwbfile_path=self.savepath, 
            metadata=metadata,
            overwrite=True,
            conversion_options=dict(Suite2pSegmentationInterface=dict(stub_test=True, 
                                                                      mask_type=None,
                                                                      include_roi_centroids=False,
                                                                      include_roi_acceptance=False,
                                                                      iterator_options=dict(display_progress=True, buffer_gb=0.05)
                                                                     ))
        )
```

### Traceback

```python
jsonschema.exceptions.ValidationError: None is not of type 'string'

Failed validating 'type' in schema['properties']['Suite2pSegmentationInterface']['properties']['mask_type']:
    {'default': 'image', 'type': 'string'}

On instance['Suite2pSegmentationInterface']['mask_type']:
    None
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.9

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

CodyCBakerPhD:
Worth mentioning we simply swapped the `mask_type` to `"pixels"` and it was much more efficient

But the problem itself happens because schema validation gets called in the Converter but not in the Interface
Interfaces should also validate these schemas but that's beside the point

Problem: The value of `None` (disabling masks entirely) has no default JSON equivalent

Two solutions:

a) adjust the `NwbMetadataEncoder` to translate `None` to `"None"` - not sure if this might cause problems anywhere else though

[...]

PR #783, opened 2024-03-18
Background Components for Caiman

### LINKED ISSUE #782: [Feature]: Update CaimanSegmentationInterface to incorporate background components

Follow-up from roiextractors [# 291](https://github.com/catalystneuro/roiextractors/pull/291)

_No response_

Yes.

### Code of Conduct

pauladkisson:
Fixes #782

Follow-up from ROIExtractors [#291](https://github.com/catalystneuro/roiextractors/pull/291)


src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -681,6 +698,7 @@ def add_plane_segmentation(
     metadata: Optional[dict],
     plane_segmentation_name: Optional[str] = None,
     plane_segmentation_index: Optional[int] = None,  # TODO: to be removed
+    background_plane_segmentation_name: Optional[str] = None,
```
CodyCBakerPhD, line 701: Why not have this include somewhere in `metadata`?

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -766,6 +784,20 @@ def add_plane_segmentation(
             f"Metadata for Plane Segmentation '{plane_segmentation_name}' not found in metadata['Ophys']['ImageSegmentation']['plane_segmentations']."
         )
 
+    background_plane_segmentation_name = (
+        background_plane_segmentation_name
+        or default_metadata["Ophys"]["ImageSegmentation"]["plane_segmentations"][1]["name"]
```
CodyCBakerPhD, line 789: Why only the second index?

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -766,6 +784,20 @@ def add_plane_segmentation(
...
+    )
+    background_plane_segmentation_metadata = None
+    for meta in metadata_copy["Ophys"]["ImageSegmentation"]["plane_segmentations"]:
+        if meta["name"] == background_plane_segmentation_name:
+            background_plane_segmentation_metadata = meta
+            break
```
CodyCBakerPhD, line 795:
```suggestion
     background_plane_segmentation_metadata = next(meta for meta in metadata_copy["Ophys"]["ImageSegmentation"]["plane_segmentations"] if meta["name"] == background_plane_segmentation_name)
```

src/neuroconv/tools/roiextractors/roiextractors.py
CodyCBakerPhD: I see, like a parallel to the main `plane_segmentation_name`

CodyCBakerPhD: But then the implication is `add_plane_segmentation` adds more than one place segmentation? Why not just contain it to be responsible for writing a single table and then write multiple tables by iterating outside of it? That way the complexity exists outside rather than inside the function

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -844,9 +915,14 @@ def add_plane_segmentation(
             )
 
         image_segmentation.add_plane_segmentation(plane_segmentations=[plane_segmentation])
+        image_segmentation.add_plane_segmentation(plane_segmentations=[background_plane_segmentation])
     return nwbfile
 
 
+def add_background_plane_segmentation():
```
CodyCBakerPhD, line 922: Exactly 😉 I would prefer this. Then perhaps an added `add_all_plane_segmentations` that calls each individual single plane segmentation addition, unless we think that's too much redirection

CodyCBakerPhD:
> pauladkisson requested review from [CodyCBakerPhD](https://github.com/CodyCBakerPhD) and [alessandratrapani](https://github.com/alessandratrapani) [13 minutes ago](https://github.com/catalystneuro/neuroconv/pull/783#event-12332730100)

@pauladkisson Tests are failing at installation - can you update the pin?

CodyCBakerPhD:
@pauladkisson Just let me know when this is ready for another review

pauladkisson:
@CodyCBakerPhD It's ready now

tests/test_ophys/test_tools_roiextractors.py
```diff
@@ -609,6 +609,23 @@ def test_none_masks(self):
...
 
+    def test_invalid_mask_type(self):
+        """Test that an invalid mask_type raises a AssertionError."""
+        segmentation_extractor = generate_dummy_segmentation_extractor(
+            num_rois=self.num_rois,
+            num_frames=self.num_frames,
+            num_rows=self.num_rows,
+            num_columns=self.num_columns,
+        )
+        with assert_raises(AssertionError):
```
CodyCBakerPhD, line 620:
```suggestion
        with assert_raises(AssertionError) as exception_info:
```

can you add an extra assertion here for the message type, especially since an `AssertionError` is a pretty common type for us

CodyCBakerPhD:
LGTM but Alessio was requesting a quick rerelease following some bug fixes to the backend configuration and we wouldn't be able to do that until the ROIExtractors pin is released

So I guess question is, how soon can we get an ROIExtractors release?

pauladkisson:
I think we can cut a release of ROIExtractors right away, but I'd like @EricThomson to take a look before we merge as well.

pauladkisson:
No response from Eric Thompson, but I don't want to leave this hanging forever, so I think we should go ahead with this and he can raise an issue if he runs into any problems with it.

CodyCBakerPhD (approved):

CodyCBakerPhD:
@pauladkisson Does look like one test case broke a while ago

pauladkisson:
Yeah, it looks like the conversion_options_schema doesn't like None for mask_type...

I tracked it down to src/neuroconv/utils/json_schema.py,
```python
annotation_json_type_map = dict(
        bool="boolean",
        str="string",
        int="number",
        float="number",
        dict="object",
        list="array",
        tuple="array",
        FilePathType="string",
        FolderPathType="string",
    )
```

Apparently None isn't a valid type?  I don't understand why we would annotate `add_to_nwbfile` as if you can provide none, if the schema doesn't actually do it. @CodyCBakerPhD, can you shed some light on this?

CodyCBakerPhD:
This has been known since https://github.com/catalystneuro/neuroconv/issues/530 and would be resolved with the Pydantic integration for handling all schema validations and signature parsing - for now I'd just comment out that one test case since we're already aware of it

EricThomson:
Sorry I was away last month when that @ came in. I'm back and can look things over now (early next week) if it would help?

pauladkisson:
> Sorry I was away last month when that @ came in. I'm back and can look things over now (early next week) if it would help?

Sure! That would be great! Just raise and issue and @ me if you run into any problems.

PR #784, opened 2024-03-19
Fix writing electrodes field in add_electrical_series

alejoe91:
Related to https://github.com/SpikeInterface/spikeinterface/issues/2588

The mask to find the electrodes corresponding to the electrical series should use both channel names and channel groups, while the current implementation only uses channel names.

Extended the test too, because we were only checking the `nwbfile.electrodes`, not the `es.electrodes`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Improvements
 
+* Fixed writing the `electrodes` field in `add_electrical_series` when multiple groups are present. [PR #784](https://github.com/catalystneuro/neuroconv/pull/784)
 * Upgraded Pydantic support to `>v2.0.0`. [PR #767](https://github.com/catalystneuro/neuroconv/pull/767)
 * Absorbed the `DatasetInfo` model into the `DatasetIOConfiguration` model. [PR #767](https://github.com/catalystneuro/neuroconv/pull/767)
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
I checked the issue on Spikeinterface and this makes sense to me. 

@alejoe91 I pushed the full case for `group_name` assignation that was missing in your code. I hope you don't mind.  The full assignation is here:

https://github.com/catalystneuro/neuroconv/blob/80e5d9c26f471006428c5c2859ab57cab280c209/src/neuroconv/tools/spikeinterface/spikeinterface.py#L277-L284

I also took the liberty of making the test naming more specific here.

Summary:
We are using a pair of `channel_name, channel_group` to define uniqueness in mapping channels in a `RecordingExtractor` object. We are using this logic within `add_electrodes` but we were not using this for creating the corresponding electrode table regions  within`add_electric_series`. The latter meant that references on the `ElectricalSeries` were wrong (less specific that they should be).

I also think that there are more bugs related to this hiding within the less used corners of `add_electrodes` here:

https://github.com/catalystneuro/neuroconv/blob/80e5d9c26f471006428c5c2859ab57cab280c209/src/neuroconv/tools/spikeinterface/spikeinterface.py#L386-L387

I think we should do a more general refactor were the different users of this uniqueness criteria use a centralized function(s) where our intent is clearly expressed and documented.

CodyCBakerPhD:
@h-mayorquin @alejoe91 Could you add a quick changelog entry and then this is good to merge?

h-mayorquin:
> @h-mayorquin @alejoe91 Could you add a quick changelog entry and then this is good to merge?

I am going to wait for @alejoe91 to checked the modifications that I did. Maybe I am missing something.

PR #793, opened 2024-03-28
Preserve order of recording extractor custom properties in `add_electrodes()`

### LINKED ISSUE #792: Preserve order of recording extractor custom properties in `add_electrodes()`

```python

```

```python
```

weiglszonja:
Resolve #792

### DIFF
--- a/tests/test_ecephys/test_tools_spikeinterface.py
+++ b/tests/test_ecephys/test_tools_spikeinterface.py
@@ -888,22 +888,44 @@ class TestAddElectrodes(TestCase):
         self.recording_1.set_property(key="custom_property_x", values=[0.1] * self.num_channels)
         self.recording_1.set_property(key="custom_property_y", values=[0.2] * self.num_channels)
         self.recording_1.set_property(key="custom_property_z", values=[0.3] * self.num_channels)
-        self.recording_1.set_property(key="custom_property_1", values=["value_1"] * self.num_channels)
 
-        add_electrodes(recording=self.recording_1, nwbfile=self.nwbfile)
+        self.recording_2.set_property(key="custom_property_2", values=["value_1"] * self.num_channels)
+
+        metadata = dict(
+            Ecephys=dict(
+                Electrodes=[
+                    dict(name="custom_property_x", description="custom description."),
+                    dict(name="custom_property_y", description="custom description."),
+                    dict(name="custom_property_z", description="custom description."),
+                    dict(name="custom_property_2", description="custom description."),
+                ]
+            )
+        )
 
         expected_columns_in_order = [
             "location",
             "group",
             "group_name",
             "channel_name",
+            "rel_x",
+            "rel_y",
             "custom_property_x",
             "custom_property_y",
             "custom_property_z",
-            "custom_property_1",
-            "rel_x",
-            "rel_y",
+            "custom_property_2",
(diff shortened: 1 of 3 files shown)


tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -881,6 +881,32 @@ def test_property_metadata_mismatch(self):
...
+            "location",
+            "group",
+            "group_name",
+            "channel_name",
+            "custom_property_x",
+            "custom_property_y",
+            "custom_property_z",
+            "custom_property_1",
+            "rel_x",
+            "rel_y",
```
CodyCBakerPhD, line 905: Why are `rel_x` and `rel_y` after the custum ones though? I would have expected them before since they are pre-defined in NWB schema

CodyCBakerPhD:
I think this behavior and all concerns could be abated if we use the `metadata["Ecephys"]["Electrodes"]: List[dict]` as the defining order of all columns, with the one condition being that all properties on the recording extractor must be defined in that metadata structure

Then we can always take the order in that list as the defining order to write, including 'reserving' a column by adding it and populating default 'empty' values for that data type even if it doesn't exist in the first recording extractor triggering the call to `add_electrodes`

weiglszonja:
> Then we can always take the order in that list as the defining order to write, including 'reserving' a column by adding it and populating default 'empty' values for that data type even if it doesn't exist in the first recording extractor triggering the call to `add_electrodes`

I'm not sure how to do this, I started refactoring to have the electrodes metadata populated (see https://github.com/catalystneuro/neuroconv/pull/793/commits/e1f2c6c8be81d5d24c857f8c621cac555de929d2) but I think I'll need help to rewrite current logic. I only managed to do it for now when adding properties by columns (see the test in https://github.com/catalystneuro/neuroconv/pull/793/commits/403b63af9b922105a12083b2de35d4b14e71be0e). 

For the Turner data these changes achieve the ordering of "custom" attributes as they appear in the electrodes metadata:
Screenshot 2024-03-29 at 13 19 15 (2)

tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -881,6 +881,54 @@ def test_property_metadata_mismatch(self):
...
+            )
+        )
+
+        expected_columns_in_order = [
+            "location",
+            "group",
+            "group_name",
+            "channel_name",
+            "rel_x",
+            "rel_y",
```
CodyCBakerPhD, line 911:
```suggestion
            "channel_name",
            "group",
            "group_name",
            "location",
            "x",
            "y",
            "z",
            "rel_x",
            "rel_y",
```

The one column we can't control the order of is naturally the `id`, but what do you think about this ordering?

Also could you make up some x/y/z values (possibly using some recent real life examples as a good baseline) to match the location?

CodyCBakerPhD:
Having the tests is a good first start

I see you have some for the single extractor case, but if you can add some for the multi extractor case you're unsure about (such that the new tests are failing) then it can help by making the problem more well defined (we know what to achieve, just not how to achieve it)

At that point I can take a look and suggest a strategy

h-mayorquin:
I am closing this as it has been gone for a while. feel free to re-open when coming back to work on this

PR #794, opened 2024-03-30
Deprecate stream_id keyword from `IntanRecordingInterface`

### LINKED ISSUE #790: `IntanRecordingInterface` should not be allowed to write streams other than the amplifier channels
From the current discussionin #789  it seems that `IntanRecordingInterface` should only write data with `stream_id=0`: the amplifier channels. Reason being that we write the data as an ElectricalSeries following the ecephys pipeline. This means that if any other stream is choosen the data will be incorrectly written.

The current signature of `IntanRecordingInterface` allows the user to specify the `stream_id` which leads to the problem described above.

https://github.com/catalystneuro/neuroconv/blob/fa6edd532f6bb6ead42ae40aa9029c7c78eead5f/src/neuroconv/datainterfaces/ecephys/intan/intandatainterface.py#L77-L95

Should we change this to avoid this scenario?
Pros and cons?

h-mayorquin:
Should fix #790.


src/neuroconv/datainterfaces/ecephys/intan/intandatainterface.py
```diff
@@ -95,6 +68,12 @@ def __init__(
         es_key : str, default: "ElectricalSeries"
         """
 
+        if stream_id is not None:
+            warnings.warn(
+                "Use of the 'stream_id' parameter is deprecated and it will be removed completly by September 2024..",
```
CodyCBakerPhD, line 73:
```suggestion
                "Use of the 'stream_id' parameter is deprecated and it will be removed after September 2024.",
```

Better to leave loose lower bound rather than right upper bound on these things since sometimes exact release schedules are outside our control

src/neuroconv/datainterfaces/ecephys/intan/intandatainterface.py
```diff
@@ -8,35 +9,6 @@
 from ....utils import FilePathType, get_schema_from_hdmf_class
 
 
-def extract_electrode_metadata_with_pyintan(file_path) -> dict:
```
CodyCBakerPhD, line 11: I'm assuming this logic has propagated up?

src/neuroconv/datainterfaces/ecephys/intan/intandatainterface.py
h-mayorquin: That makes sense.

src/neuroconv/datainterfaces/ecephys/intan/intandatainterface.py
h-mayorquin: Yes. It should be fine on neo.

src/neuroconv/datainterfaces/ecephys/intan/intandatainterface.py
```diff
@@ -95,6 +68,12 @@ def __init__(
         es_key : str, default: "ElectricalSeries"
         """
 
+        if stream_id is not None:
+            warnings.warn(
+                "Use of the 'stream_id' parameter is deprecated and it will be removed after September 2024.",
+                DeprecationWarning,
+            )
+
         self.stream_id = stream_id
```
CodyCBakerPhD, line 77: Can remove this line then, yes?

CodyCBakerPhD: Or this is a soft cycle so you still allow setting it with the warning that it's undesired behavior

CodyCBakerPhD:
One final comment then just need CHANGELOG and this is good to go

src/neuroconv/datainterfaces/ecephys/intan/intandatainterface.py
h-mayorquin: The first iteration was wrong though. Thanks for pointing it out. Should be ammended now.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Upcoming
 
+### Bug fixes
```
CodyCBakerPhD, line 3: Deprecation

h-mayorquin, line 3:
```suggestion
### Deprecations
```

CodyCBakerPhD (approved):

PR #795, opened 2024-04-01
Undo CRON for testing.yml

CodyCBakerPhD:
Dailies are triggered in a separate workflow: https://github.com/catalystneuro/neuroconv/blob/main/.github/workflows/dailies.yml#L22


h-mayorquin (approved):
Got you. So the dallies already run the cron, right?

PR #799, opened 2024-04-02
Fix write_waveforms nwbfile call

CodyCBakerPhD:
@alejoe91 any idea how to add tests for this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
 # Upcoming
 
+### Bug fixes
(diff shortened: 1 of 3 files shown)


alejoe91:
I think that we might only have tests with an existing nwbfile! I'll add one with output files

alejoe91:
> @alejoe91 any idea how to add tests for this?

Done! Added a test that writes directly to an `nwbfile_out`

Also updated a couple of small deprecations in `spikeinterface.py` (`is_extension` -> `has_extension`)

alejoe91 (approved):

PR #800, opened 2024-04-02
[Backend IVa] Add backend control to context tool

CodyCBakerPhD:
Breaking up the integration into the core classes into smaller pieces

This is the first step - adding `backend` keyword and functionality into the context tool used to write the majority of NWB files through NeuroConv

### DIFF
--- a/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
@@ -5,11 +5,11 @@ from contextlib import contextmanager
 from copy import deepcopy
 from datetime import datetime
 from pathlib import Path
-from typing import Optional
+from typing import Literal, Optional
 from warnings import warn
 
 from pydantic import FilePath
-from pynwb import NWBHDF5IO, NWBFile
+from pynwb import NWBFile
 from pynwb.file import Subject
 
 from ...utils.dict import DeepDict, load_dict_from_file
@@ -133,6 +133,7 @@ def make_or_load_nwbfile(
     nwbfile: Optional[NWBFile] = None,
     metadata: Optional[dict] = None,
     overwrite: bool = False,
+    backend: Literal["hdf5", "zarr"] = "hdf5",
     verbose: bool = True,
 ):
     """
@@ -150,11 +151,17 @@ def make_or_load_nwbfile(
     overwrite: bool, optional
         Whether to overwrite the NWBFile if one exists at the nwbfile_path.
         The default is False (append mode).
+    backend : "hdf5" or "zarr", default: "hdf5"
+        The type of backend used to create the file.
     verbose: bool, optional
         If 'nwbfile_path' is specified, informs user after a successful write operation.
         The default is True.
     """
+    from . import BACKEND_NWB_IO
+
     nwbfile_path_in = Path(nwbfile_path) if nwbfile_path else None
+    backend_io_class = BACKEND_NWB_IO[backend]
+
     assert not (nwbfile_path is None and nwbfile is None and metadata is None), (
         "You must specify either an 'nwbfile_path', or an in-memory 'nwbfile' object, "
         "or provide the metadata for creating one."
@@ -163,17 +170,21 @@ def make_or_load_nwbfile(
         "'nwbfile_path' exists at location, 'overwrite' is False (append mode), but an in-memory 'nwbfile' object was "
         "passed! Cannot reconcile which nwbfile object to write."
     )
+    if not backend_io_class.can_read(path=nwbfile_path_in):
+        raise IOError(f"The chosen backend ({backend}) is unable to read the file! Please select a different backend.")
 
     load_kwargs = dict()
     success = True
-    file_initially_exists = nwbfile_path_in.is_file() if nwbfile_path_in is not None else None
+    file_initially_exists = nwbfile_path_in.exists() if nwbfile_path_in is not None else None
     if nwbfile_path_in:
         load_kwargs.update(path=nwbfile_path_in)
         if file_initially_exists and not overwrite:
             load_kwargs.update(mode="r+", load_namespaces=True)
         else:
             load_kwargs.update(mode="w")
-        io = NWBHDF5IO(**load_kwargs)
+
+        io = backend_io_class(**load_kwargs)
+
     try:
         if load_kwargs.get("mode", "") == "r+":
             nwbfile = io.read()
--- a/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
@@ -170,18 +170,23 @@ def make_or_load_nwbfile(
         "'nwbfile_path' exists at location, 'overwrite' is False (append mode), but an in-memory 'nwbfile' object was "
         "passed! Cannot reconcile which nwbfile object to write."
     )
-    if not backend_io_class.can_read(path=nwbfile_path_in):
-        raise IOError(f"The chosen backend ({backend}) is unable to read the file! Please select a different backend.")
-
+    if overwrite is False and backend == "zarr":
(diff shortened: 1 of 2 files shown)


src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -163,17 +170,26 @@ def make_or_load_nwbfile(
...
         "passed! Cannot reconcile which nwbfile object to write."
     )
-
+    if overwrite is False and backend == "zarr":
+        # TODO: remove when https://github.com/hdmf-dev/hdmf-zarr/issues/182 is resolved
+        raise NotImplementedError("Appending a Zarr file is not yet supported!")
     load_kwargs = dict()
     success = True
-    file_initially_exists = nwbfile_path_in.is_file() if nwbfile_path_in is not None else None
+    file_initially_exists = nwbfile_path_in.exists() if nwbfile_path_in is not None else None
```
h-mayorquin, line 178: OK, this is because this covers zarr as well that is a directory, right?

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -163,17 +170,26 @@ def make_or_load_nwbfile(
...
     if nwbfile_path_in:
-        load_kwargs.update(path=nwbfile_path_in)
+        load_kwargs.update(path=str(nwbfile_path_in))
```
h-mayorquin, line 180: I should push that PR to zarr so they take paths.

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -163,17 +170,26 @@ def make_or_load_nwbfile(
...
         if file_initially_exists and not overwrite:
             load_kwargs.update(mode="r+", load_namespaces=True)
         else:
             load_kwargs.update(mode="w")
-        io = NWBHDF5IO(**load_kwargs)
+
+        if load_kwargs["mode"] == "r+" and not backend_io_class.can_read(path=str(nwbfile_path_in)):
+            raise IOError(
+                f"The chosen backend ({backend}) is unable to read the file! Please select a different backend."
```
h-mayorquin, line 188:
```suggestion
                f"The chosen backend ({backend}) is unable to read {nwbfile_path_in}! Please select a different backend."
```

Can we tell them which backend?

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -163,17 +170,26 @@ def make_or_load_nwbfile(
...
+            )
+
+        io = backend_io_class(**load_kwargs)
+
     try:
```
h-mayorquin, line 193: Does the `can_read` method eliminate the need for the try here? What else is this try doing?

h-mayorquin (dismissed):
I suggest a possible and small improvement to the error function but this looks OK.

Side comment:
Every time that I look at the `make_or_load_nwbfile` I feel that it could be simplified. Even on the name is doing too many things and the control flow at the end looks very complicated. Maybe there is no better way here if we want a context manager.

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
CodyCBakerPhD: Yup 👍

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
CodyCBakerPhD: Requested in https://github.com/hdmf-dev/hdmf-zarr/issues/181, really hoping neither of us would need to spend time on it and just let the LBNL team handle it, but who knows 🤷

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
CodyCBakerPhD: Unrelated. The `try` clause here is for the `contextlib.contextmanager`

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
CodyCBakerPhD: Sure I'll add that

CodyCBakerPhD: Added!

h-mayorquin (approved):
Thanks for the work on this.

> The most I could reduce it now is perhaps to split the 'make' and 'load' parts to separate contexts

Yes, I think the make part does not really require a context manager, right? I feel the main responsability of this -and its most common use- is to write.  But I think that mos of the complexity comes from the loading to append part.

Relevant to the discussion in #801

CodyCBakerPhD:
> Yes, I think the make part does not really require a context manager, right?

The context is to trigger write after making or loading; so they would still be `make_and_write_nwbfile` and `append_and_write_nwbfile` with duplicated switching logic everywhere

h-mayorquin:
> The context is to trigger write after making or loading; so they would still be make_and_write_nwbfile and append_and_write_nwbfile with duplicated switching logic everywhere

or passing a in-memory nwbfile, right? the commontality is still writing I feel.

I just opened the function to confirm that we have the `nwbfile` argument and then I could not figure out in 3 minutes how it is used.

I copied the code to illustrate this, I thought about it another half a minute and I can't figure out which branch do I go when I passed an nwbfile. I think is the first yield?

Oh, no, the path is optional, which means that it does not neceesarily write.

OK, forget about it. I am confused.

```
    load_kwargs = dict()
    success = True
    file_initially_exists = nwbfile_path_in.is_file() if nwbfile_path_in is not None else None
    if nwbfile_path_in:
        load_kwargs.update(path=nwbfile_path_in)
        if file_initially_exists and not overwrite:
            load_kwargs.update(mode="r+", load_namespaces=True)
        else:
            load_kwargs.update(mode="w")
        io = NWBHDF5IO(**load_kwargs)
    try:
        if load_kwargs.get("mode", "") == "r+":
            nwbfile = io.read()
        elif nwbfile is None:
            nwbfile = make_nwbfile_from_metadata(metadata=metadata)
        yield nwbfile
    except Exception as e:
        success = False
        raise e
    finally:
        if nwbfile_path_in:
            try:
                if success:
                    io.write(nwbfile)

                    if verbose:
            finally:
                io.close()

                    nwbfile_path_in.unlink()
```

CodyCBakerPhD:
> or passing a in-memory nwbfile, right? the commontality is still writing I feel.

Yes, perhaps a third function `write_nwbfile` just for that one case too

h-mayorquin:
Creation of an nwbfile does not require a context manager unless it is opened for append mode, right? Writing on the other hand does always benefit for it so the `io` is closed. 

Maybe the open and append is one of the valid use case for nested context managers:

https://discuss.codecademy.com/t/faq-context-managers-nested-context-managers/601741/2

But that could end up looking worse. No idea.

CodyCBakerPhD:
> Creation of an nwbfile does not require a context manager unless it is opened for append mode, right? Writing on the other hand does always benefit for it so the io is closed.

Hence why I'd have `make_and_write_nwbfile` coupled together as well as `write_nwbfile` available separately

A double context could be used in `append_and_write_nwbfile`, but I don't think it would be outright necessary internally

h-mayorquin:
> Hence why I'd have make_and_write_nwbfile coupled together as well as write_nwbfile available separately

Makes sense to me. But would it make more sense for the first to be `load_and_write`?

CodyCBakerPhD:
> load_and_write

Sure, though that could also be interpreted as 'read from source and write to destination' (mode="r") instead of 'read from source and rewrite to same source' (mode="a"/"r+", explicit name being 'append')

h-mayorquin:
To be honest, I kind of though that we could do the first but that does not work because we only have one path argument anyway. `load_and_append` or something else could be more precise or something better.

PR #801, opened 2024-04-02
[Backend Configuration IVc] Integrate backend control as interface method

### LINKED ISSUE #811: [Bug]: get_default_backend_config broken for mock data
### What happened?

I tried to run the dataset configuration code on a simple mock NWBFile object. See below. Am I using this right?

### Steps to Reproduce

```python
from pynwb.testing.mock.ecephys import mock_ElectricalSeries
from neuroconv.tools.nwb_helpers import (
    get_default_backend_configuration,
    get_default_dataset_io_configurations,
)

from neuroconv.tools.nwb_helpers import DATASET_IO_CONFIGURATIONS


nwbfile = mock_NWBFile()
es = mock_ElectricalSeries(nwbfile=nwbfile)


get_default_backend_configuration(nwbfile, backend="hdf5")  # fails
list(get_default_dataset_io_configurations(nwbfile, backend="hdf5"))  # fails


# works
config_class = DATASET_IO_CONFIGURATIONS["hdf5"]
config_class.from_neurodata_object(
    neurodata_object=es, dataset_name="data"
)
```

### Traceback

```python
TypeError                                 Traceback (most recent call last)
Cell In [48], line 15
     11 nwbfile = mock_NWBFile()
     12 es = mock_ElectricalSeries(nwbfile=nwbfile)
---> 15 get_default_backend_configuration(nwbfile, backend="hdf5")  # fails
     16 list(get_default_dataset_io_configurations(nwbfile, backend="hdf5"))  # fails
     19 # works

File ~/dev/neuroconv/src/neuroconv/tools/nwb_helpers/_backend_configuration.py:19, in get_default_backend_configuration(nwbfile, backend)
     15 from ..nwb_helpers import BACKEND_CONFIGURATIONS
     17 BackendConfigurationClass = BACKEND_CONFIGURATIONS[backend]
---> 19 return BackendConfigurationClass.from_nwbfile(nwbfile=nwbfile)

File ~/dev/neuroconv/src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py:45, in BackendConfiguration.from_nwbfile(cls, nwbfile)
     42 @classmethod
     43 def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
     44     default_dataset_configurations = get_default_dataset_io_configurations(nwbfile=nwbfile, backend=cls.backend)
---> 45     dataset_configurations = {
     46         default_dataset_configuration.location_in_file: default_dataset_configuration
     47         for default_dataset_configuration in default_dataset_configurations
     48     }
     50     return cls(dataset_configurations=dataset_configurations)

File ~/dev/neuroconv/src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py:45, in <dictcomp>(.0)
     42 @classmethod
     43 def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
     44     default_dataset_configurations = get_default_dataset_io_configurations(nwbfile=nwbfile, backend=cls.backend)
---> 45     dataset_configurations = {
     46         default_dataset_configuration.location_in_file: default_dataset_configuration
     47         for default_dataset_configuration in default_dataset_configurations
     48     }
     50     return cls(dataset_configurations=dataset_configurations)

File ~/dev/neuroconv/src/neuroconv/tools/nwb_helpers/_dataset_configuration.py:116, in get_default_dataset_io_configurations(nwbfile, backend)
    113         if isinstance(candidate_dataset, DataIO):
    114             continue
--> 116         dataset_io_configuration = DatasetIOConfigurationClass.from_neurodata_object(
    117             neurodata_object=column, dataset_name="data"
    118         )
    120         yield dataset_io_configuration
    121 else:
    122     # Primarily for TimeSeries, but also any extended class that has 'data' or 'timestamps'
    123     # The most common example of this is ndx-events Events/LabeledEvents types

File ~/dev/neuroconv/src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py:267, in DatasetIOConfiguration.from_neurodata_object(cls, neurodata_object, dataset_name)
    264     buffer_shape = None
    265     compression_method = None
--> 267 return cls(
    268     object_id=neurodata_object.object_id,
    269     object_name=neurodata_object.name,
    270     location_in_file=location_in_file,
    271     dataset_name=dataset_name,
    272     full_shape=full_shape,
    273     dtype=dtype,
    274     chunk_shape=chunk_shape,
    275     buffer_shape=buffer_shape,
    276     compression_method=compression_method,
    277 )

    [... skipping hidden 1 frame]

File ~/dev/neuroconv/src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py:187, in DatasetIOConfiguration.validate_all_shapes(cls, values)
    184 buffer_shape = values["buffer_shape"]
    185 full_shape = values["full_shape"]
--> 187 if len(chunk_shape) != len(buffer_shape):
    188     raise ValueError(
    189         f"{len(chunk_shape)=} does not match {len(buffer_shape)=} for dataset at location '{location_in_file}'!"
    190     )
    191 if len(buffer_shape) != len(full_shape):

TypeError: object of type 'NoneType' has no len()
```

### Operating System

macOS

### Python Executable

Python

### Python Version

3.8

### Package Versions

_No response_

### Code of Conduct

CodyCBakerPhD:
Also including in testing mixin to ensure all interfaces are able to write Zarr without a problem

Some new problems  discovered and debugged, commented in the PR below

fix #811


src/neuroconv/basedatainterface.py
```diff
@@ -75,6 +132,9 @@ def run_conversion(
         nwbfile: Optional[NWBFile] = None,
         metadata: Optional[dict] = None,
         overwrite: bool = False,
+        # TODO: when all H5DataIO prewraps are gone, introduce Zarr safely
+        # backend: Union[Literal["hdf5", "zarr"], HDF5BackendConfiguration, ZarrBackendConfiguration] = "hdf5",
```
CodyCBakerPhD, line 136: Including Zarr this early causes problems until we purge all `H5DataIO` usage across our tools, which will occur in one of the next follow-ups, after which Zarr will be re-enabled

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -260,9 +260,10 @@ def from_neurodata_object(cls, neurodata_object: Container, dataset_name: Litera
             )
             compression_method = "gzip"
         elif dtype == np.dtype("object"):  # Unclear what default chunking/compression should be for compound objects
-            chunk_shape = None
-            buffer_shape = None
-            compression_method = None
+            raise NotImplementedError(
```
CodyCBakerPhD, line 263: The other tools don't expect `None` here, nor does it make sense if it's a field that's going to be grabbed this way

To avoid configuring such a dataset, it simply shouldn't be parsed in this model to begin with

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -15,9 +16,39 @@ def configure_backend(
     """Configure all datasets specified in the `backend_configuration` with their appropriate DataIO and options."""
     nwbfile_objects = nwbfile.objects
 
+    # When a configuration is generated on an equivalent NWBFile elsewhere in memory, the object IDs will not match
+    # But a mapping can be formed to reconcile the two
+    # This is the case for the interface and converter usage of the feature (since they are separate workflow steps)
+    # As well as the NWB GUIDE (since they are separate endpoints in the Flask server)
```
CodyCBakerPhD, line 22: This block was the trickiest part of this, but a necessary one as explained

There was technically a second approach where each interface/converter is able to generate and maintain the in-memory `self.nwbfile` object to use across different steps of the workflow, but that could have all kinds of weird consequences and add more complexity to deal with inside each step, so I concluded the simplest approach was to just equivocate two configurations generated from equivalent file structures

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -36,6 +67,9 @@ def configure_backend(
         # Skip the setting of a DataIO when target dataset is a link (assume it will be found in parent)
         elif isinstance(nwbfile_object, TimeSeries) and is_dataset_linked:
             continue
+        # Skip the setting of a DataIO when target dataset is a reference type
+        elif isinstance(nwbfile_object, TimeSeriesReferenceVectorData):
```
CodyCBakerPhD, line 71: Interesting case caught by the now pretty extensive mixin testing

src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
```diff
@@ -100,18 +103,22 @@ def get_default_dataset_io_configurations(
...
-                    continue  # skip
+                    continue  # Skip
 
                 # Skip over columns that are already wrapped in DataIO
                 if isinstance(candidate_dataset, DataIO):
-                    continue
+                    continue  # Skip
+
+                # Skip over columns whose values are links, such as the 'group' of an ElectrodesTable
+                if any(isinstance(value, Container) for value in candidate_dataset):
```
CodyCBakerPhD, line 120: I guess the tools have only been used extensively for ophys projects - trying to use them on an electrodes table failed on the special `group` column

src/neuroconv/tools/nwb_helpers/_configure_backend.py
CodyCBakerPhD: I suppose an alternative approach could be to define this logic in something like a `.update_object_ids` method on the `BackendConfiguration` class - what do you think @h-mayorquin?

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -15,29 +15,64 @@ def configure_backend(
...
-        object_id = dataset_configuration.object_id
+        object_id = dataset_configuration_to_nwbfile_object_ids[dataset_configuration.object_id]
         dataset_name = dataset_configuration.dataset_name
         data_io_kwargs = dataset_configuration.get_data_io_kwargs()
 
         # TODO: update buffer shape in iterator, if present
 
-        nwbfile_object = nwbfile_objects[object_id]
-        is_dataset_linked = isinstance(nwbfile_object.fields.get(dataset_name), TimeSeries)
+        neurodata_object = nwbfile_objects[object_id]
```
CodyCBakerPhD, line 56: Noticed that the name in the other models (the ones actually exposed) are `neurodata_type` so changed this for consistency

I think all `nwbfile.objects` are technically 'neurodata objects' (instances of neurodata types) by definition, but whatevs

src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
```diff
@@ -100,17 +103,25 @@ def get_default_dataset_io_configurations(
...
                 # Skip over columns that are already wrapped in DataIO
                 if isinstance(candidate_dataset, DataIO):
+                    continue  # Skip
+
+                # Skip over columns whose values are links, such as the 'group' of an ElectrodesTable
+                if any(isinstance(value, Container) for value in candidate_dataset):
+                    continue  # Skip
+
+                # Skip when columns whose values are a reference type
+                if isinstance(column, TimeSeriesReferenceVectorData):
```
CodyCBakerPhD, line 124: This was an nteresting case caught by the now pretty extensive mixin testing - particular to the ABF interface over in `icephys` land

src/neuroconv/tools/nwb_helpers/_configure_backend.py
CodyCBakerPhD: Oops wrong place to skip it... turns out its a special kind of table column

CodyCBakerPhD:
Also caught and had to cute a plethora of `metadata` mutation issues - because the current workflow here (which again, I think is the simplest more modular approach) technically calls `add_to_nwbfile` twice on each interface, this caught some cases where the `add_to_nwbfile` method itself mutated the `metadata` dictionary. The first creation of the `NWBFile` went fine, but the second one couldn't get past metadata validation due to the mutation

src/neuroconv/basedatainterface.py
```diff
@@ -60,13 +67,63 @@ def validate_metadata(self, metadata: dict) -> None:
...
     def add_to_nwbfile(self, nwbfile: NWBFile, **conversion_options) -> None:
+        """
+        The protocol for adding this interfaces data to a PyNWB neurodata type in the in-memory pynwb.NWBFile object.
+
+        Parameters
+        ----------
+        nwbfile : pynwb.NWBFile
+            The in-memory object to add the data to.
+        **conversion_options
+            Additional keyword arguments to pass to the `.add_to_nwbfile` method.
```
bendichter, line 125:
```suggestion
```

This description provides no useful information

src/neuroconv/basedatainterface.py
```diff
@@ -60,13 +67,63 @@ def validate_metadata(self, metadata: dict) -> None:
...
+            The in-memory object with this interface's data added to it.
+        """
         nwbfile = make_nwbfile_from_metadata(metadata)
         self.add_to_nwbfile(nwbfile, metadata=metadata, **conversion_options)
         return nwbfile
 
     @abstractmethod
     def add_to_nwbfile(self, nwbfile: NWBFile, **conversion_options) -> None:
+        """
+        The protocol for adding this interfaces data to a PyNWB neurodata type in the in-memory pynwb.NWBFile object.
```
bendichter, line 118:
```suggestion
        Add data from this interface to an in-memory pynwb.NWBFile object.
```

CodyCBakerPhD: The keyword protocol adds some context that 'this is where the PyNWB code goes' relevant to describing _how_ the data of this interface _maps_ onto the chosen neurodata object representations

Perhaps

"""
Define a protocol for mapping the data from this interface to NWB neurodata objects and add those objects to the in-memory pynwb.NWBFile?
"""

bendichter:
Let's think about how we want to do this. If I understand correctly, the way you have it here, the NWBFile object is constructed twice, once to get the default dataset configs, and once again in run_conversion before write. That feels like an indirect approach and I think you should really only have to construct this object once. I believe it will also have the problem that the object IDs will be different in the two different approaches.

A more direct approach would be:

1. build the nwbfile
3. set the backend type and get default configs (maybe combined with 2)
4. update configs
5. write to disk

I think we could use the methods you already have that work directly with the nwbfile object, but this means we will need to allow for manipulation of the nwbfile object in between building and writing. So I guess that means we would need to  or provide a different pattern other than the run_conversion method for users that want to control the backend.

I think I may have said in a meeting that I wanted these methods *in* the DataInterface classes. Now I'm not so sure that's necessary. More accurately, we should define and document a workflow for how you would work with interfaces and converters, but let's try not to bend our workflow to the existing methods of the interfaces.

What about something like this?

```python
nwbfile = interface.create_nwbfile(metadata)
configs = get_default_backend_configuration(nwbfile, "hdf5") # or zarr
# update configs as desired
write_nwbfile(nwbfile, "path", configs)  # or something like that. Applies all dataset configs and then writes with the appropriate IO.
```

then you could use almost the exact same workflow for NWBConverter classes. Then the interfaces and converters are really for building the backend-agnostic version and the backend config is applied outside those classes.

CodyCBakerPhD:
> That feels like an indirect approach and I think you should really only have to construct this object once.

That was my first intuition as well, and as alluded to in https://github.com/catalystneuro/neuroconv/pull/801#discussion_r1548704523 there might be a way to do that in the current API

-> There is technically a second approach where each interface/converter is able to generate and maintain the in-memory `self.nwbfile` object as an attribute to re-use across different steps of the workflow

but my reservation from speculation...

-> is that could have all kinds of weird consequences and add more complexity to deal with inside each subsequent step of the workflow so I concluded the simplest approach was to just equivocate two configurations generated from equivalent in-memory file structures

which seems like a nifty feature to have anyway

Also worth noting that this type of multi-step back-and-forth workflow fits in really well with the GUIDE since these would be separate pages on the frontend and separate endpoints on the backend. We essentially require this kind of separation to provide a base web form for the user to interact with prior to final conversion

Also keep in mind that forming an in-memory NWBFile is a very lightweight operation - it's essentially just forming links to things that are either already lazy or already in-memory - the slow part is actually writing that in-memory object to disk, but we still only do that once anyhow


>  I believe it will also have the problem that the object IDs will be different in the two different approaches.

To confirm, the approach used by this PR has that behavior and has resolved it

The question is, is there any reason we do **not** want to support that as a general feature, regardless of how the specific interface/converter implementation details might be reworked?

That is, should a backend configuration model generated from one freshly created in-memory `pynwb.NWBFile` object apply to an 'equivalent' (all `nwbfile.object.values()` are the same) but not 'identical' (none of the `nwbfile.object.keys()` are the same) object?

If so the logic here has already been added to resolve the mapping and it wasn't too bad to do


> I think we could use the methods you already have that work directly with the nwbfile object, but this means we will need to allow for manipulation of the nwbfile object in between building and writing. So I guess that means we would need to or provide a different pattern other than the run_conversion method for users that want to control the backend.

Yeah, I've highlighted how that would work as-is (and how it _has_ recently been used in several conversions) here: https://neuroconv--802.org.readthedocs.build/en/802/user_guide/backend_control.html#interfaces-and-converters

It's not too bad either - but just note that the initial 'interactivity' step of actually knowing what object to modify and how to modify it is something that feels better outside of a context manager, and instead through a functional approach in a Ipython or Jupyter notebook type of environment as you indicate...



> I think I may have said in a meeting that I wanted these methods in the DataInterface classes. Now I'm not so sure that's necessary. 

It was also one of the cornerstones of the initial prototype: https://github.com/catalystneuro/neuroconv/pull/451/files#diff-98923ad563755c1598f8e651c54c666fe12b4790f2d5a6abaf362af06e7ded43R51

Just keep in mind that still, the biggest advantage to all of this is less about the fine-grain customization of individual fields (which is always going to require some pre-fetching and back-and-forth with the user) but more about how easy it now is to swap between HDF5 and Zarr (whereas previously we'd have had to find/replace all `H5DataIO` with `ZarrDataIO`, equivocate all compression methods and options somehow, and so on...)


> More accurately, we should define and document a workflow for how you would work with interfaces and converters,

Check out https://neuroconv--802.org.readthedocs.build/en/802/user_guide/backend_control.html#interfaces-and-converters


> but let's try not to bend our workflow to the existing methods of the interfaces.

I will point out one thing that is advantageous about the classic 'simple' workflow, one that looks like

```python
converter = MyConverter(...)

# Fetch and customize metadata
metadata  = converter.get_metadata()
metadata[...] = some_custom_metadata
...

# Fetch and customize backend
# Notice how similar this was to how we interact with metadata
backend_configuration = converter.get_backend_configuration(backend=backend)
backend_configuration[...] = some_customization
...

# Write file
converter.run_conversion(
    nwbfile_path=".../my_nwbfile.nwb",
    metadata=metadata,
    backend_configuration=backend_configuration,
)
```

I actually don't expect the majority of users to care about customizing the very specific details of individual datasets - that is very much an advanced feature for those who know and care about specific dynamic filters and their options

But I do expect more people over time to start caring about the backend type (HDF5 vs Zarr) and simply allowing all lower-level options to default to values that NeuroConv can, over time, adapt to our state-of-the-art recommendations (i.e., `wavpack` for `ElectricalSeries`, quantal flux for `PhotonSeries`, lindi video compression for `ImageSeries`, etc.)

Specifically, even if a user never directly calls `get_default_backend_configuration` or `configure_backend`, if a `backend_configuration` is not provided to `run_conversion`, then the default configuration can still be automatically generated and set from the basic call

```python
converter = MyConverter(...)

# Write file
converter.run_conversion(
    nwbfile_path=".../my_nwbfile.nwb",
    backend="zarr", 
    # backend_configuration gets fetched automatically behind the scenes just like metadata
)
```

h-mayorquin:
I also prefer @bendichter approach. I like the decoupling that is emphasized here:

> Then the interfaces and converters are really for building the backend-agnostic version and the backend config is applied outside those classes.

`run_conversion` for converters then looks like this:

```python
    def run_conversion(
        self,
        nwbfile_path: Optional[str] = None,
        nwbfile: Optional[NWBFile] = None,
        overwrite: bool = False,
        # TODO: when all H5DataIO prewraps are gone, introduce Zarr safely
        # backend: Union[Literal["hdf5", "zarr"], HDF5BackendConfiguration, ZarrBackendConfiguration] = "hdf5",
        backend: Union[Literal["hdf5"], HDF5BackendConfiguration] = "hdf5",
        **conversion_options,
    ):
    
    # Create empty nwbfile if needed
    # Iterate over the interfaces and use `add_to_nwb`
    # Get the configuration if not passed
    # Write to the provided path if provided 
```

Which allows for all the other workflows that you mentioned in your response right? You can use the `add_to_nwb` method` of those objects to get the corresponding nwbfile, get the configuration, modify it if desired and pass it. 

We can still have an object method  `get_backend_configuration` as a convenience but we make  `nwbfile` a required argument instead of creating one to avoid the necessity of keeping the mapping you created in `configure_backend`.

I like that it also works well for a functional approach with multiple interfaces. Borrowing from Ben's example

```python
nwbfile = interface.create_nwbfile(metadata)
another_interface.add_to_nwb(nwbfile)
 configs = get_default_backend_configuration(nwbfile, "hdf5") # this gets the configuration for all the data.
# update configs as desired
write_nwbfile(nwbfile, "path", configs)  # or something like that. Applies all dataset configs and then writes with the appropriate IO.
```

CodyCBakerPhD:
OK so for this particular PR, two things to change:

a) make `nwbfile` required as an in-place for interface/converter retrieval of configurations (with API method just being aliases to tools to avoid need to explicitly import the tool methods)
b) splinter off cross-NWBFile equivalency logic to separate PR (since still needed for any GUIDE implementation but not technically for here)

?

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
h-mayorquin: What is an example of something that could have this dtype? What about text?

h-mayorquin:
@CodyCBakerPhD 
That sounds good to me. 

I also made a PR to get some of the bugs that you found into main as quickly as possible:
https://github.com/catalystneuro/neuroconv/pull/814

CodyCBakerPhD:
@h-mayorquin OK ready for review

Also updated the docs to correspond to this adjustment: https://neuroconv--802.org.readthedocs.build/en/802/user_guide/backend_control.html#interfaces-and-converters

src/neuroconv/basedatainterface.py
h-mayorquin: Maybe adding an example "such as" so the user gets an idea. Or how to locate and find which properties are available? What would remain here? Things like writing to processing or acquisition, right?

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
h-mayorquin: I see. Thanks for explaining.

src/neuroconv/tools/nwb_helpers/_configure_backend.py
h-mayorquin: This makes sense to me. I wish it was in a different PR to make the diff simpler but your time is also valuable.

h-mayorquin (commented):
Took another look but did not have enough time. There are some changes to testing that I need to check in more detail.

src/neuroconv/tools/nwb_helpers/_configure_backend.py
CodyCBakerPhD: Opened https://github.com/catalystneuro/neuroconv/pull/817

src/neuroconv/tools/nwb_helpers/_configure_backend.py
CodyCBakerPhD: Moved to https://github.com/catalystneuro/neuroconv/pull/816

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -84,24 +86,54 @@ def check_metadata(self):
         validate(metadata_for_validation, schema)
         self.check_extracted_metadata(metadata)
 
-    def check_run_conversion(self, nwbfile_path: str):
```
CodyCBakerPhD, line 87: 2 things here - I thought the name `check_run_conversion` was too generic since it made other assertions inside it besides just ensuring that the call to `.run_conversion` succeeded without error

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -84,24 +86,54 @@ def check_metadata(self):
...
+    def check_no_metadata_mutation(self):
```
CodyCBakerPhD, line 89: Other thing being that we don't need to actually write a file to disk to check this behavior - it's enough to just call the `.add_to_nwbfile` protocol and check the `metadata` argument in and out

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -84,24 +86,54 @@ def check_metadata(self):
...
         metadata["NWBFile"].update(session_start_time=datetime.now().astimezone())
 
         metadata_in = deepcopy(metadata)
 
+        nwbfile = mock_NWBFile()
+        self.interface.add_to_nwbfile(nwbfile=nwbfile, metadata=metadata, **self.conversion_options)
+
+        assert metadata == metadata_in
+
+    def check_run_conversion_default_backend(self, nwbfile_path: str, backend: Literal["hdf5", "zarr"] = "hdf5"):
```
CodyCBakerPhD, line 101: Whereas for all of these we should check they can write to disk successfully since changing the `DataIO` properties in all kinds of ways can produce errors potentially during `io.write`

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -84,24 +86,54 @@ def check_metadata(self):
...
+            backend_configuration=backend_configuration,
+            **self.conversion_options,
+        )
 
     @abstractmethod
     def check_read_nwb(self, nwbfile_path: str):
         """Read the produced NWB file and compare it to the interface."""
         pass
 
+    def check_basic_zarr_read(self, nwbfile_path: str):
```
CodyCBakerPhD, line 132: Enhancing this is obvious a TODO will it is released (following complete purge of `H5DataIO`)

tests/test_on_data/test_behavior_interfaces.py
```diff
@@ -333,17 +334,36 @@ class TestDeepLabCutInterface(DeepLabCutInterfaceMixin, unittest.TestCase):
 
     save_directory = OUTPUT_PATH
 
-    def check_run_conversion(self, nwbfile_path: str):
```
CodyCBakerPhD, line 336: This particular case showing how a general name like `check_run_conversion` can be misleading - this particular one was checking entirely different things from the parent, didn't call the parent, and could have been missed otherwise here as not actually checking a certain behavior

This can be a real problem with the amount of misdirection in the current mixin setup, but given the non-concurrent I/O for various interfaces I also cannot think of a better way

src/neuroconv/basedatainterface.py
CodyCBakerPhD: @h-mayorquin Ideally, there would be a thorough tutorial about this included in the docs on constructing an interface - that would be nice all around and then here we could say "For more details and examples, see <cross link>".

src/neuroconv/tools/testing/data_interface_mixins.py
h-mayorquin: Thanks. This is indeed better.

src/neuroconv/basedatainterface.py
```diff
@@ -108,6 +159,38 @@ def run_conversion(
...
+                    nwbfile=nwbfile_out, backend=backend, **conversion_options
+                )
+
+            configure_backend(nwbfile=nwbfile_out, backend_configuration=backend_configuration)
+
+    @staticmethod
+    def get_default_backend_configuration(
+        nwbfile: NWBFile,
+        backend: Literal["hdf5", "zarr"] = "hdf5",
+        **conversion_options,
```
h-mayorquin, line 179: why have conversion options here?

h-mayorquin (approved):
This looks good to me.

src/neuroconv/basedatainterface.py
CodyCBakerPhD, line 179:
```suggestion
```

CodyCBakerPhD: Whoops - good catch. Thought I had removed all of that

src/neuroconv/basedatainterface.py
```diff
@@ -108,6 +159,37 @@ def run_conversion(
...
+            backend=backend,
             verbose=getattr(self, "verbose", False),
         ) as nwbfile_out:
-            self.add_to_nwbfile(nwbfile_out, metadata=metadata, **conversion_options)
+            if backend_configuration is None:  # Otherwise Assume the data has already been added to the NWBFile
+                # Otherwise assume the data has already been added to the NWBFile
+                self.add_to_nwbfile(nwbfile_out, metadata=metadata, **conversion_options)
+
+                backend_configuration = self.get_default_backend_configuration(
+                    nwbfile=nwbfile_out, backend=backend, **conversion_options
```
CodyCBakerPhD, line 170:
```suggestion
                    nwbfile=nwbfile_out, backend=backend
```

src/neuroconv/basedatainterface.py
```diff
@@ -108,6 +159,35 @@ def run_conversion(
             nwbfile=nwbfile,
             metadata=metadata,
             overwrite=overwrite,
+            backend=backend,
             verbose=getattr(self, "verbose", False),
         ) as nwbfile_out:
-            self.add_to_nwbfile(nwbfile_out, metadata=metadata, **conversion_options)
+            if backend_configuration is None:  # Otherwise Assume the data has already been added to the NWBFile
+                # Otherwise assume the data has already been added to the NWBFile
```
CodyCBakerPhD, line 166:
```suggestion
            if backend_configuration is None:
                # In this case, assume the relevant data has already been added to the NWBFile
```

src/neuroconv/basedatainterface.py
```diff
@@ -108,6 +159,35 @@ def run_conversion(
...
+                self.add_to_nwbfile(nwbfile_out, metadata=metadata, **conversion_options)
+
+                backend_configuration = self.get_default_backend_configuration(nwbfile=nwbfile_out, backend=backend)
+
+            configure_backend(nwbfile=nwbfile_out, backend_configuration=backend_configuration)
+
+    @staticmethod
+    def get_default_backend_configuration(
+        nwbfile: NWBFile,
+        backend: Literal["hdf5", "zarr"] = "hdf5",
```
CodyCBakerPhD, line 176:
```suggestion
        # TODO: when all H5DataIO prewraps are gone, introduce Zarr safely
        # backend: Union[Literal["hdf5", "zarr"]],
        backend: Literal["hdf5"] = "hdf5",
```

src/neuroconv/basedatainterface.py
CodyCBakerPhD: Caught a couple other missed propagations, think that should be all of them

src/neuroconv/basedatainterface.py
```diff
@@ -92,13 +134,22 @@ def run_conversion(
         overwrite : bool, default: False
             Whether to overwrite the NWBFile if one exists at the nwbfile_path.
             The default is False (append mode).
+        backend : "hdf5" or a HDF5BackendConfiguration, default: "hdf5"
+            The type of backend to use when writing the file.
```
bendichter, line 138: This typing does not match line 118

src/neuroconv/basedatainterface.py
```diff
@@ -108,6 +159,37 @@ def run_conversion(
...
+            The in-memory object with this interface's data already added to it.
+        backend : "hdf5" or "zarr", default: "hdf5"
+            The type of backend to use when creating the file.
+
+        Returns
+        -------
+        backend_configuration : HDF5BackendConfiguration or ZarrBackendConfiguration
+            The default configuration for the specified backend type.
+        """
+        return get_default_backend_configuration(nwbfile=nwbfile, backend=backend)
```
bendichter, line 195: Is the idea to include this method so that we can override it if we want?

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -12,9 +12,19 @@
...
 ) -> None:
-    """Configure all datasets specified in the `backend_configuration` with their appropriate DataIO and options."""
+    """
+    Configure all datasets specified in the `backend_configuration` with their appropriate DataIO and options.
+
+    Parameters
+    ----------
+    nwbfile : pynwb.NWBFile
+        The in-memory pynwb.NWBFile object to configure.
+    backend_configuration : "hdf5" or "zarr", default: "hdf5"
```
bendichter, line 22: The typing here does not match the typehints

src/neuroconv/basedatainterface.py
CodyCBakerPhD: That is indeed one such approach, particularly for custom lab specific interfaces - for most generic overrides (such as WavPack for Ecephys raw data) I think it would be more general to do it at the neurodata object level however, in the main dataset configuration method

src/neuroconv/tools/nwb_helpers/_configure_backend.py
CodyCBakerPhD: Good eye

https://github.com/catalystneuro/neuroconv/pull/822

src/neuroconv/basedatainterface.py
CodyCBakerPhD: Good eye

https://github.com/catalystneuro/neuroconv/pull/822

PR #802, opened 2024-04-02
[Backend Configuration Va] Basic user documentation

### LINKED ISSUE #797: [Documentation]: add docs for advanced backend support

We need docs for:

* specifying a backend
* specifying compression settings for individual datasets

Yes.

### Code of Conduct

CodyCBakerPhD:
fix  #797


docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,286 @@
...
+.. code-block:: bash
+
+    Configurable datasets identified using the hdf5 backend
+    -------------------------------------------------------
+
+    acquisition/MyTimeSeries/data
+    -----------------------------
+      dtype : float64
+      full shape of source array : (3,)
+      full size of source array : 0.00 GB
```
bendichter, line 63: I'm not crazy about printing sizes like this. I'll propose a small PR to print size with the appropriate units.

CodyCBakerPhD: A planned eventual follow-up is the auto-scaling into 'reasonable' units - feel free to take that if you want (I know I've seen you do that kind of logarithm based stuff before)

CodyCBakerPhD: But yes, it's independent of this PR which is just to get basic docs up

CodyCBakerPhD: Also planning a follow-up with HTML repr for analogous usage in notebooks, was going to let @h-mayorquin handle that due to their growing familiarity with the approach

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,286 @@
+Backend Configuration
+=====================
+
+NeuroConv offers highly convenient control over the type of file backend and the way its datasets are configured.
+
+Find out more about possible file types in the `main NWB documentation <https://nwb-overview.readthedocs.io/en/latest/faq_details/why_hdf5.html#why-use-hdf5-as-the-primary-backend-for-nwb>`_.
```
h-mayorquin, line 6:
```suggestion
Find out more about possible backend formats in the `main NWB documentation <https://nwb-overview.readthedocs.io/en/latest/faq_details/why_hdf5.html#why-use-hdf5-as-the-primary-backend-for-nwb>`_.
```

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,286 @@
...
+.. code-block:: python
+
+    from datetime import datetime
+    from uuid import uuid4
+
+    from dateutil import tz
+    from neuroconv.tools.nwb_helpers import get_default_backend_configuration
+    from pynwb import NWBFile, TimeSeries
+
+    session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=tz.gettz("US/Pacific"))
```
h-mayorquin, line 28:
```suggestion
    session_start_time = datetime(2020, 1, 1, 12, 30, 0)
```
I guess we don't need this anymore.

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,286 @@
...
+      chunk shape : (3,)
+      disk space usage per chunk : 0.00 MB
+
+      compression method : gzip
+
+
+Customization
+-------------
+
+To modify the chunking or buffering patterns and compression method or options, change those values in the ``.dataset_configurations`` object using the location of each dataset as a specifier.
```
h-mayorquin, line 92: I am confused mysefl about this. Do compression options have buffering patterns? I thought the buffering was only an iterator concept and the compressors rely only on chunks. am I confused?

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,286 @@
...
+
+    Core fields such as the maximum shape and data type of the source data cannot be altered using this method.
+
+.. note::
+
+    The ``buffer_shape`` must be a multiple of the ``chunk_shape`` along each axis.
+
+.. note::
+
+    You can see what compression methods are available on your installation by examining the following...
```
h-mayorquin, line 117: What about the options for each method? Should we add a comment on how to fetch or figure out the options for each method?

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,286 @@
...
+        folder_path=".../phy/phy_example_0"
+    )
+
+    data_interfaces = [ap_interface, phy_interface]
+    converter = ConverterPipe(data_interfaces=data_interfaces)
+
+    # Fetch available metadata
+    metadata = converter.get_metadata()
+
+    # Create the in-memory NWBFile object and retrieve a default configuration
```
h-mayorquin, line 191: I think this comment is making noise for me.

h-mayorquin: Ah, it is meant to apply to the following two lines.

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,286 @@
...
+    If you do not intend to make any alterations to the default configuration for the given backend type, then you can follow the classic workflow...
+
+    .. code-block::
+
+        converter = ConverterPipe(data_interfaces=data_interfaces)
+
+        # Fetch available metadata
+        metadata = converter.get_metadata()
+
+        # Create the in-memory NWBFile object and retrieve a default configuration
```
h-mayorquin, line 223: Here,  I think this is copied from the other structure but does not apply here, right?

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,286 @@
...
+Find out more about chunking and compression in the `advanced NWB tutorials for dataset I/O settings <https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/h5dataio.html#sphx-glr-tutorials-advanced-io-h5dataio-py>`_.
+
+Find out more about memory buffering of large source files in the `advanced NWB tutorials for iterative data write <https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/plot_iterative_write.html#sphx-glr-tutorials-advanced-io-plot-iterative-write-py>`_.
+
+
+
+Default configuration
+---------------------
+
+To retrieve a default configuration for an in-memory ``pynwb.NWBFile`` object, use the ``get_default_backend_configuration`` function...
```
h-mayorquin, line 17: I would first have a code block declaring the nwbfile as a working example, then this comment, and then the final code block:

```
    default_backend_configuration = get_default_backend_configuration(
        nwbfile=nwbfile, backend="hdf5"
    )
```

Otherwise, users have to look for the `get_default_backend_configuration` line which is almost at the end. In other words, if you are introducing this function it should come shortly after you talk about it. What do you think?

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,286 @@
...
+        converter.run_conversion(
+            nwbfile_path=nwbfile_path,
+            nwbfile=nwbfile,
+            backend=backend,
+        )
+
+    and all datasets in the NWB file will automatically use the default configuration!
+
+
+Generic tools
```
h-mayorquin, line 237: I can't make sense of this section. What does it provide that the other approaches do not? Maybe we can have that in the first paragraph. Something like:

Maybe you want to do x and that is not possible with the other approaches, to solve your problem you can use the following approach:

docs/user_guide/backend_configuration.rst
h-mayorquin: I think I will do this in general, that is, in the other sections I will create an example and say that we are working with this data, and then have a small code block where I only have the function that I want to explain.

One downside of my proposal is that is harder to copy paste but I am thinking this as an explainer rather than as a how-to so I think this makes sense.

h-mayorquin (commented):
Gave it a first read. I think this looks great and solves the issue. I am happy to accept this to merge and we can work on improving it later as it is. I do have some suggestions but I don't think they are critical.

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: At this point there is no strategy for automatic fetching or even 'figuring out' easily, so I wouldn't even mention it

Eventually maybe we can put together a few hand written schemas for these

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: What is confusing and what rephrasing would make it less confusing?

'the chunking or buffering patterns' *and* 'compression method or options'

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: What about this does not apply? This is just the bare minimum snippet to showcase the one new keyword argument for `backend : Literal["hdf5", "zarr"]` automatically configures things without custom overrides or explicit configuration passed

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: This section showcases how it can be used in conjunction with non-NeuroConv approaches (any in-memory NWBFile)

The others use the built-in class methods, not the base tool method

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: OK well see what it looks like in followup

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
...
+    time_series = TimeSeries(
+        name="MyTimeSeries",
+        description="A time series from my experiment.",
+        unit="cm/s",
+        data=[1., 2., 3.],
+        timestamps=[0.0, 0.2, 0.4],
+    )
+    nwbfile.add_acquisition(time_series)
+
+    with make_or_load_nwbfile(
```
h-mayorquin, line 268: My other suggestion here is to simplify this using the pynwb mock functions. I think that the resulting code will be simpler and that will help to emphasize the central point here, how you can use `configure_backend` with any nwbfile to determine the way in which the file is written.

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
...
+        nwbfile_path=nwbfile_path,
+        nwbfile=nwbfile,
+        overwrite=True,
+        backend="hdf5",
+        verbose=True,
```
h-mayorquin, line 273: is the verbose necessary, isn't it a default right now?  (asking out of curiosity, not prompting a change)

h-mayorquin (commented):
Great, added another round of review.

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: We're already linking to the closest possible page at the top; otherwise that info is scattered on various dynamic filter pages per method

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
...
+    If you do not intend to make any alterations to the default configuration for the given backend type, then you can follow the classic workflow...
+
+    .. code-block:: python
+
+        converter = ConverterPipe(data_interfaces=data_interfaces)
+
+        # Fetch available metadata
+        metadata = converter.get_metadata()
+
+        # Create the in-memory NWBFile object and retrieve a default configuration
```
CodyCBakerPhD, line 220:
```suggestion
        # Create the in-memory NWBFile object and apply the default configuration for HDF5
```

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: Oh I see now, the phrasing of the comment

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: I'm pretty sure this does adjust buffering and chunking values if changed on something already wrapped with an iterator, though have yet to add automatic wrapper of an iterator as a part of that

They are separate concerns; you specify chunking and buffering separately from compression. Performance and results of compression will indirectly depend on chunking pattern but does not rely on the specification directly

h-mayorquin: Do I understand well that adjusting the buffer_shape will have an effect only if the object is already wrapped in an iterator? So for example, things that we have build with `add_electrical_series`?

docs/user_guide/backend_configuration.rst
h-mayorquin: I see. OK, then for def out of scope.

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: The mocks are great for quick testing, but for a tutorial I'd rather be really explicit about how the data field is specified and all that, which is closer to how the user would create such an object (we don't want to encourage usage of mock classes for actual data conversion code)

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: right

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: right, in a follow-up extend to adjust anything sliceable at least

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
...
+
+    # Make any modifications to the configuration in this step, for example...
+    backend_configuration["acquisition/ElectricalSeriesAP/data"].compression_method = "Blosc"
+
+    # Configure and write the NWB file
+    nwbfile_path = "./my_nwbfile_name.nwb"
+    converter.run_conversion(
+        nwbfile_path=nwbfile_path,
+        nwbfile=nwbfile,
+        backend=backend,
```
bendichter, line 205: I feel like this arg shouldn't be necessary if we supply a `backend_configuration`.

docs/user_guide/backend_configuration.rst
bendichter: Yeah I agree with @CodyCBakerPhD here. It's better to not require the user to know about mocks here

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: Fair - opened https://github.com/catalystneuro/neuroconv/pull/840

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
+Backend Configuration
+=====================
+
+NeuroConv offers highly convenient control over the type of file backend and the way its datasets are configured.
```
bendichter, line 4:
```suggestion
NeuroConv offers convenient control over the type of file backend and the way each dataset is configured.
```

I prefer docs to be a bit more matter-of-fact. "highly" feels a bit sales-y to me here.

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
...
+Find out more about chunking and compression in the `advanced NWB tutorials for dataset I/O settings <https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/h5dataio.html#sphx-glr-tutorials-advanced-io-h5dataio-py>`_.
+
+Find out more about memory buffering of large source files in the `advanced NWB tutorials for iterative data write <https://pynwb.readthedocs.io/en/stable/tutorials/advanced_io/plot_iterative_write.html#sphx-glr-tutorials-advanced-io-plot-iterative-write-py>`_.
+
+
+
+Default configuration
+---------------------
+
+To retrieve a default configuration for an in-memory ``pynwb.NWBFile`` object, use the ``get_default_backend_configuration`` function...
```
bendichter, line 17:
```suggestion
To retrieve a default configuration for an in-memory ``pynwb.NWBFile`` object, use the ``get_default_backend_configuration`` function.
```

Are ellipses a common convention where docs transition to code? sklearn uses colons, e.g. [here](https://scikit-learn.org/stable/modules/linear_model.html#elastic-net). I'd be fine with that or just periods.

bendichter: Also, can we link to the API docs for get_default_backend_configuration?

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
...
+        data=[1., 2., 3.],
+        timestamps=[0.0, 0.2, 0.4],
+    )
+    nwbfile.add_acquisition(time_series)
+
+    default_backend_configuration = get_default_backend_configuration(
+        nwbfile=nwbfile, backend="hdf5"
+    )
+
+From which a printout of the contents looks like...
```
bendichter, line 47:
```suggestion
From which a printout of the contents looks like:
```

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
...
+
+.. code-block:: python
+
+    print(default_backend_configuration)
+
+.. code-block:: bash
+
+    Configurable datasets identified using the hdf5 backend
+    -------------------------------------------------------
```
bendichter, line 56:
```suggestion
    HDF5 dataset configurations
    ------------------------------
```

Could we change the title to this?

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
...
+
+    acquisition/MyTimeSeries/data
+    -----------------------------
```
bendichter, line 59:
```suggestion
    ===========================
```

minor point, but if we are going to use RST syntax for headers we might as well format it so that it gets rendered properly in RST

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
...
+
+    Core fields such as the maximum shape and data type of the source data cannot be altered using this method.
+
+.. note::
+
+    The ``buffer_shape`` must be a multiple of the ``chunk_shape`` along each axis.
+
+.. note::
+
+    You can see what compression methods are available on your installation by examining the following...
```
bendichter, line 114:
```suggestion
Core fields such as the maximum shape and data type of the source data cannot be altered using this method. The ``buffer_shape`` must be a multiple of the ``chunk_shape`` along each axis.

You can see what compression methods are available on your installation by examining the following:
```

bendichter: I like to reserve the "note" feature for side comments and gotchas. I feel like this is more essential information so I think it goes better in the main text. How would I have a dataset be stored in contiguous mode? Is that possible? Would I delete the entry from `default_backend_configuration.dataset_configurations`?

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
...
+      compression method : Zstd
+      compression options : {'clevel': 3}
+
+
+Interfaces and Converters
+-------------------------
+
+The normal workflow when writing an NWB file using a ``DataInterface`` or ``NWBConverter`` is simple to configure.
+
+The following example uses the :ref:`example data <example_data>` available from the testing repo...
```
bendichter, line 159:
```suggestion
The following example uses the :ref:`example data <example_data>` available from the testing repo:
```

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,284 @@
...
+
+    .. code-block:: bash
+
+      {'gzip': 'gzip',
+       ...
+       'Zstd': hdf5plugin._filters.Zstd}
+
+    And likewise for ``AVAILABLE_ZARR_COMPRESSION_METHODS``.
+
+We can confirm these values are saved by re-printing that particular dataset configuration...
```
bendichter, line 130:
```suggestion
We can confirm these values are saved by re-printing that particular dataset configuration:
```

It's not clear from this that these configurations made it back into the larger object. It's also not clear how you would use this configuration to modify an in-memory NWBFile object. It would also be great to write this file and then show that the written file followed the chunking and compression parameters we chose. This could be done using HDFView, or some other app or API

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: > Also, can we link to the API docs for get_default_backend_configuration?

Done

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: This line is in the python object printout in console - the `print` is not intended to be RST format, the dashes are just a way of visually separating the datasets and their contents (similar to numpy docstring style, like what would be printed from `?` or `help`)

I gave it a try out of curiosity, but since it is a code block, RST rules do not apply

image

We could take it out of the code block to apply the headers as you say, but then the appropriate header characters would also be relative to _where_ in the docs (at what level of indentation) the printout is meant to be placed

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: Good point, I moved all of these and included your latest question into an FAQ section after the main 'how to' demonstrations

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: > It's not clear from this that these configurations made it back into the larger object.

This was obscured by the odd injection of all the notes; with those moved, I put the two printouts side by side and added some spacing

image



> It would also be great to write this file and then show that the written file followed the chunking and compression parameters we chose.

The point the demo here makes is just that they get saved in the `backend_configuration` object itself - I can add something to prove that it actually writes the file in that manner (for chunks and compression anyway, can't prove buffering that way)

Though I will probably just demonstrate it with `h5py` since mentioning/showcasing HDFView could lead to annoyance because if a user wanted to try it they'd need to sign up for HDF5 - also Neurosift doesn't display this information, though I'm sure you could just request that - could also use PyNWB HTML repr after https://github.com/hdmf-dev/hdmf/pull/1100 is merged, but this section is also currently for python script / ipython usage, not specific to a jupyter notebook

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: Would be in a separate PR

docs/user_guide/backend_configuration.rst
bendichter: that's ok we can just leave it as is

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,304 @@
...
+    nwbfile_path = "./my_nwbfile_name.nwb"
+    converter.run_conversion(
+        nwbfile_path=nwbfile_path,
+        nwbfile=nwbfile,
+        backend_configuration=backend_configuration,
+    )
+
+.. note::
+
+    If you do not intend to make any alterations to the default configuration for the given backend type, then you can follow the classic workflow...
```
bendichter, line 188:
```suggestion
    If you do not intend to make any alterations to the default configuration for the given backend type, then you can follow the classic workflow:
```

docs/user_guide/backend_configuration.rst
```diff
@@ -0,0 +1,304 @@
...
+
+**Can I specify a buffer shape that incompletely spans the chunks?**
+
+The ``buffer_shape`` must be a multiple of the ``chunk_shape`` along each axis.
+
+This was found to give significant performance increases compared to previous data iterators that caused repeated I/O operations through partial chunk writes.
+
+**How do I disable chunking and compression completely?**
+
+To completely disable chunking (i.e., 'contiguous' layout), set both ``chunk_shape=None`` and ``compression_method=None``.
```
bendichter, line 302:
```suggestion
To completely disable chunking for HDF5 backends (i.e., 'contiguous' layout), set both ``chunk_shape=None`` and ``compression_method=None``.
```

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: Added a section to the FAQ for how to check that customizations (or the default for that matter) actually configure the item correctly in the resulting file

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: in #842

docs/user_guide/backend_configuration.rst
bendichter: OK great, could you please add a section on writing the data and confirming (somehow) that the data was written with the proper settings?

CodyCBakerPhD: I did, it's the last section of the FAQ. Uses the file written using the preceeding section

bendichter: OK that section of the FAQ looks great, but I still feel like it would be nice to have another snippet here showing how to apply the configs to an in-memory NWBFile object and write to disk

CodyCBakerPhD: I don't understand what you mean here - what are you imagining and how is it different from the preceding section?

docs/user_guide/backend_configuration.rst
h-mayorquin: Maybe not that important and this is what you meant but you can pass  `chunks=None` to zarr and get all the data in one file:

```
import zarr

# Create a Zarr group
zarr_folder = zarr.open_group('test_zarr.zarr', mode='w')

# Create an array and add it to the Zarr group
array_data = np.random.rand(20, 20, 20)
zarr_array = zarr_folder.create_dataset('array', data=array_data, chunks=None)
zarr_array.info
```

<table class="zarr-info"><tbody><tr><th style="text-align: left">Name</th><td style="text-align: left">/array</td></tr><tr><th style="text-align: left">Type</th><td style="text-align: left">zarr.core.Array</td></tr><tr><th style="text-align: left">Data type</th><td style="text-align: left">float64</td></tr><tr><th style="text-align: left">Shape</th><td style="text-align: left">(20, 20, 20)</td></tr><tr><th style="text-align: left">Chunk shape</th><td style="text-align: left">(20, 20, 20)</td></tr><tr><th style="text-align: left">Order</th><td style="text-align: left">C</td></tr><tr><th style="text-align: left">Read-only</th><td style="text-align: left">False</td></tr><tr><th style="text-align: left">Compressor</th><td style="text-align: left">Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)</td></tr><tr><th style="text-align: left">Store type</th><td style="text-align: left">zarr.storage.DirectoryStore</td></tr><tr><th style="text-align: left">No. bytes</th><td style="text-align: left">64000 (62.5K)</td></tr><tr><th style="text-align: left">No. bytes stored</th><td style="text-align: left">56455 (55.1K)</td></tr><tr><th style="text-align: left">Storage ratio</th><td style="text-align: left">1.1</td></tr><tr><th style="text-align: left">Chunks initialized</th><td style="text-align: left">1/1</td></tr></tbody></table>

docs/user_guide/backend_configuration.rst
h-mayorquin: OK.

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: I did not check myself (trusted @bendichter on that one)

Our config literally just passes the values on to downstream calls so I assume this would work here, in which case we can remove this line as a minor point (even though `chunks=None` is the option, and even though the array says 1/1 chunks, that's all 'equivalent' to contiguous layout in a sense)

bendichter:
I haven't had a chance to run through the code. If it all runs as advertised, this looks good to me

CodyCBakerPhD:
> If it all runs as advertised, this looks good to me

I ran through it all again - fixed a couple minor things now that the upstream ecephys change is through - confirmed everything works as advertised

We should keep in the back of our minds how to ensure this into the future however - doctest doesn't really work well for this scenario since we broke up various steps for the sake of readability (and doctest needs the variables to be in the same code block), which is more of a style that works well with using a notebook as a doc base (as PyNWB tutorials do) but we've not done that yet on NeuroConv

bendichter (approved):

PR #803, opened 2024-04-02
Do not write `offset_to_uV` and `gain_to_uV` in `Electrodes`

h-mayorquin:
As in the title. This includes a test and now I modified the code to pass most of the cases. However, there is something on the scaled tests for BlackRock and other formats which indicates that they were relying on those columns for correct gain. This is probably wrong and I will investigate further on the following days. 

Plus, I eliminate a keyword that was set to a mutable value, a well know footgun in python

### DIFF
--- a/src/neuroconv/tools/spikeinterface/spikeinterface.py
+++ b/src/neuroconv/tools/spikeinterface/spikeinterface.py
@@ -175,7 +175,12 @@ def add_electrode_groups(recording: BaseRecording, nwbfile: pynwb.NWBFile, metad
             nwbfile.create_electrode_group(**electrode_group_kwargs)
 
 
-def add_electrodes(recording: BaseRecording, nwbfile: pynwb.NWBFile, metadata: dict = None, exclude: tuple = ()):
+def add_electrodes(
+    recording: BaseRecording,
+    nwbfile: pynwb.NWBFile,
+    metadata: dict = None,
+    exclude: Optional[tuple] = None,
+):
     """
     Add channels from recording object as electrodes to nwbfile object.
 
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
> Plus, I eliminate a keyword that was set to a mutable value, a well know footgun in python

Which keyword argument was mutable?

Note that a `tuple` is not mutable

```python
def test_tuple_function(tup: tuple = ()) -> tuple:
    new_tup = tup + (2,)
    tup += (1,)  # attempt to mutate input
    return new_tup

base_tup = (0,)
returned_tup = test_tuple_function(tup=base_tup)

base_tup 
> (0,)

returned_tup
> (0, 2)
```

CodyCBakerPhD:
Let me know when tests are passing

h-mayorquin:
You are totally right about the tuple, @CodyCBakerPhD. My mistake.

CodyCBakerPhD:
It's also a lot easier schematically to deal with an argument that's only one type (even `minItems=0` in this case) due to annoyances in how `None` maps to JSON and gets loaded back into JS/TS for the GUIDE

Much simpler also to not have to toggle between `tuple` and `None` options in the web form component for fields like that, so keep an eye out and let me know if you see any signatures that are like `list | None` since I'm sure in most cases they could be simplified to just `tuple = ()`

When we refactor the schema parser to rely on Pydantic it will also become `Tuple[str] = Field(min_length=0)`

h-mayorquin:
All right, the failure comes from a bug in spikeinterface:
https://github.com/SpikeInterface/spikeinterface/pull/2661

I already pushed a fix but this will take another release. I will be changing the test here to use the pynwb API which I think it make sense for us as that is something that we care about.

PR #805, opened 2024-04-03
[Backend Configuration VIa] Removal of H5DataIO for ecephys

CodyCBakerPhD:
@h-mayorquin How do you think we should handle testing for this? Aside from removal of the current tests related to compression through those tools


h-mayorquin:
Yes, I think just removing them. We already have tests for this on the test of the configurations, right?

CodyCBakerPhD:
@h-mayorquin The integration level tests via the interfaces and converters are the way those specific behaviors get tested

h-mayorquin:
It seems that I need to take a look at the current tests again then. Let me take a look.

CodyCBakerPhD:
And by 'tested' I mean 'succeeds without error'

The unit tests for a dataset configuration stuff is what ensures the actual specification of a configuration ends up configuring the dataset on disk correctly

CodyCBakerPhD:
This is now ready I think

CodyCBakerPhD:
@h-mayorquin This one is independent of the docs PR and should be a quick and easy one to take a look at

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -570,6 +569,17 @@ def add_electrical_series(
...
     """
+    # TODO: remove completely after 10/1/2024
+    if compression is not None or compression_opts is not None:
+        warnings.warn(
+            message=(
+                "Specifying compression methods and their options at the level of tool functions has been deprecated. "
+                "Please use the `configure_backend` tool function for this purpose."
+            ),
+            category=DeprecationWarning,
+            stacklevel=2,
```
h-mayorquin, line 580: Just to confirm, how are you thinking about the `stacklevel=2` here? I understand that this will raise the warning at the level of the function that calls this one (instead of raising it when you call this function). I guess that this makes sense because we want to let know users that use this function through the interface for example, right?

h-mayorquin (commented):
Should we also change the arguments here:

https://github.com/catalystneuro/neuroconv/blob/a7b684a2337c2b1f5c7e3849ce34d3049a635093/src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py#L288-L289

And in the corresponding LFP.

Also, should we add tests here:
https://github.com/catalystneuro/neuroconv/blob/a7b684a2337c2b1f5c7e3849ce34d3049a635093/tests/test_ecephys/test_ecephys_interfaces.py#L28-L50

To check that run conversion still works the way that we expect it? (it wraps thing by default) Not so sure about this last one, what do you think?

src/neuroconv/tools/spikeinterface/spikeinterface.py
CodyCBakerPhD: Without it the printout doesn't mention what line actually triggered the warning which makes it harder to tell where and how to correct it; that's why we add `stacklevel=2` pretty much everywhere now

CodyCBakerPhD:
Good catch, modified the base interfaces

Technically everything still would have worked, just would have thrown the warning

IDK about the test, all the interfaces are already tested in that way now due to the autoconfigured default backend

h-mayorquin:
> IDK about the test, all the interfaces are already tested in that way now due to the autoconfigured default backend

In the light of the discussion here: 
https://github.com/catalystneuro/neuroconv/pull/802#discussion_r1589694694

I wouldl ike to add a test to see if `buffer_shape` is correctly propagated.

src/neuroconv/tools/spikeinterface/spikeinterface.py
h-mayorquin: And by the line that actually trigger it you mean the function that we are calling that ended up using the function where the warning is declared right?

Otherwise it just gives you the warning, right?

CodyCBakerPhD:
That's a separate (and more general) test of any backend configuration though

And this PR is only about removing compression method and options

src/neuroconv/tools/spikeinterface/spikeinterface.py
h-mayorquin: I don't know why I am asking you this, sorry. I can test this myself quickly. Thanks for indulging me : P

CodyCBakerPhD: You got it. Nothing more annoying than having to ctrl+F on a GitHub repo looking for a warning string when I should be able to click the line in an IDE

h-mayorquin:
> That's a separate (and more general) test of any backend configuration though
> 
> And this PR is only about removing compression method and options

Are we testing this somwere else now? That the buffer shape is propagated to an object that is wrapped already?

CodyCBakerPhD:
Albeit indirectly in https://github.com/catalystneuro/neuroconv/blob/main/tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_overrides.py#L45-L46, yes - because the alteration to the chunk shape would otherwise throw off the previous buffer shape

We can enhance and further specify this aspect in the follow-up that adds automatic wrapping I think

src/neuroconv/tools/spikeinterface/spikeinterface.py
h-mayorquin:
```python
import warnings

def function3(): 
    
    function2()

def function2():
    # This function calls function2
    function1()

def function1():
    # This will issue a warning with stacklevel=1
    warnings.warn("This is a stacklevel=1 warning", stacklevel=1)
    # This will issue a warning with stacklevel=2
    warnings.warn("This is a stacklevel=2 warning", stacklevel=2)

# Call function1, which in turn calls function2
function3()
```

Ouptut:
```
/tmp/ipykernel_260546/1160872622.py:13: UserWarning: This is a stacklevel=1 warning
  warnings.warn("This is a stacklevel=1 warning", stacklevel=1)
/tmp/ipykernel_260546/1160872622.py:9: UserWarning: This is a stacklevel=2 warning
  function1()
```

h-mayorquin (approved):
OK, this makes sense.

PR #806, opened 2024-04-04
Brain wide map link

h-mayorquin:
This will send to the general page of the project. I don't know what the current link was supposed to do.

### DIFF
--- a/docs/user_guide/expand_path.rst
+++ b/docs/user_guide/expand_path.rst
@@ -214,8 +214,8 @@ The information found for the first matching directory is similar.
         }
     }
 
-Example 3: `IBL Brain Wide Map Data <https://ibl.flatironinstitute.org/public>`_
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Example 3: `IBL Brain Wide Map Data <https://www.internationalbrainlab.com/data>`_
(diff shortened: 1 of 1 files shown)


CodyCBakerPhD:
It's supposed to lead to a file directory structure showing their source storage pattern

CodyCBakerPhD:
Thank goodness we copied the directory structure explicitly into the example (minus the `...`) but IDK what good the generic link will do for helping explain the use case - maybe just no link at all then?

CodyCBakerPhD:
Oh, IDK - the link still works for me: https://ibl.flatironinstitute.org/public/

h-mayorquin:
It does work for me locally. Maybe it should go on the ignore list, most likely they blocked access to bots.

CodyCBakerPhD:
The CI checker on other PRs has been flickering but not regularly failing - usually if I rerun it once it's fixed itself, so might just be slow server response times or something

h-mayorquin:
All right, let's give it some time. I think that if it keeps being annoying for a month or so, then the link should get the honor of being in the ignored list.

CodyCBakerPhD:
@h-mayorquin BTW the IBL link here has been stable for a while it seems so I think we can close

Problem lately has been SSL certificate of FicTrac: https://github.com/catalystneuro/neuroconv/actions/runs/8649412894/job/23715562511?pr=800#step:6:456

This is why I usually give it a week or more of consistent failures before even thinking about opening a PR to try to address them ourselves

h-mayorquin:
OK, a week of consistent failure seems like a good criteria to open a PR. That's great. Let's close.

PR #809, opened 2024-04-09
Add interface for single and multi-file ScanImage TIFF files

weiglszonja:
Adds specialised interface for reading multi-file (buffered) TIFF files produced via ScanImage.

Follow-up on https://github.com/catalystneuro/roiextractors/pull/297

- [x] add `ScanImageSinglePlaneMultiFileImagingInterface`
- [x] handle multi-channel metadata (naming of `TwoPhotonSeries`, links to correct `ImagingPlane` and `OpticalChannel`)
- [x] add tests
- [x] add `ScanImageMultiPlaneMultiFileImagingInterface`
- [x] add tests
- [x] add to docs


src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
```diff
@@ -86,3 +87,256 @@ def get_metadata(self) -> dict:
...
+    ----------
+    scanimage_metadata : dict
+        Dictionary of metadata extracted from a TIFF file produced via ScanImage.
+
+    Returns
+    -------
+    version : str
+        The version of ScanImage that produced the TIFF file.
+    """
+    if "SI.VERSION_MAJOR" in scanimage_metadata:
```
pauladkisson, line 337: For the most recent scan image files, the full version should be `f"{SI.VERSION_MAJOR}.{SI.VERSION_MINOR}.{SI.VERSION_UPDATE}"` but before 2020 they just had `f"{SI.VERSION_MAJOR}.{SI.VERSION_MINOR}"` and before 2019 they just had `f"{SI.VERSION_MAJOR}"` and before 2016 it was `f"{state.software.version}"` as you know (see [here](https://www.mbfbioscience.com/product-resources/scanimage-resources/scanimage-version-history) for details).

With all that in mind, let's just rename this function to `get_scanimage_major_version` so it's more clear.

tests/test_on_data/test_imaging_interfaces.py
```diff
@@ -71,6 +75,136 @@ def check_extracted_metadata(self, metadata: dict):
         )
 
 
+class TestScanImageSinglePlaneMultiFileImagingInterface(ScanImageSinglePlaneMultiFileImagingInterfaceMixin, TestCase):
+    data_interface_cls = ScanImageSinglePlaneMultiFileImagingInterface
+    interface_kwargs = [
```
pauladkisson, line 80: `interface_kwargs` should be parameterized by `parameterized.parameterize_class` -- see [discussion on slack](https://catalystneuro.slack.com/archives/C01FAC3F345/p1712161546819629)

src/neuroconv/datainterfaces/ophys/requirements.txt
```diff
@@ -1 +1 @@
-roiextractors>=0.5.5
+roiextractors @ git+https://github.com/catalystneuro/roiextractors.git@main
```
pauladkisson, line 1: Should be able to use `roiextractors>=0.5.7` now

docs/conversion_examples_gallery/conversion_example_gallery.rst
```diff
@@ -70,6 +70,7 @@ Imaging
     Miniscope <imaging/miniscope>
     Scanbox <imaging/scanbox>
     ScanImage <imaging/scanimage>
+    ScanImage Multi-File <imaging/multifilescanimage>
```
pauladkisson, line 73: For the conversion gallery, I think it makes more sense to combine multifilescanimage and single file scan image in a single page with different headings. So, instead of two different pages for scan image with two subheadings each, we would have:
- Convert single plane single file imaging data
- Convert multi-plane (volumetric) single file imaging data
- Convert single plane multi-file (buffered) imaging data
- Convert multi-plane (volumetric) multi-file (buffered) imaging data

pauladkisson (changes requested):
Overall, this looks great!

I made a suggestion about re-organizing the conversion gallery, and few nits but otherwise this is looking good.

I think while we're at it we should add interfaces for the single file `ScanImageTiffSinglePlaneImagingExtractor` and `ScanImageTiffMultiPlaneImagingExtractor` if you don't mind including that in this PR.

tests/test_on_data/test_imaging_interfaces.py
```diff
@@ -75,20 +77,26 @@ def check_extracted_metadata(self, metadata: dict):
...
+        },
+        {
+            "interface_kwargs": {
+                "folder_path": str(OPHYS_DATA_PATH / "imaging_datasets" / "ScanImage"),
+                "file_pattern": "scanimage_20240320_multifile*.tif",
+                "channel_name": "Channel 2",
+            }
+        },
+    ]
+)
```
weiglszonja, line 97: @pauladkisson tried out your suggestion but I got this error:
```python-traceback
src/neuroconv/tools/testing/data_interface_mixins.py:105 (TestScanImageSinglePlaneMultiFileImagingInterface.test_conversion_as_lone_interface)
self = <tests.test_on_data.test_imaging_interfaces.TestScanImageSinglePlaneMultiFileImagingInterface testMethod=test_conversion_as_lone_interface>

    def test_conversion_as_lone_interface(self):
>       interface_kwargs = self.interface_kwargs
E       AttributeError: 'TestScanImageSinglePlaneMultiFileImagingInterface' object has no attribute 'interface_kwargs'
```
is there something that has to be done differently for `interface_kwargs` compared to `conversion_options` to make this work?

pauladkisson: OK, I tracked down what's going wrong here. When you use `parameterized.parameterize_class` on a test case with Mixins, pytest will instantiate a class object for each parameter --> run all tests for each test class object AND instantiate an unparameterized class object and run only the Mixin tests for that object.  See dummy example below:
```python
class DebugMixin:
    def test_mixin(self):
        assert self.a

@parameterized_class(
    [
        {'a' : 1},
        {'a' : 2},
    ],
)
class TestDebug(TestCase, DebugMixin):
    def test_debug(self):
        assert self.a
```

This will generate the following output:
```bash
pytest tests/test_on_data/test_imaging_interfaces.py -k TestDebug -s --verbose                                                                                                                      1 ↵
============================================================================================ test session starts ============================================================================================
platform darwin -- Python 3.9.18, pytest-8.0.0, pluggy-1.4.0 -- /opt/anaconda3/envs/neuroconv_env/bin/python
cachedir: .pytest_cache
rootdir: /Users/pauladkisson/Documents/CatalystNeuro/NWB/neuroconv
configfile: pyproject.toml
plugins: cov-4.1.0
collecting ... Running GIN tests locally!
collected 61 items / 56 deselected / 5 selected                                                                                                                                                             

tests/test_on_data/test_imaging_interfaces.py::TestDebug_0::test_debug PASSED
tests/test_on_data/test_imaging_interfaces.py::TestDebug_0::test_mixin PASSED
tests/test_on_data/test_imaging_interfaces.py::TestDebug_1::test_debug PASSED
tests/test_on_data/test_imaging_interfaces.py::TestDebug_1::test_mixin PASSED
tests/test_on_data/test_imaging_interfaces.py::TestDebug::test_mixin FAILED

================================================================================================= FAILURES ==================================================================================================
___________________________________________________________________________________________ TestDebug.test_mixin ____________________________________________________________________________________________

self = <tests.test_on_data.test_imaging_interfaces.TestDebug testMethod=test_mixin>

    def test_mixin(self):
>       assert self.a
E       AttributeError: 'TestDebug' object has no attribute 'a'

tests/test_on_data/test_imaging_interfaces.py:81: AttributeError
============================================================================================= warnings summary ==============================================================================================
../../../../../../opt/anaconda3/envs/neuroconv_env/lib/python3.9/site-packages/hdmf/container.py:11
  /opt/anaconda3/envs/neuroconv_env/lib/python3.9/site-packages/hdmf/container.py:11: DeprecationWarning: 
  Pyarrow will become a required dependency of pandas in the next major release of pandas (pandas 3.0),
  (to allow more performant data types, such as the Arrow string type, and better interoperability with other libraries)
  but was not found to be installed on your system.
  If this would cause problems for you,
  please provide us feedback at https://github.com/pandas-dev/pandas/issues/54466
          
    import pandas as pd

src/neuroconv/tools/yaml_conversion_specification/_yaml_conversion_specification.py:7
src/neuroconv/tools/yaml_conversion_specification/_yaml_conversion_specification.py:7
  /Users/pauladkisson/Documents/CatalystNeuro/NWB/neuroconv/src/neuroconv/tools/yaml_conversion_specification/_yaml_conversion_specification.py:7: DeprecationWarning: jsonschema.RefResolver is deprecated as of v4.18.0, in favor of the https://github.com/python-jsonschema/referencing library, which provides more compliant referencing behavior as well as more flexible APIs for customization. A future release will remove RefResolver. Please file a feature request (on referencing) if you are missing an API for the kind of customization you need.
    from jsonschema import RefResolver, validate

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========================================================================================== short test summary info ==========================================================================================
FAILED tests/test_on_data/test_imaging_interfaces.py::TestDebug::test_mixin - AttributeError: 'TestDebug' object has no attribute 'a'
```

Notice how the mix in test is run one extra time with no `a` parameter, which causes the test to fail.

tests/test_on_data/test_imaging_interfaces.py
```diff
@@ -71,6 +77,142 @@ def check_extracted_metadata(self, metadata: dict):
...
+class TestScanImageSinglePlaneMultiFileImagingInterface(ScanImageSinglePlaneMultiFileImagingInterfaceMixin, TestCase):
+    data_interface_cls = ScanImageSinglePlaneMultiFileImagingInterface
+    save_directory = OUTPUT_PATH
```
pauladkisson, line 100:
```suggestion
@parameterized_class(
    [
        {"channel_name": "Channel 1"},
        {"channel_name": "Channel 2"},
    ],
)
class TestScanImageSinglePlaneMultiFileImagingInterface(ScanImageSinglePlaneMultiFileImagingInterfaceMixin, TestCase):
    data_interface_cls = ScanImageSinglePlaneMultiFileImagingInterface
    save_directory = OUTPUT_PATH
    interface_kwargs = dict(
        folder_path=str(OPHYS_DATA_PATH / "imaging_datasets" / "ScanImage"),
        file_pattern="scanimage_20240320_multifile*.tif",
    )
    channel_name = "Channel 1"

    def __init__(self, *args, **kwargs) -> None:
        self.interface_kwargs["channel_name"] = self.channel_name
        super().__init__(*args, **kwargs)
```

pauladkisson: This is how I would suggest to fix the problem.

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -796,3 +796,65 @@ def check_read_nwb(self, nwbfile_path: str):
             imaging_extractor = self.interface.imaging_extractor
             times_from_extractor = imaging_extractor._times
             assert_array_equal(one_photon_series.timestamps, times_from_extractor)
+
+
+class ScanImageSinglePlaneMultiFileImagingInterfaceMixin(DataInterfaceTestMixin, TemporalAlignmentMixin):
+    def check_read_nwb(self, nwbfile_path: str):
+        with NWBHDF5IO(nwbfile_path, "r") as io:
```
pauladkisson, line 803:
```suggestion
        if isinstance(self.interface_kwargs, dict):
            self.interface_kwargs = [self.interface_kwargs]
        with NWBHDF5IO(nwbfile_path, "r") as io:
```

pauladkisson (changes requested):

src/neuroconv/tools/testing/data_interface_mixins.py
pauladkisson: I also noticed this bug with the `check_read_nwb` function

tests/test_on_data/test_imaging_interfaces.py
weiglszonja: I see now, thank you for the explanation!

tests/test_on_data/test_imaging_interfaces.py
weiglszonja: Thank you for this suggestion!

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
```diff
@@ -1,88 +1,522 @@
...
+    def __init__(
+        self,
+        file_path: FilePathType,
+        fallback_sampling_frequency: Optional[float] = None,
+        verbose: bool = True,
+    ):
+        warnings.warn(
+            "ScanImageImagingInterface will be deprecated. Use ScanImageSinglePlaneImagingInterface instead.",
+            DeprecationWarning,
+        )
```
weiglszonja, line 521: what do you think about this @pauladkisson ?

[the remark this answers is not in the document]
pauladkisson: Just following the pattern from roiextractors here.

pauladkisson (changes requested):
Small change with the deprecation warning, otherwise looking great!

src/neuroconv/datainterfaces/__init__.py
```diff
@@ -137,7 +142,10 @@
     SbxImagingInterface,
     TiffImagingInterface,
     Hdf5ImagingInterface,
-    ScanImageImagingInterface,
```
CodyCBakerPhD, line 140: I would keep the classic ScanImageImagingInterface around for both compatibility and convenience - it can be the 'best' type of autodetector duck typer we can make (has fields for both `file_path`, `folder_path` but not both -> if `file_path` given, uses single file interface and tries to detect if the file is multiplane or not, otherwise `folder_path`and again whatever best attempt we can do to guess if it's multiplane)

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterface.py
CodyCBakerPhD, line 1: I'm personally partial to splitting these across files - or at the very least adding plural to this file to indicate it contains more than one interface

src/neuroconv/datainterfaces/__init__.py
CodyCBakerPhD: Or it looks like you are

https://github.com/catalystneuro/neuroconv/pull/809/files#diff-161e12dc1cb66083907ea7e54e4fdc8f149a15b354e08d61246384d677f82131R497

without the convenience idea

so I'd still keep the import around until it's completely deprecated?

CodyCBakerPhD: Or perhaps simpler to just restrict the `ScanImageImagingInterface` to be `file_path` routing to single vs. multi plane

then a `ScanImageMultiFileImagingInterface` that takes `folder_path` and route to single plane or multi-plane within multi-file

then no need to deprecate and should be simpler to add than my first idea

pauladkisson (commented):
LGTM

pauladkisson:
> I would keep the classic ScanImageImagingInterface around for both compatibility and convenience

I'm realizing that in our current implementations we still need `ScanImageTiffImagingExtractor` for reading legacy data (<=v3.8). So unless we refactor the new ScanImage classes to be able to support that, we won't ever be able to remove it...

Maybe we should take off the deprecation warnings? I'm not sure what the best practice is here.

CodyCBakerPhD:
If the current `ScanImageImagingExtractor` reads legacy data, then I'd rename the interface to `LegacyScanImageImagingInterface` (name of extractor I leave up to you)

This is the approach we took went with OpenEphys

CodyCBakerPhD:
And in that case we also kept the blanket name `OpenEphysRecordingInterface` to serve as routing between legacy and others: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ecephys/openephys/openephysdatainterface.py#L19

pauladkisson:
> If the current `ScanImageImagingExtractor` reads legacy data, then I'd rename the interface to `LegacyScanImageImagingInterface` (name of extractor I leave up to you)
> And in that case we also kept the blanket name OpenEphysRecordingInterface to serve as routing between legacy and others

That sounds good to me, and the router version of `ScanImageImagingInterface` would be designed such that it doesn't break back-compatibility right?

weiglszonja:
Thank you for the feedback @CodyCBakerPhD and @pauladkisson, I tried to organise the commits to be easier to follow the changes; let me know what you think. Now we have `ScanImageImagingInterface` and `ScanImageMultiFileImagingInterface` that are routing classes to single vs multi-plane (volumetric).

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -851,3 +851,114 @@ def check_read_nwb(self, nwbfile_path: str):
...
+        super().__init__(*args, **kwargs)
+
+    def check_read_nwb(self, nwbfile_path: str):
+        with NWBHDF5IO(nwbfile_path, "r") as io:
+            nwbfile = io.read()
+
+            assert self.imaging_plane_name in nwbfile.imaging_planes
+            assert self.photon_series_name in nwbfile.acquisition
+            two_photon_series = nwbfile.acquisition[self.photon_series_name]
+            assert two_photon_series.data.shape == (6, 256, 528, 2)
```
CodyCBakerPhD, line 948: Aren't these particular values specific to a single session in the example data? Shouldn't the "expected" values then be set in the main test suite?

If so, then this mixin could have attributes or a dict attribute for each expected value that might change, like `self.expected_two_photon_series_data_shape`, etc.

CodyCBakerPhD:
Down to one minor question on the testing setup, otherwise this is looking lovely

src/neuroconv/tools/testing/data_interface_mixins.py
weiglszonja: Ah yes, you're right.

weiglszonja: Okay, I think I managed to simplify the mixin classes with this.

tests/test_on_data/test_imaging_interfaces.py
```diff
@@ -53,22 +201,130 @@ def check_extracted_metadata(self, metadata: dict):
...
+    photon_series_name = "TwoPhotonSeriesChannel1"
+    imaging_plane_name = "ImagingPlaneChannel1"
+
+    def __init__(self, *args, **kwargs) -> None:
+        self.interface_kwargs["channel_name"] = self.channel_name
+        super().__init__(*args, **kwargs)
+
+    @classmethod
+    def setUpClass(cls):
+        cls.expected_two_photon_series_data_shape = (30, 512, 512)
```
CodyCBakerPhD, line 276: This could be written along with all other class attributes at the top, right?

tests/test_on_data/test_imaging_interfaces.py
```diff
@@ -53,22 +201,130 @@ def check_extracted_metadata(self, metadata: dict):
...
+    data_interface_cls = ScanImageMultiFileImagingInterface
+    interface_kwargs = dict(
+        folder_path=str(OPHYS_DATA_PATH / "imaging_datasets" / "ScanImage"),
+        file_pattern="scanimage_20240320_multifile*.tif",
+    )
+    save_directory = OUTPUT_PATH
+
+    channel_name = "Channel 1"
+    photon_series_name = "TwoPhotonSeriesChannel1"
+    imaging_plane_name = "ImagingPlaneChannel1"
```
CodyCBakerPhD, line 268:
```suggestion
    imaging_plane_name = "ImagingPlaneChannel1"
    
    expected_two_photon_series_data_shape = (30, 512, 512)
```

Maybe more readable by not needing the class method wrapper and `setUpClass`?

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterfaces.py
```diff
@@ -0,0 +1,687 @@
...
+from typing import Optional
+
+from dateutil.parser import parse as dateparse
+
+from ..baseimagingextractorinterface import BaseImagingExtractorInterface
+from ....utils import FilePathType, FolderPathType
+
+
+class ScanImageImagingInterface(BaseImagingExtractorInterface):
+    """Interface for ScanImage TIFF files."""
```
pauladkisson, line 13: I think it could be useful to give a quick description of the routing path here in the docstring.

pauladkisson (changes requested):

weiglszonja:
@CodyCBakerPhD I noticed the [API docs](https://neuroconv--809.org.readthedocs.build/en/809/api/interfaces.ophys.html#scanimage-imaging) is empty for ScanImage, do you have any idea how to fix this?

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterfaces.py
weiglszonja: Thanks, how about this: https://github.com/catalystneuro/neuroconv/pull/809/commits/20ee52eb7064fb48ddfb9c07b9a15f07372fc0c3

CodyCBakerPhD:
This looks great to me! Letting @pauladkisson make final decision

Fixed the API docs

weiglszonja:
> This looks great to me! Letting @pauladkisson make final decision
> 
> Fixed the API docs

Ah I see now, thank you!

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterfaces.py
```diff
@@ -0,0 +1,701 @@
...
+
+        file_paths = natsorted(Path(folder_path).glob(file_pattern))
+        first_file_path = file_paths[0]
+
+        image_metadata = image_metadata or extract_extra_metadata(file_path=first_file_path)
+        self.image_metadata = image_metadata
+
+        version = get_scanimage_major_version(scanimage_metadata=image_metadata)
+        if version == "3.8":
+            raise ValueError("ScanImage version 3.8 is not supported. Please use ScanImageImagingInterface instead.")
```
pauladkisson, line 369: this ValueError needs to be tested

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterfaces.py
```diff
@@ -0,0 +1,701 @@
...
+        parsed_metadata = parsed_metadata or parse_metadata(metadata=self.image_metadata)
+
+        if parsed_metadata["num_planes"] == 1:
+            raise ValueError(
+                "Only one plane detected. For single plane imaging data use ScanImageSinglePlaneImagingInterface instead."
+            )
+
+        avaliable_channels = parsed_metadata["channel_names"]
+        if channel_name is None:
+            if len(avaliable_channels) > 1:
```
pauladkisson, line 263: This case where the channel isn't properly specified needs to be tested

pauladkisson (changes requested):
Looking great! A couple missed test cases that I spotted from CodeCov, but after that gets fixed, this should be good to go.

PR #812, opened 2024-04-11
[Backend Configuration VIIa] Fix model schemas for GUIDE

CodyCBakerPhD:
cc @garrettmflynn thanks for catching this, looks like I overlooked certain programmatic typing and just falsely assumed Pydantic would automatically exclude them

Easy to specify the exclusion ourselves though

I still need to propagate to the other class but I'll try to get this ready today


garrettmflynn:
Started working with this on the GUIDE branch. Works as much as I need it to for now!

Though why are the `$defs` referring to the dataset properties placed inside `additional_properties` instead of just `properties`? And why is `dtype` omitted but other non-editable properties included?

CodyCBakerPhD:
> Though why are the $defs referring to the dataset properties placed inside additional_properties instead of just properties?

This is all auto-generated from Pydantic (who spend _a lot_ more of their time figuring out Python to JSON mappings than we do lol) so to understand anything about _how_ they make those decisions would have to peruse the docs and such over there

> And why is dtype omitted but other non-editable properties included?

This one I know - this is actually the cause of the previous error, because in the model it is strictly a `InstanceOf[np.dtype]` (not the `str` of that; strict instead of duck typing helps simplify the code in the various functions) - because that instance is a Pythonic object it doesn't here map to anything in JSON (gets skipped) - though I suppose a condition for decoding that particular field as a str could be possible if desired

garrettmflynn:
Gotcha thanks for the clarification! No dtype is fine, just noticed it was slightly inconsistent with my expectation

CodyCBakerPhD:
@h-mayorquin Basically, the current JSON schema retrieval from the Pydantic models fails on the Python-specific annotation types

Simple fix to the base classes and added tests for it all

bendichter:
@CodyCBakerPhD could you provide examples of where the current schema generation fails?

CodyCBakerPhD:
One such example, representative of the others, is the `compression_method` field of the `HDF5DatasetIOConfiguration` model, which allows any of

- `tuple(AVAILABLE_HDF5_COMPRESSION_METHODS.keys())`
  - The main way of selecting the method from a dropdown through the GUIDE, and the main `enum` exposed to the JSON schema
- `InstanceOf[h5py._hl.filters.FilterRefBase]`
  - Allowed for advanced programmatic users to specify the HDF5 class explicitly in their code if they wish - such as following the PyNWB tutorial on dynamic filters
  - This is the problematic one since a Python class has no JSON equivalent
- `None`
  - No compression

CodyCBakerPhD:
@h-mayorquin Also your attn

bendichter:
OK, I see. I think this is a representative example of how the in-memory classes and API are often not 1-1. Let's keep this in mind and maybe next time not use pydantic classes for base classes internally.

bendichter (approved):

h-mayorquin:
Sorry @CodyCBakerPhD . I had a busy last two weeks, glad @bendichter took a look at it. I think he is more familar with schema issues anyway.

PR #814, opened 2024-04-11
Avoid modifying metadata in-place in the `VideoInterface` and `AbfInterface`

h-mayorquin:
This was discovered by @CodyCBakerPhD in #801

I am pushing her to get this fix to the main branch as quick as possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 ### Bug fixes
 * Fixed writing waveforms directly to file [PR #799](https://github.com/catalystneuro/neuroconv/pull/799)
-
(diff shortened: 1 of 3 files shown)


CodyCBakerPhD:
What would you think about a test to add to the general mixin for ensuring metadata in equals metadata out after calling `add_to_nwbfile`?

CodyCBakerPhD:
Also CHANGELOG

h-mayorquin:
OK, I think even better. Let's start discussing how that would look like here:
https://github.com/catalystneuro/neuroconv/pull/815

Maybe it would be easier than what I think.

CodyCBakerPhD (approved):

h-mayorquin:
Thanks for changing the title!

PR #815, opened 2024-04-11
Add general test to avoid in-place modification of metadata by interfaces 

h-mayorquin:
OK, maybe something like this?

### DIFF
--- a/src/neuroconv/tools/testing/data_interface_mixins.py
+++ b/src/neuroconv/tools/testing/data_interface_mixins.py
@@ -137,11 +137,11 @@ class DataInterfaceTestMixin:
                 self.interface = self.data_interface_cls(**self.test_kwargs)
 
                 metadata = self.interface.get_metadata()
-                metadata_in = deepcopy(metadata)
                 metadata["NWBFile"].update(session_start_time=datetime.now().astimezone())
+
+                metadata_in = deepcopy(metadata)
                 nwbfile = self.interface.create_nwbfile(metadata=metadata)
 
-                # Test that no modification took place
                 assert metadata == metadata_in
 
 
--- a/src/neuroconv/tools/testing/data_interface_mixins.py
+++ b/src/neuroconv/tools/testing/data_interface_mixins.py
@@ -84,12 +84,18 @@ class DataInterfaceTestMixin:
         validate(metadata_for_validation, schema)
         self.check_extracted_metadata(metadata)
 
-    def run_conversion(self, nwbfile_path: str):
+    def check_run_conversion(self, nwbfile_path: str):
         metadata = self.interface.get_metadata()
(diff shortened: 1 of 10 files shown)


src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -123,6 +123,27 @@ def test_conversion_as_lone_interface(self):
                 # Any extra custom checks to run
                 self.run_custom_checks()
 
+    def test_add_to_nwb(self):
```
CodyCBakerPhD, line 126:
```suggestion
    def test_metadata_immutability(self):
```

CodyCBakerPhD: or 'test_no_metadata_mutation'

CodyCBakerPhD:
Yeah this is exactly what I was thinking

src/neuroconv/tools/testing/data_interface_mixins.py
h-mayorquin: I like both.

h-mayorquin: Did something like the second suggestion and added interface.

h-mayorquin:
There is some redundancy in this test as we don't need to check this for every case I feel. We could use our mock interfaces for some cases to avoid that excess but this should suffice for now.

CodyCBakerPhD:
> There is some redundancy in this test as we don't need to check this for every case I feel. 

An interface could override `get_metadata` (as they often do) and some unintentional aspect of that could do the same problem anywhere at any time, so I'd be in favor of exhaustive testing

CodyCBakerPhD (approved):

CodyCBakerPhD:
Oh my, looks like more fixes are needed

h-mayorquin:
OK, this is is a bigger problem that I don't have time to solve. Some of the formats (EDF) don't allow for concurrent access and either within the same test or because of pytest concurrency this is happening. Any suggestions so we can delay having to solve this at the format level?

CodyCBakerPhD:
> OK, this is is a bigger problem that I don't have time to solve. Some of the formats (EDF) don't allow for concurrent access and either within the same test or because of pytest concurrency this is happening. Any suggestions so we can delay having to solve this at the format level?

Interesting!

Maybe make it a `check_` instead of a `test_` and inject it into the main mixing test call?

h-mayorquin:
> > OK, this is is a bigger problem that I don't have time to solve. Some of the formats (EDF) don't allow for concurrent access and either within the same test or because of pytest concurrency this is happening. Any suggestions so we can delay having to solve this at the format level?
> 
> Interesting!
> 
> Maybe make it a `check_` instead of a `test_` and inject it into the main mixing test call?

Ended up doing something like this. I don't like it much but practice beats purity I guess.

PR #816, opened 2024-04-12
[Backend Configuration VIIb] Allow NWBFile equivalency for GUIDE

CodyCBakerPhD:
GUIDE requires separate endpoints on the backend for retrieving a serialized default backend configuration for a user to modify and thenr eturn before deserializing, configuring, and writing the file

This requires two separate NWBFiles that are
- **equivalent**: all nwbfile.object.values() are the same
- but not **identical**: none of the nwbfile.object.keys() are the same


CodyCBakerPhD:
Should probably add a few tests for this now that the main interface/converter approach doesn't rely on it

CodyCBakerPhD:
Take note of https://github.com/catalystneuro/neuroconv/pull/801/#discussion_r1548704523

garrettmflynn:
@CodyCBakerPhD Need some help merging some changes from `main`

CodyCBakerPhD:
@h-mayorquin OK, tests are passing so should be ready

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -11,6 +11,86 @@
...
+            dataset_configuration.location_in_file: dataset_configuration.object_id
+            for dataset_configuration in new_default_backend_configuration.dataset_configurations.values()
+        }
+
+        dataset_configuration_to_nwbfile_object_ids = {
+            dataset_configuration.object_id: locations_in_file_to_new_object_ids[dataset_configuration.location_in_file]
+            for dataset_configuration in backend_configuration.dataset_configurations.values()
+        }
+
+    nwbfile_object_ids = [nwbfile_object_id for nwbfile_object_id in nwbfile_objects.keys()]
```
h-mayorquin, line 66: Is this repeated from above?

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -111,6 +115,17 @@ def check_run_conversion_with_backend(self, nwbfile_path: str, backend: Literal[
...
+    def check_configure_backend_for_equivalent_nwbfiles(self, backend: Literal["hdf5", "zarr"] = "hdf5"):
+        metadata = self.interface.get_metadata()
+        if "session_start_time" not in metadata["NWBFile"]:
+            metadata["NWBFile"].update(session_start_time=datetime.now().astimezone())
+
+        nwbfile_1 = self.interface.create_nwbfile(metadata=metadata, **self.conversion_options)
+        nwbfile_2 = self.interface.create_nwbfile(metadata=metadata, **self.conversion_options)
+
+        backend_configuration = get_default_backend_configuration(nwbfile=nwbfile_1, backend=backend)
+        configure_backend(nwbfile=nwbfile_2, backend_configuration=backend_configuration)
```
h-mayorquin, line 127: What are we testing here for? I am looking for an assertion, is it just that it does not fail when you try to re-map? In other words, how would this test fai?

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -11,6 +11,86 @@
...
+        dataset_configuration.object_id in nwbfile_object_ids
+        for dataset_configuration in backend_configuration.dataset_configurations.values()
+    ]
+
+    # Default to identity
+    dataset_configuration_to_nwbfile_object_ids = {
+        dataset_configuration_object_id: dataset_configuration_object_id
+        for dataset_configuration_object_id in dataset_configuration_object_ids
+    }
+    if not all(
```
h-mayorquin, line 50: I prefer to use specific named variables for this long boolean conditions. Can we call the thing in parenthesis something like `all_ids_in_dataset_configuration_are_in_nwbfile`?

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -27,18 +107,20 @@ def configure_backend(
     is_ndx_events_installed = is_package_installed(package_name="ndx_events")
     ndx_events = sys.modules.get("ndx_events", None)
 
-    nwbfile_objects = nwbfile.objects
+    dataset_configuration_to_nwbfile_object_ids = _resolve_nwbfile_object_ids(
```
h-mayorquin, line 110: I would rather avoid having to call this for the most common neuroconv case. Given that this is an edge case can we have this as an external method that is called in those conditions. I don't think that is worth it to add more complexity to the function for this edge case that can be handled outside it.

h-mayorquin (commented):
Maybe we can add the nwbfile.identifier as an attribute to the backend configuration. That way we can easily test if the backend configuration might require a remapping and it "works" for the nwbfile that is to be configured.

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Test is if it doesn't error, like the other checks in the mixin final to file write

It would fail if some object in the `backend_configuration` was not in the `nwbfile_2` or had any other problem throughout the process

src/neuroconv/tools/nwb_helpers/_configure_backend.py
CodyCBakerPhD: Yep, cleaned

src/neuroconv/tools/nwb_helpers/_configure_backend.py
CodyCBakerPhD: Done

src/neuroconv/tools/nwb_helpers/_configure_backend.py
CodyCBakerPhD: I'd rather not waterbed this condition up to the converter; a function named `configure_backend` ought to do what it says minus the relatively minor detail of exactly equal Pydantic model structures minus randomly chosen object IDs

src/neuroconv/tools/testing/data_interface_mixins.py
h-mayorquin: I see, do you think that we would like to add a test where this fails to catch that it fails in the way that we intend it to fail (a location in the nwbfile is not present in the backend configuration)?

h-mayorquin (approved):
This is good to me:
1) I wonder whether we want to add a tes that makes this fails as mentioned in the comment.
2) I have a suggestion for an improvement for decoupling the logic of `configure_backend` from that of the re-mapping to handle with the edge case. The proposal is in #863

src/neuroconv/tools/nwb_helpers/_configure_backend.py
h-mayorquin: I think the proposal of #863 decouples backend_configuration from the re-maping one extra step.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -93,7 +93,7 @@ class DatasetIOConfiguration(BaseModel, ABC):
     model_config = ConfigDict(validate_assignment=True)  # Re-validate model on mutation
 
     # Immutable fields about the dataset
-    object_id: str = Field(description="The UUID of the neurodata object containing the dataset.", frozen=True)
+    object_id: str = Field(description="The UUID of the neurodata object containing the dataset.")
```
CodyCBakerPhD, line 96: Is this still needed anymore? @h-mayorquin

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -29,6 +29,10 @@ class BackendConfiguration(BaseModel):
         )
     )
 
+    nwbfile_identifier: str = Field(
```
CodyCBakerPhD, line 32: Is this still needed anymore?

h-mayorquin: I don't think so, it is nice provenance but our (edge case) prove it less useful to identify whether files require remapping. At the end, we do the full check.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -63,4 +67,73 @@ def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
...
+        new_backend_configuration = backend_configuration_class.from_nwbfile(nwbfile=nwbfile)
+
+        for dataset_configuration in new_backend_configuration.dataset_configurations.values():
+            location_in_new_nwbfile = dataset_configuration.location_in_file
+            if location_in_new_nwbfile not in location_to_former_configuration:
+                raise ValueError(f"Configuration for object in the following {location_in_new_nwbfile} not found.")
+
+            # Mapping the configuration through locations in the new NWBFile to the former configuration
+            former_configuration = location_to_former_configuration[location_in_new_nwbfile]
+            former_configuration.object_id = dataset_configuration.object_id
```
CodyCBakerPhD, line 134: Eh, OK, I see the mutation still happens here

I don't think it's necessary but I can see how it makes the syntax a bit easier on the final setter two lines below

CodyCBakerPhD: It's fine 👍

CodyCBakerPhD: Or maybe not fine depending on if the `former_configuration` is a reference of the source configuration instead of a local copy (which I thought at first)

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
h-mayorquin: This is needed:

https://github.com/catalystneuro/neuroconv/blob/7744c669cbb49078da40ae89650b93f2910bcdb1/src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py#L132-L135

Now that I look at it this looks as it is modifying the old configuration.

Let me fix this. Do you mind if I push here or would you rather have me make another PR?

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -63,4 +67,73 @@ def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
...
+
+        Notes
+        -----
+        This function creates a new configuration object. The original configuration remains unchanged.
+        """
+
+        location_to_former_configuration = {
+            dataset_configuration.location_in_file: dataset_configuration
+            for dataset_configuration in self.dataset_configurations.values()
+        }
```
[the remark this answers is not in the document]
h-mayorquin: Yes, I realized after I looked at it : /

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -63,4 +67,73 @@ def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
...
+        bool
+            True if all NWB object IDs in `nwbfile` are present in this backend configuration's
+            dataset configurations, False otherwise.
+        """
+
+        ids_in_backend_configuration = {
+            dataset_configuration.object_id for dataset_configuration in self.dataset_configurations.values()
+        }
+        ids_in_nwbfile = set(nwbfile.objects.keys())
+        return ids_in_nwbfile.issubset(ids_in_backend_configuration)
```
CodyCBakerPhD, line 92: Shouldn't some part of this logic deal with compatibility at the level of `location_in_file`?

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
CodyCBakerPhD: OK can you remove it in that case? If we don't need it yet, but we'll keep it in mind

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: We could write it here, this is done now in the other function where if a location in the new file does not appear on the old it will tell you exactly where a backend configuration is missing.

CodyCBakerPhD: > here if a location in the new file does not appear on the old it will tell you exactly where a backend configuration is missing.

Might be looser than that, right? And only iterate over locations in the new file that also appear in the old file, but object IDs might differ?

This is also more general than I actually need at the moment; the main intention of this feature right now is one-to-one between two perfectly equivalent NWB files

h-mayorquin: I don't how could it be looser. You need the location in the new file to be present in the old file to do the mapping.

So right there are two checks
1) `is_compatible_with_nwbfile` is just checking if the ids in the new nwbfile are all in the configuration. It tells if you if a re-mapping might be necessary.
2) The second function `build_remapped_backend_to_nwbfile` tries to re-map and it fails if it not possible telling you more precisely why.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: Sure.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
h-mayorquin: Did it

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
CodyCBakerPhD: The `issubset` is what makes it loose currently, if you want to make it strict then check that they have exactly the same number of objects and also the same exact set of `location_in_file` over all the objects

h-mayorquin: Yeah, but that's fine, a configuration that has all the ids of an nwbfile can be used to set it up. 

So, why I want is a function that tells me whether I might require a remapping and then do the strict check when we actually try to remap. Maybe the negative of `is_compatible_with_nwbfile` should be named `might_require_remapping` and we can have that?  What other suggestions you have?

All of my effort is concentrated on trying to take this edge case out of the main branch of configure_backend:

https://github.com/catalystneuro/neuroconv/blob/263fe5390089039c540b18472ef03fb27537a46f/src/neuroconv/tools/nwb_helpers/_configure_backend.py#L14-L71

Make it clear that this is an edge case that most developers don't need to think about it. Hide the complications away from the most common path.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: Should be done.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: The easier way would be to have the code out of the main path but I can see how this would be inconvenient for you on the guide. So I am trying to reach a compromise here.

CodyCBakerPhD: > might_require_remapping 

`requires_remapping` is a lot clearer, yes

_might_ shouldn't play into it - it either does or it does not, and that should be determined before actually doing it

**if** `requires_remapping` **then** `remap` makes sense

h-mayorquin: The thing is that the check that you want, requires building all the locations for the nwbfile. 
That is not necessary in most cases. 

Maybe some default computational waste is OK to settle this.

h-mayorquin: >You wanted to add this check to isolate the remapping from the rest of the logic, but I'm saying that just asking the object IDs subset is not a specific enough condition to ensure a remapping is actually needed or even appropriate

No, everything that you are requesting makes sense.  You are correct that the check is not complete as it is. To me the most important thing is that the remapping is isloated as much as possible from the main and common flow. I think we accomplished that. Now I am just whining to myself that I wish that there was a test that did not require that extra computation.

But you are completlely right that the computation is not that heavy and we should just do that. It is better to have simple, clear and readable code that is a bit slower. If we ever need that optimization then we implement it.

CodyCBakerPhD: Sounds like we align - so last 2 things on this PR

a) is to make this logical check more robust (albeit more expensive; I can add a suggestion for memoization once I see the new structure)

b) add a test(s) highlighting expected failure cases of this new feature

h-mayorquin: Yeah, I do think there is a simpler approach, check #871.

CodyCBakerPhD:
@h-mayorquin OK I added a bunch of tests and some minor refactors

Take another look then we can proceed

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -64,3 +64,90 @@ def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
...
+        with updated `object_id` fields.
+        """
+        # Use a fresh default configuration to get mapping of object IDs to locations in file
+        default_configurations = list(get_default_dataset_io_configurations(nwbfile=nwbfile, backend=self.backend))
+
+        if len(default_configurations) != len(self.dataset_configurations):
+            raise ValueError(
+                f"The number of default configurations ({len(default_configurations)}) does not match the number of "
+                f"specified configurations ({len(self.dataset_configurations)})!"
+            )
```
CodyCBakerPhD, line 102: Added this as an earlier more verbose catch of what was otherwise a dict KeyError for a UUID during the `nwbfile.objects` iteration (not very obvious what was wrong)

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -64,3 +64,90 @@ def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
...
+
+            former_configuration = self.dataset_configurations[location_in_file]
+            former_object_id = former_configuration.object_id
+
+            if former_object_id == object_id:
+                continue
+
+            remapped_configuration = former_configuration.model_copy(update={"object_id": object_id})
+            objects_requiring_remapping[location_in_file] = remapped_configuration
+
```
CodyCBakerPhD, line 125: Flattened out the logic of this entire section to reduce cyclomatic complexity

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -64,3 +64,90 @@ def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
...
+        for dataset_configuration in default_configurations:
+            location_in_file = dataset_configuration.location_in_file
+            object_id = dataset_configuration.object_id
+
+            location_cannot_be_remapped = location_in_file not in self.dataset_configurations
+            if location_cannot_be_remapped:
+                raise KeyError(
+                    "Unable to remap the object IDs for object at location '{location_in_file}'! This "
+                    "usually occurs if you are attempting to configure the backend for two files of "
+                    "non-equivalent structure."
```
CodyCBakerPhD, line 114: Added this as an earlier more verbose catch of what was otherwise a dict KeyError for a UUID during the nwbfile.objects iteration (not very obvious what was wrong)

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
CodyCBakerPhD, line 102: This is for some of those bidirectional cases you see in the tests; otherwise would have been silent about a case that this really should not apply

src/neuroconv/tools/testing/data_interface_mixins.py
CodyCBakerPhD: Added tests

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: Yep, nice, I think that you are missing the f of the f-string in the error though?

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: OK. Yes, you did mention this in our discussion and then I forgot about it. Good.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_equivalency.py
```diff
@@ -0,0 +1,151 @@
...
+def array_2(seed: int = 1) -> NWBFile:
+    integer_array = _generate_integer_array(seed=seed)
+
+    return integer_array
+
+
+@pytest.mark.parametrize("backend", ["hdf5", "zarr"])
+def test_configure_backend_equivalency(
+    tmpdir: Path,
+    array_1: NWBFile,
```
h-mayorquin, line 53: Is the type of `array_` NWBFile? Just double checking?

h-mayorquin (commented):
This looks good to me, just some f-string that I think is missing and we will be ready to go.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_equivalency.py
CodyCBakerPhD: Oops

So I started off attempting to use `nwbfile`'s as fixtures, which you might notice would have reduced a lot of duplication in places

But they had mutation issues that caused really weird issues (individual tests passed if you ran them in isolation but not the entire suite after the first passing check)

CodyCBakerPhD: Though maybe I just need to try again with `scope="function"`...

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: This is better.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_equivalency.py
CodyCBakerPhD: Nah, I think that provides too much redirection - it's useful to see the slightly different `NWBFile` structures laid out side by side but the irrelevant detail of the integer array abstracted to the top

h-mayorquin: Could be, I remember having the same type of trouble when using fixtures in the spikeinterface tests.

h-mayorquin (approved):

PR #820, opened 2024-04-20
add fictract docs to ignore links as it is not working temporarily

h-mayorquin:
It seems that this will take a bit longer than we expected.

https://github.com/rjdmoore/fictrac/issues/30

I will also remember that there is another broken link.


h-mayorquin:
OK, @CodyCBakerPhD. I think it is better to link to the repo as the repo as the other link that is failing and it will be updated to some point. I also fix the other links that were broken. This is ready to go on my side.

CodyCBakerPhD:
@h-mayorquin Agreed, thanks for handling this

CodyCBakerPhD (approved):

PR #821, opened 2024-04-22
Upgrade spikeinterface tools to `SortingAnalyzer`

alejoe91:
and remove `WaveformExtractor`

Let's keep this as a draft until SI v0.101.0 is released


alejoe91:
Temporarily pinned SI to 0.101.0rc0, we can switch to >=0.101.0 when released

alejoe91:
Thanks @h-mayorquin 

Yes then I would suggest to merge this first! :)

requirements-rtd.txt
```diff
@@ -5,5 +5,5 @@ readthedocs-sphinx-search==0.1.2
 sphinx-toggleprompt==0.2.0
 sphinx-copybutton==0.5.0
 roiextractors  # Needed for the API documentation
-spikeinterface  # Needed for the API documentation
+spikeinterface==0.101.0  # Needed for the API documentation
```
h-mayorquin, line 8: Why pinning instead of lower `>=` ?

CHANGELOG.md
```diff
@@ -110,6 +110,9 @@
 ### Testing
 * Add general test for metadata in-place modification by interfaces. [PR #815](https://github.com/catalystneuro/neuroconv/pull/815)
 
+### Dependencies
+* Updated `spikeinterface` to `v0.101.0` to use `SortingAnalyzer` instead of `WaveformExtractor` in the `spikeinterface` tools
```
h-mayorquin, line 114: Can you add that we are deprecating waveform extractor?

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1464,8 +1468,8 @@ def add_waveforms(
     )
 
 
-def write_waveforms(
```
h-mayorquin, line 1467: Can you keep these functions with the previous signature and they  throw and informative error of when was the last version of neuroconv where they were supported if they are called? I don't want to throw an import error without context if someone is relying on this.

h-mayorquin (changes requested):
OK, just let's not add some nice deprecations and add these information into the changelog.

requirements-rtd.txt
alejoe91: I followed the other pins, used >= for the main requirements

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1563,6 +1568,33 @@ def write_waveforms(
         )
 
 
+def write_waveforms(
```
h-mayorquin, line 1571: Cool, two things and this is done:
1) Add one exactly like this for `add_waveforms`
2) We have a six months deprecation in neuroconv. Write in the NotImplementedError that the function will be completley removed at some point after February 2025

alejoe91: done in last commit!

h-mayorquin (approved):
This is ready to go. My modularize later by breaking it into at least the following two blocks:
- Add templates (mean, std)
- Add quality metrics

PR #823, opened 2024-04-24
Debug `None` segment `_t_start` in SortingInterface alignment

CodyCBakerPhD:
@garrettmflynn Give this a try

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,14 @@
 # Upcoming
 
 ### Deprecations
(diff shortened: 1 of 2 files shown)


garrettmflynn (approved):
Looks good!

garrettmflynn:
To follow up, this change allows for the submission of a start time for a Sorting Interface without an associated recording interface—though this still doesn't return a timestamps array that can be displayed.

Is that the desired behavior?

CodyCBakerPhD:
Yes, because there are no timestamps to return

I'll correct any reliance on that on the GUIDE side when I get to that PR

PR #804, opened 2024-04-03
[Backend Configuration IVd] Integrate backend methods with NWBConverter

CodyCBakerPhD:
Similar to #801 but for NWBConverters

Need to add some tests and ensure it works well with format specific subconverters

PR #825, opened 2024-04-26
Path as string in FicTrac test

h-mayorquin:
Discussed in #804.


CodyCBakerPhD:
@h-mayorquin Any idea what's up with the commit history here?

h-mayorquin:
No idea, do you want me to squash it?

h-mayorquin:
No, I set it to the base PR here then it changed automatically when the other one was merged. Sure, I can do the other thing though.

CodyCBakerPhD:
> No, I set it to the base PR here then it changed automatically when the other one was merged. Sure, I can do the other thing though.

I see, that would have the same effect due the upstream squash

PR #829, opened 2024-04-29
Debug scanimage tests

### LINKED ISSUE #828: [Bug]: Failures in imaging plane name on new ScanImage tests
### What happened?

Really not sure what happened to break it on main post merge (since tests had to have been passing at time of merge...) but this block just started erroring on other PRs

### Steps to Reproduce

```python
Caught in CI: https://github.com/catalystneuro/neuroconv/actions/runs/8868996493/job/24349189441

traced to lines: https://github.com/catalystneuro/neuroconv/pull/827/files#diff-8a26fe5d073ef0f17a94da85ec6792dfcfb6f070abd2a8673c8c7e8e3a4899a6L52
```

### Traceback

_No response_

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.8

### Package Versions

_No response_

### Code of Conduct

pauladkisson:
Fixes #828

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Global Setup
         run: |
           python -m pip install -U pip  # Official recommended way
-          pip install pytest-xdist<=3.5.0
+          pip install pytest-xdist==3.5.0
           git config --global user.email "CI@example.com"
           git config --global user.name "CI Almighty"
           pip install wheel # Needed for scan image
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Global Setup
         run: |
           python -m pip install -U pip  # Official recommended way
-          pip install pytest-xdist
+          pip install pytest-xdist<=3.5.0
           git config --global user.email "CI@example.com"
           git config --global user.name "CI Almighty"
           pip install wheel # Needed for scan image
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Global Setup
         run: |
           python -m pip install -U pip  # Official recommended way
-          pip install pytest-xdist==3.5.0
+          pip install pytest-xdist
           git config --global user.email "CI@example.com"
           git config --global user.name "CI Almighty"
           pip install wheel # Needed for scan image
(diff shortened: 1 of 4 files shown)


weiglszonja:
So in `tests.test_on_data.test_imaging_interfaces.TestScanImageImagingInterfaceMultiPlaneCase_1_Channel_4` that should be parameterised with  
```python
{"channel_name": "Channel 4", ..."imaging_plane_name": "ImagingPlaneChannel4"}
```
what we have instead in `interface_kwargs` is:
```python
{'file_path': '/Users/weian/ophys_testing_data/imaging_datasets/ScanImage/scanimage_20220923_roi.tif', 'channel_name': 'Channel 1'}
```
But `imaging_plane_name` is correctly set to `ImagingPlaneChannel4`.
So it seems the interface_kwargs are not getting updated correctly, still not sure why :she

.github/workflows/testing.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Global Setup
         run: |
           python -m pip install -U pip  # Official recommended way
-          pip install pytest-xdist
+          pip install pytest-xdist==3.5.0
```
weiglszonja, line 35: I'm not sure this is needed
```suggestion
          pip install pytest-xdist
```

weiglszonja:
I think the [failed tests](https://github.com/catalystneuro/neuroconv/actions/runs/8892765001/job/24417573787?pr=829) are not related to ScanImage anymore

CodyCBakerPhD:
So pytest-xdist 3.6.1 is what broke this? But 3.6.0 was yanked https://pypi.org/project/pytest-xdist/#history

Thanks for digging into this everyone

The EDF thing is an eternal annoyance and sometimes goes away if you rerun it enough times

CodyCBakerPhD (approved):

pauladkisson:
Looks like we can use xdist 3.6.1 if we pass in interface_kwargs directly like szonja mentioned.

weiglszonja:
> So pytest-xdist 3.6.1 is what broke this? But 3.6.0 was yanked https://pypi.org/project/pytest-xdist/#history

I'm not sure, but before installing `pytest-xdist` I wasn't able to replicate the error in the first place. The issue that I pointed out in https://github.com/catalystneuro/neuroconv/pull/829#issuecomment-2083664907 is not exactly related, but looks like `pytest-xdist` has a history of "weirdness" with using parameterized tests.

PR #831, opened 2024-04-30
improve printing of byte sizes

bendichter:


bendichter:
Note that I am using Binary Prefix (Base 1024), not Decimal Prefix (Base 1000). Both are commonly used, though I believe Binary Prefix is more common when talking about RAM usage.

src/neuroconv/utils/str_utils.py
```diff
@@ -0,0 +1,47 @@
...
+        A human-readable string representation of the size.
+
+    Examples
+    --------
+    >>> human_readable_size(123)
+    '123 B'
+    >>> human_readable_size(1234)
+    '1.21 KB'
+    >>> human_readable_size(123456789)
+    '117.74 MB'
```
CodyCBakerPhD, line 24: With default of `binary=True` these would be `iB` - left the last one to demonstrate that usage

```suggestion
    >>> human_readable_size(123)
    '123 B'
    >>> human_readable_size(1234)
    '1.21 KiB'
    >>> human_readable_size(123456789, binary=False)
    '123.46 MB'
```

src/neuroconv/utils/str_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    # Define the suffixes for each size unit
+    suffixes = ["", "K", "M", "G", "T", "P", "E", "Z", "Y"]
+    i = 0
+    # Convert bytes to a float for division
+    size_float = float(size_bytes)
+
+    base = 1000 if not binary else 1024
+
+    # Find the appropriate unit
+    while size_float >= base and i < len(suffixes) - 1:
```
CodyCBakerPhD, line 39: I thought there was a non-iterate one-shot calculation you had using a log of the binary (or decimal) base?

CodyCBakerPhD:
> Both are commonly used, though I believe Binary Prefix is more common when talking about RAM usage.

Hardware often deals with binary because that's literally how much space can exist in RAM or on disk as well - software, including what is displayed to users, is often the more intuitive decimal usage since it is less confusing to an average user to specify something like `123456 bytes` and get `117 MiB` out of it.

Also a tad more intuitive to easily to calculate decimal version from things like shape of array and itemsize of dtype since no correction factor is needed

For that reason I'd be inclined to use `binary=False` in backend configurations, but `binary=True` in memory errors

But this is a thing that is usually a personal preference

bendichter:
Yeah, I was reading up a bit on conventions for this. Sometimes MB means binary and sometimes it means decimal. There isn't an easy way to know. I chose to go with binary base with the "i" in the middle since that is explicit, but I agree it's less intuitive. I think you're right that it may be better to have memory config in decimal

CodyCBakerPhD:
Looks good to me

CodyCBakerPhD (approved):

h-mayorquin:
> Note that I am using Binary Prefix (Base 1024), not Decimal Prefix (Base 1000). Both are commonly used, though I believe Binary Prefix is more common when talking about RAM usage.

I also prefer base 1024 in general, and I think that the 'i' in between makes it clear.

But I remember that @CodyCBakerPhD told me something along those lines:

> Hardware often deals with binary because that's literally how much space can exist in RAM or on disk as well - software, including what is displayed to users, is often the more intuitive decimal usage since it is less confusing to an average user to specify something like 123456 bytes and get 117 MiB out of it.

Just to be concrete, this problems happens because the `GenericDataChunkIterator` takes units on base 10, right? That's the part of the ecosystem where the user is expected to input units in base 10?

PR #834, opened 2024-05-03
Fix latest pynwb tests

CodyCBakerPhD:
These started failing on dev a few days ago, just now getting around to fixing them

### DIFF
--- a/requirements-minimal.txt
+++ b/requirements-minimal.txt
@@ -4,9 +4,9 @@ jsonschema>=3.2.0
 PyYAML>=5.4
 scipy>=1.4.1
 h5py>=3.9.0
-hdmf>=3.12.2
-hdmf_zarr>=0.4.0
-pynwb>=2.3.2;python_version>='3.8'
+hdmf>=3.13.0
+hdmf_zarr>=0.7.0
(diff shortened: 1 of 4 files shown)


h-mayorquin (approved):
OK. Is this because the timezone does not get added now? Is not returned?

CodyCBakerPhD:
Timezone is no longer required or set automatically; I think it can also take `date` (not exact timestamp) now too

h-mayorquin:
Thanks man! Maybe we should simplify the conversion gallery?

CodyCBakerPhD:
I think the conversion gallery can still serve the purpose of _encouraging_ people to set exact times (with explicit, not assumed, timezones) as best practice

But I suppose a detail should be added somewhere about how if exact timestamps aren't known a basic date can suffice

bendichter:
The thing is, that code is exactly the same for every interface, so maybe we could add a page to the conversion gallery that demonstrates how to create "DANDI-compatible NWB files"

h-mayorquin:
I like that last idea. I also conceive the gallery as a minimal example "this is the fastest way of doing a conversion" so I want as little frictions as possible.

PR #836, opened 2024-05-04
enhance metadata section of User Guide

bendichter:


docs/user_guide/datainterfaces.rst
```diff
@@ -45,35 +45,118 @@ but generally will not read the underlying data.
...
+
+.. code-block:: python
+
+    metadata["NWBFile"].update(
+        experimenter="Darwin, Charles",
+        lab="Evolutionary Biology",
+        institution="University of Cambridge",
+        experiment_description="Observation of finches on the Galapagos Islands",
+        session_id="session_1",
+        identifier="unique_identifier",  # will be automatically assigned if not provided
```
CodyCBakerPhD, line 112:
```suggestion
```

Towards goal of absolute minimization

Also should we pair `session_description` closer to `experiment_description` and maybe move them both to the top?

docs/user_guide/datainterfaces.rst
```diff
@@ -45,35 +45,117 @@ but generally will not read the underlying data.
...
+    metadata["NWBFile"]["session_start_time"] = metadata["NWBFile"]["session_start_time"].replace(tzinfo=tz.gettz("US/Pacific"))
+
+NWB Best Practices also recommends several other fields that are rarely present in the extracted metadata. The metadata dictionary is the place to add this information.
+
+.. code-block:: python
+
+    metadata["NWBFile"].update(
+        session_id="session_1",
+        session_description="Observation of finches on the Galapagos Islands",
+        experiment_description="Observation of finches on the Galapagos Islands",
```
CodyCBakerPhD, line 109:
```suggestion
        session_description="Observations of desert plants and reptiles on the island of San Cristobal.",
        experiment_description="Observations of wildlife across the Galapagos Islands.",
```

A user might wonder why these are duplicated - though we sometimes do that in practice, I think this could be an opportunity to highlight how they are different (with experiment being broader in scope than just this one session)

CodyCBakerPhD (approved):

PR #838, opened 2024-05-04
add neuralynx_nvt to toctree

bendichter:


docs/conversion_examples_gallery/conversion_example_gallery.rst
```diff
@@ -93,6 +93,7 @@ Behavior
     DeepLabCut <behavior/deeplabcut>
     FicTrac <behavior/fictrac>
     LightningPose <behavior/lightningpose>
+    Neuralynx NVT <behavior/neuralynx_nvt>
```
CodyCBakerPhD, line 96: Should we move it further up closer to the other Neuralynx one? It was kind of hard to spot otherwise

bendichter: It's currently in alphabetical order. We could add a note in the other Neuralynx page that points to this one

CodyCBakerPhD: Oh, hmm... I see now on the side

I was referring to the main nav tree on the left bar

image

but I guess that's just the flattened version?

Should definitely have a pointer since someone might have clicked the first Neuralynx one on the left

bendichter: yeah, I think we should get rid of that side bar, but that's a separate issue

CodyCBakerPhD (approved):

PR #842, opened 2024-05-06
Style suggestions from other PR

CodyCBakerPhD:
From https://github.com/catalystneuro/neuroconv/pull/802#discussion_r1590476370

PR #844, opened 2024-05-06
trim backend docs

bendichter:


docs/user_guide/backend_configuration.rst
```diff
@@ -40,22 +40,22 @@ To retrieve a default configuration for an in-memory ``pynwb.NWBFile`` object, u
...
+    print(backend_configuration)
 
 returns:
 
 .. code-block:: bash
 
-    HDF5 dataset configurations
-    ---------------------------
+    HDF5 backend configuration
+    --------------------------
```
CodyCBakerPhD, line 58: Please move this over to https://github.com/catalystneuro/neuroconv/pull/842 so we can keep them in sync

docs/user_guide/backend_configuration.rst
```diff
@@ -133,10 +134,10 @@ Then we can use this configuration to write the NWB file:
 
     from neuroconv.tools.nwb_helpers import configure_backend, BACKEND_NWB_IO
 
-    dataset_configurations["acquisition/MyTimeSeries/data"] = dataset_configuration
+    backend_configuration.dataset_configurations["acquisition/MyTimeSeries/data"] = dataset_configuration
```
CodyCBakerPhD, line 137: This is to keep the horizontal width from getting too large in the central pydata window

bendichter: Does it actually work if you don't assign it back to the `backend_configuration` object though?

CodyCBakerPhD: Yes, because the `dataset_configurations` variable is a pointer to that part of the `backend_configuration`, not a copy

docs/user_guide/backend_configuration.rst
bendichter, line 58:
```suggestion
    HDF5 dataset configurations
    ---------------------------
```

docs/user_guide/backend_configuration.rst
CodyCBakerPhD: In particular, this is what happens without breaking it up into separate lines

image

Though it also increases readability in general IMO

bendichter: OK I have confirmed that this is true, but then I think it would be clearer if we printed out `print(backend_configuration)` to demonstrate this

CodyCBakerPhD: But then you show the unneeded information about the unmodified `timestamps` field

I don't think most people will be confused by this minor detail

bendichter: I think it might, but I'm OK with keeping it as is for now

bendichter, line 137:
```suggestion
    dataset_configurations["acquisition/MyTimeSeries/data"] = dataset_configuration
```

CodyCBakerPhD (approved):

PR #845, opened 2024-05-08
Pin Zarr for tests

CodyCBakerPhD:
Tests just failed today due to a surprise break from Zarr release yesterday - some issue with decoding Blosc for our custom backend configuration stuff

Shouldn't effect actual typical usage of the backend feature aside from Blosc I assume, since none of the other compressor tests complain

[Filed issue with hdmf-zarr](https://github.com/hdmf-dev/hdmf-zarr/issues/192) so team knows about it, and if it's an issue with the way they specify the read from the IO level (I doubt it, but never know)

Downpinning in tests temporarily to fix the CI, will check over time if resolved automagically

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -40,7 +40,7 @@ jobs:
 
 
       - name: Install neuroconv with minimal requirements
-        run: pip install .
+        run: pip install .[test]
       - name: Run import tests
         run: |
           pytest tests/imports.py::TestImportStructure::test_top_level
--- a/requirements-minimal.txt
+++ b/requirements-minimal.txt
@@ -6,7 +6,6 @@ scipy>=1.4.1
 h5py>=3.9.0
 hdmf>=3.13.0
 hdmf_zarr>=0.7.0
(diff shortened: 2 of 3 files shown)


h-mayorquin:
I think that we should put upper bounds everywhere so people who install our package don't get the version that does not work, right?

CodyCBakerPhD:
> I think that we should put upper bounds everywhere so people who install our package don't get the version that does not work, right?

Thing is, _most_ of the package works, just perhaps not this one particular case that showed up in one corner of our tests - so I'd wait to see if this gets fixed before totally restricting the latest upper bound in general

Maybe circle back in a week?

CodyCBakerPhD:
Filed an issue on HDMF for the dev tests

h-mayorquin:
Waiting a week seems fine to me.

h-mayorquin:
You still want this merge to have the tests working?

CodyCBakerPhD:
> You still want this merge to have the tests working?

Yes please

h-mayorquin (approved):

PR #848, opened 2024-05-10
add convenience function: configure_and_write_nwbfile

bendichter:
also shuffled imports around a bit to solve the circularity issues so you no longer have to do fancy internal imports


src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -226,3 +232,75 @@ def make_or_load_nwbfile(
 
                 if not success and not file_initially_exists:
                     nwbfile_path_in.unlink()
+
+
+def resolve_backend(
```
CodyCBakerPhD, line 237:
```suggestion
def _resolve_backend(
```

Fine to abstract this out to reduce duplication, but the phrase `resolve_backend` is a bit ambiguous

So I'd recommend marking as private (also since it's not expose to `nwb_helpers`, which is also good) just to make it clear this is an internal tool for us and our argument parsing only

CodyCBakerPhD:
Fine by me - small comment on marking intermediate helper as private, then good to go

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
bendichter: good idea

bendichter:
@CodyCBakerPhD I'm not sure how to get past this

CodyCBakerPhD:
> I'm not sure how to get past this

Get past what?

The core tests are passing fine (this PR is just shifting code around for reduction and simplicity)

Dev tests have been failing past 2 days due to HDMF issue that the team is aware of, last I heard they planned on reverting a recent merge at which point it should stabilize our end

CodyCBakerPhD (approved):

PR #851, opened 2024-05-10
detect the wrong number of timestamps

bendichter:


CodyCBakerPhD:
?

Does ROIExtractors handle that for us?

bendichter:
I found this check on the `set_times()` method so no need to add it here

PR #852, opened 2024-05-11
use abstract class DatasetIOConfiguration for typing

bendichter:
Replaces `Union[HDF5DatasetIOConfiguration, ZarrDatasetIOConfiguration]` with `DatasetIOConfiguration`. This doesn't change much, but this is the standard way of specifying multiple subclasses that share a base class, and it readies the repo for additional backends if they should ever come


CodyCBakerPhD:
This is fine for the testing suite (since general tests can and should apply to any child of a `DatasetIOConfiguration`), however

> Replaces Union[HDF5DatasetIOConfiguration, ZarrDatasetIOConfiguration] with DatasetIOConfiguration. This doesn't change much, but this is the standard way of specifying multiple subclasses that share a base class

The annotations do say different things, and anywhere in the core that references `Union[HDF5DatasetIOConfiguration, ZarrDatasetIOConfiguration]` is explicitly stating that only those two backends have been approved to work with this feature, not any other arbitrary custom child class that someone else may have made

If there ever is a third backend, we'd want to thoroughly ensure the backend configuration feature plays well with it before updating annotations

CodyCBakerPhD (approved):

PR #854, opened 2024-05-12
Support Path objects in source_data schema validation

h-mayorquin:
Is that something that we would want? I always find it annoying when working on conversations that I have to stringify at the signature or before.


CodyCBakerPhD:
Yes, but the broader fix (in anticipation of the slew of future annotations) is still to use Pydantic for the schema validation instead of JSON

CodyCBakerPhD:
I'd appreciate more work on the broader issue, feels less satisfying to add a bunch of fluff that will get removed inclemently

But if you really want it then OK

h-mayorquin:
Yes please, I would like to have this as I think is a small change for a quality of life improvement and I am not sure when I will have time to work in the broader issue.

CodyCBakerPhD (approved):

PR #858, opened 2024-05-13
Use zoneinfo instead of dateutil

### LINKED ISSUE #599: [Proposal] Use `ZoneInfo` instead of `dateutil` for time zone specification
Since python 3.9 zoneinfo has been accepted as the standard way of handling time zones and has been included in the standard library:
https://docs.python.org/3/library/zoneinfo.html

However, in the gallery and in some conversions were are still using dateutil for this (example):

https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/recording/openephys.html

We still are supporting python 3.8 so we can not just use `ZoneInfo` as it is. Fortunately, there is a backport for this specific purpose:|
https://pypi.org/project/backports.zoneinfo/

Which we can use that right now to start using the standard solution and avoid the installation of an extra dependency for users that use python 3.9.

h-mayorquin:
Should close #599.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,6 @@
 * Improve printing of bytes [PR #831](https://github.com/catalystneuro/neuroconv/pull/831)
 * Added `configure_and_write_nwbfile` and optimized imports in `tools.nwb_helpers` module. [PR #848](https://github.com/catalystneuro/neuroconv/pull/848)
 * Support for pathlib in source data schema validation [PR #854](https://github.com/catalystneuro/neuroconv/pull/854)
-* Use ZoneInfo instead of dateutil.tz [PR #858](https://github.com/catalystneuro/neuroconv/pull/858)
 
 ### Bug fixes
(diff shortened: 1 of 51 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 * Improve printing of bytes [PR #831](https://github.com/catalystneuro/neuroconv/pull/831)
 * Added `configure_and_write_nwbfile` and optimized imports in `tools.nwb_helpers` module. [PR #848](https://github.com/catalystneuro/neuroconv/pull/848)
 * Support for pathlib in source data schema validation [PR #854](https://github.com/catalystneuro/neuroconv/pull/854)
+* Use ZoneInfo instead of dateutil.tz [PR #858](https://github.com/catalystneuro/neuroconv/pull/858)
```
CodyCBakerPhD, line 20: Not really a feature, can you move to improvements and specify it's on the conversion gallery?

h-mayorquin: OK. Done

CodyCBakerPhD (approved):

PR #860, opened 2024-05-15
Attempt to fix GUIDE electrode table rendering

CodyCBakerPhD:
Of an unneeded field in the Converter case (specific to the NIDQ portion)


CodyCBakerPhD:
cc @garrettmflynn

garrettmflynn (approved):
Works great. Thanks for working through this!

PR #861, opened 2024-05-15
Pass progress_bar_class to HDMF

garrettmflynn:
To properly pass `progress_Bar_class` to HDMF (https://github.com/hdmf-dev/hdmf/pull/1110) through the GUIDE in https://github.com/NeurodataWithoutBorders/nwb-guide/pull/778, the `SpikeInterfaceRecordingDataChunkIterator` needs to pass another argument to the parent class.


CodyCBakerPhD:
Also swapped codespell to pre-commit now that they support that (and it was being problematic to a known issue in latest update)

PR #863, opened 2024-05-16
Marginal improvement to #816

h-mayorquin:
A suggestion to improve #816 by having the remapping return a backend configuration instead.

Advantages:
* Returning the backend configuration type instead of the dictionary is more specific (we know more about the output!)
* The function is a little bit more general at the same time. 
* We don't leak details of the mapping to `backend_configuration`


h-mayorquin:
> The logic is failing to map proper IDs here
> 
> But I'll consider the general suggestion and think on how to refactor #816
> 
> Another approach might be a direct `update_object_ids` method on the `DatasetIOConfiguration` model

This points to some deeper issue that sometimes the returned nwbfiles have the same identifier. Look at this. I opened the debuger at these lines for a failing test (FicTrac):

```python
    def check_configure_backend_for_equivalent_nwbfiles(self, backend: Literal["hdf5", "zarr"] = "hdf5"):
        metadata = self.interface.get_metadata()
        if "session_start_time" not in metadata["NWBFile"]:

        nwbfile_1 = self.interface.create_nwbfile(metadata=metadata, **self.conversion_options)
        nwbfile_2 = self.interface.create_nwbfile(metadata=metadata, **self.conversion_options)

        backend_configuration = get_default_backend_configuration(nwbfile=nwbfile_1, backend=backend)
        configure_backend(nwbfile=nwbfile_2, backend_configuration=backend_configuration)
```

```python
nwbfile_1.identifier
'8a86e132-9384-4688-82b8-c2c2943009e2'
nwbfile_2.identifier
'8a86e132-9384-4688-82b8-c2c2943009e2'
```

So the mappign was not performed.

h-mayorquin:
Yeah, I thought about the approach of updating directly but given that this objects are not too memory heavy I went for the heuristic that functions should not have side effects.

h-mayorquin:
> The logic is failing to map proper IDs here
> 
> But I'll consider the general suggestion and think on how to refactor #816
> 
> Another approach might be a direct `update_object_ids` method on the `DatasetIOConfiguration` model

Oh, moreover, that's a froozen attribute. So a new instance is indeed required.

h-mayorquin:
> > The logic is failing to map proper IDs here
> > But I'll consider the general suggestion and think on how to refactor #816
> > Another approach might be a direct `update_object_ids` method on the `DatasetIOConfiguration` model
> 
> Oh, moreover, that's a froozen attribute. So a new instance is indeed required.

Ah, not if we change the whole configuration for the whole location. Anyway. Tests are passing now.

CodyCBakerPhD:
> This points to some deeper issue that sometimes the returned nwbfiles have the same identifier. Look at this. I opened the debuger at these lines for a failing test (FicTrac):

Actually yeah; that's a good point. They were made using the same `metadata` (which is what specifies the `identifier`) so maybe the file identifier in-memory cannot be trusted

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -1,74 +1,100 @@
...
 from pynwb import NWBFile, TimeSeries
 
+from ._configuration_models._base_backend import BackendConfiguration
 from ._configuration_models._hdf5_backend import HDF5BackendConfiguration
 from ._configuration_models._zarr_backend import ZarrBackendConfiguration
 from ..importing import is_package_installed
 
 
-def _resolve_nwbfile_object_ids(
+def _remap_backend_configuration_to_nwbfile(
```
CodyCBakerPhD, line 15: I do like this approach more, but what would you think about making it an instance method on the `backend_configuration`? Like `.equivocate_object_ids(self, nwbfile: pynwb.NWBFile)`

CodyCBakerPhD: We could also unfreeze the object ID field and adjust the `self: BackendConfiguration` in-place - WDYT?

h-mayorquin: I was also thinking about those two but did not want to rock or boat or delay your other PR.

I think that both make sense!

CodyCBakerPhD: Sounds good, did you want to do that in this PR?

h-mayorquin: Yeah, sure.

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -85,16 +25,16 @@ def configure_backend(
         The configuration model to use when configuring the datasets for this backend.
     """
     is_ndx_events_installed = is_package_installed(package_name="ndx_events")
-    ndx_events = sys.modules.get("ndx_events", None)
+    ndx_events = importlib.import_module("ndx_events") if is_ndx_events_installed else None
```
h-mayorquin, line 28: This was giving me trouble somehow...

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -65,3 +68,88 @@ def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
...
+        -------
+        bool
+            True if all NWB object IDs in `nwbfile` are present in this backend configuration's
+            dataset configurations, False otherwise.
+        """
+
+        ids_in_backend_configuration = {
+            dataset_configuration.object_id for dataset_configuration in self.dataset_configurations.values()
+        }
+        ids_in_nwbfile = set(nwbfile.objects.keys())
```
h-mayorquin, line 91: I don't know if you want to restrict this check to files that can be setted (don't have an io?). Just wondering.

src/neuroconv/tools/nwb_helpers/_configure_backend.py
h-mayorquin: Done.

src/neuroconv/tools/nwb_helpers/_configure_backend.py
CodyCBakerPhD: Might have to do with the change a while back to use `find_spec` in `is_package_installed` instead of attempting to import the module, which would have added it to the `sys.modules`

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -65,3 +68,88 @@ def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
...
+            The NWBFile object to check compatibility against.
+
+        Returns
+        -------
+        bool
+            True if the backend configuration is NOT compatible with `nwbfile`, False otherwise.
+        """
+        return not self.is_compatible_with_nwbfile(nwbfile)
+
+    def remap_to_new_nwbfile(self, nwbfile: NWBFile) -> Self:
```
h-mayorquin, line 110: Not happy with the naming because it kind of sounds that it is doing it in place.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
CodyCBakerPhD: Dealing with an `nwbfile` loaded from an IO is tricky for this feature. Right now really only intended it for items that are still purely in-memory and only dealing with first write

Should probably add skip and test for this in a followup

h-mayorquin: Makes sense.

src/neuroconv/tools/nwb_helpers/_configure_backend.py
h-mayorquin: I see.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -65,3 +68,88 @@ def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
...
+            dataset configurations, False otherwise.
+        """
+
+        ids_in_backend_configuration = {
+            dataset_configuration.object_id for dataset_configuration in self.dataset_configurations.values()
+        }
+        ids_in_nwbfile = set(nwbfile.objects.keys())
+        return ids_in_nwbfile.issubset(ids_in_backend_configuration)
+
+    def is_incompatible_with_nwbfile(self, nwbfile: NWBFile) -> bool:
```
CodyCBakerPhD, line 94: I'd rather not have a dedicated method for doing reversed logic. Anyone can add `not is_compatible_with_nwbfile` (which is just as long to type) if they want to adjust their logic somewhere

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
CodyCBakerPhD: And you'd rather still return a new object rather than relax the mutability of the object IDs?

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: All right. I am more concerned with expressivness than length. 

Between:
1) ` if not is_compatible_with_nwbfile:` 
2) ` if is_incompatible_with_nwbfile:`

I prefer the second.

I don't feel strong about this, I can remove it. But this is where I am coming from, I usually don't mind lines of code if they make the thing more expressive, same for adding methods and attributes.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: Yeah, I think if we can we should avoid mutation in place. Someone might be passing configuration then reusing it somewhere else and having this side-effect inadvertently introduce a bug in their flow.. The usual cost of non-mutability is performance but I think here the cost is low.

How do you think about this? What are the advantages of mutation?

CodyCBakerPhD: Just avoiding user confusion

How about `get_equivalent_backend_configuration(self, nwbfile...)`?

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
CodyCBakerPhD: I find 1. easier to read due to the spacing between the logic modifier and the method name

In fact, if I don't pay really close attention to notice the 'in` in 2., I'd think they were the same method

Plus then you don't have to juggle the extra possible logic combinations inside your head; with both you have 4 possiblities that someone could write in code

`is_compatible_with_nwbfile`
`not is_compatible_with_nwbfile`
`is_incompatible_with_nwbfile`
`not is_incompatible_with_nwbfile`

whereas with only one method, you'd only have two possible which is cleaner

P.S., even in our own code, you could make lines more readable in your view by defining an intermediate variable

`is_incompatible_with_nwbfile = not is_compatible_with_nwbfile` before using it - I just don't see what it needs to be a whole extra method

h-mayorquin: But the point is that I don't need to jugle with the four of them. I have the two so I never need the negation. That is, I only need to use  `is_compatible_with_nwbfile`  or `if is_incompatible_with_nwbfile` depending on the situation.

Anyway, I don't feel very strong about this. I will keep the positive compatibility method. I think is marginal effect whatever way it goes.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: I think that's better than mine. Let me sleep on it to see if we can come up with something better.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: Ok, removed.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
h-mayorquin: At the end I just went fully verbose. This is ready on my side.

PR #866, opened 2024-05-17
Propagate photon series type to imaging interfaces

alessandratrapani:
Following #847
Add photon_series_type input arg to imaging interfaces that do not set automatically metadata specific to TwoPhotonSeries (such as ScanImage interfaces) or specific to OnePhotonSeries (as Miniscope interface).


src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py
```diff
@@ -44,7 +45,16 @@ def __init__(
         self.verbose = verbose
         self.photon_series_type = photon_series_type
 
-    def get_metadata_schema(self) -> dict:
+    def get_metadata_schema(self, photon_series_type: Literal["OnePhotonSeries", "TwoPhotonSeries"] = None) -> dict:
```
CodyCBakerPhD, line 48:
```suggestion
    def get_metadata_schema(self, photon_series_type: Optional[Literal["OnePhotonSeries", "TwoPhotonSeries"]] = None) -> dict:
```

I know it's just temporary; still including

src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py
```diff
@@ -87,7 +97,17 @@ def get_metadata_schema(self) -> dict:
         fill_defaults(metadata_schema, self.get_metadata())
         return metadata_schema
 
-    def get_metadata(self) -> DeepDict:
+    def get_metadata(self, photon_series_type: Literal["OnePhotonSeries", "TwoPhotonSeries"] = None) -> DeepDict:
```
CodyCBakerPhD, line 100:
```suggestion
    def get_metadata(self, photon_series_type: Optional[Literal["OnePhotonSeries", "TwoPhotonSeries"]] = None) -> DeepDict:
```

CodyCBakerPhD:
@alessandratrapani Could you add at least one test or otherwise modify the imaging extractor mixin to use this new structure so the CI doesn't get the deprecation warnings?

Should be the last thing to do on this PR then it's good to go

alessandratrapani:
> @alessandratrapani Could you add at least one test or otherwise modify the imaging extractor mixin to use this new structure so the CI doesn't get the deprecation warnings?
> 
> Should be the last thing to do on this PR then it's good to go

I add a test to trigger the warning. I don't think the CI gets the deprecation warning because it does not pass the photn_series_type argument, but maybe I am wrong

CodyCBakerPhD (approved):

PR #869, opened 2024-05-20
Updated DataInterface Documentation

tabedzki:
Updated documentation to reflect the change from `run_conversion` to `add_to_nwbfile`


CodyCBakerPhD:
Good catch @tabedzki, thanks!

PR #871, opened 2024-05-20
Avoid double computation for backend check

h-mayorquin:
From the discussion in https://github.com/catalystneuro/neuroconv/pull/816/files#r1607028944 from #816

OK, this is I think a simpler approach. breaks the previous construction into two functions. The first part finds all the locations that require-remapping and the second part uses pydantic machinery to create a copy of the model and modify only what requires modification.

I think this should be faster and is more flexible. What would be missing is some testing. Let me know what you think.


CodyCBakerPhD (approved):

CodyCBakerPhD:
Looks great to me; I'll enhance tests over the next couple of days

Issue #772, opened 2024-03-11
[Bug]: rmv WaveformExtractor


bendichter:
### What happened?

The latest version of spikeinterface has removed `WaveformExtractor`, which is now causing an import error for the dev installation. See https://github.com/catalystneuro/neuroconv/actions/runs/8225815354/job/22491395514?pr=771

Let's look at refactoring NeuroConv to match so that we can benefit from the latest SpikeInterface features.

@h-mayorquin , you might be a good person to handle this.

### Steps to Reproduce

```python
See above
```

### Operating System

Linux

### Python Executable

Conda

### Python Version

3.8

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

CodyCBakerPhD:
Already on the radar from https://github.com/catalystneuro/neuroconv/pull/768#issuecomment-1981525510

One question might be how smoothly we want to remove it; some current pipelines (as I recall, the ones Alessio setup for Allen Insistute?) might depend on it otherwise and we wouldn't want to break them

Perhaps `if hasattr(spikeinterface, WaveformExtractor)` then define write methods as we do now, otherwise don't define until we add support for the SortingAnalyzer?

h-mayorquin:
All right, I will take a look.

Some context for provenance:
* The `SortyingAnalyser` that will replace `WaveformExtractor` is now on dev and according to @alejoe91 in Slack it will be released on May. 
* Other relevant issue here:
https://github.com/SpikeInterface/spikeinterface/pull/2546

CodyCBakerPhD:
This has since been resolved I believe, right @h-mayorquin?

h-mayorquin:
This is done.

PR #878, opened 2024-05-25
fix typo in ecephys `waveform_sds` to `waveform_stds`

h-mayorquin:
I don't know how this flew over us for so long.

I added a deprecation only on `add_units` because all the other uses were introduced yesterday (today being 2024-05-25)

[Edit] Now that I looked at this this probably was intentional by @alejoe91 as the signature of `nwbfile.add_unit` is called `waveform_sd` and he just pluralized the two. 

https://pynwb.readthedocs.io/en/stable/pynwb.file.html#pynwb.file.NWBFile.add_unit

I don't know what is better.

From my conversation with @alejoe91 these arguments are only used in one pipeline, but now that are [moving towards adding the sorting analyzer ](#772 ) [here](https://github.com/catalystneuro/neuroconv/pull/821) this will be (hopefully) used more.

So this is a good chance to choose a good naming scheme?

We have now:
`waveform_means`
`waveform_sds` 

And the signature of `add_unit` in pynwb is:
`waveform_mean`
`waveform_sd`

This PR has:
`waveform_means`
`waveform_stds`

Could also be:
`waveforms_mean`
`waveforms_std`

Maybe the best is just to stick with pynwb so "we get out what we put in" in terms of parameters. 

any opinions @CodyCBakerPhD @alejoe91 ?


alejoe91:
Waveform_sd is the nwb name though...

CodyCBakerPhD:
I leave this to you and @alejoe91

h-mayorquin:
> Waveform_sd is the nwb name though...

Right. And this is the the argument for `add_unit` which only adds a row. I think that your choice of having the plural name in our function that adds all rows makes sense.

Closing this.

PR #883, opened 2024-05-30
MedpcInterface

pauladkisson:
This PR adds a datainterface for MedPC output files.


src/neuroconv/datainterfaces/behavior/medpc/medpc_helpers.py
```diff
@@ -0,0 +1,146 @@
...
+    if start_line is None:
+        raise ValueError(
+            f"Could not find the start variable ({start_variable}) of the session with conditions {session_conditions}"
+        )
+    session_lines = lines[start_line:end_line]
+    return session_lines
+
+
+def read_medpc_file(
+    file_path: str,
```
bendichter, line 84: FilePathType

src/neuroconv/datainterfaces/behavior/medpc/medpcdatainterface.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class MedPCInterface(BaseDataInterface):
+    """Data Interface for MedPC output files"""
+
+    keywords = ["behavior"]
+
+    def __init__(
+        self,
+        file_path: str,
```
bendichter, line 22: FilePathType

pauladkisson:
@CodyCBakerPhD, do you know whats going on with these doctests? The documentation look fine to me on RTD.

pauladkisson:
> do you know whats going on with these doctests? The documentation look fine to me on RTD.

nm, it was the >>> that pytest didnt like

CodyCBakerPhD:
@pauladkisson Good for full review now?

pauladkisson:
> @pauladkisson Good for full review now?

Yep, this is all ready to go.

docs/api/interfaces.behavior.rst
```diff
@@ -28,3 +28,11 @@ SLEAP
 Video
 -----
 .. automodule:: neuroconv.datainterfaces.behavior.video.videodatainterface
+
+MedPC
+-----
+.. automodule:: neuroconv.datainterfaces.behavior.medpc.medpcdatainterface
+
+MedPC Helpers
```
CodyCBakerPhD, line 36: We don't usually included such submodules in the API docs

docs/conversion_examples_gallery/behavior/medpc.rst
```diff
@@ -0,0 +1,86 @@
+MedPC data conversion
+-----------------------------
```
CodyCBakerPhD, line 2:
```suggestion
MedPC data conversion
--------------------------
```

docs/conversion_examples_gallery/behavior/medpc.rst
```diff
@@ -0,0 +1,86 @@
...
+            "name": "right_nose_poke_times",
+            "description": "Right nose poke times",
+        },
+        {
+            "name": "right_reward_times",
+            "description": "Right reward times",
+        },
+        {
+            "name": "footshock_times",
+            "description": "Footshock times",
```
CodyCBakerPhD, line 72:
```suggestion
            "description": "Left nose poke times.",
        },
        {
            "name": "left_reward_times",
            "description": "Left reward times.",
        },
        {
            "name": "right_nose_poke_times",
            "description": "Right nose poke times.",
        },
        {
            "name": "right_reward_times",
            "description": "Right reward times.",
        },
        {
            "name": "footshock_times",
            "description": "Footshock times.",
```

CodyCBakerPhD: Proper sentences to encourage the practice for nicer rendering in tooltips on things like Neurosift

docs/conversion_examples_gallery/behavior/medpc.rst
```diff
@@ -0,0 +1,86 @@
...
+        },
+    ]
+    metadata["MedPC"]["IntervalSeries"] = [
+        {
+            "name": "reward_port_intervals",
+            "description": "Interval of time spent in reward port (1 is entry, -1 is exit)",
```
CodyCBakerPhD, line 78:
```suggestion
            "description": "Interval of time spent in reward port (1 is entry, -1 is exit).",
```

src/neuroconv/datainterfaces/behavior/medpc/medpc_helpers.py
```diff
@@ -0,0 +1,173 @@
+import numpy as np
+
+from neuroconv.utils import FilePathType
+
+
+def get_medpc_variables(file_path: FilePathType, variable_names: list) -> dict:
```
CodyCBakerPhD, line 6: IMO should mark these (and parent file) as private to indicate we are simply using them in our own interface code, not to be advertised or exposed for public use since often the return structures (such as the `dict` here) are specific to our own needs

src/neuroconv/datainterfaces/behavior/medpc/medpc_helpers.py
```diff
@@ -0,0 +1,173 @@
...
+                    session_dict[output_name] = np.array([], dtype=float)
+                elif type(session_dict[output_name]) == str:  # not a multiline variable
+                    raise ValueError(
+                        f"Expected {output_name} to be a multiline variable, but found a single line variable."
+                    )
+                else:
+                    session_dict[output_name] = np.array(session_dict[output_name], dtype=float)
+                    session_dict[output_name] = np.trim_zeros(
+                        session_dict[output_name], trim="b"
+                    )  # MEDPC adds extra zeros to the end of the array
```
CodyCBakerPhD, line 172: Especially with all this parsing; a generic package for reading MedPC files would be something to support all that, but that's way too much for us to claim to support - instead, just a collection of internal helper functions for getting the data into an NWB file in a particular way

src/neuroconv/datainterfaces/behavior/medpc/medpcdatainterface.py
```diff
@@ -0,0 +1,248 @@
...
+
+from neuroconv.basetemporalalignmentinterface import BaseTemporalAlignmentInterface
+from neuroconv.tools import get_package, nwb_helpers
+from neuroconv.utils import DeepDict, FilePathType
+
+from .medpc_helpers import read_medpc_file
+
+
+class MedPCInterface(BaseTemporalAlignmentInterface):
+    """Data Interface for MedPC output files.
```
CodyCBakerPhD, line 15:
```suggestion
    """
    Data Interface for MedPC output files.
```

We tend to use this style on NeuroConv (when using `?` or `help` it also helps console clutter by adding a first layer of newline padding to the top)

src/neuroconv/datainterfaces/behavior/medpc/requirements.txt
```diff
@@ -0,0 +1 @@
+ndx-events
```
CodyCBakerPhD, line 1: Which version ranges? This is a very particular extension in that regard

src/neuroconv/datainterfaces/behavior/medpc/medpcdatainterface.py
```diff
@@ -0,0 +1,248 @@
...
+
+    keywords = ["behavior"]
+    display_name = "MedPC"
+    info = "Interface for handling MedPC output files."
+    associated_suffixes = ".txt"
+    default_medpc_name_to_info_dict = {}
+    default_events = []
+    default_interval_series = []
+    default_module_name = "behavior"
+    default_module_description = "Operant behavioral data from MedPC."
```
CodyCBakerPhD, line 45: It's certainly an interesting style to have all of these set as class attributes

I could understand it a little more if the reason was to be able to reference and control those defaults across multiple methods (such as `get_metadata` as well as `add_to_nwbfile`) but currently they only seem to be used in the `add_to_nwbfile` method, which would really imply that they are actually conversion options rather than metadata or init values

Actually, the 'medpc_name_to_info_dict' occurring in both the `__init__` signature and the instance attribute (as a default) is very confusing

I'd recommend removing these as attributes and localizing their usage within particular steps or defining a JSON 'default' in the metadata schema

CodyCBakerPhD:
Overall looks pretty good, certainly appreciate all the documentation for it

src/neuroconv/datainterfaces/behavior/medpc/medpc_helpers.py
pauladkisson: yeah I think a separate package for medpc is outside of our scope.

src/neuroconv/datainterfaces/behavior/medpc/medpc_helpers.py
pauladkisson: It makes sense for `_get_session_lines` to be private since it really only supposed to be used by `read_medpc_file`.  But I used `get_medpc_variables` directly in the Lerner Lab conversion to obtain a list of all the sessions I wanted to convert, so that should be public.  Similarly, I think `read_medpc_file` should be public in case any future conversions want to just read the medpc data and do some operation before writing to nwb.

src/neuroconv/datainterfaces/behavior/medpc/medpcdatainterface.py
pauladkisson: I know it's a bit clunky, but I needed a way to pass in the Events dictionary, IntervalSeries dictionary, and `medpc_name_to_info_dict` without using metadata (where they normally come from).  So, I built a backdoor with the default dicts.

I would much rather simply pass these dicts in with the metadata, but all of the checks in `test_all_conversion_checks` grab metadata from the interface `metadata = self.interface.get_metadata()`.  These dictionaries can't come from `get_metadata()` bc they are specified by the user -- not the data itself.

CodyCBakerPhD: That sounds like a `conversion_option` like I say above

Could you move them to that role so they are scoped and defined better?

CodyCBakerPhD: Also, I'm not understanding

> I would much rather simply pass these dicts in with the metadata, but all of the checks in test_all_conversion_checks grab metadata from the interface metadata = self.interface.get_metadata().

If you need to override something in the checks or define a special mixin for the new class, that's not unusual

CodyCBakerPhD: (which you already have anyway)

src/neuroconv/datainterfaces/behavior/medpc/medpcdatainterface.py
```diff
@@ -0,0 +1,254 @@
...
+            session_conditions=self.source_data["session_conditions"],
+            start_variable=self.source_data["start_variable"],
+        )
+        aligned_timestamps_dict = self.get_timestamps()
+        for name, aligned_timestamps in aligned_timestamps_dict.items():
+            session_dict[name] = aligned_timestamps
+
+        # Add behavior data to nwbfile
+        module_name = metadata["MedPC"].get("module_name", module_name)
+        module_description = metadata["MedPC"].get("module_description", module_description)
```
CodyCBakerPhD, line 209: It looks like, from the argument setting above, that this call to `.get` would by default return 'None' if the metadata field is not specified, correct?

I would remove the duplication of this information (module name and description) by choosing it to be specified fully in either `metadata` or as `conversion_options` (not both)

I would also set a static default

src/neuroconv/datainterfaces/behavior/medpc/medpcdatainterface.py
```diff
@@ -0,0 +1,254 @@
...
+            description = event_dict["description"]
+            event_data = session_dict[name]
+            if len(event_data) > 0:
+                event = ndx_events.Events(
+                    name=name,
+                    description=description,
+                    timestamps=event_data,
+                )
+                behavior_module.add(event)
+        interval_dicts = metadata["MedPC"].get("IntervalSeries", interval_series)
```
CodyCBakerPhD, line 228: Similar to the field above, I would recommend choosing one or the other (metadata or conversion options) for how to fetch/specify this data to avoid confusion about which to use

CodyCBakerPhD (approved):

CodyCBakerPhD:
Looks great, thanks for working on this!

bendichter:
Yes, thanks for working on this, @pauladkisson!

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -907,6 +907,239 @@ def test_interface_alignment(self):
                 self.check_nwbfile_temporal_alignment()
 
 
+class MedPCInterfaceMixin(DataInterfaceTestMixin, TemporalAlignmentMixin):
```
h-mayorquin, line 910: @CodyCBakerPhD @pauladkisson 

Hey guys I am working on testing and looking at this.

Looks like mose of these methods were copied directly from DataInterfaceTestMixing and they were not necessary.  Looking at `check_configure_backend_for_equivalent_nwbfiles` for example, the metadata line was removed but the original function should handle that fine.

Was this re-defined  for other purpose that I am missing?

pauladkisson: Yes, these tests were redefined to work on metadata passed in from the user (which is critical for MedPCInterface) rather than just interface.get_metadata()

PR #884, opened 2024-05-30
[Bug] Fix stub times in recording interface

CodyCBakerPhD:
@h-mayorquin Found this problem via the GUIDE

How would you like it to be tested?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,8 +32,9 @@
 * Added `XX` and `XO` to the base metadata schema. [PR #833](https://github.com/catalystneuro/neuroconv/pull/833)
 * `BaseImagingExtractor.add_to_nwbfile()` is fixed in the case where metadata is not supplied. [PR #849](https://github.com/catalystneuro/neuroconv/pull/849)
 * Prevent `SpikeGLXConverterPipe` from setting false properties on the sub-`SpikeGLXNIDQInterface`. [PR #860](https://github.com/catalystneuro/neuroconv/pull/860)
-* Fix bug when adding ragged arrays to the electrode and units table [PR #870](https://github.com/catalystneuro/neuroconv/pull/870)
-* Fix bug where `write_recording` will call an empty nwbfile when passing a path [PR #877](https://github.com/catalystneuro/neuroconv/pull/877)
+* Fixed a bug when adding ragged arrays to the electrode and units table. [PR #870](https://github.com/catalystneuro/neuroconv/pull/870)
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
(if not obvious from changelog; if you do any kind of alignment to the recording interface then try to run conversion in stub mode, you do not observe the alignment as being applied)

CodyCBakerPhD:
cc @garrettmflynn

h-mayorquin:
Is this a Spikeinterface problem? The times that were set are not propagated to `ConcatenateSegmentRecording`?

CodyCBakerPhD:
> Is this a Spikeinterface problem? The times that were set are not propagated to ConcatenateSegmentRecording?

I wouldn't call it a 'bug' since it seems to be intended behavior (with both default `ignore_times=True` and `False`): https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/segmentutils.py#L125-L133

Now maybe it handles the segment t start or whatever it's called, but we're setting the actual timestamps of the vector as you can see

CodyCBakerPhD:
Looks like there's already a test that maybe just needs fixing and a little enhancement?

https://github.com/catalystneuro/neuroconv/blob/fix_recording_interface_stub_times/tests/test_ecephys/test_ecephys_interfaces.py#L39-L42

CodyCBakerPhD:
Any idea why `set_times` on the concatenated recording is complaining about the size mismatch? From the logic, they should be equivalent right?

h-mayorquin:
> I wouldn't call it a 'bug' since it seems to be intended behavior (with both default ignore_times=True and False): https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/segmentutils.py#L125-L133

> Now maybe it handles the segment t start or whatever it's called, but we're setting the actual timestamps of the vector as you can see

Mmm I see.

> Any idea why `set_times` on the concatenated recording is complaining about the size mismatch? From the logic, they should be equivalent right?

Where? Or do you have a small script?

h-mayorquin:
Ah, you mean on the testing -sorry-.

I will take a look at this!

h-mayorquin:
@CodyCBakerPhD I pushed a small fix. I does fix the problem but I am not sure if it will break something else. I think that my fix is the correct operation that should be carried out.

CodyCBakerPhD:
> @CodyCBakerPhD I pushed a small fix. I does fix the problem but I am not sure if it will break something else. I think that my fix is the correct operation that should be carried out.

Thank you 🙏

LGTM, pushing changelog then good to go

h-mayorquin (approved):

PR #890, opened 2024-06-06
Debug conditional addition of watermark

CodyCBakerPhD:
@garrettmflynn Minor typo, prevented GUIDE watermark from applying


garrettmflynn (approved):
Whoops! Good catch

PR #894, opened 2024-06-10
Add backend skip for zero length axes

### LINKED ISSUE #891: [Bug]: Object of type `LabMetaDataExtension` is not JSON serializable
### What happened?

I am save a custom `.mat` matlab data file using nwb extension called tank-lab-to-nwb  ([repo](https://github.com/catalystneuro/tank-lab-to-nwb)) that uses the ndx-tank-metadata ([repo](https://github.com/catalystneuro/ndx-tank-metadata), [pip](https://pypi.org/project/ndx-tank-metadata/)) on github. 

I am able to properly load in example .mat data and add the information to an empty in-memory nwbfile by doing
```python
empty_nwb = mock_NWBFile()
interface.add_to_nwbfile(empty_nwb, metadata = metadata)
````
where interface is the custom VirmenDataInterface from the tank_lab_to_nwb repo. 

The alternative of `interface.create_nwbfile(metadata = metadata)` also safely creates an in-memory NWB file. The error comes when I execute `interface.run_conversion(nwbfile_path="./fdf13.nwb", metadata = metadata, overwrite=True)` , I end up with the error below.

Some useful information about the variables in question. 

@CodyCBakerPhD  since you had worked on this previously, do you have an idea of which variable type this might be failing?

The `obj` value when the code fails 

```python
obj
LabMetaData abc.LabMetaDataExtension at 0x5790013840
Fields:
  commit_id: 3611d165cf09db453e272b63e3f460cb95d3c10f New stationary task, copy velocity-position
  experiment_name: jessejorge_pwm_v2
  location: 165I-miniVR-T-6
  mazes: mazes <class 'ndx_tank_metadata.maze_extension.MazeExtension'>
  num_trials: 200
  protocol_name: pwm_jessejorge_v2
  rig: rig abc.RigExtension at 0x5790020608
Fields:
  arduinoPort: COM6
  ballCircumference: 63.8
  colorAdjustment: [0.  0.4 0.5]
  hasDAQ: 1
  hasSyncComm: 0
  laserChannel: 3
  leftPuffChannel: 1
  minIterationDT: 0.01
  nidaqDevice: 1
  nidaqLines: [ 0 11]
  nidaqPort: 0
  rewardChannel: 0
  rewardDuration: 0.025
  rewardSize: 0.004
  rig: 165I-miniVR-T-6
  rightPuffChannel: 2
  sensorDotsPerRev: [2310.1 2310.1 2310.1 2310.1]
  simulationMode: 0
  soundAdjustment: 0.2
  syncClockChannel: 5
  syncDataChannel: 6
  toroidXFormP1: 0.3879
  toroidXFormP2: 0.392

  session_end_time: 2024-03-22T12:09:54.779000
  stimulus_bank_path: C:\Experiments\ViRMEn\experiments\protocols\stimulus_trains_pwm_jessejorge_v2.mat
  world_file_name: jessejorge_pwm_v2
```

```python
obj.mazes
mazes ndx_tank_metadata.maze_extension.MazeExtension at 0x5790015136
Fields:
  colnames: ['world' 'lStart' 'lCue' 'lMemory' 'cueDuration' 'cueVisibleAt'
 'cueProbability' 'cueDensityPerM' 'antiFraction' 'nCueSlots'
 'tri_turnHint' 'color' 'numTrials' 'numTrialsPerMin' 'criteriaNTrials'
 'warmupNTrials' 'numSessions' 'performance' 'maxBias' 'warmupMaze'
 'warmupPerform' 'warmupBias' 'warmupMotor' 'easyBlock' 'easyBlockNTrials'
 'numBlockTrials' 'blockPerform']
  columns: (
    world <class 'hdmf.common.table.VectorData'>,
    lStart <class 'hdmf.common.table.VectorData'>,
    lCue <class 'hdmf.common.table.VectorData'>,
    lMemory <class 'hdmf.common.table.VectorData'>,
    cueDuration <class 'hdmf.common.table.VectorData'>,
    cueVisibleAt <class 'hdmf.common.table.VectorData'>,
    cueProbability <class 'hdmf.common.table.VectorData'>,
    cueDensityPerM <class 'hdmf.common.table.VectorData'>,
    antiFraction <class 'hdmf.common.table.VectorData'>,
    nCueSlots <class 'hdmf.common.table.VectorData'>,
    tri_turnHint <class 'hdmf.common.table.VectorData'>,
    color <class 'hdmf.common.table.VectorData'>,
    numTrials <class 'hdmf.common.table.VectorData'>,
    numTrialsPerMin <class 'hdmf.common.table.VectorData'>,
    criteriaNTrials <class 'hdmf.common.table.VectorData'>,
    warmupNTrials <class 'hdmf.common.table.VectorData'>,
    numSessions <class 'hdmf.common.table.VectorData'>,
    performance <class 'hdmf.common.table.VectorData'>,
    maxBias <class 'hdmf.common.table.VectorData'>,
    warmupMaze <class 'hdmf.common.table.VectorData'>,
    warmupPerform <class 'hdmf.common.table.VectorData'>,
    warmupBias <class 'hdmf.common.table.VectorData'>,
    warmupMotor <class 'hdmf.common.table.VectorData'>,
    easyBlock <class 'hdmf.common.table.VectorData'>,
    easyBlockNTrials <class 'hdmf.common.table.VectorData'>,
    numBlockTrials <class 'hdmf.common.table.VectorData'>,
    blockPerform <class 'hdmf.common.table.VectorData'>
  )
  description: description of the mazes
  id: id <class 'hdmf.common.table.ElementIdentifiers'
```

```python
obj.rig
rig abc.RigExtension at 0x5790020608
Fields:
  arduinoPort: COM6
  ballCircumference: 63.8
  colorAdjustment: [0.  0.4 0.5]
  hasDAQ: 1
  hasSyncComm: 0
  laserChannel: 3
  leftPuffChannel: 1
  minIterationDT: 0.01
  nidaqDevice: 1
  nidaqLines: [ 0 11]
  nidaqPort: 0
  rewardChannel: 0
  rewardDuration: 0.025
  rewardSize: 0.004
  rig: 165I-miniVR-T-6
  rightPuffChannel: 2
  sensorDotsPerRev: [2310.1 2310.1 2310.1 2310.1]
  simulationMode: 0
  soundAdjustment: 0.2
  syncClockChannel: 5
  syncDataChannel: 6
  toroidXFormP1: 0.3879
  toroidXFormP2: 0.392
```

### Steps to Reproduce

```python
# %% The script used to get this bug
"""Author: Christian"""
# from tank_lab_to_nwb import VirmenDataInterface
import sys
import os
sys.path.append(os.getcwd())
print(os.getcwd())

from pathlib import Path
from isodate import duration_isoformat
from datetime import timedelta, datetime

import neuroconv

from tank_lab_to_nwb.convert_towers_task.virmenbehaviordatainterface import VirmenDataInterface



# %%
base_path = Path("/Users/ct5868/code/tank-lab-to-nwb/")
virmen_file_path = base_path / "behavior_local" / "jessejorge_pwm_jessejorge_pwm_pilot_165I-miniVR-T-6_jyanar_ya011_T_20240322.mat"


interface = VirmenDataInterface(file_path = virmen_file_path)

# %%
print(metadata)

# %%
metadata["Subject"] = dict(
    subject_id="M001",
    sex="M",
    age="P30D",
    species="Mus musculus",
)


# %%
interface.get_metadata_schema()

# %%
from datetime import datetime
from zoneinfo import ZoneInfo

metadata["NWBFile"]["session_start_time"] = datetime(2021, 1, 1, 12, 0, 0, tzinfo=ZoneInfo("US/Pacific"))

# %%
interface.get_metadata_schema()

# %%
empty_nwb = mock_NWBFile()

print(metadata)


# %%


interface.create_nwbfile(metadata = metadata)
interface.run_conversion(nwbfile_path="./mariokart.nwb", metadata = metadata, overwrite=True)
# interface.run_conversion(nwbfile_path="./fdf13.nwb", metadata = metadata, overwrite=True)

# %%
```

### Traceback

```python
Exception has occurred: TypeError
Object of type LabMetaDataExtension is not JSON serializable
  File "/Users/ct5868/code/neuroconv/src/neuroconv/utils/json_schema.py", line 34, in default
    return super().default(obj)
  File "/Users/ct5868/code/neuroconv/src/neuroconv/utils/json_schema.py", line 350, in validate_metadata
    serialized_metadata = encoder.encode(metadata)
  File "/Users/ct5868/code/neuroconv/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py", line 70, in make_nwbfile_from_metadata
    validate_metadata(metadata=metadata, schema=base_metadata_schema)
  File "/Users/ct5868/code/neuroconv/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py", line 231, in make_or_load_nwbfile
    raise e
  File "/Users/ct5868/code/neuroconv/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py", line 231, in make_or_load_nwbfile
    raise e
  File "/Users/ct5868/code/neuroconv/src/neuroconv/basedatainterface.py", line 167, in run_conversion
    with make_or_load_nwbfile(
  File "/Users/ct5868/code/tank-lab-to-nwb/notebooks/attempting behavior datainterfaces--debugging.py", line 62, in <module>
    interface.run_conversion(nwbfile_path="./mariokart.nwb", metadata = metadata, overwrite=True)
TypeError: Object of type LabMetaDataExtension is not JSON serializable
LabMetaData
```

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.10

### Package Versions

[environment_for_issue.txt](https://github.com/user-attachments/files/15742688/environment_for_issue.txt)

### Code of Conduct

CodyCBakerPhD:
fix #891


tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_zero_length_axes.py
```diff
@@ -0,0 +1,135 @@
...
+    low = np.iinfo(dtype).min
+    high = np.iinfo(dtype).max
+    return random_number_generator.integers(low=low, high=high, size=shape, dtype=dtype)
+
+
+@pytest.fixture(scope="session")
+def integer_array_with_zero_length_axis(
+    seed: int = 1,
+    dtype: np.dtype = np.dtype("int16"),
+    shape: Tuple[int, int] = (12, 0),
```
h-mayorquin, line 45:
```suggestion
    shape: Tuple[int, int] = (12, 0),  # 12 so it matches the dimension of the other column
```
Is this the reason?

h-mayorquin (commented):
The fix makes sense. Thanks for the tests.

 Question:
This comes all the way from the `GenericDataChunkIterator` passess through `DatasetIOConfiguration` and then it surfaces where we do the fixes here.

Would it make sense to fix it deeper  yet? It seems that at least propagating this fix to the level of `from_neurodata_object` as we already have the full shape there and  I see that both full shape and chunk shape are optional arguments.

Do you think that could work?

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_zero_length_axes.py
CodyCBakerPhD: Yep

CodyCBakerPhD:
> Would it make sense to fix it deeper yet? It seems that at least propagating this fix to the level of from_neurodata_object as we already have the full shape there and I see that both full shape and chunk shape are optional arguments.
> 
> Do you think that could work?

I debated how deeply to fix it but decided this shallow level is where it should be (alongside all the other skip conditions too)

The thing is, we really shouldn't be applying an iterator to something with a zero-length axis (or at least, if we wanted to, adding support for that at the top-level HDMF would be a much bigger development)

Likewise, we want `DatasetIOConfiguration.from_neurodata_object(...)` to error (rather than return `None`) in situations where it shouldn't apply, and setting control of zero-length parts of a chunk is broadly not something we've ever supported or tested (not that it couldn't be, but again would need some top-level effort and propagation)

So skipping it in the helper function for forming the entire backend configuration makes most sense to me

CodyCBakerPhD:
We should probably wait for @tabedzki to confirm this works for them before merging

@tabedzki 

Please try

```bash
pip install git+https://github.com/catalystneuro/neuroconv.git@skip_dataset_configuration_zero_length_axis
```

and let me know if that works with the workflow I outlined in the issue

h-mayorquin:
OK, 
> Likewise, we want `DatasetIOConfiguration.from_neurodata_object(...)` to error (rather than return `None`) in situations where it shouldn't apply, and setting control of zero-length parts of a chunk is broadly not something we've ever supported or tested (not that it couldn't be, but again would need some top-level effort and propagation)

I was thinking that we should set chunk_size=None and buffer_size=None in those cases. My understanding is that is equivalent to skipping. Am I wrong? But, having an explicit error at that level makes sense to me as well. 

OK on waiting btw. This is ready to me, so let's just wait for @tabedzki to check if the fix works.

CodyCBakerPhD:
> I was thinking that we should set chunk_size=None and buffer_size=None in those cases.

The issue with that is (while that is accurate to how it ends up on disk) from an API perspective we make it _seem_ like those values could be changed when in reality they cannot; we've not built mutation validation conditions for attempts to set non-`None` chunks on objects that we otherwise don't support chunking in general on

h-mayorquin:
> > I was thinking that we should set chunk_size=None and buffer_size=None in those cases.
> 
> The issue with that is (while that is accurate to how it ends up on disk) from an API perspective we make it _seem_ like those values could be changed when in reality they cannot; we've not built mutation validation conditions for attempts to set non-`None` chunks on objects that we otherwise don't support chunking in general on

OK, thanks for explaining. Can I summarize this point as things in configuration should have mutable properties for compression and iterative writing parameters but here we can't relally change them (they can only be None) so better not to include them in the dictionary?

This makes sense.

tabedzki:
Thank you! I'll be able to test this later and I'll let you know how it goes

tabedzki:
I think it works as the error hasn't appeared again but I cannot 100% say as I still have some errors converting the Matlab file. I think one of the fields within the matlab struct is not being properly handled. Will report back later

h-mayorquin (approved):
Thanks @tabedzki.

@CodyCBakerPhD I think we should merge as this solves a bug.

tabedzki:
@CodyCBakerPhD  I was able to confirm that using this branch in lieu of the main branch does indeed address this issue.

Thank you!

PR #897, opened 2024-06-11
some edits to the backend configuration tutorial

bendichter:
Reading this page over again, I feel the first section, [Default configuration](https://neuroconv.readthedocs.io/en/main/user_guide/backend_configuration.html#default-configuration), does a fine job of showing how you would use these tools outside the context of `DataInterface` and `NWBConverter` classes. From what I can tell now that we have enhanced that first section the only thing this last section is adding is the use of `make_or_load_nwbfile`. Since we are providing an in-memory NWB File in that section, it must be writing a new file, which is the same action that is being taken in the first section, and I think it would be confusing to show two different ways of doing the exact same thing.

### DIFF
--- a/docs/user_guide/backend_configuration.rst
+++ b/docs/user_guide/backend_configuration.rst
@@ -142,7 +142,10 @@ Then we can use this configuration to write the NWB file:
 Interfaces and Converters
 -------------------------
 
-The normal workflow when writing an NWB file using a ``DataInterface`` or ``NWBConverter`` is simple to configure.
+A similar workflow can be used when writing an NWB file using a ``DataInterface`` or ``NWBConverter`` is simple to configure.
+
+Having get_default_backend_configuration as a method of the DataInterface and NWBConverter classes allows descending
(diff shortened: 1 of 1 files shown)


docs/user_guide/backend_configuration.rst
```diff
@@ -173,13 +176,11 @@ The following example uses the :ref:`example data <example_data>` available from
...
 
-    # Create the in-memory NWBFile object and retrieve a default configuration
-    backend="hdf5"
-
+    # Create the in-memory NWBFile object and retrieve a default configuration for the backend
     nwbfile = converter.create_nwbfile(metadata=metadata)
     backend_configuration = converter.get_default_backend_configuration(
         nwbfile=nwbfile,
-        backend=backend,
+        backend="hdf5,
```
CodyCBakerPhD, line 183:
```suggestion
        backend="hdf5",
```

CodyCBakerPhD (approved):

PR #902, opened 2024-06-12
[Cloud Deployment IIa] Rclone docker image extension for config file

CodyCBakerPhD:
Similar to the YAML based dockerfile, it's smoothest for AWS integration to just pass the file stream via an environment variable, and since the config file will never be larger than the size limit on environment variables this is fine

Any alternative approach would need to deal with file transfer of the configuration via separate operations, which adds more complexity


CodyCBakerPhD:
OK, I finally got this working!

It's probably best if you review this @h-mayorquin so you can learn how the Docker stuff works for maintenance going forward. I mean, hopefully it would never break, but maybe updating various tokens will be required at some point, or updated documentation, etc.

h-mayorquin (approved):
All right we went throug this together yesterday. I read it once more again to see if everything clicks and it makes sense to me. 

Approving.

PR #904, opened 2024-06-13
Add packaging explicitly

CodyCBakerPhD:
No clue why no CI anywhere in our ecosystem catches this during minimal installation and imports

But adding it explicitly anyway; it is a third-party package and up to this point I figure we just assumed something upstream relied on it implicitly

cc @rly @laurelrr

### DIFF
--- a/requirements-minimal.txt
+++ b/requirements-minimal.txt
@@ -15,3 +15,4 @@ pandas
 parse>=1.20.0
 click
 docstring-parser
+packaging  # Issue 903


h-mayorquin (commented):
I think the CI migth be missing it because some of the test dependencies install it implicitly.

Could a from neuroconv import * with install plain neuroconv as a test?

Would that work with the current importing structure or we would need to try to import everything case by case to see if it fails?

CodyCBakerPhD:
> Could a from neuroconv import * with install plain neuroconv as a test?

`import neuroconv` is all we would need since it isn't nested as a lazy dependency anywhere (seems to assume it ends up in the third party dependency structure somewhere somehow)

Added a step to the basic testing workflow to ensure that without any extra's specified

h-mayorquin:
Ups, I tried to push this to another branch for testing but I accidentally pushed to yours, sorry @CodyCBakerPhD .

h-mayorquin:
Anyway, this looks good to me. I want to see if the new test fails without this and then I will rever the last change and is good to go.

CodyCBakerPhD:
Or rather, it would on the users original setup and device, for whatever reason. But the way in the systems of the CI this wasn't and still isn't a problem

CodyCBakerPhD:
(still doesn't hurt in general to have that import test though)

h-mayorquin:
Ok, packaging is installed indirectly through `pytest-xdist` on these lines:

https://github.com/catalystneuro/neuroconv/blob/14f0b48430ab5264d7c34f132c6878b0226ce93c/.github/workflows/testing.yml#L32-L38

Which in turn has `pytest` as a dependency

https://github.com/pytest-dev/pytest-xdist/blob/c7b4f6114479d4053f0ef8411ed5e0a410b207f4/pyproject.toml#L35-L38

Which has packaging:
https://github.com/pytest-dev/pytest/blob/2effd8cb2c91abf4782201c431f19d0b22517920/pyproject.toml#L48-L54

That's the reason our CI is not detecting this failure.

We should move pytest-xdist and wheel to the testing dependencies file I think.

h-mayorquin (approved):
Let's merge this, I will take care of moving the rest.

PR #911, opened 2024-06-13
Fix overwrite on corrupt file

### LINKED ISSUE #910: [Bug]: run_conversion(overwrite=True) does not overwrite an improperly formatted NWB file
### What happened?

During debugging of `add_to_nwbfile` in my code, I would often crash or quit the python process part of the way through writing the NWB file, leaving it improperly formatted. 

So running the following code twice (once that you stop part way through and once after that) will yield the errors below. Is there any way to handle this error? 

### Steps to Reproduce

```python
interface.run_conversion(
    # nwbfile_path=nwbfile_path,
    nwbfile_path="./fdf19.nwb",
    metadata=metadata,
    overwrite=True)
```

### Traceback

```python
TypeError                                 Traceback (most recent call last)
Cell In[9], line 22
     16 print(metadata)
     19 # interface.create_nwbfile(metadata=metadata)
     20 # interface.create_nwbfile(metadata=metadata)
---> 22 interface.run_conversion(
     23     # nwbfile_path=nwbfile_path,
     24     nwbfile_path="./fdf19.nwb",
     25     metadata=metadata,
     26     overwrite=True
     27     )

File ~/code/neuroconv/src/neuroconv/basedatainterface.py:167, in BaseDataInterface.run_conversion(self, nwbfile_path, nwbfile, metadata, overwrite, backend, backend_configuration, **conversion_options)
    164 if metadata is None:
    165     metadata = self.get_metadata()
--> 167 with make_or_load_nwbfile(
    168     nwbfile_path=nwbfile_path,
    169     nwbfile=nwbfile,
    170     metadata=metadata,
    171     overwrite=overwrite,
    172     backend=backend,
    173     verbose=getattr(self, "verbose", False),
    174 ) as nwbfile_out:
    175     if no_nwbfile_provided:
    176         self.add_to_nwbfile(nwbfile=nwbfile_out, metadata=metadata, **conversion_options)

File /opt/anaconda3/envs/neuroconv_3.12/lib/python3.12/contextlib.py:137, in _GeneratorContextManager.__enter__(self)
    135 del self.args, self.kwds, self.func
    136 try:
--> 137     return next(self.gen)
    138 except StopIteration:
    139     raise RuntimeError("generator didn't yield") from None

File ~/code/neuroconv/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py:209, in make_or_load_nwbfile(nwbfile_path, nwbfile, metadata, overwrite, backend, verbose)
    203 else:
    204     load_kwargs.update(mode="w")
    206 backends_that_can_read = [
    207     backend_name
    208     for backend_name, backend_io_class in BACKEND_NWB_IO.items()
--> 209     if backend_io_class.can_read(path=str(nwbfile_path_in))
    210 ]
    211 # Future-proofing: raise an error if more than one backend can read the file
    212 assert (
    213     len(backends_that_can_read) <= 1
    214 ), "More than one backend is capable of reading the file! Please raise an issue describing your file."

File /opt/anaconda3/envs/neuroconv_3.12/lib/python3.12/site-packages/pynwb/__init__.py:272, in NWBHDF5IO.can_read(path)
    270 try:
    271     with h5py.File(path, "r") as file:   # path is HDF5 file
--> 272         return get_nwbfile_version(file)[1][0] >= 2    # Major version of NWB >= 2
    273 except IOError:
    274     return False

TypeError: 'NoneType' object is not subscriptable
```

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.12

### Package Versions

_No response_

### Code of Conduct

CodyCBakerPhD:
fix #910

### DIFF
--- a/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
@@ -195,7 +195,7 @@ def make_or_load_nwbfile(
 
     load_kwargs = dict()
     file_initially_exists = nwbfile_path_in.exists() if nwbfile_path_in is not None else None
-    if nwbfile_path_in:
+    if nwbfile_path_in is not None:
         load_kwargs.update(path=str(nwbfile_path_in))
 
         append_mode = file_initially_exists and not overwrite
@@ -233,22 +233,23 @@ def make_or_load_nwbfile(
         nwbfile_loaded_succesfully = False
         raise load_error
     finally:
-        if nwbfile_path_in:
+        if nwbfile_path_in is not None and nwbfile_loaded_succesfully:
             nwbfile_written_succesfully = True
             try:
-                if nwbfile_loaded_succesfully:
-                    io.write(nwbfile)
+                io.write(nwbfile)
 
-                    if verbose:
-                        print(f"NWB file saved at {nwbfile_path_in}!")
+                if verbose:
+                    print(f"NWB file saved at {nwbfile_path_in}!")
             except Exception as write_error:
                 nwbfile_written_succesfully = False
                 raise write_error
             finally:
                 io.close()
 
-                if not nwbfile_written_succesfully and not file_initially_exists:
-                    nwbfile_path_in.unlink()
+        any_load_or_write_error = not nwbfile_loaded_succesfully or not nwbfile_written_succesfully
+        attempt_to_remove_existing_file = any_load_or_write_error and not file_initially_exists
+        if attempt_to_remove_existing_file:
+            nwbfile_path_in.unlink()
 
 
 def _resolve_backend(
--- a/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
@@ -10,9 +10,10 @@ from pathlib import Path
 from typing import Literal, Optional
 from warnings import warn
 
+import h5py
 from hdmf_zarr import NWBZarrIO
 from pydantic import FilePath
-from pynwb import NWBHDF5IO, NWBFile
+from pynwb import NWBHDF5IO, NWBFile, get_nwbfile_version
 from pynwb.file import Subject
 
 from . import BackendConfiguration, configure_backend, get_default_backend_configuration
@@ -200,23 +201,23 @@ def make_or_load_nwbfile(
         load_kwargs.update(path=str(nwbfile_path_in))
         if file_initially_exists and not overwrite:
             load_kwargs.update(mode="r+", load_namespaces=True)
+
+            backends_that_can_read = [
+                backend_name
+                for backend_name, backend_io_class in BACKEND_NWB_IO.items()
+                if backend_io_class.can_read(path=str(nwbfile_path_in))
+            ]
+            # Future-proofing: raise an error if more than one backend can read the file
+            assert (
+                len(backends_that_can_read) <= 1
+            ), "More than one backend is capable of reading the file! Please raise an issue describing your file."
+            if backend not in backends_that_can_read:
+                raise IOError(
+                    f"The chosen backend ('{backend}') is unable to read the file! "
+                    f"Please select '{backends_that_can_read[0]}' instead."
+                )
         else:
             load_kwargs.update(mode="w")
-
-        backends_that_can_read = [
-            backend_name
-            for backend_name, backend_io_class in BACKEND_NWB_IO.items()
-            if backend_io_class.can_read(path=str(nwbfile_path_in))
-        ]
-        # Future-proofing: raise an error if more than one backend can read the file
-        assert (
-            len(backends_that_can_read) <= 1
-        ), "More than one backend is capable of reading the file! Please raise an issue describing your file."
-        if load_kwargs["mode"] == "r+" and backend not in backends_that_can_read:
-            raise IOError(
-                f"The chosen backend ('{backend}') is unable to read the file! "
-                f"Please select '{backends_that_can_read[0]}' instead."
-            )
 
         io = backend_io_class(**load_kwargs)
 
--- a/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
@@ -147,6 +147,17 @@ def add_device_from_metadata(nwbfile: NWBFile, modality: str = "Ecephys", metada
             nwbfile.create_device(**dict(defaults, **device_metadata))
 
 
+def _cleanup_existing_nwbfile(nwbfile_path: FilePath) -> None:
+    try:
+        nwbfile_path.unlink()
+    # Windows in particular can encounter errors at this step
+    except PermissionError:  # pragma: no cover
+        message = (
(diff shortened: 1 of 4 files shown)


tests/test_minimal/test_tools/test_context_tools.py
```diff
@@ -197,3 +199,16 @@ def test_make_or_load_nwbfile_pass_nwbfile(self):
             nwbfile_out = io.read()
             assert "test1" in nwbfile_out.acquisition
             assert "test2" in nwbfile_out.acquisition
+
+
+def test_make_or_load_nwbfile_on_corrupt_file(tmpdir: Path) -> None:
```
h-mayorquin, line 204: I think it would be beetter if this test were to import h5py explictly instead of doing it implicitly through

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -10,9 +10,10 @@
 from typing import Literal, Optional
 from warnings import warn
 
+import h5py
```
h-mayorquin, line 13: Remove unused imports.

h-mayorquin (commented):
I think this makes sense even outside of the scope of the bug, right?

That is, the test for backend matching only makes sense in the context of append mode, right? If you are going to ovewrite the file then no test is really necessary.

CodyCBakerPhD:
> That is, the test for backend matching only makes sense in the context of append mode, right? If you are going to ovewrite the file then no test is really necessary.

That's the idea - I think we had it outside previously to reduce cyclomatic complexity, but it might actually deserve to be nested there logically

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
CodyCBakerPhD: Yeah I started down this path initially, but found it was the wrong direction; you cleaned up in the other PR though

tests/test_minimal/test_tools/test_context_tools.py
CodyCBakerPhD: Can you clarify?

h-mayorquin: The h5py was not imported in the test module. But you added already. When I removed the h5py import in the PR  module the tests were failing.

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -194,53 +194,63 @@ def make_or_load_nwbfile(
...
                     if verbose:
                         print(f"NWB file saved at {nwbfile_path_in}!")
+            except Exception as write_error:
+                nwbfile_written_succesfully = False
+                raise write_error
             finally:
                 io.close()
 
-                if not success and not file_initially_exists:
+                if not nwbfile_loaded_succesfully and not file_initially_exists:  # Not sure we need this
```
h-mayorquin, line 250: Should we moreve this as you suggested? I  don't see how we ever get here.

h-mayorquin: This is required because apparently the file is created by the IO and not by write (as I thought):

```python
from pynwb import NWBHDF5IO

nwb_file_path = Path("./test.nwb")
nwb_file_path.unlink(missing_ok=True)
assert not nwb_file_path.exists()
NWBHDF5IO(path=nwb_file_path, mode="w")
assert nwb_file_path.exists()
```

CodyCBakerPhD: Nice catch; I expanded the logic for the one cleanup condition and made it more explicit about when we'd actually want to cleanup

h-mayorquin:
So what remains is the windows problem for the beahvior of unlik. I wonder if this is caused by our parallel suite? Maybe creating a temporary directory within the test itself will solve this problem.

CodyCBakerPhD:
The file is only used by a single process at a time, this is a pretty common Windows thing

h-mayorquin:
But something else must be using the file, right? I know the bug because we encountered it on spikeinterface. Linux allows you to mark something for deletion if it is being used but windows does not...  so it seems to me that something should be using when you try to unlik it?

CodyCBakerPhD:
Right but to that point we've closed all known handles to best of our ability; occasionally comes down to garbage collection and whatnot, some spare thing in the kernel hasn't released it yet (but resetting the kernel usually does the trick)

Usually a special pain in notebooks and the like

I wouldn't read too much into it. Maybe it'll disappear when we refactor this function into the distinct parts in a future follow-up

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -194,54 +194,73 @@ def make_or_load_nwbfile(
...
+        any_load_or_write_error = not nwbfile_loaded_succesfully or not nwbfile_written_succesfully
+        file_was_freshly_created = (
+            not file_initially_exists and nwbfile_path_in is not None and nwbfile_path_in.exists()
+        )
+        attempt_to_cleanup = any_load_or_write_error and file_was_freshly_created
+        if attempt_to_cleanup:
+            try:
+                nwbfile_path_in.unlink()
+            # Windows in particular can encounter errors at this step
+            except PermissionError:  # pragma: no cover
```
h-mayorquin, line 259: Seems like a sensible solution

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -194,54 +194,73 @@ def make_or_load_nwbfile(
...
+                nwbfile_written_succesfully = False
+                raise write_error
             finally:
                 io.close()
-
-                if not success and not file_initially_exists:
-                    nwbfile_path_in.unlink()
+                del io
+
+        any_load_or_write_error = not nwbfile_loaded_succesfully or not nwbfile_written_succesfully
```
h-mayorquin, line 250: question: why did indenting this in the second finally did not work out?

h-mayorquin (approved):
Seems good to me.

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
CodyCBakerPhD: Not entirely sure what you're asking

There could be a way of including it in the parent nearby `if` of line 237 but thinking about all the interacting logic gets a bit messy and I figured this was more readable if not more typing

h-mayorquin: The block starting in 250.
I have this impression that first it was within the finally of 246 but that it was moved out.
I am asking if there was a reason for that out of curiosity?

CodyCBakerPhD: I've gone back and forth; it can fit in various levels depending on the logical conditions

I expanded the logic to make a bit more sense in each case, even at the minor cost of repeating the operation multiple times (no harm due to early exit)

h-mayorquin: Ah, OK.  This makes sense.

h-mayorquin:
Thanks for this improvement @CodyCBakerPhD

PR #912, opened 2024-06-13
add logic to erase file if writing fails

h-mayorquin:
Discussion from issue https://github.com/catalystneuro/neuroconv/issues/910. @CodyCBakerPhD, this is what I had in mind.

### DIFF
--- a/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
@@ -10,10 +10,9 @@ from pathlib import Path
 from typing import Literal, Optional
 from warnings import warn
 
-import h5py
 from hdmf_zarr import NWBZarrIO
 from pydantic import FilePath
-from pynwb import NWBHDF5IO, NWBFile, get_nwbfile_version
+from pynwb import NWBHDF5IO, NWBFile
 from pynwb.file import Subject
 
 from . import BackendConfiguration, configure_backend, get_default_backend_configuration
@@ -195,13 +194,15 @@ def make_or_load_nwbfile(
         raise NotImplementedError("Appending a Zarr file is not yet supported!")
 
     load_kwargs = dict()
-    success = True
     file_initially_exists = nwbfile_path_in.exists() if nwbfile_path_in is not None else None
     if nwbfile_path_in:
         load_kwargs.update(path=str(nwbfile_path_in))
-        if file_initially_exists and not overwrite:
+
+        append_mode = file_initially_exists and not overwrite
+        if append_mode:
             load_kwargs.update(mode="r+", load_namespaces=True)
 
+            # Check if the selected backend is the backend of the file in nwfile_path
             backends_that_can_read = [
                 backend_name
                 for backend_name, backend_io_class in BACKEND_NWB_IO.items()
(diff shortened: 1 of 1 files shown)


src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -221,27 +222,35 @@ def make_or_load_nwbfile(
...
                     if verbose:
                         print(f"NWB file saved at {nwbfile_path_in}!")
+            except Exception as write_error:
+                nwbfile_written_succesfully = False
+                raise write_error
             finally:
                 io.close()
 
-                if not success and not file_initially_exists:
+                if not nwbfile_loaded_succesfully and not file_initially_exists:  # Not sure we need this
```
CodyCBakerPhD, line 250: If no file initially exists then there is nothing to unlink?

h-mayorquin: Makes sense.

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -221,27 +222,35 @@ def make_or_load_nwbfile(
...
         if load_kwargs.get("mode", "") == "r+":
             nwbfile = io.read()
         elif nwbfile is None:
             nwbfile = make_nwbfile_from_metadata(metadata=metadata)
         yield nwbfile
-    except Exception as e:
-        success = False
-        raise e
+    except Exception as load_error:
+        nwbfile_loaded_succesfully = False
```
CodyCBakerPhD, line 233: Could have been a problem with `metadata` in `make_nwbfile_from_metadata` that is getting caught in this block, separate from file read

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -221,27 +222,35 @@ def make_or_load_nwbfile(
...
+                nwbfile_written_succesfully = False
+                raise write_error
             finally:
                 io.close()
 
-                if not success and not file_initially_exists:
+                if not nwbfile_loaded_succesfully and not file_initially_exists:  # Not sure we need this
+                    nwbfile_path_in.unlink()
+
+                if not nwbfile_written_succesfully and not file_initially_exists:
```
CodyCBakerPhD, line 253: Aside from better variable naming, the key difference in this PR is this line by limiting the global notion of 'success' to just 'success during write process' and if file didn't initially exist but write started and failed, try to clean it up

Emphasis on 'try' since not guaranteed to succeed without error

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -221,27 +222,35 @@ def make_or_load_nwbfile(
...
             try:
-                if success:
+                if nwbfile_loaded_succesfully:
                     io.write(nwbfile)
 
                     if verbose:
                         print(f"NWB file saved at {nwbfile_path_in}!")
+            except Exception as write_error:
+                nwbfile_written_succesfully = False
+                raise write_error
```
CodyCBakerPhD, line 246: I see, you also added a capture block here to further scope the success of the operation

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
h-mayorquin: Yep. I think that could be that's why I am calling ti "load_error" instead of "create_error" to cover both read and create_with_metadata errors.

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
h-mayorquin: Right. I think this is the core. Separating errors from loading or creating from errors of writing.

But as you were mentioning maybe us handling unlinking is not a good idea ...

CodyCBakerPhD: In this case we're attempting to leave the system as we left it, which is fine

The case I am against is what you don't have here

```python
if any_error_condition and file_initially_exists:
    nwbfile_path_in.unlink()
```

CodyCBakerPhD (approved):

PR #913, opened 2024-06-13
Move testing dependencies on the CI to testing requirements.

h-mayorquin:
From the discussion on https://github.com/catalystneuro/neuroconv/pull/904 were some dependencies were installed implictly on the CI through testing libraries:

pytest-xdist and wheel.

I am moving the first to the testing-requirements file and leaving wheel out to see exactly where we neeed it. Either at the level of scanimage requirements or as a general testing dependency.

### DIFF
--- a/.github/workflows/dev-testing.yml
+++ b/.github/workflows/dev-testing.yml
@@ -37,9 +37,6 @@ jobs:
           python -m pip install -U pip  # Official recommended way
           git config --global user.email "CI@example.com"
           git config --global user.name "CI Almighty"
-          pip install wheel # Needed for scan image
-
-
 
       - name: Install full requirements
         run: pip install --no-cache-dir .[full,test]
--- a/.github/workflows/dev-testing.yml
+++ b/.github/workflows/dev-testing.yml
@@ -35,7 +35,6 @@ jobs:
       - name: Global Setup
         run: |
           python -m pip install -U pip  # Official recommended way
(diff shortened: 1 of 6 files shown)


.github/workflows/dev-testing.yml
```diff
@@ -35,7 +35,6 @@ jobs:
       - name: Global Setup
         run: |
           python -m pip install -U pip  # Official recommended way
-          pip install pytest-xdist
           git config --global user.email "CI@example.com"
           git config --global user.name "CI Almighty"
           pip install wheel # Needed for scan image
```
CodyCBakerPhD, line 40: Keeping `wheel` in dev but not elsewhere?

h-mayorquin: Not necessary since the latest version of scan image finally decided to dish setup.py:

https://gitlab.com/vidriotech/scanimagetiffreader-python/-/merge_requests/2/diffs

CodyCBakerPhD: So are you intending to remove it from this line or still keeping it?

h-mayorquin: Ah, sorry, I misread, I will remove it here as well : )

h-mayorquin: Done and if test pass this should be ready to go.

CodyCBakerPhD (approved):

Issue #915, opened 2024-06-17
[Feature]: Option for multiple DLC inputs


vigji:
### What would you like to see added to NeuroConv?

I am trying to load multiple videos and DLC files from an experiment where different cameras with different frame rates etc look at different parts of the animal. as a result, I have something like:
```
animal1data
   |_video1.avi
   |_video1DLCetc.h5
   |_video2.avi
   |_video2DLCetc.h5
   
DLC_models_dir
  |_model_for_video1/config.yaml
  |_model_for_video2/config.yaml
  
```

I can load multiple camera videos passing a list to the `VideoInterface`. Already here it feels like I'm stretching the class' domain, as from what I understand this would be oriented toward multiple segments *in time*, correct?

On the other hand, it is not clear to me if it is possible to load multiple DLC tracking results. I tried to have multiple `DeeplabcutInterface` objects passed to the `ConverterPipe` combinator but it complains about having multiple data of the same category (`ValueError: 'PoseEstimation' already exists in ProcessingModule 'behavior'`)

### Is your feature request related to a problem?

I cannot find a way to implement multiple DLC labelling in the same file.

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

CodyCBakerPhD:
We rely on DLC2NWB for this functionality; I see you've already started contributing some code to that

Would you be willing to submit PRs on their end to allow control over the name of the `PostEstimation` object at this line (https://github.com/DeepLabCut/DLC2NWB/blob/main/dlc2nwb/utils.py#L223)?

As shown in the README for `ndx-pose`: https://github.com/DeepLabCut/DLC2NWB/blob/main/dlc2nwb/utils.py#L223

vigji:
Open to try anything but 

> allow control over the name of the PostEstimation object at this line 
I think you will have to elaborate a bit more as I did not get this point and how I should procede - I did not find pointers in the `ndx-pose` README (probably you pasted twice the same link?)

CodyCBakerPhD:
> I did not find pointers in the ndx-pose README (probably you pasted twice the same link?)

Whoops,  updated

Basically just inject a keyword argument for `post_estimation_name` that gets set at that line as

[...]

PR #917, opened 2024-06-19
add name parameter to the `DeepLabCutInterface`

vigji:
This PR should fix the issue described in  #915.

It consists in the small addition of a name argument to the `DeepLabCutInterface` that allows users to specify names different from the `nwb-pose` default `PoseEstimation` (https://github.com/rly/ndx-pose/blob/a847ad4be75e60ef9e413b8cbfc99c616fc9fd05/spec/ndx-pose.extensions.yaml#L71).

To work, the call to `dlc2nwb. _write_pes_to_nwbfile()` should accept additional arguments, as described in https://github.com/DeepLabCut/DLC2NWB/issues/24. At the moment, and until `dlc2nwb ` people give feedback on that issue, this functionality would require a different fork of `dlc2nwb`: https://github.com/vigji/DLC2NWB.

### DIFF
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -6,4 +6,3 @@ ndx-miniscope
 spikeinterface[qualitymetrics]>=0.100.0
 zarr<2.18.0  # Error with Blosc (read-only during decode) in numcodecs on May 7; check later if resolved
 pytest-xdist
-dlc2nwb @ git+https://github.com/vigji/dlc2nwb.git@main
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -6,3 +6,4 @@ ndx-miniscope
 spikeinterface[qualitymetrics]>=0.100.0
 zarr<2.18.0  # Error with Blosc (read-only during decode) in numcodecs on May 7; check later if resolved
 pytest-xdist
+dlc2nwb @ git+https://github.com/vigji/dlc2nwb.git@main
--- a/src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
+++ b/src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
@@ -15,6 +15,7 @@ def write_subject_to_nwb(
     individual_name: str,
     config_file: FilePathType,
     timestamps: Optional[Union[List, np.ndarray]] = None,
+    name: Optional[str] = "PoseEstimation",
 ):
     """
     Given, subject name, write h5file to an existing nwbfile.
@@ -32,6 +33,8 @@ def write_subject_to_nwb(
         Path to a project config.yaml file
     timestamps : list, np.ndarray or None, default: None
         Alternative timestamps vector. If None, then use the inferred timestamps from DLC2NWB
+    name : str or None, default: PoseEstimation
+        Description of the pose estimation procedure and output. Default from npx-pose.
     Returns
     -------
     nwbfile : pynwb.NWBFile
@@ -45,7 +48,7 @@ def write_subject_to_nwb(
 
     df_animal = df.groupby(level="individuals", axis=1).get_group(individual_name)
     return dlc2nwb.utils._write_pes_to_nwbfile(
-        nwbfile, individual_name, df_animal, scorer, video, paf_graph, timestamps, exclude_nans=False
+        nwbfile, individual_name, df_animal, scorer, video, paf_graph, timestamps, exclude_nans=False, name=name,
     )
 
 
@@ -133,6 +136,7 @@ class DeepLabCutInterface(BaseTemporalAlignmentInterface):
         self,
         nwbfile: NWBFile,
         metadata: Optional[dict] = None,
+        name: Optional[str] = "PoseEstimation",
     ):
         """
         Conversion from DLC output files to nwb. Derived from dlc2nwb library.
@@ -151,4 +155,5 @@ class DeepLabCutInterface(BaseTemporalAlignmentInterface):
             individual_name=self.subject_name,
             config_file=str(self.source_data["config_file_path"]),
             timestamps=self._timestamps,
+            name=name,
         )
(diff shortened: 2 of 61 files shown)


src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -15,6 +15,7 @@ def write_subject_to_nwb(
     individual_name: str,
     config_file: FilePathType,
     timestamps: Optional[Union[List, np.ndarray]] = None,
+    name: Optional[str] = "PoseEstimation",
```
CodyCBakerPhD, line 18:
```suggestion
    container_name: Optional[str] = "PoseEstimation",
```

Would probably want to be a little more explicit since this technically makes a container of many neurodata objects

vigji:
Reasonable suggestion! Let me know if there is anything else I should do

CodyCBakerPhD:
Last thing to do would be to add a new test case that uses this new feature after the current test case: https://github.com/catalystneuro/neuroconv/blob/main/tests/test_on_data/test_behavior_interfaces.py#L317

Which would also require setting developer dependencies (in the form of `git+https://github.com/vigji/dlc2nwb.git@main`) to install the version of the DLC tools that allow for it so we can make sure both packages work well together

vigji:
> Last thing to do would be to add a new test case that uses this new feature after the current test case: https://github.com/catalystneuro/neuroconv/blob/main/tests/test_on_data/test_behavior_interfaces.py#L317

Hope I did this correctly

> Which would also require setting developer dependencies (in the form of `git+https://github.com/vigji/dlc2nwb.git@main` to https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/behavior/deeplabcut/requirements.txt) to install the version of the DLC tools that allow for it so we can make sure both packages work well together

I changed this in `requirements-testing.txt`, is this correct?

requirements-testing.txt
CodyCBakerPhD, line 1: > > Which would also require setting developer dependencies (in the form of git+https://github.com/vigji/dlc2nwb.git@main to https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/behavior/deeplabcut/requirements.txt) to install the version of the DLC tools that allow for it so we can make sure both packages work well together
>
> I changed this in requirements-testing.txt, is this correct?

The `datainterfaces/behavior/deeplabcut/requirements.txt` file, right next to the interface

CodyCBakerPhD:
Test looks good! Will need to wait a bit to merge upstream before getting this through

Thanks for all the work

requirements-testing.txt
vigji: ok done now

vigji:
Perfect, let me know if there's anything more I need to take care of, and thank you :)

vigji:
After having tested this in parallel with the `VideoInterface` for the conversion project, I have to say I am slightly annoyed by the different approaches to solve the same issue: an argument to the object instantiation there, an optional parameter to the conversion function call here. Should I change this to be consistent with `VideoInterface`? What would be your approach? Am I missing something?

CodyCBakerPhD:
The usual excuse for that is that various decisions were made for the sake of maintaining back-compatibility; the video interfaces is one of our oldest and most used interfaces after all (since it is so generic) - so if we were to change something fundamental about it, it might cause problems for a number of conversion pipelines

That said, we do try to standardize these things with slow deprecation cycles every now and again. Enhancing the video interfaces is something we've wanted to do for a while now but no one has had the time to implement all improvements

vigji:
Perfectly reasonable! So the strategy used here for the `DeeplacutInterface` is the way to go, correct?

CodyCBakerPhD:
Yes, something like a container name seems to me something more relevant as a conversion option (this is also the case in ophys/ecephys modules here)

Just waiting on the upstream repo to respond to PR and we can get this through. I'll inject myself there soon otherwise just to get this done

requirements-testing.txt
```diff
@@ -6,3 +6,4 @@ ndx-miniscope
 spikeinterface[qualitymetrics]>=0.100.0
 zarr<2.18.0  # Error with Blosc (read-only during decode) in numcodecs on May 7; check later if resolved
 pytest-xdist
+dlc2nwb @ git+https://github.com/vigji/dlc2nwb.git@main
```
CodyCBakerPhD, line 9:
```suggestion
```

No longer necessary with the rebase

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -151,4 +163,5 @@ def add_to_nwbfile(
             individual_name=self.subject_name,
             config_file=str(self.source_data["config_file_path"]),
             timestamps=self._timestamps,
+            container_name=container_name,
```
CodyCBakerPhD, line 166:
```suggestion
            pose_estimation_container_kwargs=dict(name=container_name),
```

I exposed it in a bit more of a controlled way when I ported the code

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -133,6 +144,7 @@ def add_to_nwbfile(
         self,
         nwbfile: NWBFile,
         metadata: Optional[dict] = None,
+        container_name: Optional[str] = "PoseEstimation",
```
CodyCBakerPhD, line 147:
```suggestion
        container_name: str = "PoseEstimation",
```

I think we can drop the `None` case here to be explicit about the value of the default string

src/neuroconv/datainterfaces/behavior/deeplabcut/requirements.txt
```diff
@@ -1,4 +1,4 @@
-dlc2nwb>=0.3
+dlc2nwb @ git+https://github.com/vigji/dlc2nwb.git@main
```
CodyCBakerPhD, line 1:
```suggestion
```

CodyCBakerPhD:
@vigji Sorry this took so long; we're just taking matters into our own hands at this point

There should be a box somewhere on this PR about giving reviewers permission to push code or something like that - if you check that box I'd be happy to make the required changes to get this through

vigji:
Great! thanks! Could not find such button but I should have applied all changes.

CodyCBakerPhD:
@vigji There is a remaining merge conflict which must be resolved via the CLI (be sure to try to merge `port_dlc2nwb` on our end)

CodyCBakerPhD (approved):

CodyCBakerPhD:
Thanks again @vigji !

PR #922, opened 2024-06-22
Fix SpikeGLXConverter conversion option schema

### LINKED ISSUE #919: [Bug]: SpikeGLXConverterPipe cannot generate stub preview NWB files
### What happened?

In NWB GUIDE, generating a preview file with a `SpikeGLXConverterPipe` converts all the data to NWB instead of a stub. It looks like `stub_test` is not being set / passed in `SpikeGLXConverterPipe`. The preview NWB file is the same size as the full conversion NWB file, and the preview takes forever to run. 

### Steps to Reproduce

```python
Use GUIDE with test data using single SpikeGLXConverterPipe. Neurosift shows ElectricalSeriesAP has shape (90000, 384) which is the same as the final data size.
```

### Traceback

_No response_

### Operating System

macOS

### Python Executable

Python

### Python Version

3.12

### Package Versions

_No response_

### Code of Conduct

CodyCBakerPhD:
fix #919 

cc @rly If you dev install with this does it help?

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
+++ b/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
@@ -79,11 +79,13 @@ class SpikeGLXConverterPipe(ConverterPipe):
             if "nidq" in stream:
                 file_path = folder_path / f"{folder_path.stem}_t0.nidq.bin"
                 interface = SpikeGLXNIDQInterface(file_path=file_path)
-            data_interfaces.update({stream: interface})
+            data_interfaces.update({str(stream): interface})  # Without str() casting, is a numpy string
 
         super().__init__(data_interfaces=data_interfaces, verbose=verbose)
 
     def get_conversion_options_schema(self) -> dict:
-        return {
-            name: interface.get_conversion_options_schema() for name, interface in self.data_interface_objects.items()
-        }
+        conversion_options_schema = super().get_conversion_options_schema()
+        conversion_options_schema["properties"].update(
+            {name: interface.get_conversion_options_schema() for name, interface in self.data_interface_objects.items()}
+        )
+        return conversion_options_schema
--- a/src/neuroconv/nwbconverter.py
+++ b/src/neuroconv/nwbconverter.py
@@ -1,6 +1,7 @@
+"""Contains core class definitions for the NWBConverter and ConverterPipe."""
+
 import json
 from collections import Counter
-from copy import deepcopy
 from pathlib import Path
 from typing import Dict, List, Literal, Optional, Tuple, Union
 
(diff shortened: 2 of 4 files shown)


rly:
That worked! Thanks for the quick fix.

CodyCBakerPhD:
OK I'll get a minimal test added tomorrow

CodyCBakerPhD:
@h-mayorquin @bendichter This should be good to go

h-mayorquin:
The diff turned out to be confusing because of the movements. I think that a lot of stuff here is only re-ordering of code right? What is the core change?

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
```diff
@@ -79,11 +79,13 @@ def __init__(
...
         super().__init__(data_interfaces=data_interfaces, verbose=verbose)
 
     def get_conversion_options_schema(self) -> dict:
-        return {
-            name: interface.get_conversion_options_schema() for name, interface in self.data_interface_objects.items()
-        }
+        conversion_options_schema = super().get_conversion_options_schema()
+        conversion_options_schema["properties"].update(
+            {name: interface.get_conversion_options_schema() for name, interface in self.data_interface_objects.items()}
+        )
```
CodyCBakerPhD, line 90: @h-mayorquin This is the main diff

src/neuroconv/nwbconverter.py
```diff
@@ -53,26 +54,23 @@ def get_source_schema(cls) -> dict:
...
-                {interface_name: unroot_schema(data_interface.get_conversion_options_schema())}
-            )
-        return conversion_options_schema
-
     @classmethod
     def validate_source(cls, source_data: Dict[str, dict], verbose: bool = True):
         """Validate source_data against Converter source_schema."""
         cls._validate_source_data(source_data=source_data, verbose=verbose)
 
+    def _validate_source_data(self, source_data: Dict[str, dict], verbose: bool = True):
```
CodyCBakerPhD, line 62: @h-mayorquin Just seems like this should be closer to the method above rather than distant from it

src/neuroconv/nwbconverter.py
```diff
@@ -53,26 +54,23 @@ def get_source_schema(cls) -> dict:
             source_schema["properties"].update({interface_name: unroot_schema(data_interface.get_source_schema())})
         return source_schema
 
-    def get_conversion_options_schema(self) -> dict:
```
CodyCBakerPhD, line 56: @h-mayorquin This was inconvenient to locate since it was defined after class methods, before `__init__` and isn't something that gets called until before `add_to_nwbfile`

I think it was left here because it maybe used to have been a class method

src/neuroconv/nwbconverter.py
```diff
@@ -257,28 +258,15 @@ def get_default_backend_configuration(
 
 
 class ConverterPipe(NWBConverter):
-    """Takes a list or dict of pre-initialized interfaces as arguments to build an NWBConverter class"""
-
-    def get_conversion_options_schema(self) -> dict:
```
CodyCBakerPhD, line 262: Same issue as above; odd that this is the first method defined even before any class methods or `__init__`

src/neuroconv/nwbconverter.py
```diff
@@ -257,28 +258,15 @@ def get_default_backend_configuration(
...
-                {interface_name: unroot_schema(data_interface.get_conversion_options_schema())}
-            )
-        return conversion_options_schema
+    """Takes a list or dict of pre-initialized interfaces as arguments to build an NWBConverter class."""
 
-    def get_source_schema(self) -> dict:
-        raise NotImplementedError("Source data not available with previously initialized classes")
+    @classmethod
+    def get_source_schema(cls) -> dict:
+        raise NotImplementedError("Source data not available with previously initialized classes.")
```
CodyCBakerPhD, line 265: Doesn't make much difference, but it's a class method in the parent after all

CodyCBakerPhD:
> The diff turned out to be confusing because of the movements. I think that a lot of stuff here is only re-ordering of code right? What is the core change?

Commented

The code that was moved was unchanged, it just made debugging (finding the actual source of the issue) easier

h-mayorquin:
Thank you, yes, I trust your sense of organization. I just wanted to make my job easier by focusing at the actual diff so I would not get distracted. Thanks for pointing it out explicitly. Taking a look.

CodyCBakerPhD:
See https://github.com/NeurodataWithoutBorders/nwb-guide/pull/870 for source issue on GUIDE (it was a 2-step fix)

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxconverter.py
h-mayorquin: Ok, this makes sense, I am thinking about the structure of the tiff analogous, do you think it would make sense to follow the logic from here:

https://github.com/catalystneuro/neuroconv/blob/1efe156ef005921c623672b8815eabd3021f6bf7/src/neuroconv/datainterfaces/ophys/brukertiff/brukertiffconverter.py#L28-L30

h-mayorquin (approved):
OK, so in general for interfaces and converters `get_conversion_options_schema` should return a dictionary with the schema. But in the case of the `SpikeGLXConverter` it was returning a dictionary with {stream: conversion_schema} which was rendered empty by the unroot routine.

The fix just makes this equal to the other ones, is this correct?

CodyCBakerPhD:
> Ok, this makes sense, I am thinking about the structure of the tiff analogous, do you think it would make sense to follow the logic from here:

I didn't check others; the one you point to might work, not sure - it's also a question if you want to expose the options per interface of a subconverter or apply a blanket statement

PR #926, opened 2024-06-24
Fix conditional existence of xml path

CodyCBakerPhD:
Should fix https://github.com/NeurodataWithoutBorders/nwb-guide/issues/881

I'll add a test case tomorrow


CodyCBakerPhD:
@bendichter OK should be fixed now

bendichter (approved):

PR #914, opened 2024-06-17
multi videos

luiztauffer:
Fix #905

PR #928, opened 2024-06-25
Docstring  improvements

h-mayorquin:
Docstring improvements to #914.

Is this how this should work?

The thing that I am confused about is, when will add multiple videos in one interface vs separate videos.


CodyCBakerPhD:
> The thing that I am confused about is, when will add multiple videos in one interface vs separate videos.

Multiple videos together comprising multiple segments of a single data stream from a single camera -> multiple files to a single interface -> a single neurodata object created with those links

Different cameras, therefore different data streams -> different interface -> different neurodata object

luiztauffer:
@h-mayorquin looks good to me! merging

PR #934, opened 2024-07-02
Propagate annotations to `IntanRecordingInterface`

h-mayorquin:
Also included in #933 but probably easier to make this quick improvement here.

This  propagates the annotations of intan that were made available on https://github.com/NeuralEnsemble/python-neo/pull/1477


CodyCBakerPhD (approved):

CodyCBakerPhD:
So the default for this is normally `False`?

Would it be worth exposing to our `__init__` at any point?

h-mayorquin:
Yes, the default on spikeinterface is False. At some point Ben asked Julia to extract as much annotations as possible from neo which she did, but I think Sam did not want more metadata by default on spikeinterface so it is False on spikeinterface.

Now, all of those properties are loaded by default as properties which then our spikeinterface functions send to the electrodes table.

I don't see any scenario where we would want to limit the amount of metadata that ends in the nwbfile and the functional approach to interfaces with a pipe allows for removing properties from the extractor if needed for a specific case.

So my intuition is that we should not expose it.

h-mayorquin:
Thanks for the quick review @CodyCBakerPhD

PR #945, opened 2024-07-10
Test ruff for import removal

CodyCBakerPhD:
Curious what this will do here

Looking for a way to automatically remove unused imports since isort doesn't cover that

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,6 +41,8 @@ exclude = [
 select = ["F401", "I"]  # TODO: eventually, expand to other 'F' linting
 fixable = ["ALL"]
 
+[tool.ruff.lint.isort]
+default-section = "first-party"
 
 
 [tool.codespell]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,8 @@ select = ["F401", "I"]  # TODO: eventually, expand to other 'F' linting
 fixable = ["ALL"]
 
 [tool.ruff.lint.isort]
-default-section = "first-party"
+relative-imports-order = "closest-to-furthest"
+known-first-party = ["neuroconv"]
 
 
 [tool.codespell]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,10 @@ exclude = [
   "*/__init__.py"
 ]
 
+[tool.ruff.lint]
+select = ["F401", "I"]  # TODO: eventually, expand to other 'F' linting
+fixable = ["ALL"]
+
 
 
 [tool.codespell]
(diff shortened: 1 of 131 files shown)


tests/test_minimal/test_tools/globus_transfer_tools.py
```diff
@@ -15,14 +16,9 @@
     transfer_globus_content,
 )
 
-try:
```
CodyCBakerPhD, line 18: This was the only thing that could not be fixed automatically; but it's a good catch

CodyCBakerPhD:
This looks pretty good and sets us up for future improvements (though many are not automatically fixable though still good things to inform developers about)

`ruff` extends `isort`, so we can do a complete replacement here

tests/test_minimal/test_tools/globus_transfer_tools.py
h-mayorquin: This is better for the debugger as well. When I debug I need to skip over those assertions so the change here is better.

h-mayorquin:
I think ruff also extends black, right?

CodyCBakerPhD:
Not a pure extension, no, but rather is tightly compatible (as long as certain `.toml` options are not at odds)

`ruff` without tight selection like I have here is more like a classic `flake8` linter -only certain things can it fix automatically in a safe manner

h-mayorquin:
SHould this have a changelog? (not sure)

CodyCBakerPhD:
> Should this have a changelog? (not sure)

I opted for no, since nothing (aside from very minor globus test) in source actually changed

pyproject.toml
```diff
@@ -19,19 +30,22 @@ extend-exclude = '''
...
-minversion = "6.0"
-addopts = "-ra --doctest-glob='*.rst'"
-testpaths = [
-"docs/conversion_examples_gallery/",
-"tests"
+
+
+[tool.ruff]
+exclude = [
+  "*/__init__.py"
```
h-mayorquin, line 37: how comes we exclude these ones?

h-mayorquin (approved):
I think this is good. So now this also removes unused imports in pre-commit, right?

I am unsure about which rule is removing the f in the f-strings. Do you know that one?

pyproject.toml
CodyCBakerPhD: Beacuse we don't fully use sklearn convention with exposed imports and that flags unfixable F401's since the things we import in those files aren't in `__all__` for that submodule

CodyCBakerPhD:
> I am unsure about which rule is removing the f in the f-strings. Do you know that one?

Ah, that was a push from a previous commit on this PR that enabled all auto-fixes, not just F401 (the imports)

I think it is also a good one, but will leave it's re-activation for another time

h-mayorquin:
Thanks for your explanations.

Issue #936, opened 2024-07-03
Schema bug when using write_sorting directly


h-mayorquin:
I  don't have time to dig right now but I wanted to report it:

```python
from pathlib import Path
from spikeinterface.core import NumpySorting
from neuroconv.tools.spikeinterface import write_sorting
import tempfile
import numpy as np

spikes_dict1 =  {"A": np.asarray([0, 1, 2]), "B": np.asarray([3, 4, 5])}
with tempfile.TemporaryDirectory() as tmpdir:
    temporary_folder = Path(tmpdir)
    nwbfile_path = temporary_folder / "a_chunking_error.nwb"

    sorting1 = NumpySorting.from_unit_dict(spikes_dict1, sampling_frequency=30_000.0)

    write_sorting(sorting=sorting1, nwbfile_path=nwbfile_path)

{
	"name": "ValidationError",
	"message": "None is not of type 'object'

Failed validating 'type' in schema:
    {'$id': 'base_metafile.schema.json',
     '$schema': 'http://json-schema.org/draft-07/schema#',
     'description': 'Base schema for the metafile',
     'properties': {'NWBFile': {'additionalProperties': False,
                                'properties': {'data_collection': {'description': 'Notes '
                                                                                  'about '
                                                                                  'data '
                                                                                  'collection '
                                                                                  'and '
                                                                                  'analysis.',
                                                                   'type': 'string'},
                                               'experiment_description': {'description': 'general '
                                                                                         'description '
                                                                                         'of '
                                                                                         'the '
                                                                                         'experiment',
                                                                          'type': 'string'},
                                               'experimenter': {'description': 'Name '
                                                                               'of '
                                                                               'person/people '
                                                                               'who '
                                                                               'performed '
                                                                               'experiment',
                                                                'items': {'title': 'experimenter',
                                                                          'type': 'string'},
                                                                'type': 'array'},
                                               'identifier': {'description': 'A '
                                                                             'unique '
                                                                             'text '
                                                                             'identifier '
                                                                             'for '
                                                                             'the '
                                                                             'file. '
                                                                             'If '
                                                                             'one '
                                                                             'is '
                                                                             'not '
                                                                             'provided '
                                                                             'it '
                                                                             'will '
                                                                             'be '
                                                                             'randomly '
                                                                             'assigned',
                                                              'type': 'string'},
                                               'institution': {'description': 'Institution(s) '
                                                                              'where '
                                                                              'experiment '
                                                                              'is '
[...]

PR #948, opened 2024-07-11
context `make_or_load_nwbfile` to raise an error when nwbfile needs to be created but metadata is None

h-mayorquin:
Related to the issue #936

### DIFF
--- a/tests/test_minimal/test_tools/test_context_tools.py
+++ b/tests/test_minimal/test_tools/test_context_tools.py
@@ -228,4 +228,4 @@ def test_raise_error_when_metadata_is_empty_and_creation_is_needed(tmpdir):
 
     with pytest.raises(ValueError):
         with make_or_load_nwbfile(nwbfile_path=nwbfile_path, metadata=None, overwrite=True) as nwbfile:
-            nwbfile
+            pass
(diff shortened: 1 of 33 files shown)


tests/test_minimal/test_tools/test_context_tools.py
```diff
@@ -220,3 +221,11 @@ def test_make_or_load_nwbfile_on_corrupt_file(tmpdir: Path) -> None:
...
         time_series = mock_TimeSeries()
         nwbfile.add_acquisition(time_series)
+
+
+def test_raise_error_when_metadata_is_empty_and_creation_is_needed(tmpdir):
+    nwbfile_path = tmpdir / "test_make_or_load_nwbfile_empty_metadata.nwb"
+
+    with pytest.raises(ValueError):
+        with make_or_load_nwbfile(nwbfile_path=nwbfile_path, metadata=None, overwrite=True) as nwbfile:
+            nwbfile
```
CodyCBakerPhD, line 231:
```suggestion
            pass
```

I think you can just do this?

h-mayorquin: Yes both work, do you prefer the latter?

h-mayorquin:
Dandi seems to be asking us to upgrade?

tests/test_minimal/test_tools/test_context_tools.py
CodyCBakerPhD: I prefer the latter unless you want to do something with the object (print, assert something, etc)

CodyCBakerPhD:
> Dandi seems to be asking us to upgrade?

They did a release today and it might have put their side out of sync. We'll keep an eye on it but it shouldn't have anything to do with our side of things

tests/test_minimal/test_tools/test_context_tools.py
h-mayorquin: OK, let's change it. I am adding a changelog and changing to pass.

h-mayorquin: Done.

CodyCBakerPhD (approved):

PR #949, opened 2024-07-11
Add `validate_metadata` to interface.run_conversion

h-mayorquin:
Making some progress on https://github.com/catalystneuro/neuroconv/issues/42


CodyCBakerPhD:
Ahh, has been a long time coming

This one is scary though so thanks for taking charge handling it

src/neuroconv/nwbconverter.py
```diff
@@ -101,13 +101,20 @@ def get_metadata(self) -> DeepDict:
...
         # The encoder produces a serialized object, so we deserialized it for comparison
         serialized_metadata = encoder.encode(metadata)
         decoded_metadata = json.loads(serialized_metadata)
-        validate(instance=decoded_metadata, schema=self.get_metadata_schema())
+
+        metadata_schema = self.get_metadata_schema()
+        if append_mode:
+            # Eliminate required from NWBFile
+            nwbfile_schema = metadata_schema["properties"]["NWBFile"]
+            nwbfile_schema.pop("required", None)
```
CodyCBakerPhD, line 115: However, in truth, the entirety of `NWBFile` level metadata will be ignored if specified in this case (also `Subject`)

So any reason not to pop the entirety of the subschema for those?

h-mayorquin: By ignore you mean that it won't be used? Those things are only used when creating the NWBFile, right? 

I think it is OK that if those values are passed in the metadata the check is performed. I see no lost from it save some small performance thing. I also feel that if I drop that version of the schema I might have to do the same operation in the metadata which seems dirtier.

CodyCBakerPhD (approved):

src/neuroconv/nwbconverter.py
```diff
@@ -101,13 +101,20 @@ def get_metadata(self) -> DeepDict:
             metadata = dict_deep_update(metadata, interface_metadata)
         return metadata
 
-    def validate_metadata(self, metadata: Dict[str, dict]):
+    def validate_metadata(self, metadata: Dict[str, dict], append_mode: bool = False):
```
CodyCBakerPhD, line 104: Just something to keep in mind here

The argument on `run_conversion` is `overwrite` and it defaults to `False` (append being enabled by default if file exists)

I know you detect the file existence and whatnot below, but there's some slight API inconsistency here to think about

h-mayorquin: Good point.

PR #950, opened 2024-07-11
Deprecate `converter.run_conversion` without nwbfile_path 

h-mayorquin:
Following the unification in #42 and the precedent in the run conversion for data interfaces:

https://github.com/catalystneuro/neuroconv/blob/f87955aa61f16f8531da97eda3e3b2afc1e98a66/src/neuroconv/basedatainterface.py#L155-L160

### DIFF
--- a/src/neuroconv/nwbconverter.py
+++ b/src/neuroconv/nwbconverter.py
@@ -1,6 +1,7 @@
 """Contains core class definitions for the NWBConverter and ConverterPipe."""
 
 import json
+import warnings
 from collections import Counter
 from pathlib import Path
 from typing import Dict, List, Literal, Optional, Tuple, Union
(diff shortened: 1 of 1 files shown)


src/neuroconv/nwbconverter.py
```diff
@@ -203,6 +204,14 @@ def run_conversion(
             Similar to source_data, a dictionary containing keywords for each interface for which non-default
             conversion specification is requested.
         """
+
+        if nwbfile_path is None:
```
CodyCBakerPhD, line 208: And so too should remove the default in the signature?

h-mayorquin: Shouldn't we do that once the deadline is met?

CodyCBakerPhD: Oh, I see, we never actually warned about this level only the interface level

Nevermind

CodyCBakerPhD (approved):

PR #951, opened 2024-07-11
Remove `interface.run_conversion` without nwbfile_path

h-mayorquin:
The deadline for the deprecation has expired.

### DIFF
--- a/src/neuroconv/basedatainterface.py
+++ b/src/neuroconv/basedatainterface.py
@@ -115,7 +115,7 @@ class BaseDataInterface(ABC):
 
     def run_conversion(
         self,
-        nwbfile_path: Union[str, Path],
+        nwbfile_path: str,
         nwbfile: Optional[NWBFile] = None,
         metadata: Optional[dict] = None,
         overwrite: bool = False,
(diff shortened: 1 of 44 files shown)


src/neuroconv/basedatainterface.py
```diff
@@ -116,7 +115,7 @@ def add_to_nwbfile(self, nwbfile: NWBFile, **conversion_options) -> None:
 
     def run_conversion(
         self,
-        nwbfile_path: Optional[str] = None,
+        nwbfile_path: Union[str, Path],
```
CodyCBakerPhD, line 118: Do we actually support `Path` here generally yet? IDK if PyNWB/HDMF/HDMF-Zarr want `str` only at IO time

h-mayorquin: I have always assumed we do. PYNWB does not but I am less sure about zarr. Let's revert to restrictive then just out of caution.

h-mayorquin: Done.

CodyCBakerPhD:
Looks like a few straggler tests that still used the deprecated approach😆

CodyCBakerPhD (approved):

PR #955, opened 2024-07-11
Move towards pyproject.toml standard

h-mayorquin:
I am packacking the dicarlo conversion and this is a good time to address this long standing issue and getting more familiar with the packaging and dependency managment process.

This is the first attempt at moving our static configuration to pyproject. The dynamic configuration is the spiky part and I expect this to fail until I tune that:

https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

Let's see


CodyCBakerPhD:
Thanks for taking this on

Good luck!

I'd also be willing to reopen the old discussion about whether or not we should only have local `requirements.txt` files per data format, that magically get parsed during a very programmatic setup call, instead of having one central list of all optional dependencies

The benefit of a single list is (i) it's easier to check when cutting a release if we have any dev branch pins, (ii) it's easier to spot conflicts in version ranges across the repo

Granted though, I wonder, is there a way in `pyproject.toml` to also declare `full` (or other keys) to be set-based intersection of other optional dependencies? Or must it all be fully explicit?

h-mayorquin:
> Good luck!

x D

> I'd also be willing to reopen the old discussion about whether or not we should only have local requirements.txt files per data format, that magically get parsed during a very programmatic setup call, instead of having one central list of all optional dependencies

> The benefit of a single list is (i) it's easier to check when cutting a release if we have any dev branch pins, (ii) it's easier to spot conflicts in version ranges across the repo

That was my preference then as is my preference now. It will be a lot of work to change though. Back then also Ben pointed out that this was better to achieve a pluggin architecture. Not sure I understood completly but I thought that it was about only doing local changes.

> Granted though, I wonder, is there a way in pyproject.toml to also declare full (or other keys) to be set-based intersection of other optional dependencies? Or must it all be fully explicit?

You can do stuff like this but I am not sure it works as intended.
```
dev = [
    "spikeinterface[test]",
    "spikeinterface[test_core]",
    "spikeinterface[docs]",
    "black",
    "pre-commit",
]
```

h-mayorquin:
Relevant:

https://setuptools.pypa.io/en/latest/build_meta.html#backend-wrapper

h-mayorquin:
OK, the installation part with pip locally works quite well. We have to ensure that the build process for sdist and wheels works.

h-mayorquin:
This should be ready @CodyCBakerPhD

setup.py
```diff
@@ -4,93 +4,52 @@
...
-extras_require = defaultdict(list)
+def read_requirements(file):
+    with open(root / file) as f:
+        return f.readlines()
+
 
-extras_require["dandi"].append("dandi>=0.58.1")
-extras_require["full"].extend(extras_require["dandi"])
+extras_require = defaultdict(list)
+extras_require["full"] = ["dandi>=0.58.1", "hdf5plugin"]
```
CodyCBakerPhD, line 18: Hmm so we have to duplicate these extras here and in the `.toml`?

CodyCBakerPhD:
Interesting, so you're saying we can mix custom Python logic for finding extra dependencies in `setup.py` as well as explicitly written extras in the `.toml`?

setup.py
h-mayorquin: Yes, for the moment. I think there is a way of reducing this but this is a big change already and I would like to do it step-wise.

h-mayorquin:
> Interesting, so you're saying we can mix custom Python logic for finding extra dependencies in `setup.py` as well as explicitly written extras in the `.toml`?

I am kind of suprised it works tbh. I was expecting that defining some optional dependencies in the pyproject would be exclusive or with extras_require with the setup.py but at last this does not seem to be the case.

PR #962, opened 2024-07-15
Testing alternative rclone transfer strategy

CodyCBakerPhD:
Testing alternative rclone strategy

Just curious if I can get this to work directly instead of building the custom image (which is currently working)


CodyCBakerPhD:
Yeah, not working but with no obvious failure output. Will proceed with what works for now but will keep this in mind

PR #964, opened 2024-07-15
Stop casting ints to floats

### LINKED ISSUE #960: recasting unit properties as floats
I am trying to add integer-valued properties to a sortingextractor and convert to NWB, but they are converted to float in the process. I'd really rather write them as integers. It looks like this line is responsible:

https://github.com/catalystneuro/neuroconv/blame/5986a72a2d3fa5a0cfc20b4ad5fa6d3954d096a9/src/neuroconv/tools/spikeinterface/spikeinterface.py#L1169-L1170

I don't see why we would want to automatically convert all integers to floats. @h-mayorquin it looks like you wrote these lines, do you remember why we did this? It's from 2 years ago so fine if not, but I would rather us be able to maintain integer data types where possible.

I am pretty sure I can find another way around this, so it's not blocking me atm, but I think it should be addressed as it can lead to sub-optimal NWB files

bendichter:
fix #960


tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -1020,6 +1020,16 @@ def test_property_addition(self):
         expected_properties_in_units_table = ["", "", "added_value", "added_value", "added_value", "added_value"]
         self.assertListEqual(properties_in_units_table, expected_properties_in_units_table)
 
+    def test_preserve_int_property(self):
+        """Add a property only available in a second sorting."""
```
bendichter, line 1024:
```suggestion
```

h-mayorquin:
This breaks backwards compatbility at the level of `add_sorting` and `add_units_table` . The current works in main but does not work in your branch:

```python
from spikeinterface.core import generate_sorting


nwbfile = mock_NWBFile()

sorting1 = generate_sorting(num_units=4)
sorting1 = sorting1.rename_units(new_unit_ids=["a", "b", "c", "d"])

sorting2 = generate_sorting(num_units=4)
sorting2 = sorting2.rename_units(new_unit_ids=["c", "d", "e", "f"])
sorting2.set_property(key="an_integer_property", values=[1, 2, 3, 4])

from neuroconv.tools.spikeinterface import add_sorting 

add_sorting(sorting=sorting1, nwbfile=nwbfile)
add_sorting(sorting=sorting2, nwbfile=nwbfile)
```

I am kind of surprise that we don't have a test for this.

h-mayorquin:
More comments:
* The keyword should be implemented at the level of `add_sorting` and `add_units_table` so the same workflow can be achieved for those functions. 

* About the interfaceI also think that the keyword should be something like `make_table_expandable` because this is really what you sacrifce when you turn off this setting.  Then, a warning can be thrown to the user when casting indicating that writing integers values is only achievable when writing non expandable tables and that they can set this of by `make_table_expandable=False`. There might be other limitations of making table expandable and it would be nice to switch all of them with only one keyword instead of having to add an extra keyword argument per situation. Plus, if at some point in the future we change hdmf to add [missing integer values like pandas](https://pandas.pydata.org/pandas-docs/stable/user_guide/gotchas.html#missing-value-representation-for-numpy-types) then `cast_int_to_float` stops making sense. A user friendly error should be produce in the opposite situation that someone did not cast and then they try to expand. Currently, `ValueError: cannot convert float NaN to integer` is obscure and not very useful.

bendichter:
@h-mayorquin thanks for looking into this.

As I asked [here](https://github.com/catalystneuro/neuroconv/issues/960#issuecomment-2227497855), how often does this use-case really come up where we'd have units that have different properties in the same Unis table?

What if we catch this edge case, and recast to floats only when we need to and throw a warning indicating that we are doing this?

bendichter:
I think a more natural use-case for expanding a table would be to add rows that have columns that are consistent with the existing table. That is doable without casting to floats, so I `make_table_expandable` would be a bit misleading. In the majority of cases the table is already expandable. I guess I would change my mind if you could provide a common use-case that would require this, but I can't think of a single one where I would prefer stacking units with different columns rather than creating multiple Units tables.

h-mayorquin:
The example I mentioned before was working on consensus algorithms: adding sorting results from different sorters with different properties (one of which is type int).  

> I think a more natural use-case for expanding a table would be to add rows that have columns that are consistent with the existing table. That is doable without casting to floats, so I think make_table_expandable would be a bit misleading

Yes, I thought about it: the table is expandable in many cases that don't require the casting. What do you think of `ensure_expandable_table`? 

One thing that is not clear how to handle is compatbility across calls of `add_units_table`.  Making this behavior a parameter creates an implicit coupling between the first call of `add_units_table` and any subsequent use which is a bad practice. Probably, we can check the type of the previous entries for the same property and just use that type. So, the parameter is really only respected on the first call to `add_units_table`. 

A consideration that just made me change my mind to support the default that you want is that the current approach is loosy for int64. int32 can be safely cast to float 64 (the mantissa of the later has enough bytes) by doubling the memory cost but if the type is int64 we will be [loosing information in most systems](https://numpy.org/doc/stable/user/basics.types.html#extended-precision) (asuming hdmf has the types of numpy). Anyway, we don't have the logic to handle any of those considerations so we are currently doing loosy conversions for int64. In other words the current behavior has a bug.

I really wish that we had a nullable value for int types but is unlikely that hdf would support that

h-mayorquin:
Yes, as I wrote above I changed my mind about the default. Given that we can guarantee lossless casting for in64 it is probably better to re-think the whole thing. So my current position is:

Let's not cast by default and produce error friendly messages when the need to have nullable values of int arises which is:
* When a sorting or recording is added to an already existing table and there are missing properties.
* When a sorting or recording is added for the first time but they have missing properties.

I probably think that focusing on the missing values and what to do with them boht as a use case an in the interface (the name of the keywords) is a more fruitful approach. **The edge case is the need to have nullable ints** and I think is easier to think around that.

PR #965, opened 2024-07-17
GuppyInterface

pauladkisson:
This PR adds 

1. A DataInterface for [GuPPy](https://github.com/LernerLab/GuPPy) output (processed fiber photometry analysis), which depends on [ndx-guppy](https://github.com/catalystneuro/ndx-guppy)
2. A data interface for behavioral events from the TDT system, which depends on ndx-events version 0.2.2 --> now peeled off to #1751
3. A converter that combines the TDT fiber photometry interface, its new TDT event interface, and the new GuPPy data interface. 

Adding support for other data acquisition formats like raw CSV, Doric, and Neurophotometrics will be left to future work. 

See https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/46 for updated testing data.

For now, the pyproject.toml dependency chain pins the GitHub version of ndx-guppy since it's not released on PyPI yet. Once this PR gets reviewed, I will cut a release of NDX Guppy and update the dependencies accordingly. I just wanted to hold off on that step since I anticipate that review might surface some issues with ndx-guppy itself that we will want to propagate upstream.


h-mayorquin:
@pauladkisson quick question: could we peel off `TDTEventsInterface` into its own PR? Looks like it's the only piece here that stands on its own (no dependency on the fiber photometry table or the converter), so it could land independently of the GuPPy/ndx-guppy work. What do you think?

pauladkisson:
> @pauladkisson quick question: could we peel off `TDTEventsInterface` into its own PR? Looks like it's the only piece here that stands on its own (no dependency on the fiber photometry table or the converter), so it could land independently of the GuPPy/ndx-guppy work. What do you think?

Sure, no problem. See #1751

pauladkisson:
This PR has been updated to reference the new TDT events structure and is now ready for review. @h-mayorquin 

Failing tests should be fixed by #1769 I think.

pauladkisson:
@h-mayorquin I switched to using mock guppy data, so https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/46 should be obsolete

pauladkisson:
Depends on hdmf-dev/hdmf#1533.

The `GuppyEventsTable.events` object reference to a `pynwb.event.EventsTable` can't be written directly on current hdmf (hdmf-dev/hdmf#1532: a reference column whose targets are `DynamicTable`s gets a spurious rank-3 `maxshape` and fails the write). It's shimmed here with `H5DataIO(maxshape=(None,))`, marked `TODO(hdmf#1532)` at each site. Once #1533 merges and lands in an hdmf release, the shim should be removed.

pauladkisson:
Guppy converter split off to a separate PR as requested. #1783

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
```diff
@@ -0,0 +1,1380 @@
...
+            artifact_removal_method = guppy_parameters.get("artifactsRemovalMethod")
+            if artifact_removal_method is None:
+                warnings.warn(
+                    "GuPPy parameters do not specify 'artifactsRemovalMethod' but artifact removal "
+                    "intervals were found; defaulting to 'concatenate' (GuPPy's UI default).",
+                    UserWarning,
+                )
+                artifact_removal_method = "concatenate"
+
+        self.folder_path = folder_path
```
h-mayorquin: I think most of this should be private arguments. So we keep the public facing side of the inteface to a minimum.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
```diff
@@ -0,0 +1,1380 @@
...
+_PREFIX_TO_TRACE_TYPE = dict(cntrl_sig_fit="control_fit", dff="dff", z_score="z_score")
+# Per-window peak/area metric row prefixes in the peak_AUC_*.h5 DataFrame index.
+_BIN_COLUMN_PATTERN = re.compile(r"bin_\((\d+(?:\.\d+)?)-(\d+(?:\.\d+)?)\)$")
+
+
+class GuppyInterface(BaseTemporalAlignmentInterface):
+    """
+    Data Interface for converting GuPPy (Guided Photometry Analysis in Python) processed outputs.
+
+    GuPPy is a processing tool, not an acquisition system. This interface writes the derived products
```
h-mayorquin: I think a bullet point list would be a better format to describe what the interface adds.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
```diff
@@ -0,0 +1,1380 @@
...
+            PeakAUCs=peak_aucs_metadata,
+        )
+        return metadata
+
+    def get_metadata_schema(self) -> dict:
+        """Return the metadata schema for this interface."""
+        trace_type_schema = dict(type="string", enum=list(self._TRANSIENT_FEATURES))
+        metadata_schema = super().get_metadata_schema()
+        metadata_schema["properties"].setdefault("Ophys", get_base_schema(tag="Ophys"))
+        metadata_schema["properties"]["Ophys"]["properties"]["Guppy"] = dict(
```
h-mayorquin: shouldn't this be nested under "FiberPhotometry: instead of "Ophys"?

pauladkisson: Yes, thank you, done.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
pauladkisson: Done.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
pauladkisson: Done.

pauladkisson:
Tests are failing because ndx-guppy is not yet on PyPI.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
```diff
@@ -0,0 +1,1385 @@
...
+            "through a converter, e.g. TDTFiberPhotometryGuppyConverter)."
+        )
+        return fiber_photometry_lab_meta_data.fiber_photometry_table
+
+    def add_to_nwbfile(
+        self,
+        nwbfile: NWBFile,
+        metadata: dict,
+        *,
+        fiber_photometry_table_region_indices: dict[str, list[int]],
```
h-mayorquin: I find the name `fiber_photometry_table_region_indices` confusing. it reads like a flat list of row indices, but it is actually a mapping from region label to that region's rows in the `FiberPhotometryTable` (thanks for adding the typing). A name that surfaces the dict, something like `region_to_fiber_photometry_table_rows`, would remove the first-read confusion.

h-mayorquin: Stepping back on why this argument exists at all: the derived traces (and the `GuppyRegionsTable`) link into the acquisition `FiberPhotometryTable` by integer row index, but `GuppyInterface` neither builds that table nor knows its row order, since that is an acquisition-side fact. So the converter has to compute this map and inject it, and that injection is also what forces the interface to be non-standalone. We could avoid the argument entirely if the `FiberPhotometryTable` carried the source stream identifier on each row: with a `stream_name` column `GuppyInterface` matches each region's signal and control stores (the `store_id`s from `storesList.csv`) against the table itself and derives the region-to-row map, instead of having it handed in:

```python
stream_name_to_row = {row.stream_name: index for index, row in enumerate(fiber_photometry_table)}
region_to_indices = {
    region: sorted(stream_name_to_row[store_id] for store_id in stores.values())  # signal + control
    for region, stores in self.region_to_store_names.items()
}
```

This is worth doing on its own merits, the same way the electrodes table carries `channel_name`. Today the `FiberPhotometryTable` has no way to record which acquisition stream a row came from, so two rows with the same location, wavelengths, and devices are indistinguishable except by order. A `stream_name` column lets not only GuPPy but any other aggregator disambiguate rows and tell which acquisition stream each one is.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
```diff
@@ -0,0 +1,1385 @@
...
+
+    def set_aligned_starting_time(self, aligned_starting_time: float) -> None:
+        """Shift every region's timestamps by ``aligned_starting_time``."""
+        region_to_timestamps = self.get_timestamps()
+        self.set_aligned_timestamps(
+            {region: timestamps + aligned_starting_time for region, timestamps in region_to_timestamps.items()}
+        )
+
+    @staticmethod
+    def _get_fiber_photometry_table(nwbfile: NWBFile):
```
h-mayorquin: should not this be a stand-alone method in `tools.fiber_phometry.py`? Are not we using something similar there?

h-mayorquin:
Some feedback about ndx-guppy:

Following up on [this PR comment](https://github.com/catalystneuro/neuroconv/pull/965#discussion_r3580713011) about carrying a `stream_name` identifier on the `FiberPhotometryTable` as a column. I think having the `raw_store_name` on the `GuppyRegionsTable` is also lossy in its current form: a region spans two streams (signal and isosbestic control), so the scalar column keeps the signal store and drops the control, while the `fiber_photometry_table_region` link already reaches both. So the region table would need no store column at all. Maybe we should drop it from the `GuppyRegionsTable`?

Separately, I feel the most common interpretation of `region` is anatomical but this is not the case: labels must be unique in Guppy, so two fibers in the same anatomical structure become two "regions" in the GuPPy sense. Plus, dual-color recording in the same anatomical area splits into two regions as well. It seems to me that the right unit here is a recording site for GuPPy to analyze, so maybe we should improve the docstring to make this more emphatic, or rename as such `region` -> `recording_site` if you'd rather the schema be self-explanatory. I would prefer the second but that would take it away from GuPPy's internal vocabulary, which is undesirable. But if you are open to rename I would also do `GuppyRegionsTable` -> `GuppyRecordingSitesTable` to avoid the confusion completely.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
```diff
@@ -0,0 +1,1385 @@
...
+        event types were merged into one table, that table carries an ``event_type`` discriminator
+        column; the fallback finds the table whose discriminator lists ``event_name``.
+        """
+        if nwbfile.events is None:
+            return None
+        table = nwbfile.events.get(_to_table_object_name(event_name))
+        if table is not None:
+            return table
+        for events_table in nwbfile.events.values():
+            if "event_type" in events_table.colnames and event_name in list(events_table["event_type"][:]):
```
h-mayorquin: `_resolve_events_table` untangles two event layouts: a solo `EventsTable` per type (matched by CamelCased name) and a merged table with an `event_type` discriminator column (matched by scanning that column). Worth flagging what happens in the merged case: it returns the whole merged table, and `GuppyEventsTable.events` then object-references it. So the provenance link degrades to "your event is somewhere in this table" . In short, you cannot get from a product to its actual event rows without filtering `event_type` yourself. In that case the chain `product -> GuppyEventsTable row -> events` breaks down to a coarse table pointer.

Now that events are per-occurrence tables in pynwb 4.0, I think this can be sidestepped by linking the PSTH (and the other event-aligned products) directly to the `EventsTable` occurrence rows. A PSTH trial is one occurrence, so its per-trial `event` reference can be a `DynamicTableRegion` into the specific rows. That is precise in both layouts (it points at the actual rows, not the containing table), and it is a row reference rather than an object reference to a table, so it also avoids the hdmf#1532/#1533 dependency that the current `events` object reference relies on.

h-mayorquin:
A question about how the event-aligned products are structured. Looking at how the PSTHs are built, we are stacking them across events: GuPPy writes one PSTH file per `(event, region, feature)`, but the interface groups by `(region, feature, baseline)` and concatenates across events, so a single `GuppyPSTH` holds several events' trials side by side, distinguished by a per-trial `event` column. (The same grouping happens for `GuppyPeakAUC` and `GuppyCrossCorrelation`.)

What is the intent behind stacking them rather than keeping one object per `(region, event, feature)`, matching GuPPy's files? A per-event object would be homogeneous: its `event` becomes a single reference instead of a per-trial one and quite a lot of machinery for stacking would go away making the interface way simpler. Curios on how you made the decision.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
```diff
@@ -0,0 +1,1385 @@
...
+                return events_table
+        return None
+
+    def _add_events_table(self, ndx_guppy, processing_module, nwbfile: NWBFile):
```
h-mayorquin:
```suggestion
    def _add_guppy_events_table(self, ndx_guppy, processing_module, nwbfile: NWBFile):
```

h-mayorquin: Feel free to ignore but I would prefer to follow the pattern that we use everywhere of naming things like _add_guppy_events_table_to_nwbfile just for the sake of concistency. And along thoes lines, add such methods for the machiinery that is lacking them such as psths, transients, etc. I think the general add_to_nwbfile would become more modular and readable that way.

pauladkisson:
Good comments all around and a lot to work through...

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
```diff
@@ -0,0 +1,1385 @@
...
+                entries=entries,
+                peak_auc_metadata=peak_auc_metadata,
+                region_reference=region_reference,
+                event_reference=event_reference,
+                bin_basis=bin_basis,
+            )
+            processing_module.add(peak_auc)
+
+        # Valid-signal intervals.
+        if self._valid_signal_intervals_by_region:
```
h-mayorquin: Another design question: 

Valid-signal intervals are a per-region property: GuPPy marks artifacts once per region (per signal + control pair) and applies the same windows to that region's signal, its control, and every derived trace computed from them. `GuppyRegionsTable` rows are exactly those entities. NWB core already has a pattern for per-entity valid time windows: the `Units` table stores per-unit [`obs_intervals`](https://pynwb.readthedocs.io/en/stable/pynwb.misc.html#pynwb.misc.Units) (a ragged `[start, stop]` column, with its `obs_intervals_index`) directly on the entity's own row, with the contract that all of that unit's `spike_times` fall inside them. The direct analog here is a ragged `obs_intervals`-style column on `GuppyRegionsTable`, where each region row carries its own valid `[start, stop]` intervals.

The current design instead adds a separate `GuppyValidSignalIntervals` table whose rows carry a `region` `DynamicTableRegion` pointing back at `GuppyRegionsTable`, i.e. a per-region fact stored one indirection away from the region it describes. The only extra it carries is the artifact-removal method, which already sits structurally on `GuppyParameters` (`artifacts_removal_method`) and is just re-embedded here as a description string, so the separate object adds no information beyond the per-region intervals themselves. I have used this same separate-object pattern in `ndx-spikesorting`, though there each object was one-per-metric and linked a specific column rather than the whole entity: the metric columns ([`UnitVectorData`](https://github.com/catalystneuro/ndx-spikesorting/blob/883c25692d3fdcd272c944c397acfd3cb85b02c9/spec/ndx-spikesorting.extensions.yaml#L18-L35)) each carry a `time_support` reference to the specific window they were computed over, and the valid periods themselves live in a dedicated [`ValidUnitPeriods`](https://github.com/catalystneuro/ndx-spikesorting/blob/883c25692d3fdcd272c944c397acfd3cb85b02c9/spec/ndx-spikesorting.extensions.yaml#L397-L408) object. 

I am curious what drove the separate object here, is this about how guppy is used?

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
```diff
@@ -0,0 +1,1385 @@
...
+        return kwargs
+
+    def _peak_windows(self) -> tuple[np.ndarray, np.ndarray]:
+        """Return the real (non-NaN-padded) peak window start/stop arrays from the parameters."""
+        start_points = np.asarray(self._guppy_parameters.get("peak_startPoint"), dtype=np.float64)
+        end_points = np.asarray(self._guppy_parameters.get("peak_endPoint"), dtype=np.float64)
+        valid = ~np.isnan(start_points)
+        return start_points[valid], end_points[valid]
+
+    def get_metadata(self) -> DeepDict:
```
h-mayorquin, line 479: The `get_metadata` has two purposes: (1) tell the user what the format natively provides and (2) expose a natural handle for the user to edit fields such as names, descriptions, etc. Currently, however, we have fields like `region`, `trace_basename`, and `event_names` that should not be user editable: they are internal handles `add_to_nwbfile` reads back to find and link data, so editing them crashes the conversion. For example, `region="dms"` is a join key, a user editing it to `"DMS (dorsomedial striatum)"` (assuming it's a label) gets a `KeyError` in `region_to_row_index`; `trace_basename` is the `.hdf5` filename to open, so editing it gets a `FileNotFoundError`.

The internal bookkeeping and the editable metadata fields should be segregated.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
```diff
@@ -0,0 +1,1385 @@
...
+        """Return metadata pre-populated from the GuPPy outputs and parameters file."""
```
h-mayorquin, line 480: The five product families (`Traces`, `Transients`, `CrossCorrelations`, `PSTHs`, `PeakAUCs`) are stored as **lists** of name-carrying dicts (`PSTHs: [ {name: "psth_dms_z_score", ...}, ... ]`), but the direction since [#269](https://github.com/catalystneuro/neuroconv/issues/269) has been **dicts keyed by `metadata_key`** (the "tag" in that issue), and the `BaseFiberPhotometryInterface` this pairs with is already keyed. So a new interface on lists is adopting the shape we're removing. The transform is mechanical: `[ {name: X, ...}, ... ]` becomes `{ X: {name: X, ...}, ... }` (and the schema's `type: "array"` blocks become `additionalProperties` object blocks), and override becomes a direct path (`...["PSTHs"]["psth_dms_z_score"]["description"] = ...`) instead of a list scan.

On multiple instances: I'd scope the whole `Guppy` block under a per-interface `metadata_key` so everything the interface writes lives under one key and two GuPPy interfaces in one file don't collide, the disambiguation that `metadata_key` was introduced for in [#269](https://github.com/catalystneuro/neuroconv/issues/269). I don't think anything is truly linkable across interfaces here (regions, events, and parameters are all per session), so the `metadata_key` only plays that top-level disambiguation role, not any sharing one.

h-mayorquin (changes requested):
It think this is mostly ready, the only things that I would really required are the metadata changes as those should conform to neuroconv standards.

The rests are questions about design or suggestions that can be dropped with justification so we document our decisions somewhere. 

After the metadata is ready I will review the testing part which is what I have not done today.

Thanks for the big effort and for the wait.

pauladkisson:
> I find the name `fiber_photometry_table_region_indices` confusing. it reads like a flat list of row indices, but it is actually a mapping from region label to that region's rows in the `FiberPhotometryTable` (thanks for adding the typing). A name that surfaces the dict, something like `region_to_fiber_photometry_table_rows`, would remove the first-read confusion.
> Stepping back on why this argument exists at all: the derived traces (and the `GuppyRegionsTable`) link into the acquisition `FiberPhotometryTable` by integer row index, but `GuppyInterface` neither builds that table nor knows its row order, since that is an acquisition-side fact. So the converter has to compute this map and inject it, and that injection is also what forces the interface to be non-standalone. We could avoid the argument entirely if the `FiberPhotometryTable` carried the source stream identifier on each row: with a `stream_name` column `GuppyInterface` matches each region's signal and control stores (the `store_id`s from `storesList.csv`) against the table itself and derives the region-to-row map, instead of having it handed in:

> 
> ```python
> stream_name_to_row = {row.stream_name: index for index, row in enumerate(fiber_photometry_table)}
> region_to_indices = {
>     region: sorted(stream_name_to_row[store_id] for store_id in stores.values())  # signal + control
>     for region, stores in self.region_to_store_names.items()
> }
> ```
> 
> This is worth doing on its own merits, the same way the electrodes table carries `channel_name`. Today the `FiberPhotometryTable` has no way to record which acquisition stream a row came from, so two rows with the same location, wavelengths, and devices are indistinguishable except by order. A `stream_name` column lets not only GuPPy but any other aggregator disambiguate rows and tell which acquisition stream each one is.

> Some feedback about ndx-guppy:
> 
> Following up on [this PR comment](https://github.com/catalystneuro/neuroconv/pull/965#discussion_r3580713011) about carrying a `stream_name` identifier on the `FiberPhotometryTable` as a column. I think having the `raw_store_name` on the `GuppyRegionsTable` is also lossy in its current form: a region spans two streams (signal and isosbestic control), so the scalar column keeps the signal store and drops the control, while the `fiber_photometry_table_region` link already reaches both. So the region table would need no store column at all. Maybe we should drop it from the `GuppyRegionsTable`?
> 
> Separately, I feel the most common interpretation of `region` is anatomical but this is not the case: labels must be unique in Guppy, so two fibers in the same anatomical structure become two "regions" in the GuPPy sense. Plus, dual-color recording in the same anatomical area splits into two regions as well. It seems to me that the right unit here is a recording site for GuPPy to analyze, so maybe we should improve the docstring to make this more emphatic, or rename as such `region` -> `recording_site` if you'd rather the schema be self-explanatory. I would prefer the second but that would take it away from GuPPy's internal vocabulary, which is undesirable. But if you are open to rename I would also do `GuppyRegionsTable` -> `GuppyRecordingSitesTable` to avoid the confusion completely.


Ok, so these are a lot of interrelated points that I think are worth splitting out and handling separately. Roughly in reverse order,

1. **The "regions" concept:** after a recent PR, we have an established [glossary](https://guppy.readthedocs.io/reference/glossary.html) of what a store, store_id, and store_label mean in Guppy. We did not establish what the region refers to, and maybe we should have. The current usage in Guppy is a single fiber recording site shared by a single signal/control channel pair. This situation seems like a reasonable starting place to me, but certainly is open to criticism. In particular, the multicolor situation seems like a relevant shortcoming. If we want to rename the types in ndx-guppy, it is certainly worth the trouble to make the names clear in the GuPPy processing tool itself. 
2. **Updating the `FiberPhotometryTable` in ndx-fiber-photometry:** This option is definitely on the table, but I'm not so sure what adding a string name to each row of the table really accomplishes. Unlike the electrodes table, there is not always a straightforward mapping between a hardware configuration and a "channel". We've also run into similar issues with the `FiberPhotometryTable` in the past, and so it's worth considering them together [here](https://github.com/catalystneuro/ndx-fiber-photometry/issues/54).
3. **Should Guppy be a stand-alone interface?**: Because of the linkages between the ndx-guppy types and the source tables (FiberPhotometryTable and EventsTables), Guppy is _not_ standalone (although the linkages themselves are optional, so it can optionally stand alone). This awkwardness is the price that we pay for the rich linkage. Perhaps it's not worth it, and we can just let the Guppy types stand on their own without explicitly linking to the source acquisition types. That's a discussion we could certainly have. If we want to keep the linkages, though, I think the converter is absolutely the best place to specify that rather than it being automatic. I am not confident that automatic specification would work reliably. In particular, what happens if Guppy is run on an existing NWB file with fiber photometry data?
4. **The name `fiber_photometry_table_region_indices` is confusing:** Fair enough. I am more than happy to update it.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
pauladkisson: There is no such dedicated helper, but it seems like it would be good to have one.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
pauladkisson: I think this is a desirable situation to have a DynamicTableRegion that references these specific rows that the events-based tables refer to. However, there is one major problem to work through. The tables like `GuppyPSTH` aggregate event instances across all event types into a single matrix (num_samples, num_trials; where num_trials = num_event1 + num_event2 + ... + num_eventN). This was done to keep a lid on the number of neurodata types involved, which grows quite large as the number of event types grows. 

So I can't just add a single `DynamicTableRegion` because there is more than one table to reference. Maybe I could add a list of `DynamicTableRegion` objects? Something for me to look into...

pauladkisson:
> A question about how the event-aligned products are structured. Looking at how the PSTHs are built, we are stacking them across events: GuPPy writes one PSTH file per `(event, region, feature)`, but the interface groups by `(region, feature, baseline)` and concatenates across events, so a single `GuppyPSTH` holds several events' trials side by side, distinguished by a per-trial `event` column. (The same grouping happens for `GuppyPeakAUC` and `GuppyCrossCorrelation`.)
> 
> What is the intent behind stacking them rather than keeping one object per `(region, event, feature)`, matching GuPPy's files? A per-event object would be homogeneous: its `event` becomes a single reference instead of a per-trial one and quite a lot of machinery for stacking would go away making the interface way simpler. Curios on how you made the decision.

This is a great question, and I am happy to explain. This situation is true for a variety of the ndx-guppy types, but without loss of generality, I will just refer to GuppyPSTH.

Originally, I designed the extension to match Guppy's outputs in exactly the way that you mention: one Guppy PSTH object per Guppy psth output file. However, I noticed that this created quite a few GuppyPSTH objects, since the total number is equal to the product of `num_regions * num_trace_types * num_events`. For two regions and three events, this is already 12 GuppyPSTH objects (num_trace_types is always 2). I was concerned that for a larger number of regions and events, the number of neurodata objects created by this extension could balloon to a problematic level. I know that for a large number of neurodata objects, it can start to slow down read speeds, which is a situation I wanted to avoid (although I'm not sure exactly how many objects you need before this becomes a problem).

So I concatenated them together along the event axis.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
pauladkisson: This is not an option that I really considered, since I had envisioned the Regions table as a sort of pure registry, but I think it's a very reasonable consolidation. I'm going to go ahead and implement it.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
h-mayorquin: Thanks.

h-mayorquin:
> Updating the FiberPhotometryTable in ndx-fiber-photometry: This option is definitely on the table, but I'm not so sure what adding a string name to each row of the table really accomplishes. Unlike the electrodes table, there is not always a straightforward mapping between a hardware configuration and a "channel". We've also run into similar issues with the FiberPhotometryTable in the past, and so it's worth considering them together https://github.com/catalystneuro/ndx-fiber-photometry/issues/54.

To clarify what I meant: I wasn't proposing to change the `FiberPhotometryTable` schema in ndx-fiber-photometry. Since it's a `DynamicTable`, I only meant adding an extra column at conversion time in the base fiber photometry interface just enough for the interface to have a per-row handle to join on. Exactly the way that we do in the electrodes table where channel name is not in the spec.

Thanks for pulling up #54, I'd forgotten we'd already worked through this on the call. That's actually the more important question here: are we still planning to move the derived response series onto option 4 from that discussion, i.e. keep `fiber_photometry_table_region` on the signal row and add the optional reference field for the isosbestic (we left the name open, "reference" vs "baseline")?

h-mayorquin:
> Originally, I designed the extension to match Guppy's outputs in exactly the way that you mention: one Guppy PSTH object per Guppy psth output file. However, I noticed that this created quite a few GuppyPSTH objects, since the total number is equal to the product of `num_regions * num_trace_types * num_events`. For two regions and three events, this is already 12 GuppyPSTH objects (num_trace_types is always 2). I was concerned that for a larger number of regions and events, the number of neurodata objects created by this extension could balloon to a problematic level. I know that for a large number of neurodata objects, it can start to slow down read speeds, which is a situation I wanted to avoid (although I'm not sure exactly how many objects you need before this becomes a problem).
> 
> So I concatenated them together along the event axis.

I see, yes, many objects make pynwb slow. It is a shame that we are introducing complexity and making decisions here for something that is not a limitation of the schema or the backend but of pynwb.

> I think this is a desirable situation to have a DynamicTableRegion that references these specific rows that the events-based tables refer to. However, there is one major problem to work through. The tables like GuppyPSTH aggregate event instances across all event types into a single matrix (num_samples, num_trials; where num_trials = num_event1 + num_event2 + ... + num_eventN). This was done to keep a lid on the number of neurodata types involved, which grows quite large as the number of event types grows.
>
> So I can't just add a single DynamicTableRegion because there is more than one table to reference. Maybe I could add a list of DynamicTableRegion objects? Something for me to look into...

Regarding the linking of the events, another solution is to defer the linking to the converter time (this is where the events will be added anyway) and add the events as a merged table with all the event_types. Given that the GuPPy interface is aggregating across event types I think this is in-line with that and would solve the problem of having to link multiple tables, what do you think? Given that, as you mentioned above, GuPPy is not stand-alone, doing this at converter time makes sense and would sidestep the multiple-tables problem entirely (one merged table means one `DynamicTableRegion`), and, since it becomes a row reference rather than an object reference to a table, the hdmf#1532/#1533 dependency the current `events` link is blocked on. It would also retire the `_resolve_events_table` name/discriminator guessing, since the converter provides the mapping directly. And I'm already working on a solution for #1786 (writing a value column shared across event types merged into one `EventsTable`), which is exactly the base support this merged table needs, so this direction lines up with that.

h-mayorquin:
> **The "regions" concept:** after a recent PR, we have an established glossary of what a store, store_id, and store_label mean in Guppy. We did not establish what the region refers to, and maybe we should have. The current usage in Guppy is a single fiber recording site shared by a single signal/control channel pair. This situation seems like a reasonable starting place to me, but certainly is open to criticism. In particular, the multicolor situation seems like a relevant shortcoming. If we want to rename the types in ndx-guppy, it is certainly worth the trouble to make the names clear in the GuPPy processing tool itself.

Agreed, and since you're open to making the concept clear in GuPPy itself, I'd go for the rename: `region` -> `recording_site` and `GuppyRegionsTable` -> `GuppyRecordingSitesTable`, coordinated across GuPPy and ndx-guppy, with a glossary-grade definition (one signal/control analysis unit for a single fiber and indicator). That removes my only hesitation (renaming ndx-guppy alone would open a GuPPy-says-region / ndx-says-recording_site seam), and since the concept was never in the glossary, this is defining it correctly rather than breaking an established term.

> **Should Guppy be a stand-alone interface?** Because of the linkages between the ndx-guppy types and the source tables (FiberPhotometryTable and EventsTables), Guppy is not standalone (although the linkages themselves are optional, so it can optionally stand alone). [...] If we want to keep the linkages, I think the converter is absolutely the best place to specify that rather than it being automatic. I am not confident that automatic specification would work reliably. In particular, what happens if Guppy is run on an existing NWB file with fiber photometry data?

Agreed, the linkage belongs in the converter, not in automatic derivation. And embracing that actually simplifies things rather than adding to them: once the converter owns the linkage (it's the piece that knows both the acquisition rows and the GuPPy regions, and it injects the mapping), we don't need the automatic self-derivation path at all, and the machinery that would have supported it falls away. So the `stream_name` idea reduces to just a provenance column (the `channel_name` analog), worth adding on its own merits but no longer doing any linking work.

pauladkisson:
> > Updating the FiberPhotometryTable in ndx-fiber-photometry: This option is definitely on the table, but I'm not so sure what adding a string name to each row of the table really accomplishes. Unlike the electrodes table, there is not always a straightforward mapping between a hardware configuration and a "channel". We've also run into similar issues with the FiberPhotometryTable in the past, and so it's worth considering them together [catalystneuro/ndx-fiber-photometry#54](https://github.com/catalystneuro/ndx-fiber-photometry/issues/54).
> 
> To clarify what I meant: I wasn't proposing to change the `FiberPhotometryTable` schema in ndx-fiber-photometry. Since it's a `DynamicTable`, I only meant adding an extra column at conversion time in the base fiber photometry interface just enough for the interface to have a per-row handle to join on. Exactly the way that we do in the electrodes table where channel name is not in the spec.
> 
> Thanks for pulling up #54, I'd forgotten we'd already worked through this on the call. That's actually the more important question here: are we still planning to move the derived response series onto option 4 from that discussion, i.e. keep `fiber_photometry_table_region` on the signal row and add the optional reference field for the isosbestic (we left the name open, "reference" vs "baseline")?

I'm not sure. I wanted to get some buy-in/consensus from the rest of the team before making changes to ndx-fiber-photometry so I don't have to roll them back later. And it turned out that I didn't strictly need a DFF series because I had flexible referencing in the GuppyRegions table.

pauladkisson:
> Regarding the linking of the events, another solution is to defer the linking to the converter time (this is where the events will be added anyway) and add the events as a merged table with all the event_types. Given that the GuPPy interface is aggregating across event types I think this is in-line with that and would solve the problem of having to link multiple tables, what do you think? Given that, as you mentioned above, GuPPy is not stand-alone, doing this at converter time makes sense and would sidestep the multiple-tables problem entirely (one merged table means one DynamicTableRegion), and, since it becomes a row reference rather than an object reference to a table, the hdmf#1532/https://github.com/catalystneuro/neuroconv/pull/1533 dependency the current events link is blocked on. It would also retire the _resolve_events_table name/discriminator guessing, since the converter provides the mapping directly. And I'm already working on a solution for https://github.com/catalystneuro/neuroconv/issues/1786 (writing a value column shared across event types merged into one EventsTable), which is exactly the base support this merged table needs, so this direction lines up with that.

This is a great idea! So if I'm understanding correctly, the Guppy Converter would enforce a specification on top of the base events system in NeuroConv, such that all event types land in a single EventsTable. It would then construct the Guppy events table registry to reference those various events, and then the Guppy interface would be able to point to that registry. 

One thing to note is this fix pushes the issue out of scope of this PR to the Guppy converter: #1783

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
pauladkisson: Done!

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
pauladkisson: Done!

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
pauladkisson: Metadata is updated!

Here is an example YML file for your reference of the structure. 

<Details>

```yaml
# metadata['FiberPhotometry']['Guppy'] from a 2-region (dms/dls) mock GuPPy output.
# Families are dicts keyed by object name; entries hold only editable fields.

FiberPhotometry:
  Guppy:
    ProcessingModule:
      name: fiber_photometry
      description: GuPPy-derived fiber photometry processing outputs (GuPPy version 2.0.0a7).
    Traces:
      cntrl_sig_fit_dms:
        name: cntrl_sig_fit_dms
        description: GuPPy fitted control trace for region 'dms'.
      dff_dms:
        name: dff_dms
        description: GuPPy ΔF/F trace for region 'dms'.
      z_score_dms:
        name: z_score_dms
        description: GuPPy z-scored trace for region 'dms'.
      cntrl_sig_fit_dls:
        name: cntrl_sig_fit_dls
        description: GuPPy fitted control trace for region 'dls'.
      dff_dls:
        name: dff_dls
        description: GuPPy ΔF/F trace for region 'dls'.
      z_score_dls:
        name: z_score_dls
        description: GuPPy z-scored trace for region 'dls'.
    Transients:
      transients_dms_z_score:
        name: transients_dms_z_score
        description: GuPPy-detected transient peaks in the 'z_score' trace for region 'dms'.
      transients_dms_dff:
        name: transients_dms_dff
        description: GuPPy-detected transient peaks in the 'dff' trace for region 'dms'.
      transients_dls_z_score:
        name: transients_dls_z_score
        description: GuPPy-detected transient peaks in the 'z_score' trace for region 'dls'.
      transients_dls_dff:
        name: transients_dls_dff
        description: GuPPy-detected transient peaks in the 'dff' trace for region 'dls'.
    TransientSummary:
      name: transient_summary
      description: 'Per-(region, trace_type) GuPPy transient summary: events/min and mean peak amplitude.'
    CrossCorrelations:
      cross_correlation_dff_dls_dms:
        name: cross_correlation_dff_dls_dms
        description: GuPPy peri-event cross-correlation of the 'dff' trace between regions 'dls' and 'dms'.
      cross_correlation_z_score_dls_dms:
        name: cross_correlation_z_score_dls_dms
        description: GuPPy peri-event cross-correlation of the 'z_score' trace between regions 'dls' and
          'dms'.
    PSTHs:
      psth_dms_z_score:
        name: psth_dms_z_score
        description: GuPPy peri-event PSTH of the 'z_score' trace for region 'dms' (baseline-corrected).
      psth_dms_z_score_baseline_uncorrected:
        name: psth_dms_z_score_baseline_uncorrected
        description: GuPPy peri-event PSTH of the 'z_score' trace for region 'dms' (baseline-uncorrected).
      psth_dms_dff:
        name: psth_dms_dff
        description: GuPPy peri-event PSTH of the 'dff' trace for region 'dms' (baseline-corrected).
      psth_dms_dff_baseline_uncorrected:
        name: psth_dms_dff_baseline_uncorrected
        description: GuPPy peri-event PSTH of the 'dff' trace for region 'dms' (baseline-uncorrected).
      psth_dls_z_score:
        name: psth_dls_z_score
        description: GuPPy peri-event PSTH of the 'z_score' trace for region 'dls' (baseline-corrected).
      psth_dls_z_score_baseline_uncorrected:
        name: psth_dls_z_score_baseline_uncorrected
        description: GuPPy peri-event PSTH of the 'z_score' trace for region 'dls' (baseline-uncorrected).
      psth_dls_dff:
        name: psth_dls_dff
        description: GuPPy peri-event PSTH of the 'dff' trace for region 'dls' (baseline-corrected).
      psth_dls_dff_baseline_uncorrected:
        name: psth_dls_dff_baseline_uncorrected
        description: GuPPy peri-event PSTH of the 'dff' trace for region 'dls' (baseline-uncorrected).
    PeakAUCs:
      peak_auc_dms_z_score:
        name: peak_auc_dms_z_score
        description: GuPPy peak/area summary of the 'z_score' PSTH for region 'dms'.
      peak_auc_dms_dff:
        name: peak_auc_dms_dff
        description: GuPPy peak/area summary of the 'dff' PSTH for region 'dms'.
      peak_auc_dls_z_score:
        name: peak_auc_dls_z_score
        description: GuPPy peak/area summary of the 'z_score' PSTH for region 'dls'.
      peak_auc_dls_dff:
        name: peak_auc_dls_dff
        description: GuPPy peak/area summary of the 'dff' PSTH for region 'dls'.
```

</Details>

pauladkisson:
> > Regarding the linking of the events, another solution is to defer the linking to the converter time (this is where the events will be added anyway) and add the events as a merged table with all the event_types. Given that the GuPPy interface is aggregating across event types I think this is in-line with that and would solve the problem of having to link multiple tables, what do you think? Given that, as you mentioned above, GuPPy is not stand-alone, doing this at converter time makes sense and would sidestep the multiple-tables problem entirely (one merged table means one DynamicTableRegion), and, since it becomes a row reference rather than an object reference to a table, the hdmf#1532/#1533 dependency the current events link is blocked on. It would also retire the _resolve_events_table name/discriminator guessing, since the converter provides the mapping directly. And I'm already working on a solution for #1786 (writing a value column shared across event types merged into one EventsTable), which is exactly the base support this merged table needs, so this direction lines up with that.
> 
> This is a great idea! So if I'm understanding correctly, the Guppy Converter would enforce a specification on top of the base events system in NeuroConv, such that all event types land in a single EventsTable. It would then construct the Guppy events table registry to reference those various events, and then the Guppy interface would be able to point to that registry.
> 
> One thing to note is this fix pushes the issue out of scope of this PR to the Guppy converter: #1783

Done!

pauladkisson:
@h-mayorquin I do believe I have responded to all of your feedback. If I missed anything, please let me know. Otherwise, I'd like to cut a release of ndx-guppy and get this work merged in! 

One thing to note is that I slimmed down the registry tables now that they're being constructed at the converter level and pulled the artifact tracking GuppyValidSignalIntervals table back out into its own type. I think this creates a clean separation of responsibilities, but if you disagree, please let me know.

h-mayorquin:
> @h-mayorquin I do believe I have responded to all of your feedback. If I missed anything, please let me know. Otherwise, I'd like to cut a release of ndx-guppy and get this work merged in!
> 
> One thing to note is that I slimmed down the registry tables now that they're being constructed at the converter level and pulled the artifact tracking GuppyValidSignalIntervals table back out into its own type. I think this creates a clean separation of responsibilities, but if you disagree, please let me know.

Hi, I don't have further suggestions about the ndx-guppy so go ahead and make a release but I realized that I did not review the time alignment model here in this PR and I am thinking about it now at the moment. 

If you need this to go through you can separate the time alignment part in a separate PR and you can merge this one (this will also make my review easier) but otherwise, I will review that today.

pauladkisson:
> > @h-mayorquin I do believe I have responded to all of your feedback. If I missed anything, please let me know. Otherwise, I'd like to cut a release of ndx-guppy and get this work merged in!
> > One thing to note is that I slimmed down the registry tables now that they're being constructed at the converter level and pulled the artifact tracking GuppyValidSignalIntervals table back out into its own type. I think this creates a clean separation of responsibilities, but if you disagree, please let me know.
> 
> Hi, I don't have further suggestions about the ndx-guppy so go ahead and make a release but I realized that I did not review the time alignment model here in this PR and I am thinking about it now at the moment.

Great, I will go ahead and make that release. 

> If you need this to go through you can separate the time alignment part in a separate PR and you can merge this one (this will also make my review easier) but otherwise, I will review that today.

No, I probably want to just keep everything together. Thanks for all the hard work on the reviews.

h-mayorquin (changes requested):
I am fine with `GuppyValidSignalIntervals` as a stand-alone type, no objection. Curious what made you revert it though. Just having a class specifically for that is what you mean by separation of responsibilities? 

Ok, I reviewed the time alignment part.

Three points to discuss. I am assuming that the time basis is per recording site, and that either we set the timestamps of that recording site (`set_aligned_timestamps`) or we offset them (`set_aligned_starting_time`).

* I think `trial_onset_times` in PSTHs, cross-correlations and peak-AUCs should be aligned. `set_aligned_starting_time` moves the traces, the transients and the valid intervals, but not the PSTH / cross-correlation / peak-AUC. The spec says `trial_onset_times` is "Absolute onset time of each trial in seconds (session clock)", so I think they should be moved if their timestamps move, right?

* I think the `GuppyValidSignalIntervals` intervals should follow the same logic as the transients. They are per-recording-site times exactly like the transient peaks, so they should be interpolated against that site's time basis in the same way. Why would they follow a different logic if they indicate which part of the recording site is valid signal?

* When writing the `GuppyDerivedResponseSeries` we should do what we do when writing any TimeSeries child in this library. We should check whether the timestamps are regular with `calculate_regular_series_rate` and based on that write either `starting_time` + `rate` or `timestamps`. 

And a question on `bin_edges`: the spec says they are "interpreted as trial-index ranges or time ranges according to the bin_basis attribute", but it does not say whether those time ranges are absolute session time or relative to something. `trial_onset_times` and `peri_event_time` are both explicit ("seconds (session clock)" and "seconds (relative to event onset)"), so this one is the odd field out. Which is it? If they are absolute then they are a session coordinate too and would need the same treatment as the onsets above.

tests/test_modalities/test_guppy/test_guppy_interface.py
```diff
@@ -0,0 +1,622 @@
...
+                        "PrtR": "port_entries",
+                    },
+                    # Signal store id per recording site, from the default mock TDT topology.
+                    expected_recording_site_to_store_id={"dms": "Dv2A", "dls": "Dv4B"},
+                ),
+                id="mock_isosbestic_two_recording_sites",
+            ),
+        ]
+    )
+    def case(self, request, tmp_path):
```
h-mayorquin: We should add a scope class at the class or module level here so this is not re-generated per test.

h-mayorquin (changes requested):
Two comments about testing.
* The data generating fixture should use the pytest scope function to avoid re-creating it one very test.
* We should test the `get_metadata_schema` function.

h-mayorquin:
One more thing I forgot to raise. As we discussed above, this is not really meant to be used stand-alone but in the context of the `GuppyConverter`. Would it make sense to make it private then? If the interface is private we can change it whenever we need to without breaking anyone, and we can put the effort into properly documenting the converter's interface and keeping that one stable instead.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
h-mayorquin: You added the `metadata_key` under `metadata["FiberPhotometry"][metadata_key]` but what I mean is to add it under `metadata["FiberPhotometry"]["Guppy"][metadata_key]` instead. Otherwise the key sits at the same level as the actual fiber photometry categories such as `FiberPhotometryTable` and `FiberPhotometryIndicators`, and nothing stops someone passing `metadata_key="FiberPhotometryTable"` and clobbering that block. The idea is to have any guppy instance nested under "FiberPhotometry"/"Guppy"

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py
h-mayorquin: Thanks I think this is done as well. @pauladkisson . Resolving.

pauladkisson:
> I am fine with GuppyValidSignalIntervals as a stand-alone type, no objection. Curious what made you revert it though. Just having a class specifically for that is what you mean by separation of responsibilities?

Yes, I reverted it so that the Guppy converter wouldn't have to go digging around into the Guppy output files (the only one it looks at is storesList.csv), since I'm envisioning that responsibility as being primarily held by the Guppy interface.

pauladkisson:
> And a question on `bin_edges`: the spec says they are "interpreted as trial-index ranges or time ranges according to the bin_basis attribute", but it does not say whether those time ranges are absolute session time or relative to something. `trial_onset_times` and `peri_event_time` are both explicit ("seconds (session clock)" and "seconds (relative to event onset)"), so this one is the odd field out. Which is it? If they are absolute then they are a session coordinate too and would need the same treatment as the onsets above. 

When the bin edges are time ranges (i.e., the bin basis is equal to 'time'), the bin edges are absolute time (rounded integer minutes) relative to the session start time.

pauladkisson:
> One more thing I forgot to raise. As we discussed above, this is not really meant to be used stand-alone but in the context of the `GuppyConverter`. Would it make sense to make it private then? If the interface is private we can change it whenever we need to without breaking anyone, and we can put the effort into properly documenting the converter's interface and keeping that one stable instead.

Yes, I think this is a good idea.

pauladkisson:
> Ok, I reviewed the time alignment part.

Okay, so I was thinking a little bit more critically about this, and I think it's a bit of a can of worms. 

First of all, in order for the Guppy analyses to work properly, the events stream needs to be already aligned with the fiber photometry stream, prior to its ingestion into Guppy. Systems like TDT already work in one central clock, so not a huge constraint, but it's there. 

Second, if an experimenter needed to align a Guppy clock, which includes fiber photometry, events, and Guppy output, to another stream (ex. video), they would need to align the fiber photometry timestamps, the events timestamps, and all of the various different timestamps in the Guppy output to that stream. 

But the events interfaces don't include any mechanism for temporal alignment, and I don't think that they should, since the event timestamps are basically the whole data.

So as it stands, the Guppy converter triplet (FP + Events + Guppy) is not temporally alignable, even if we implement temporal alignment in the Guppy interface.

With all that said, I think maybe the best thing to do is to drop temporal alignment entirely. What do you think?

h-mayorquin:
> > Ok, I reviewed the time alignment part.
> 
> Okay, so I was thinking a little bit more critically about this, and I think it's a bit of a can of worms.
> 
> First of all, in order for the Guppy analyses to work properly, the events stream needs to be already aligned with the fiber photometry stream, prior to its ingestion into Guppy. Systems like TDT already work in one central clock, so not a huge constraint, but it's there.
> 
> Second, if an experimenter needed to align a Guppy clock, which includes fiber photometry, events, and Guppy output, to another stream (ex. video), they would need to align the fiber photometry timestamps, the events timestamps, and all of the various different timestamps in the Guppy output to that stream.
> 
> But the events interfaces don't include any mechanism for temporal alignment, and I don't think that they should, since the event timestamps are basically the whole data.
> 
> So as it stands, the Guppy converter triplet (FP + Events + Guppy) is not temporally alignable, even if we implement temporal alignment in the Guppy interface.
> 
> With all that said, I think maybe the best thing to do is to drop temporal alignment entirely. What do you think?

Yes, I think this is difficult and I think is the fault of our alignment model. I really like what you did here and it inspired me to think more about it:

https://neuroconv--1793.org.readthedocs.build/en/1793/user_guide/temporal_alignment.html


I think to defer is a good call then do the alignment once we have the converter and an alignment model for events so we can get this feature in. Can you the metadata request above?

tests/test_modalities/test_guppy/test_guppy_interface.py
pauladkisson: I ended up scoping the actual output folder to the class level, but this parameterized case fixture still gets scoped per function, which I think is fine.

pauladkisson:
All feedback addressed!

h-mayorquin (approved):
LGTM

PR #966, opened 2024-07-22
Remove date tags & debug release trigger for docker images

CodyCBakerPhD:
The dev date tags are needlessly flooding the GHCR; apparently it keeps around any images that have a unique tag (which would apply to all date tags) and only overrides based on that (such as static `dev` tag)

Also wondered why the builds weren't triggering for latest release, realized it was a naming issue in the GitHub Action

### DIFF
--- a/.github/workflows/build_and_upload_docker_image_dev.yml
+++ b/.github/workflows/build_and_upload_docker_image_dev.yml
@@ -35,7 +35,7 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           push: true  # Push is a shorthand for --output=type=registry
-          tags: ghcr.io/catalystneuro/neuroconv:dev,ghcr.io/catalystneuro/neuroconv:${{ steps.date.outputs.date_tag }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build_and_upload_docker_image_dev.yml
```diff
@@ -2,7 +2,7 @@ name: Build and Upload Docker Image of Current Dev Branch to GHCR
 
 on:
   schedule:
-    - cron: "0 16 * * 1"  # Weekly at noon EST on Monday
+    - cron: "0 13 * * *" # Daily at 9 EST
```
h-mayorquin, line 5: Os Is this temporary or you want to push more often?

h-mayorquin (approved):
OK, should we erase the old ones?

.github/workflows/build_and_upload_docker_image_dev.yml
CodyCBakerPhD: More often when it replaces rather than piles on is better

However, doing a fresh build on merge to main would actually be best

Will have separate PR for that

PR #972, opened 2024-07-29
Fix suffixes for GUIDE

CodyCBakerPhD:
@pauladkisson Wow, I completely missed this on review of your PR

GUIDE needs these to be a tuple, even of a single element

Since you're taking over DevOps, it might be a good idea to setup dev testing of NWB GUIDE on NeuroConv to ensure this kind of thing doesn't happen silently and slowly creep up over there


pauladkisson (approved):

pauladkisson:
> Since you're taking over DevOps, it might be a good idea to setup dev testing of NWB GUIDE on NeuroConv to ensure this kind of thing doesn't happen silently and slowly creep up over there

Will add to my TODOs. Also, specifically in this case looks like we need a more thorough check of the GUIDE attrs (should be a tuple).

CodyCBakerPhD:
Yes, I suppose an intermediate to full testing would be to make some timple pytest based NeuroConv-wide assertions on attribute structure for consistency

PR #974, opened 2024-07-29
More Detailed Testing of GUIDE attrs

pauladkisson:
See #972 for motivation.

### DIFF
--- a/tests/test_minimal/test_tools/test_importing.py
+++ b/tests/test_minimal/test_tools/test_importing.py
@@ -11,5 +11,21 @@ def test_guide_attributes():
     for name, summary in summaries.items():
         for key, value in summary.items():
             assert value is not None, f"{name} is missing GUIDE related attribute {key}."
-            if isinstance(value, tuple):
+            if key == "keywords":
+                assert isinstance(
+                    value, (list, tuple)
+                ), f"{name} incorrectly specified GUIDE-related attribute 'keywords' (must be list or tuple)."
+            elif key == "display_name":
+                assert isinstance(
+                    value, str
+                ), f"{name} incorrectly specified GUIDE-related attribute 'display_name' (must be str)."
+            elif key == "info":
+                assert isinstance(
+                    value, str
+                ), f"{name} incorrectly specified GUIDE-related attribute 'info' (must be str)."
+            elif key == "associated_suffixes":
+                assert isinstance(
+                    value, (list, tuple)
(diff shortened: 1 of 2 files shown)


tests/test_minimal/test_tools/test_importing.py
```diff
@@ -11,5 +11,21 @@ def test_guide_attributes():
     for name, summary in summaries.items():
         for key, value in summary.items():
             assert value is not None, f"{name} is missing GUIDE related attribute {key}."
-            if isinstance(value, tuple):
+            if key == "keywords":
+                assert isinstance(
+                    value, (list, tuple)
+                ), f"{name} incorrectly specified GUIDE-related attribute 'keywords' (must be list or tuple)."
```
CodyCBakerPhD, line 17: Do we have current cases where it is a list instead of a tuple? In general I would move to have it be immutable (and thus constructed by partial copy if you want to 'inherit' from a parent attribute)

tests/test_minimal/test_tools/test_importing.py
```diff
@@ -11,5 +11,21 @@ def test_guide_attributes():
...
+                    value, str
+                ), f"{name} incorrectly specified GUIDE-related attribute 'display_name' (must be str)."
+            elif key == "info":
+                assert isinstance(
+                    value, str
+                ), f"{name} incorrectly specified GUIDE-related attribute 'info' (must be str)."
+            elif key == "associated_suffixes":
+                assert isinstance(
+                    value, (list, tuple)
+                ), f"{name} incorrectly specified GUIDE-related attribute 'associated_suffixes' (must be list or tuple)."
```
CodyCBakerPhD, line 29: Same as above; list should not be allowed (and I hope it currently isn't used anywhere)

tests/test_minimal/test_tools/test_importing.py
pauladkisson: At least MedPC uses a list, not sure about others.

pauladkisson: we shall see from the tests

CodyCBakerPhD (approved):

CodyCBakerPhD:
Lovely - thank you!

CodyCBakerPhD:
re-triggering RTD

CodyCBakerPhD:
Well, I have no idea why RTD isn't reporting correctly but it is running the build: https://readthedocs.org/projects/neuroconv/builds/25147260/

Issue #958, opened 2024-07-12
[Bug]: Dimension mismatch for SpikeGLX data with more than one electrical recording


tabedzki:
### What happened?

Approaches taken:

- Successfully enter the conversion routine for both imec folders individually if I use two SpikeGLXRecordingInterface s within the definition of my NWBConverter (one for the LF and one for the recording itself)
  - I had no problems doing an individual LFP conversion, an individual recording conversion and no problems for a combined LF/Recording conversion
- However, if I point to the root folder ya008_20240525_g0 for use with the SpikeGLXConverterPipe , I get a shape mismatch warning on line 90 in [ecephys.py](https://github.com/NeurodataWithoutBorders/pynwb/blob/20c5d75a6ce0b2deb729b81803c9c6440cff029c/src/pynwb/ecephys.py#L90) where the len(args_to_set['electrodes'].data) = 672 and data_shape = (117364459, 384).
  - We have two probes, each with 384 channels/electrodes
  - However, they are being combined in a way that causes a shape/dimensionality mismatch.

### Steps to Reproduce

```python
1. Use the SpikeGLXConverterPipe as an interface for NWBConverter in lieu of SpikeGLXRecordingInterface
```

### Traceback

```python
data_shape=(117364459, 384)
len(args_to_set['electrodes'].data)=672
/opt/anaconda3/envs/neuroconv_3.12/lib/python3.12/site-packages/pynwb/ecephys.py:90: UserWarning: ElectricalSeries 'ElectricalSeriesAPIMEC0': The second dimension of data does not match the length of electrodes. Your data may be transposed.
  warnings.warn("%s '%s': The second dimension of data does not match the length of electrodes. "
/opt/anaconda3/envs/neuroconv_3.12/lib/python3.12/site-packages/pynwb/ecephys.py:90: UserWarning: ElectricalSeries 'ElectricalSeriesLFIMEC0': The second dimension of data does not match the length of electrodes. Your data may be transposed.
  warnings.warn("%s '%s': The second dimension of data does not match the length of electrodes. "
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.12

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

h-mayorquin:
[...]

PR #957, opened 2024-07-11
Fix default electrical series naming for SpikeGLXConverter

CodyCBakerPhD:
@tabedzki Does this fix your issue from Slack?

PR #977, opened 2024-08-02
fix metadata when writing more than one probe

h-mayorquin:
This is a draft. It should be working but reuquires tests and some cleaning.

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglx_utils.py
+++ b/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglx_utils.py
@@ -7,25 +7,6 @@ from pathlib import Path
 from ....utils import FilePathType
 
 
-def add_recording_extractor_properties(recording_extractor) -> None:
-    """Automatically add shank group_name and shank_electrode_number for spikeglx."""
-    probe = recording_extractor.get_probe()
-    channel_ids = recording_extractor.get_channel_ids()
-
-    if probe.get_shank_count() > 1:
-        group_name = [contact_id.split("e")[0] for contact_id in probe.contact_ids]
-        shank_electrode_number = [int(contact_id.split("e")[1]) for contact_id in probe.contact_ids]
-    else:
-        shank_electrode_number = recording_extractor.ids_to_indices(channel_ids)
-        group_name = ["s0"] * len(channel_ids)
-
-    recording_extractor.set_property(key="shank_electrode_number", ids=channel_ids, values=shank_electrode_number)
-    recording_extractor.set_property(key="group_name", ids=channel_ids, values=group_name)
-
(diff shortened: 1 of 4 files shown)


CodyCBakerPhD:
This all looks to me like a larger ProbeInterface metadata enhancement for this interface; while that's fine in general (and you're the person to do that) I was thinking all we really need for a simple fix to 'get the parent PR through' is to override the group names property of the extractors in the `SpikeGLXConverterPipe` (maybe like `imec_{probe_index}_s0`), as well as the `metadata` to match

Surely that's only a few lines there?

h-mayorquin:
The key thing is to write `group_name` using the probeinstead of shanks. No need to reference the shanks with "s0" though. I added the shank id because that information is not in the probe group anymore. The name of the devices and the electrodes also needs to be changed for the metadata to match. That would be what you need to add to the parent PR if you want to solve https://github.com/catalystneuro/neuroconv/issues/958 and have proper metadata in the parent.

Adding the following quantities is an addition and I can add that into another PR though:
- contact ids, probe_port and probe_slot

CodyCBakerPhD:
As I understand, our current write tools are still based on the use of `group_name` as a property on the extractor, with or without a probe attached

A 'group' in the classical sense being equivalent to a global shank ID and the 'device' being the probe

What you say probably sounds like all the better ways of representing (though TBH it's probably best to just do it when you integrate `ndx-extracellular-channels` into our write tools here) but not necessarily a minimal fix to get what we need for now

How should we proceed with #957 to get the core fix in for the user?

h-mayorquin:
In the current state of neuroconv the global id of an electrode is "group_name" + "channel_name". See:
https://github.com/catalystneuro/neuroconv/blob/8912133ea17e39fe0b184dd77ad391ce632049c3/src/neuroconv/tools/spikeinterface/spikeinterface.py#L354-L362

https://github.com/catalystneuro/neuroconv/blob/8912133ea17e39fe0b184dd77ad391ce632049c3/src/neuroconv/tools/spikeinterface/spikeinterface.py#L668-L677


The minimal fix that will allow the user to write data from two probes right metadata needs either of the following options:
* The group name (and therefore the ElectrodeGroup) [needs](https://github.com/catalystneuro/neuroconv/blob/8912133ea17e39fe0b184dd77ad391ce632049c3/src/neuroconv/tools/spikeinterface/spikeinterface.py#L290-L310) to be based on the probe  and not in the shank.
* Change the logic for the global id as described above so it does not use the "group name" but another yet to be added attribute that reflects the probe.

h-mayorquin:
Oh, now that you mention it, it does seem that you can create a group_name that contains **both** the probe and shank as you mention above:

> SpikeGLXConverterPipe (maybe like imec_{probe_index}_s0), as well as the metadata to match

I think that should work.

PR #979, opened 2024-08-02
Second fix for NWB GUIDE nested converter stubbing

CodyCBakerPhD:
OK, this should finally finish https://github.com/NeurodataWithoutBorders/nwb-guide/pull/870 @rly 

cc @h-mayorquin long-term it might be nice to think about aligning this minor discrepancy between how conversion options are passed between data interfaces and converters


CodyCBakerPhD:
@h-mayorquin OK tests passing now, and left comments in the code explaining routing pattern for different styles of subconverters

src/neuroconv/nwbconverter.py
```diff
@@ -158,10 +159,24 @@ def create_nwbfile(self, metadata: Optional[dict] = None, conversion_options: Op
...
         conversion_options = conversion_options or dict()
-        for interface_name, data_interface in self.data_interface_objects.items():
-            data_interface.add_to_nwbfile(
-                nwbfile=nwbfile, metadata=metadata, **conversion_options.get(interface_name, dict())
-            )
+        for interface_key, data_interface in self.data_interface_objects.items():
+            if isinstance(data_interface, NWBConverter):
+                subconverter_kwargs = dict(nwbfile=nwbfile, metadata=metadata)
+
+                # Certain subconverters fully expose control over their interfaces conversion options
```
h-mayorquin, line 166: Which interfaces?

src/neuroconv/nwbconverter.py
```diff
@@ -158,10 +159,24 @@ def create_nwbfile(self, metadata: Optional[dict] = None, conversion_options: Op
 
     def add_to_nwbfile(self, nwbfile: NWBFile, metadata, conversion_options: Optional[dict] = None) -> None:
         conversion_options = conversion_options or dict()
-        for interface_name, data_interface in self.data_interface_objects.items():
-            data_interface.add_to_nwbfile(
-                nwbfile=nwbfile, metadata=metadata, **conversion_options.get(interface_name, dict())
-            )
+        for interface_key, data_interface in self.data_interface_objects.items():
```
h-mayorquin, line 162: I think interface name is a better variable name, why change this?

h-mayorquin (changes requested):
I would like to see some examples to think better about this.

src/neuroconv/nwbconverter.py
CodyCBakerPhD: Bruker/Miniscope to name a few. Haven't compiled a complete list

src/neuroconv/nwbconverter.py
CodyCBakerPhD: It was never the name of the interface, it was whatever key the individual assigned to the entries of the `data_interface_classes` (or autogenerated from a `ConverterPipe`)

Neither name is perfect though since it can be either an interface or a subconverter

So maybe just `key`? or `item_key`?

CodyCBakerPhD:
The example is the SpikeGLXConverterPipe, the source of the issue from GUIDE: https://github.com/NeurodataWithoutBorders/nwb-guide/pull/870

src/neuroconv/nwbconverter.py
h-mayorquin: Are you coming from the angle that a class name is **only**  class.__name__? If so, fine. I think that interface_key is OK if you want to make that distinction.  I had a more general that name is whatever we use to refer to the interface but this is less important. I would not go for generic key or item_key os the current is OK.

CodyCBakerPhD: I'm coming at the angle that is it a key in a dictionary and does not need to match or even correlate with `interface.__name__`

CodyCBakerPhD: In the GUIDE we refer to this as the more verbose `interface_or_subconverter_key`

h-mayorquin: In my mind the name of an interface is not `interface.__name__`. I think that's  fine identification to have though so I think interface_key is OK.

I don't like the thing we did with the passing converter pipes as interfaces here tbh, the naming is wrong and I think it will create problems in the future.

CodyCBakerPhD: How else would we perform a single `run_conversion` call of both multi-probe SpikeGLX (using the Converter) and Phy sorting (single interface)?

Or multi-plane Bruker with PoseEstimation and ROI responses?

We made prolific use of this in the NWB GUIDE, where you can easily compose multi-stream converters with single-stream interfaces. In terms of their APIs, this issue with `conversion_options` is the only difference to reconcile

h-mayorquin:
Can you add some tests? I think I would need that to judge this well.

CodyCBakerPhD:
Sure, I can try

src/neuroconv/nwbconverter.py
h-mayorquin: Yeah, I did not realize you went that way until I started reading this.

src/neuroconv/nwbconverter.py
CodyCBakerPhD: Ah, but then we lose any differences that would have resulted from special `get_metadata` or `add_to_nwbfile` (including conversion option exposure) overrides in the subcoverters

h-mayorquin: I will check the examples you mentioned to think better about this.

tests/test_on_data/test_format_converters/test_spikeglx_converter.py
```diff
@@ -96,6 +96,26 @@ class TestConverter(NWBConverter):
         expected_session_start_time = datetime(2020, 11, 3, 10, 35, 10).astimezone()
         self.assertNWBFileStructure(nwbfile_path=nwbfile_path, expected_session_start_time=expected_session_start_time)
 
+    def test_single_probe_spikeglx_converter_preview(self):
```
CodyCBakerPhD, line 99: @h-mayorquin Test added

tests/test_on_data/test_format_converters/test_spikeglx_converter.py
```diff
@@ -96,6 +96,30 @@ class TestConverter(NWBConverter):
...
         self.assertNWBFileStructure(nwbfile_path=nwbfile_path, expected_session_start_time=expected_session_start_time)
 
+    def test_single_probe_spikeglx_converter_in_converter_preview(self):
+        class TestConverter(NWBConverter):
+            data_interface_classes = dict(SpikeGLX=SpikeGLXConverterPipe)
+
+        source_data = dict(SpikeGLX=dict(folder_path=str(SPIKEGLX_PATH / "Noise4Sam_g0")))
+        converter = TestConverter(source_data=source_data)
+
+        conversion_options = dict(
```
h-mayorquin, line 106: Thanks.
Let me think a bit on this.

One question:
Do you expect the user to know that the conversion options of each of the sub interfaces are set with their stream names? if so, how how would they? Or are they supposed to be handled by automatic tools?

CodyCBakerPhD: All information for specifying the `conversion_options` dictionary is available in the `conversion_options_schema`

So a user has several options:

a) Read the JSON schema returned by the highest-level `converter.get_conversion_options_schema()`. Not the easiest thing to do but this is the intended route (principle applies to source data schema and metadata schema too)

b) If they defined a custom `NWBConverter` to be nested inside another, they already know their own key references by definition

c) If they used a `ConverterPipe` or core-supported subconverter based on a `ConverterPipe`, then they could find the automatically generated key references via minimal syntax `converter[subconverter_key].data_interface_objects.keys()`. We could even expose a convenience retrieval for it.

The GUIDE of course simplifies all this by automating construction based on the schema structure

h-mayorquin: OK, thanks a lot fo bearing with me so far.

Second question:
Can we just pass conversion_options to all the interfaces os a pipe? Here in this case, they will be the same, right? Is there a case where they will differ? 

I know that they can differ in practice, like, one could theoretically want a different writing buffer or progress bar for one interface of a sub-converter but my question is: is there a place where it would be critical that they are separated?

CodyCBakerPhD: Take a look at bruker/miniscope for that style. There, one common option specifies stubbing. Which is mildly convenient for that but therefore the GUIDE does not support in-file progress updates for those converters, whereas it does for SpikeGLX (because we then have control over each TQDM bar for each data stream in the converter)

This is exactly the kind of question I think is fine for you to consider in the long-term; but for now we really just need this bug fixed and released for one of the more popular formats and uses on the GUIDE.

h-mayorquin: OK, then there is no need to generalize yet.

Move the modified `add_to_nwb` in this branch from to the SpikeGLX converter. This will fix your bug and will avoid us to modify one of the most central functions in the libraryfor a specific case. If we find a better way in the future then we only need to break the SipkeGLX converter.

h-mayorquin: Ok, thinking more about it I think that what bruker does is wrong. Overiding add_to_nwb means that some other utilities like setting the backend configurarion are not propagated in `run_conversion`:

https://github.com/catalystneuro/neuroconv/blob/a970fd44456b7508f43822dc5cf23efdb3285699/src/neuroconv/datainterfaces/ophys/brukertiff/brukertiffconverter.py#L98-L122

Let's make the converters behave like converters and take conversion_options as a dict instead of the "kwargs" arguments.

Check out  https://github.com/catalystneuro/neuroconv/pull/987

Any other converter that comes to mind that is overwriting `add_to_nwb`?

h-mayorquin: Comment still stands though @CodyCBakerPhD 

Let's fix this for nwb converters locally and wait for @weiglszonja on the other PR to do a more general solution that involves changing the central functions.

CodyCBakerPhD: What I'm saying is nothing is outright *wrong* with the current approaches - they are both valid choices in their own respects

- a subconverter that exposes granular control over individual interfaces
- a subconverter that aggregates common commands, like `stub_test` and applies them all through a simpler top-level command

I'm saying that choosing between these, or choosing to keep both at the same time, is the decision that is outside of the scope of this PR, can take a longer time, and receive input from the rest of the team (in particular Szonja who put a lot of those together)

This PR makes a minimal modification to fix a particular bug to allow both styles for flexibility since that reflects the actual current state of things (including any custom converters defined outside of NeuroConv)

I'd rather not delay this fix more than needed since it is rather important for the GUIDE and we're looking to do a release for it soon

h-mayorquin: It is not a minimal modification if it modifies `add_to_nwb` of the `NWBConverter`. Modifying the converter that trigger the bug directly is the minimal fix as I suggested. What's wrong with that?

CodyCBakerPhD: What minimal fix to SpikeGLXConverter are you suggesting that satisfies all the other concerns about its broader use (in particular progress updates)?

h-mayorquin: The function that you added here `add_to_nwb`. 

Add it the converters that need it. That's a localized change.

I thought that it was only spikeglx but it sems that ther is more of them.

CodyCBakerPhD: This is the main converter class that needs it.

As demonstrated in the test, this is for an `NWBConverter` that contains a `SpikeGLXConverter`.

There is no fix to the `SpikeGLXConverter` specifically outside of the general standardization of how to style subconverter conversion options, which is a broader question if it should be standardized at all, and if so which standard should be chosen.

I have added the change directly to the GUIDE in https://github.com/NeurodataWithoutBorders/nwb-guide/pull/908 so I suppose this PR is not necessary if you would prefer to close it - just be aware that programmatic usage of NeuroConv without this fix will have the same issue going forward

h-mayorquin:
OK, closing as this was fixed at the GUIDE level.

PR #980, opened 2024-08-05
[Cloud Deployment IIIa]: pre-PR for removal of unused helpers

CodyCBakerPhD:
A pre-PR to #384 to simplify the changelog

As mentioned over there, these functions aren't really used in practice and are likely both out of date and inaccurate (they certainly aren't verified)

This PR can serve as an easy reference if we ever want to add them back in refactored form


CodyCBakerPhD:
(I don't think CHANGELOG is needed)

src/neuroconv/tools/data_transfers/__init__.py
```diff
@@ -1,14 +1,6 @@
...
-from ._helpers import estimate_total_conversion_runtime
 
-__all__ = [
-    "estimate_s3_conversion_cost",
-    "automatic_dandi_upload",
-    "get_globus_dataset_content_sizes",
-    "transfer_globus_content",
-    "estimate_total_conversion_runtime",
-]
+__all__ = ["automatic_dandi_upload", "get_globus_dataset_content_sizes", "transfer_globus_content"]
```
h-mayorquin, line 6:
```suggestion
__all__ = ["automatic_dandi_upload", "get_globus_dataset_content_sizes", "transfer_globus_content",]
```

Let's keep this thing to keep the diffs simpler if you don't mind (trailing coma)

h-mayorquin (approved):
Makes sense. I like trailing comas in list because I think that coupled with black directives it makes diff easier.

Issue #982, opened 2024-08-06
Bug when adding new property in second table


h-mayorquin:
I am working on https://github.com/catalystneuro/neuroconv/issues/960 but found the following bug that I want to solve

```python
from pynwb.testing.mock.file import mock_NWBFile
from spikeinterface.core import generate_recording
from neuroconv.tools.spikeinterface import add_electrodes

recording1 = generate_recording(num_channels=3)
recording1 = recording1.rename_channels(new_channel_ids=["a", "b", "c"])
recording1.set_property(key="group_name", values=["group1"] * 3)
recording2 = generate_recording(num_channels=3)
recording2 = recording2.rename_channels(new_channel_ids=["a", "b", "c"])
recording2.set_property(key="group_name", values=["group2"] * 3)

recording2.set_property(key="second_recording_property", values=["value"] * 3)

nwbfile = mock_NWBFile()

add_electrodes(recording=recording1, nwbfile=nwbfile)
add_electrodes(recording=recording2, nwbfile=nwbfile)

nwbfile.electrodes.to_dataframe()

```

The output is:
![image](https://github.com/user-attachments/assets/96178ef4-d5e0-41f5-89a1-bb7f3c68419a)

but the new property should be on second group.

PR #983, opened 2024-08-06
[Refactor] Centralize logic of global electrode id 

h-mayorquin:
Refactor of `add_electrodes_table`. I will push fixes for #958 and #982 with this.

Basically, I am defining the mapping betweeen the recording and the electrodes table in a centralized place instead of having it decoupled in two different places as we have in main.

There is nothing new added. I moved one test a bit and that shows in the diff but nothing is new.


h-mayorquin:
Uh, some of the failing tests here are passing locally for me. This is strange.

h-mayorquin:
OK, I discovered what I think is failing here:

https://github.com/catalystneuro/neuroconv/blob/31fca672d8836b25634e9cdb753828df0eec754f/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglx_utils.py#L18-L23

Sometimes the property "shank electrode number"  is int and sometimes the property is str. That does no texplain the odd pattern of failing with python version but should be fix.

h-mayorquin:
OK, this is ready, it was an condition inside and if loop that I did not handle well.

pyproject.toml
```diff
@@ -30,8 +30,8 @@ classifiers = [
 ]
 requires-python = ">=3.9"
 dependencies = [
-    "numpy>=1.22.0; python_version <= '3.11'",
-    "numpy>=1.26.0; python_version >= '3.12'",
+    "numpy>=1.22.0, <2.0.0; python_version <= '3.11'",
```
CodyCBakerPhD, line 33: What is status on 2.0 support?

CodyCBakerPhD (approved):

pyproject.toml
h-mayorquin: I think pynwb recently added it, right? I added it to most of spikeinterface but just being cautios here.

CodyCBakerPhD: I thought I heard they fixed it

PR #986, opened 2024-08-07
Improve spikeglx metadata

h-mayorquin:
In place of https://github.com/catalystneuro/neuroconv/pull/977
- Add shank id when available
- Add contact ids which contain a reference to the shank when it exists
- Remove references to shank when the probe does not have shanks in the probeinterface sense.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 * The `OpenEphysBinaryRecordingInterface` now uses `lxml` for extracting the session start time from the settings.xml file and does not depend on `pyopenephys` anymore. [PR #971](https://github.com/catalystneuro/neuroconv/pull/971)
 * Swap the majority of package setup and build steps to `pyproject.toml` instead of `setup.py`. [PR #955](https://github.com/catalystneuro/neuroconv/pull/955)
 * The `DeeplabcutInterface` now skips inferring timestamps from movie when timestamps are specified, running faster. [PR #967](https://github.com/catalystneuro/neuroconv/pull/967)
+* Improve metadata wirintg to SpikeGLX data interface. Added contact ids, shank ids and remove references to shanks for neuropixels 1.0. Also deprecated the previous neuroconv exclusive property "electrode_shank_number` [PR #986](https://github.com/catalystneuro/neuroconv/pull/986)
+
 
 ### Dependencies
 * Updated `spikeinterface` to `v0.101.0` to use `SortingAnalyzer` instead of `WaveformExtractor` (discontinued) in the `spikeinterface` tools
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * The `OpenEphysBinaryRecordingInterface` now uses `lxml` for extracting the session start time from the settings.xml file and does not depend on `pyopenephys` anymore. [PR #971](https://github.com/catalystneuro/neuroconv/pull/971)
 * Swap the majority of package setup and build steps to `pyproject.toml` instead of `setup.py`. [PR #955](https://github.com/catalystneuro/neuroconv/pull/955)
 * The `DeeplabcutInterface` now skips inferring timestamps from movie when timestamps are specified, running faster. [PR #967](https://github.com/catalystneuro/neuroconv/pull/967)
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 * The `OpenEphysBinaryRecordingInterface` now uses `lxml` for extracting the session start time from the settings.xml file and does not depend on `pyopenephys` anymore. [PR #971](https://github.com/catalystneuro/neuroconv/pull/971)
 * Swap the majority of package setup and build steps to `pyproject.toml` instead of `setup.py`. [PR #955](https://github.com/catalystneuro/neuroconv/pull/955)
 * The `DeeplabcutInterface` now skips inferring timestamps from movie when timestamps are specified, running faster. [PR #967](https://github.com/catalystneuro/neuroconv/pull/967)
+* Improve metadata wirintg to SpikeGLX data interface. Added contact ids, shank ids and remove references to shanks for neuropixels 1.0. Also deprecated the previous neuroconv exclusive property "electrode_shank_number` [PR #986](https://github.com/catalystneuro/neuroconv/pull/986)
```
CodyCBakerPhD, line 10:
```suggestion
* Improve metadata writing for SpikeGLX data interface. Added contact ids, shank ids and, remove references to shanks for neuropixels 1.0. Also deprecated the previous neuroconv exclusive property "electrode_shank_number` [PR #986](https://github.com/catalystneuro/neuroconv/pull/986)
```

Surprised codespell did not complain

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglx_utils.py
```diff
@@ -105,6 +107,7 @@ def get_device_metadata(meta) -> dict:
             "21": "NP2.0(1-shank)",
             "24": "NP2.0(4-shank)",
             "1030": "NP1.0 NHP",
+            "2013": "NP2.0(4-shank)",
```
CodyCBakerPhD, line 110: Is there any example data on GIN we can use to test this one?

h-mayorquin: Not sure.

The soluton that we went for here of building a local table is bad. We should do this from probeinterface anyway.

CHANGELOG.md
h-mayorquin: Thanks!

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglx_utils.py
CodyCBakerPhD: It would be better at ProbeInterface level, yes

CodyCBakerPhD: Would be nice to have test data for this shank logic, just like how I added the multi probe data to demonstrate the new device adding logic

h-mayorquin: Let me remove this line as it is totally incidental.

h-mayorquin: Ok, moved this to another PR.

https://github.com/catalystneuro/neuroconv/pull/990

CodyCBakerPhD: I'm saying it would be fine to include it this line if you also add multishank specific tests on multishank example data, to prove your improvements to the shank metadata and leave of record of what we expect that metadata to look like

Right now it only proves that the single-shank data doesn't include the unitary shank details per probe

Basically asking for tests just like what we did for the multi probe case, but for multi shank instead

Then we can say and prove SpikeGLX metadata is in a good place

h-mayorquin: We already have tests for multishank neuropixel 2.0 here. It was a good suggestion as I added tests for the new added attributes.

https://github.com/catalystneuro/neuroconv/blob/5ddffc6ff0eb24b12f594cd323bc9d7a71c197b5/tests/test_on_data/test_metadata/test_spikeglx_metadata.py#L27-L51

CodyCBakerPhD (approved):

PR #987, opened 2024-08-07
Make Bruker converters `add_to_nwb` and `run_conversion` consistent with all the other converters

h-mayorquin:
Converters core methods `add_to_nwb` should be consistent. This PR deprecates the kwargs mechanism of the bruker converters and substitutes it for passing a dictionary as in all the other cases.

This will avoid the need to do introspection in https://github.com/catalystneuro/neuroconv/pull/979


h-mayorquin:
This is not going to be good for converters that a lot of interfaces (altought I think that should not happen) so we should provide some utility function for that.

weiglszonja:
> This is not going to be good for converters that a lot of interfaces (altought I think that should not happen) so we should provide some utility function for that.

Can you elaborate this? I'm not sure I understand the issue enough here

PR #988, opened 2024-08-07
Delay import of DandiIcephysMetadata

h-mayorquin:
Imports should be done at __init__.

This is better for performance.

### DIFF
--- a/src/neuroconv/datainterfaces/icephys/baseicephysinterface.py
+++ b/src/neuroconv/datainterfaces/icephys/baseicephysinterface.py
@@ -11,14 +11,6 @@ from ...utils import (
     get_schema_from_method_signature,
 )
 
-try:
-    from ndx_dandi_icephys import DandiIcephysMetadata
-
-    HAVE_NDX_DANDI_ICEPHYS = True
-except ImportError:
-    DandiIcephysMetadata = None
-    HAVE_NDX_DANDI_ICEPHYS = False
-
 
 class BaseIcephysInterface(BaseExtractorInterface):
     """Primary class for all intracellular NeoInterfaces."""
@@ -33,6 +25,16 @@ class BaseIcephysInterface(BaseExtractorInterface):
         return source_schema
 
     def __init__(self, file_paths: list):
+
+        try:
+            from ndx_dandi_icephys import DandiIcephysMetadata
+
+            self.DandiIcephysMetadata = DandiIcephysMetadata
+            self.HAVE_NDX_DANDI_ICEPHYS = True
+        except ImportError:
+            self.DandiIcephysMetadata = None
+            self.HAVE_NDX_DANDI_ICEPHYS = False
+
         from ...tools.neo import get_number_of_electrodes, get_number_of_segments
 
         super().__init__(file_paths=file_paths)
@@ -49,8 +51,8 @@ class BaseIcephysInterface(BaseExtractorInterface):
 
     def get_metadata_schema(self) -> dict:
         metadata_schema = super().get_metadata_schema()
-        if DandiIcephysMetadata:
-            metadata_schema["properties"]["ndx-dandi-icephys"] = get_schema_from_hdmf_class(DandiIcephysMetadata)
+        if self.DandiIcephysMetadata:
+            metadata_schema["properties"]["ndx-dandi-icephys"] = get_schema_from_hdmf_class(self.DandiIcephysMetadata)
         metadata_schema["properties"]["Icephys"] = get_metadata_schema_for_icephys()
         return metadata_schema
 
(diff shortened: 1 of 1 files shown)


src/neuroconv/datainterfaces/icephys/baseicephysinterface.py
```diff
@@ -33,6 +25,16 @@ def get_source_schema(cls) -> dict:
         return source_schema
 
     def __init__(self, file_paths: list):
+
+        try:
+            from ndx_dandi_icephys import DandiIcephysMetadata
```
CodyCBakerPhD, line 30: While we're here, can we replace the antiquated try/except block with `if < importlib? >.find_spec`?

h-mayorquin: I am lazy to do it but is the right way.

h-mayorquin: OK, changed to an importlib findspec based solution.

src/neuroconv/datainterfaces/icephys/baseicephysinterface.py
```diff
@@ -33,6 +26,17 @@ def get_source_schema(cls) -> dict:
         return source_schema
 
     def __init__(self, file_paths: list):
+        # Check if the ndx_dandi_icephys module is available
+        dandi_icephys_spec = importlib.util.find_spec("ndx_dandi_icephys")
+        if dandi_icephys_spec is not None:
+            dandi_icephys_module = importlib.util.module_from_spec(dandi_icephys_spec)
+            dandi_icephys_spec.loader.exec_module(dandi_icephys_module)
+            self.DandiIcephysMetadata = getattr(dandi_icephys_module, "DandiIcephysMetadata")
```
CodyCBakerPhD, line 34:
```suggestion
            from ndx_dandi_icephys import DandiIcephysMetadata
```

I mean just to determine if the package is installed; classic import syntax is cleaner than the rest once you know it's safe to call

src/neuroconv/datainterfaces/icephys/baseicephysinterface.py
```diff
@@ -33,6 +26,17 @@ def get_source_schema(cls) -> dict:
...
+            self.HAVE_NDX_DANDI_ICEPHYS = True
+        else:
+            self.DandiIcephysMetadata = None
+            self.HAVE_NDX_DANDI_ICEPHYS = False
```
CodyCBakerPhD, line 38: Curious what you would think about setting these as the default at the class level (next to `ExtractorModuleName`)?

src/neuroconv/datainterfaces/icephys/baseicephysinterface.py
h-mayorquin: Sure.

src/neuroconv/datainterfaces/icephys/baseicephysinterface.py
h-mayorquin: I don't like it really. They are determined at import times, how we could make attributes of them?

CodyCBakerPhD (approved):

src/neuroconv/datainterfaces/icephys/baseicephysinterface.py
```diff
@@ -33,6 +26,15 @@ def get_source_schema(cls) -> dict:
         return source_schema
 
     def __init__(self, file_paths: list):
+        # Check if the ndx_dandi_icephys module is available
+        dandi_icephys_spec = importlib.util.find_spec("ndx_dandi_icephys")
+        if dandi_icephys_spec is not None:
+
```
CodyCBakerPhD, line 32:
```suggestion
        if dandi_icephys_spec is not None:
            from ndx_dandi_icephys import DandiIcephysMetadata
            
            self.DandiIcephysMetadata = DandiIcephysMetadata

```

src/neuroconv/datainterfaces/icephys/baseicephysinterface.py
```diff
@@ -49,8 +51,8 @@ def __init__(self, file_paths: list):
 
     def get_metadata_schema(self) -> dict:
         metadata_schema = super().get_metadata_schema()
-        if DandiIcephysMetadata:
-            metadata_schema["properties"]["ndx-dandi-icephys"] = get_schema_from_hdmf_class(DandiIcephysMetadata)
+        if self.DandiIcephysMetadata:
```
CodyCBakerPhD, line 54:
```suggestion
        if self.DandiIcephysMetadata is not None:
```

src/neuroconv/datainterfaces/icephys/baseicephysinterface.py
```diff
@@ -109,11 +111,11 @@ def add_to_nwbfile(
             nwbfile = make_nwbfile_from_metadata(metadata)
 
         if (
-            HAVE_NDX_DANDI_ICEPHYS
+            self.HAVE_NDX_DANDI_ICEPHYS
```
CodyCBakerPhD, line 114:
```suggestion
            self.HAVE_NDX_DANDI_ICEPHYS is not None
```

src/neuroconv/datainterfaces/icephys/baseicephysinterface.py
h-mayorquin: Thanks for fixing this. Went to eat and then got into a meeting.

PR #989, opened 2024-08-07
Drop casting to int in units table and introduce fine control over null values in `units_table`

h-mayorquin:
Should close #https://github.com/catalystneuro/neuroconv/issues/960 should come after https://github.com/catalystneuro/neuroconv/pull/985


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -1144,12 +1144,16 @@ def add_units_table(
         Waveform standard deviation for each unit. Shape: (num_units, num_samples, num_channels).
     unit_electrode_indices : list of lists of int, optional
         For each unit, a list of electrode indices corresponding to waveform data.
+    null_values_for_properties: dict
```
bendichter, line 1147:
```suggestion
    null_values_for_properties: dict, optional
```

tests/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -1487,20 +1546,22 @@ def test_write_recordingless(self):
...
-            sorting_analyzer=self.single_segment_analyzer,
-            nwbfile_path=self.nwbfile_path,
-            write_electrical_series=True,
-            metadata=metadata,
-        )
-        with NWBHDF5IO(self.nwbfile_path, "r") as io:
-            nwbfile = io.read()
-            self._test_analyzer_write(self.single_segment_analyzer, nwbfile)
-            self.assertIn("ElectricalSeriesRaw", nwbfile.acquisition)
+    # def test_write_sorting_analyzer_to_file(self):
```
CodyCBakerPhD, line 1549: What's the plan here?

h-mayorquin: Move those to `add_methods`. There is an error there that I can't really pin down but the test is too wide for me to figure it out.

CodyCBakerPhD: So to clarify, your source code changes here break a current test for `write_sorting_analyzer`, but we don't know why so the plan is to comment it out and debug separately?

h-mayorquin: Exactly. I will solve that after The function write_sorting_analyzer is doing too many things. I want to break appart those tests to the format of the other adds where we test the construction of the nwbfile separated from the wriging and not all bundled up like we do here.

CodyCBakerPhD (approved):

PR #990, opened 2024-08-07
Get default description for neuropixels probes 

h-mayorquin:
Coming from #986.

As discussed there there are many more probe types that we don't have:

https://github.com/SpikeInterface/probeinterface/blob/3728499714f68cfadaf76ef51b19eb8cc813b011/src/probeinterface/io.py#L1081-L1108

This is a temporary patch so the interfaces does not stop in its tracks when this information is missing.

Eventually, we should fetch this from probeinterface.

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglx_utils.py
+++ b/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglx_utils.py
@@ -107,7 +107,7 @@ def get_device_metadata(meta) -> dict:
             "1030": "NP1.0 NHP",
         }
         probe_type = str(meta["imDatPrb_type"])
-        probe_type_description = probe_type_to_probe_description.get(probe_type, "")
+        probe_type_description = probe_type_to_probe_description.get(probe_type, "Unknown SpikeGLX probe type.")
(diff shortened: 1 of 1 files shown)


src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglx_utils.py
```diff
@@ -107,7 +107,7 @@ def get_device_metadata(meta) -> dict:
             "1030": "NP1.0 NHP",
         }
         probe_type = str(meta["imDatPrb_type"])
-        probe_type_description = probe_type_to_probe_description[probe_type]
+        probe_type_description = probe_type_to_probe_description.get(probe_type, "")
```
CodyCBakerPhD, line 110:
```suggestion
        probe_type_description = probe_type_to_probe_description.get(probe_type, "Unknown SpikeGLX probe type.")
```

How about this? Missing or empty descriptions are a best practice violation, and this at least indicates that we tried to look it up

h-mayorquin: Sounds better! I will try to extract this from probeinterface as soon as I can : )

CodyCBakerPhD (approved):

PR #1003, opened 2024-08-13
Repack Nwb Files

### LINKED ISSUE #892: [Feature]: repack NWB file

It would be nice to have a workflow that could take in an NWB file that has already been saved to disk and repack it with recommended chunking and compression.

The first step would be to fetch the current backend configuration from existing Datasets. Maybe this could be a function in `_dataset_configuration.py`:

`get_existing_backend_configuration(nwbfile) -> BackendConfiguration`

where the nwbfile must be linked to an on-disk NWB File. The backend should be automatically extracted, so no need to have that as a separate arg.

Then we should have a way to get the recommended configuration for that file. This works already in some cases with `get_default_backend_configuration(nwbfile, backend)`, but does not work in all cases. If you have an `ImageSeries` with an external file and a (0,0,0) dataset, this triggers an error when the dataset in an h5py Dataset:

```python
from neuroconv.tools.nwb_helpers import get_default_backend_configuration
from pynwb.image import ImageSeries

nwbfile = mock_NWBFile()

im_series = ImageSeries(
    name="my_video",
    external_file=["my_video.mp4"],
    starting_frame=[0],
    format="external",
    rate=30.0
)
nwbfile.add_acquisition(im_series)

from pynwb import NWBHDF5IO

with NWBHDF5IO("this_test4.nwb", "w") as io:
    io.write(nwbfile)

io = NWBHDF5IO("this_test4.nwb", "r")
nwbfile = io.read()

get_default_backend_configuration(nwbfile, "hdf5")
```

```pytb
ZeroDivisionError                         Traceback (most recent call last)
Cell In[22], line 1
----> 1 get_default_backend_configuration(nwbfile, "hdf5")

File ~/dev/neuroconv/src/neuroconv/tools/nwb_helpers/_backend_configuration.py:19, in get_default_backend_configuration(nwbfile, backend)
     16 """Fill a default backend configuration to serve as a starting point for further customization."""
     18 BackendConfigurationClass = BACKEND_CONFIGURATIONS[backend]
---> 19 return BackendConfigurationClass.from_nwbfile(nwbfile=nwbfile)

File ~/dev/neuroconv/src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py:61, in BackendConfiguration.from_nwbfile(cls, nwbfile)
     58 @classmethod
     59 def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
     60     default_dataset_configurations = get_default_dataset_io_configurations(nwbfile=nwbfile, backend=cls.backend)
---> 61     dataset_configurations = {
     62         default_dataset_configuration.location_in_file: default_dataset_configuration
     63         for default_dataset_configuration in default_dataset_configurations
     64     }
     66     return cls(dataset_configurations=dataset_configurations)

File ~/dev/neuroconv/src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py:61, in <dictcomp>(.0)
     58 @classmethod
     59 def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
     60     default_dataset_configurations = get_default_dataset_io_configurations(nwbfile=nwbfile, backend=cls.backend)
---> 61     dataset_configurations = {
     62         default_dataset_configuration.location_in_file: default_dataset_configuration
     63         for default_dataset_configuration in default_dataset_configurations
     64     }
     66     return cls(dataset_configurations=dataset_configurations)

File ~/dev/neuroconv/src/neuroconv/tools/nwb_helpers/_dataset_configuration.py:154, in get_default_dataset_io_configurations(nwbfile, backend)
    151 if isinstance(candidate_dataset, np.ndarray) and candidate_dataset.size == 0:
    152     continue
--> 154 dataset_io_configuration = DatasetIOConfigurationClass.from_neurodata_object(
    155     neurodata_object=neurodata_object, dataset_name=known_dataset_field
    156 )
    158 yield dataset_io_configuration

File ~/dev/neuroconv/src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py:272, in DatasetIOConfiguration.from_neurodata_object(cls, neurodata_object, dataset_name)
    270     compression_method = "gzip"
    271 elif dtype != np.dtype("object"):
--> 272     chunk_shape = SliceableDataChunkIterator.estimate_default_chunk_shape(
    273         chunk_mb=10.0, maxshape=full_shape, dtype=np.dtype(dtype)
    274     )
    275     buffer_shape = SliceableDataChunkIterator.estimate_default_buffer_shape(
    276         buffer_gb=0.5, chunk_shape=chunk_shape, maxshape=full_shape, dtype=np.dtype(dtype)
    277     )
    278     compression_method = "gzip"

File ~/dev/neuroconv/src/neuroconv/tools/hdmf.py:38, in GenericDataChunkIterator.estimate_default_chunk_shape(chunk_mb, maxshape, dtype)
     35 chunk_bytes = chunk_mb * 1e6
     37 min_maxshape = min(maxshape)
---> 38 v = tuple(math.floor(maxshape_axis / min_maxshape) for maxshape_axis in maxshape)
     39 prod_v = math.prod(v)
     40 while prod_v * itemsize > chunk_bytes and prod_v != 1:

File ~/dev/neuroconv/src/neuroconv/tools/hdmf.py:38, in <genexpr>(.0)
     35 chunk_bytes = chunk_mb * 1e6
     37 min_maxshape = min(maxshape)
---> 38 v = tuple(math.floor(maxshape_axis / min_maxshape) for maxshape_axis in maxshape)
     39 prod_v = math.prod(v)
     40 while prod_v * itemsize > chunk_bytes and prod_v != 1:

ZeroDivisionError: division by zero
```

We should either adjust this function so it works for this purpose, or create a different function for this specific purpose.

Then, finally, we would need a way to write this to a new file, probably using the `export` function in pynwb.

It would be nice to have two usage modes, one that completely automates everything, and one that would allow users to repack specific datasets with specific parameters.

This workflow should also allow the user to switch from one backend to another.

It's somewhat common for other users to upload sub-optimal NWB files. This would also be a suitable workflow for when users create NWB Files in MatNWB and don't know how to configure the datasets properly there.

No.

### Code of Conduct

pauladkisson:
Fixes #892

### DIFF
--- a/src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py
+++ b/src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py
@@ -3,9 +3,11 @@
 from typing import Any, Dict, Literal, Union
 
 import h5py
+from hdmf import Container
 from pydantic import Field, InstanceOf
+from typing_extensions import Self
 
-from ._base_dataset_io import DatasetIOConfiguration
+from ._base_dataset_io import DatasetIOConfiguration, _find_location_in_memory_nwbfile
 from ...importing import is_package_installed
 
 _base_hdf5_filters = set(h5py.filters.decode)
@@ -78,3 +80,28 @@ class HDF5DatasetIOConfiguration(DatasetIOConfiguration):
             compression_bundle = dict(compression=self.compression_method, compression_opts=compression_opts)
 
         return dict(chunks=self.chunk_shape, **compression_bundle)
+
+    @classmethod
+    def from_existing_neurodata_object(
+        cls, neurodata_object: Container, dataset_name: Literal["data", "timestamps"]
+    ) -> Self:
+        location_in_file = _find_location_in_memory_nwbfile(neurodata_object=neurodata_object, field_name=dataset_name)
+        full_shape = getattr(neurodata_object, dataset_name).shape
+        dtype = getattr(neurodata_object, dataset_name).dtype
+        chunk_shape = getattr(neurodata_object, dataset_name).chunks
+        buffer_shape = getattr(neurodata_object, dataset_name).maxshape
+        compression_method = getattr(neurodata_object, dataset_name).compression
+        compression_opts = getattr(neurodata_object, dataset_name).compression_opts
+        compression_options = dict(compression_opts=compression_opts)
+        return cls(
+            object_id=neurodata_object.object_id,
+            object_name=neurodata_object.name,
+            location_in_file=location_in_file,
+            dataset_name=dataset_name,
+            full_shape=full_shape,
+            dtype=dtype,
+            chunk_shape=chunk_shape,
+            buffer_shape=buffer_shape,
+            compression_method=compression_method,
+            compression_options=compression_options,
+        )
--- a/src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
+++ b/src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
@@ -172,3 +172,63 @@ def get_default_dataset_io_configurations(
                 )
 
                 yield dataset_io_configuration
+
+
+def get_existing_dataset_io_configurations(
+    nwbfile: NWBFile,
+    backend: Literal["hdf5", "zarr"],
+) -> Generator[DatasetIOConfiguration, None, None]:
+
+    DatasetIOConfigurationClass = DATASET_IO_CONFIGURATIONS[backend]
+
+    known_dataset_fields = ("data", "timestamps")
+    for neurodata_object in nwbfile.objects.values():
+        if isinstance(neurodata_object, DynamicTable):
+            dynamic_table = neurodata_object  # For readability
+
+            for column in dynamic_table.columns:
+                candidate_dataset = column.data  # VectorData object
+
+                # Skip over columns whose values are links, such as the 'group' of an ElectrodesTable
+                if any(isinstance(value, Container) for value in candidate_dataset):
+                    continue  # Skip
+
+                # Skip when columns whose values are a reference type
+                if isinstance(column, TimeSeriesReferenceVectorData):
+                    continue
+
+                # Skip datasets with any zero-length axes
+                dataset_name = "data"
+                candidate_dataset = getattr(column, dataset_name)
+                full_shape = get_data_shape(data=candidate_dataset)
+                if any(axis_length == 0 for axis_length in full_shape):
+                    continue
+
+                dataset_io_configuration = DatasetIOConfigurationClass.from_existing_neurodata_object(
+                    neurodata_object=column, dataset_name=dataset_name
+                )
+
+                yield dataset_io_configuration
+        elif isinstance(neurodata_object, NWBContainer):
+            for known_dataset_field in known_dataset_fields:
+                # Skip optional fields that aren't present
+                if known_dataset_field not in neurodata_object.fields:
+                    continue
+
+                candidate_dataset = getattr(neurodata_object, known_dataset_field)
+
+                # Skip edge case of in-memory ImageSeries with external mode; data is in fields and is empty array
+                if isinstance(candidate_dataset, np.ndarray) and candidate_dataset.size == 0:
+                    continue
+
+                # Skip datasets with any zero-length axes
+                candidate_dataset = getattr(neurodata_object, known_dataset_field)
+                full_shape = get_data_shape(data=candidate_dataset)
+                if any(axis_length == 0 for axis_length in full_shape):
+                    continue
+
+                dataset_io_configuration = DatasetIOConfigurationClass.from_existing_neurodata_object(
+                    neurodata_object=neurodata_object, dataset_name=known_dataset_field
+                )
+
+                yield dataset_io_configuration
--- a/temp_test.py
+++ b/temp_test.py
@@ -1,0 +1,61 @@
+import shutil
+from datetime import datetime
+from pathlib import Path
+
+from lerner_lab_to_nwb.seiler_2024.seiler_2024_convert_session import session_to_nwb
+
+
+def main():
+    # Parameters for conversion
+    data_dir_path = Path("/Volumes/T7/CatalystNeuro/Lerner/raw_data")
+    output_dir_path = Path("/Volumes/T7/CatalystNeuro/Lerner/conversion_nwb")
+    stub_test = False
+
+    if output_dir_path.exists():
+        shutil.rmtree(
+            output_dir_path, ignore_errors=True
+        )  # ignore errors due to MacOS race condition (https://github.com/python/cpython/issues/81441)
+
+    # Fiber Photometry session
+    experiment_type = "FP"
+    experimental_group = "PS"
+    subject_id = "112.283"
+    start_datetime = datetime(2019, 6, 20, 9, 32, 4)
+    session_conditions = {
+        "Start Date": start_datetime.strftime("%m/%d/%y"),
+        "Start Time": start_datetime.strftime("%H:%M:%S"),
+    }
+    start_variable = "Start Date"
+    behavior_file_path = (
+        data_dir_path
+        / f"{experiment_type} Experiments"
+        / "Behavior"
+        / f"{experimental_group}"
+        / f"{subject_id}"
+        / f"{subject_id}"
+    )
+    fiber_photometry_folder_path = (
+        data_dir_path
+        / f"{experiment_type} Experiments"
+        / "Photometry"
+        / f"Punishment Sensitive"
+        / f"Early RI60"
+        / f"Photo_{subject_id.split('.')[0]}_{subject_id.split('.')[1]}-190620-093542"
+    )
+    session_to_nwb(
+        data_dir_path=data_dir_path,
+        output_dir_path=output_dir_path,
+        behavior_file_path=behavior_file_path,
+        fiber_photometry_folder_path=fiber_photometry_folder_path,
+        has_demodulated_commanded_voltages=False,
+        subject_id=subject_id,
+        session_conditions=session_conditions,
+        start_variable=start_variable,
+        experiment_type=experiment_type,
+        experimental_group=experimental_group,
+        stub_test=stub_test,
+    )
+
+
+if __name__ == "__main__":
+    main()
--- a/temp_test.py
+++ b/temp_test.py
@@ -5,8 +5,8 @@ import numpy as np
 from pynwb import NWBHDF5IO, H5DataIO, TimeSeries
 from pynwb.testing.mock.file import mock_NWBFile
 
-from neuroconv.tools.nwb_helpers._configuration_models._hdf5_dataset_io import (
-    HDF5DatasetIOConfiguration,
+from neuroconv.tools.nwb_helpers._dataset_configuration import (
+    get_existing_dataset_io_configurations,
 )
 
 
@@ -21,7 +21,9 @@ def write_nwbfile(nwbfile_path: Path):
         description="an example time series",
         data=H5DataIO(data=data, compression="gzip", chunks=(1,), compression_opts=2),
         unit="m",
-        timestamps=timestamps,
+        timestamps=H5DataIO(
+            timestamps, compression="gzip", chunks=(1,), compression_opts=2
+        ),  # TODO: add support for uncompressed timestamps
     )
     nwbfile.add_acquisition(time_series_with_timestamps)
     with NWBHDF5IO(nwbfile_path, mode="w") as io:
@@ -33,15 +35,9 @@ def main():
     write_nwbfile(nwbfile_path)
     with NWBHDF5IO(nwbfile_path, mode="r") as io:
         nwbfile = io.read()
-        for neurodata_object in nwbfile.objects.values():
-            print(neurodata_object.name)
-            if isinstance(neurodata_object, TimeSeries):
-                config = HDF5DatasetIOConfiguration.from_existing_neurodata_object(
-                    neurodata_object=neurodata_object, dataset_name="data"
-                )
-                print(f"{config.chunk_shape = }")
-                print(f"{config.compression_method = }")
-                print(f"{config.compression_options = }")
+        existing_dataset_io_configurations = get_existing_dataset_io_configurations(nwbfile, backend="hdf5")
+        for dataset_io_configuration in existing_dataset_io_configurations:
+            print(dataset_io_configuration)
 
 
 if __name__ == "__main__":
--- a/temp_test.py
+++ b/temp_test.py
@@ -1,60 +1,47 @@
-import shutil
-from datetime import datetime
+import os
 from pathlib import Path
 
-from lerner_lab_to_nwb.seiler_2024.seiler_2024_convert_session import session_to_nwb
+import numpy as np
+from pynwb import NWBHDF5IO, H5DataIO, TimeSeries
+from pynwb.testing.mock.file import mock_NWBFile
+
+from neuroconv.tools.nwb_helpers._configuration_models._hdf5_dataset_io import (
+    HDF5DatasetIOConfiguration,
+)
+
+
+def write_nwbfile(nwbfile_path: Path):
+    if nwbfile_path.exists():
+        os.remove(nwbfile_path)
+    nwbfile = mock_NWBFile()
+    timestamps = np.arange(10.0)
+    data = np.arange(100, 200, 10)
+    time_series_with_timestamps = TimeSeries(
+        name="test_timeseries",
+        description="an example time series",
+        data=H5DataIO(data=data, compression="gzip", chunks=(1,), compression_opts=2),
+        unit="m",
+        timestamps=timestamps,
+    )
+    nwbfile.add_acquisition(time_series_with_timestamps)
+    with NWBHDF5IO(nwbfile_path, mode="w") as io:
+        io.write(nwbfile)
 
 
 def main():
-    # Parameters for conversion
-    data_dir_path = Path("/Volumes/T7/CatalystNeuro/Lerner/raw_data")
-    output_dir_path = Path("/Volumes/T7/CatalystNeuro/Lerner/conversion_nwb")
-    stub_test = False
-
-    if output_dir_path.exists():
-        shutil.rmtree(
-            output_dir_path, ignore_errors=True
-        )  # ignore errors due to MacOS race condition (https://github.com/python/cpython/issues/81441)
-
-    # Fiber Photometry session
-    experiment_type = "FP"
-    experimental_group = "PS"
-    subject_id = "112.283"
-    start_datetime = datetime(2019, 6, 20, 9, 32, 4)
-    session_conditions = {
-        "Start Date": start_datetime.strftime("%m/%d/%y"),
-        "Start Time": start_datetime.strftime("%H:%M:%S"),
-    }
-    start_variable = "Start Date"
-    behavior_file_path = (
-        data_dir_path
-        / f"{experiment_type} Experiments"
-        / "Behavior"
-        / f"{experimental_group}"
-        / f"{subject_id}"
-        / f"{subject_id}"
-    )
-    fiber_photometry_folder_path = (
-        data_dir_path
-        / f"{experiment_type} Experiments"
-        / "Photometry"
-        / f"Punishment Sensitive"
-        / f"Early RI60"
-        / f"Photo_{subject_id.split('.')[0]}_{subject_id.split('.')[1]}-190620-093542"
-    )
-    session_to_nwb(
-        data_dir_path=data_dir_path,
-        output_dir_path=output_dir_path,
-        behavior_file_path=behavior_file_path,
-        fiber_photometry_folder_path=fiber_photometry_folder_path,
-        has_demodulated_commanded_voltages=False,
-        subject_id=subject_id,
-        session_conditions=session_conditions,
-        start_variable=start_variable,
-        experiment_type=experiment_type,
-        experimental_group=experimental_group,
-        stub_test=stub_test,
-    )
+    nwbfile_path = Path("/Volumes/T7/CatalystNeuro/temp.nwb")
+    write_nwbfile(nwbfile_path)
+    with NWBHDF5IO(nwbfile_path, mode="r") as io:
+        nwbfile = io.read()
+        for neurodata_object in nwbfile.objects.values():
+            print(neurodata_object.name)
+            if isinstance(neurodata_object, TimeSeries):
+                config = HDF5DatasetIOConfiguration.from_existing_neurodata_object(
+                    neurodata_object=neurodata_object, dataset_name="data"
+                )
+                print(f"{config.chunk_shape = }")
+                print(f"{config.compression_method = }")
+                print(f"{config.compression_options = }")
 
 
 if __name__ == "__main__":


src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py
```diff
@@ -78,3 +80,28 @@ def get_data_io_kwargs(self) -> Dict[str, Any]:
             compression_bundle = dict(compression=self.compression_method, compression_opts=compression_opts)
 
         return dict(chunks=self.chunk_shape, **compression_bundle)
+
+    @classmethod
+    def from_existing_neurodata_object(
```
CodyCBakerPhD, line 85:
```suggestion
    def from_neurodata_object(
```

IMO don't need the 'existing' part in the name. Evident enough that if it applies to a neurodata object, that object must already exist

However, it is important to note (and test appropriate skip conditions) the application to neurodata objects that are in-memory instead of written/read to/from disk, which can change the way a couple of things behave

src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py
```diff
@@ -78,3 +80,28 @@ def get_data_io_kwargs(self) -> Dict[str, Any]:
...
+        cls, neurodata_object: Container, dataset_name: Literal["data", "timestamps"]
+    ) -> Self:
+        location_in_file = _find_location_in_memory_nwbfile(neurodata_object=neurodata_object, field_name=dataset_name)
+        full_shape = getattr(neurodata_object, dataset_name).shape
+        dtype = getattr(neurodata_object, dataset_name).dtype
+        chunk_shape = getattr(neurodata_object, dataset_name).chunks
+        buffer_shape = getattr(neurodata_object, dataset_name).maxshape
+        compression_method = getattr(neurodata_object, dataset_name).compression
+        compression_opts = getattr(neurodata_object, dataset_name).compression_opts
+        compression_options = dict(compression_opts=compression_opts)
```
CodyCBakerPhD, line 95: Huh, but we already have the generic method name and code for applying to in-memory rather than on-disk: https://github.com/catalystneuro/neuroconv/blob/c33dfbfbb367bf978e99a85ce3130383233eeca4/src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py#L248

So what about making this a simpler router like so?

```

        if is_object_dataset_written_to_disk:
            location_in_file = _find_location_in_memory_nwbfile(neurodata_object=neurodata_object, field_name=dataset_name)
            full_shape = getattr(neurodata_object, dataset_name).shape
            dtype = getattr(neurodata_object, dataset_name).dtype
            chunk_shape = getattr(neurodata_object, dataset_name).chunks
            buffer_shape = getattr(neurodata_object, dataset_name).maxshape
            compression_method = getattr(neurodata_object, dataset_name).compression
            compression_opts = getattr(neurodata_object, dataset_name).compression_opts
            compression_options = dict(compression_opts=compression_opts)
        else:
            super().from_neurodata_object(...)
            
        return ...
```

temp_test.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def main():
+    nwbfile_path = Path("/Volumes/T7/CatalystNeuro/temp.nwb")
+    write_nwbfile(nwbfile_path)
+    with NWBHDF5IO(nwbfile_path, mode="r") as io:
+        nwbfile = io.read()
+        existing_dataset_io_configurations = get_existing_dataset_io_configurations(nwbfile, backend="hdf5")
+        for dataset_io_configuration in existing_dataset_io_configurations:
+            print(dataset_io_configuration)
```
CodyCBakerPhD, line 40: What are you thinking about how the API should look like for how to use this repack helper?

src/neuroconv/tools/nwb_helpers/_backend_configuration.py
```diff
@@ -17,3 +18,17 @@ def get_default_backend_configuration(
...
+
+
+def get_existing_backend_configuration(nwbfile: NWBFile) -> Union[HDF5BackendConfiguration, ZarrBackendConfiguration]:
+    """Fill an existing backend configuration to serve as a starting point for further customization."""
+
+    read_io = nwbfile.read_io
+    if isinstance(read_io, NWBHDF5IO):
+        backend = "hdf5"
+    elif isinstance(read_io, NWBZarrIO):
+        backend = "zarr"
```
bendichter, line 30: Very minor, but you could use this dictionary, which would allow this method to automatically support any new backends added to that dictionary.

https://github.com/catalystneuro/neuroconv/blob/40d786af408e34efaea66edb387fde3d4b8cf32f/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py#L22

In general, it's better to define this mapping in one single spot. (though you would need to reverse the dictionary first (`reversed_dict = {value: key for key, value in original_dict.items()}`)

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_default_backend_configuration.py
CodyCBakerPhD, line 1: Can you move these tests to a standalone location, such as `/tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_repack.py`?

To indicate how distinct this usage is intended to be from the other workflows (and to avoid accidental mixing of intended test cases)

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_default_backend_configuration.py
```diff
@@ -262,3 +263,22 @@ def test_complex_zarr(zarr_nwbfile_path):
...
+
+    with NWBHDF5IO("test.nwb", "w") as io:
+        io.write(nwbfile)
+
+    io = NWBHDF5IO("test.nwb", "r")
+    nwbfile = io.read()
+    print(nwbfile.acquisition["my_video"])
+
+    backend_config = get_default_backend_configuration(nwbfile, "hdf5")
+    print(backend_config)  # TODO: Figure out why this doesn't throw an error like Ben said it did
```
bendichter, line 284: Can we add an integration test where we write a new NWB file with a different backend configuration?

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -370,3 +375,51 @@ def configure_and_write_nwbfile(
...
+
+    IO = BACKEND_NWB_IO[backend_configuration.backend]
+    nwbfile.set_modified()
+    with IO(export_nwbfile_path, mode="w") as io:
+        io.export(nwbfile=nwbfile, src_io=nwbfile.read_io, write_args=dict(link_data=False))
+
+
+def repack_nwbfile(
+    *,
+    nwbfile: NWBFile,
```
CodyCBakerPhD, line 395: I think the signature of this function is intended to work on an input NWB file path (some `.nwb` file that has already been written to disk, albeit with non-ideal chunking/compression)

That's also why I suggested started by drafting test structure first to show how you 'expect' the API for the function to feel like before deciding on implementation strategy (though I am aware that both can inform each other in an iterative review process)

Basically, I think our goal is to make something analogous to `h5repack` but much more performant and with much better default behavior by setting more optimal chunk shapes and compression methods

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -370,3 +375,51 @@ def configure_and_write_nwbfile(
...
+    export_nwbfile_path: Path,
+    backend_configuration: BackendConfiguration,
+) -> None:
+    configure_backend(nwbfile=nwbfile, backend_configuration=backend_configuration)
+
+    IO = BACKEND_NWB_IO[backend_configuration.backend]
+    nwbfile.set_modified()
+    with IO(export_nwbfile_path, mode="w") as io:
+        io.export(nwbfile=nwbfile, src_io=nwbfile.read_io, write_args=dict(link_data=False))
```
CodyCBakerPhD, line 390: Can you explain what `link_data=False` does here?

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_default_backend_configuration.py
```diff
@@ -262,3 +263,22 @@ def test_complex_zarr(zarr_nwbfile_path):
 
 """
     assert stdout.getvalue() == expected_print
+
+
+def test_000_ImageSeries():
```
CodyCBakerPhD, line 268: I would like to see many more tests here; starting with more basic types like `TimeSeries`, `DynamicTable`, etc. before working our way up to the edge case that is an ImageSeries (external vs. internal mode too on that)

CodyCBakerPhD: Keep in mind the intended point of this function; to start with objects that are either uncompressed, unchuncked, or badly chunked; and then we create a new copy of the file that has better compression and chunking on all applicable datasets

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_default_backend_configuration.py
CodyCBakerPhD: Hmm... You mean something like 'make a Zarr-backend copy of this existing HDF5-backed NWB file'? Or same type of backend (such as both HDF5) but different configuration?

bendichter: good question. Both, I guess. I just want to make sure this works end-to-end and does not require us to do anything funky in practice

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
bendichter: Yes, that is what we want, but I do see an advantage of creating this API first and having the file reading in a separate function. Then we can do this reading from disk or reading from cloud without changing the internal logic of this function at all.

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
pauladkisson: `link_data=False` ensures that the cache of builders is cleared for each neurodata object so that they can be overwritten with the new backend information.

bendichter: It seems like the code is almost exactly the same as `configure_and_write_nwbfile`. Would it be better to include this in that function by including an `export` flag?

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
bendichter: I'm thinking something like:

```python
def repack_nwbfile(
    filepath_in: str,
    filepath_out: str
    backend_in: Literal[None, "hdf5", "zarr"] = None,  # if None, infer backend
    backend_out: Literal[None, "hdf5", "zarr"] = None,  # if None, infer from backend_configuration_changes. if that is None, use same as input
    use_default_configuration=True,  # if False, use existing backend
    backend_configuration_changes=None,
)
```

and then it would be great to also have a CLI for this. The backend_configuration_changes might be hard to specify with a CLI but we could try to do something analogous to h5repack

pauladkisson: Wouldn't we always want to infer the `backend_in`?

bendichter: yeah, probably. I was thinking there might be edge cases where it might be difficult to infer or where supplying it will allow us to skip the inference step, which might speed this up, but I think currently our backend inference should be able to work very fast and accurately so for now I think it would be fine to not include this

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -370,3 +375,46 @@ def configure_and_write_nwbfile(
...
+    IO = BACKEND_NWB_IO[backend]
+    with IO(nwbfile_path, mode="r") as io:
+        nwbfile = io.read()
+        if use_default_backend_configuration:
+            backend_configuration = get_default_backend_configuration(nwbfile=nwbfile, backend=backend)
+        else:
+            backend_configuration = get_existing_backend_configuration(nwbfile=nwbfile)
+        dataset_configurations = backend_configuration.dataset_configurations
+
+        backend_configuration_changes = backend_configuration_changes or dict()
```
bendichter, line 412: minor stylistic point: I would consider this line input parsing and would generally put it first in the function,- here, before the line `IO = BACKEND_NWB_IO[backend]`. Another reason is this line does not need to be in the context it is in, so it should be moved outside. If you can move a line of code outside of an indent, you usually should.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -56,11 +59,16 @@ def model_json_schema(cls, **kwargs) -> Dict[str, Any]:
         return super().model_json_schema(mode="validation", schema_generator=PureJSONSchemaGenerator, **kwargs)
 
     @classmethod
-    def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
-        default_dataset_configurations = get_default_dataset_io_configurations(nwbfile=nwbfile, backend=cls.backend)
+    def from_nwbfile(cls, nwbfile: NWBFile, mode: Literal["default", "existing"] = "default") -> Self:
```
bendichter, line 62: Are there more than 2 modes possible here? If not, I would prefer using a boolean flag here

src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py
```diff
@@ -78,3 +82,37 @@ def get_data_io_kwargs(self) -> dict[str, Any]:
...
+        if use_default_dataset_io_configuration:
+            return super().from_neurodata_object(neurodata_object=neurodata_object, dataset_name=dataset_name)
+
+        location_in_file = _find_location_in_memory_nwbfile(neurodata_object=neurodata_object, field_name=dataset_name)
+        full_shape = getattr(neurodata_object, dataset_name).shape
+        dtype = getattr(neurodata_object, dataset_name).dtype
+        chunk_shape = getattr(neurodata_object, dataset_name).chunks
+        buffer_chunk_shape = chunk_shape or full_shape
+        buffer_shape = SliceableDataChunkIterator.estimate_default_buffer_shape(
+            buffer_gb=0.5, chunk_shape=buffer_chunk_shape, maxshape=full_shape, dtype=np.dtype(dtype)
```
bendichter, line 102: buffer_gb should be settable

src/neuroconv/tools/nwb_helpers/_configuration_models/_zarr_dataset_io.py
```diff
@@ -130,3 +134,36 @@ def get_data_io_kwargs(self) -> dict[str, Any]:
             compressor = False
 
         return dict(chunks=self.chunk_shape, filters=filters, compressor=compressor)
+
+    @classmethod
+    def from_neurodata_object(
```
bendichter, line 139: there's a lot of repeated code here with the hdf5 version of this function. Would there be any way to reduce the repetition?

src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
```diff
@@ -172,3 +172,82 @@ def get_default_dataset_io_configurations(
                 )
 
                 yield dataset_io_configuration
+
+
+def get_existing_dataset_io_configurations(
+    nwbfile: NWBFile,
+    backend: Literal["hdf5", "zarr"],
```
bendichter, line 179: maybe we could make the default that you use the same backend as the original file

src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py
pauladkisson: This is how it's done in the base currently, so i think it should be addressed as a separate issue -- see here: #1299

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_default_backend_configuration.py
pauladkisson: > Can we add an integration test where we write a new NWB file with a different backend configuration?

Done!

src/neuroconv/tools/nwb_helpers/_configuration_models/_zarr_dataset_io.py
pauladkisson: Yep, added a staticmethod to the base class that should help de-duplicate.

bendichter:
Could you please add a chapter in the user guide?

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -44,14 +46,29 @@ def configure_backend(
 
         neurodata_object = nwbfile.objects[object_id]
         is_dataset_linked = isinstance(neurodata_object.fields.get(dataset_name), TimeSeries)
+        if isinstance(neurodata_object.fields.get(dataset_name), AbstractDataChunkIterator):
+            data_chunk_iterator_class = None
+            data_chunk_iterator_kwargs = dict()
+        else:
+            data_chunk_iterator_class = DataChunkIterator
+            data_chunk_iterator_kwargs = dict(buffer_size=np.prod(dataset_configuration.buffer_shape))
```
h-mayorquin, line 54: use math and not np.prod as that will overflow in most cases and screw up your buffer case. 

Are these changes related to repacking?

pauladkisson: > use math and not np.prod as that will overflow in most cases and screw up your buffer case.

Ok, will do.

> Are these changes related to repacking?

Yes, the DataChunkIterator solves this issue: https://github.com/hdmf-dev/hdmf/issues/1170

h-mayorquin: Ah, thanks, OK. so hdmf does not support re-setting the chunking without copying so we do this through an iterator.

Mmm, you are solving your problem and this is great but in this PR as well as deteting compound data types I kind of feel that hdmf is kicking its complexity to us for something that it should provide. I guess that we are faster : P

pauladkisson: Speaking of which, I ran into another issue with hdmf-zarr when it comes to compound dtypes -- see here: https://github.com/hdmf-dev/hdmf-zarr/issues/272

src/neuroconv/tools/nwb_helpers/_configuration_models/_zarr_dataset_io.py
```diff
@@ -127,3 +130,28 @@ def get_data_io_kwargs(self) -> dict[str, Any]:
...
+        cls,
+        neurodata_object: Container,
+        dataset_name: Literal["data", "timestamps"],
+        builder: BaseBuilder | None = None,
+        use_default_dataset_io_configuration: bool = True,
+    ) -> Self:
+        if use_default_dataset_io_configuration:
```
bendichter: add docstring please

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -298,7 +304,7 @@ def from_neurodata_object(
                 number_of_channels=number_of_channels, number_of_frames=number_of_frames, dtype=dtype
             )
 
-            buffer_shape = None  # This is the non-iterative path
+            buffer_shape = full_shape
```
h-mayorquin: why did we change this?

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -82,6 +82,9 @@ def _infer_dtype_of_list(list_: list[int | float | list]) -> np.dtype:
 def _infer_dtype(dataset: h5py.Dataset | zarr.Array) -> np.dtype:
     """Attempt to infer the dtype of the contained values of the dataset."""
     if hasattr(dataset, "dtype"):
+        if isinstance(dataset.dtype, list):
```
h-mayorquin: why do we have this change?

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -56,11 +59,31 @@ def model_json_schema(cls, **kwargs) -> dict[str, Any]:
         return super().model_json_schema(mode="validation", schema_generator=PureJSONSchemaGenerator, **kwargs)
 
     @classmethod
-    def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
-        default_dataset_configurations = get_default_dataset_io_configurations(nwbfile=nwbfile, backend=cls.backend)
+    def from_nwbfile(cls, nwbfile: NWBFile, use_default_dataset_io_configurations: bool = True) -> Self:
```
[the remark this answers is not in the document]
pauladkisson: There are two paths now to generate a `BackendConfiguration`: from default (old path which specifies recommended/default chunking and compression options for every neurodata_object) and from existing (new path that reads the existing chunking and compression options for every neurodata object). 

`use_default_dataset_io_configurations` specifies which path to take.

docs/user_guide/backend_configuration.rst
```diff
@@ -256,6 +284,85 @@ You can use the :py:meth:`~neuroconv.tools.nwb_helpers._configuration_models._ba
...
+By default, repack_nwbfile will use the default backend configuration,
+but you can also specify changes with an appropriately formatted dictionary.
+
+.. code-block:: python
+
+    repack_nwbfile(
+        nwbfile_path="uncompressed_nwbfile.nwb",
+        export_nwbfile_path="repacked_nwbfile.nwb",
+        backend="hdf5",
+        backend_configuration_changes={
```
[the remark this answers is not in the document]
pauladkisson: It works like a diff: yes. However, by default, repack_nwbfile will propagate the default backend configuration for all objects that aren't specified in backend_configuration_changes. If a user wants to make a targeted change to only a few of the objects, they can set the use_default_backend_configuration flag to false. See docs below.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
pauladkisson: Because `dataset.dtype` is not formatted properly for NumPy in the case of compound data types like pixel_mask. 

```python
dataset.dtype = ['uint32', 'uint32', 'float32']
data_type = dtype([('f0', '<u4'), ('f1', '<u4'), ('f2', '<f4')])
```

If, for example, I removed this change, I would get the following error: 
```python
src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py:493: in repack_nwbfile
    configure_and_write_nwbfile(
src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py:425: in configure_and_write_nwbfile
    configure_backend(nwbfile=nwbfile, backend_configuration=backend_configuration)
src/neuroconv/tools/nwb_helpers/_configure_backend.py:35: in configure_backend
    locations_to_remap = backend_configuration.find_locations_requiring_remapping(nwbfile=nwbfile)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py:119: in find_locations_requiring_remapping
    default_configurations = list(get_default_dataset_io_configurations(nwbfile=nwbfile, backend=self.backend))
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
src/neuroconv/tools/nwb_helpers/_dataset_configuration.py:145: in get_default_dataset_io_configurations
    dataset_io_configuration = DatasetIOConfigurationClass.from_neurodata_object(
src/neuroconv/tools/nwb_helpers/_configuration_models/_hdf5_dataset_io.py:110: in from_neurodata_object
    return super().from_neurodata_object(
src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py:290: in from_neurodata_object
    dtype = _infer_dtype(dataset=candidate_dataset)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

dataset = <hdmf.backends.hdf5.h5_utils.ContainerH5TableDataset object at 0x10c733890>

    def _infer_dtype(dataset: h5py.Dataset | zarr.Array) -> np.dtype:
        """Attempt to infer the dtype of the contained values of the dataset."""
        if hasattr(dataset, "dtype"):
            # if isinstance(dataset.dtype, list):
            #     data_type = np.dtype(", ".join(dataset.dtype))
            #     return data_type
>           data_type = np.dtype(dataset.dtype)
                        ^^^^^^^^^^^^^^^^^^^^^^^
E           TypeError: Field elements must be 2- or 3-tuples, got ''uint32''

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py:88: TypeError
```

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
pauladkisson: See https://github.com/catalystneuro/neuroconv/pull/1003#discussion_r2487885896

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -372,6 +375,7 @@ def configure_and_write_nwbfile(
     nwbfile_path: FilePath | None = None,
     backend: Literal["hdf5", "zarr"] | None = None,
     backend_configuration: BackendConfiguration | None = None,
+    export: bool = False,
```
h-mayorquin: Do we need this? I think that we can detect automatically whether the file was read form disk and use the appropriate method below, can't we?

docs/user_guide/backend_configuration.rst
```diff
@@ -139,6 +139,34 @@ Then we can use this configuration to write the NWB file:
     configure_and_write_nwbfile(nwbfile=nwbfile, backend_configuration=backend_configuration, nwbfile_path="output.nwb")
 
 
+Existing configuration
```
h-mayorquin: I think this is the core feature that is added in this PR. This allow the user to get the current configuration, modify what they need to modify and then write to nwbfile. The feature is quite flexible in that way.

On top of that repack is a wrapper of this behavior that works for a sub-set of use cases with fewer lines.

docs/user_guide/backend_configuration.rst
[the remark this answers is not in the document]
pauladkisson: Yes, that's right.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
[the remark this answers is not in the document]
pauladkisson: I don't know if I agree. I think both behaviors are reasonable. from_default tells users how they _should_ apply chunking/compression options, and from_existing tells users what chunking and compression options have already been specified.

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
pauladkisson: Maybe it's possible? I will have to look into it.

pauladkisson: Tracking down the history of this. Originally, I had it as a `configure_and_export_nwbfile` function, but then that was collapsed to an option inside `configure_and_write_nwbfile` based on this [comment](https://github.com/catalystneuro/neuroconv/pull/1003#discussion_r1719885740). 

But I think you're right. We shouldn't need the option at all because if the read_io is not None, then you're definitely going to want to export the nwbfile, and if you try to write the nwbfile when the read_io is not None, it actually throws an error. So there's no sensible case where you would want to do that. 

<details><summary>Details</summary>
<p>

```python
  File "/Users/pauladkisson/Documents/CatalystNeuro/Neuroconv/neuroconv/temp_test.py", line 34, in <module>
    main()
    ~~~~^^
  File "/Users/pauladkisson/Documents/CatalystNeuro/Neuroconv/neuroconv/temp_test.py", line 28, in main
    configure_and_write_nwbfile(nwbfile=read_nwbfile, nwbfile_path=nwbfile_path)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pauladkisson/Documents/CatalystNeuro/Neuroconv/neuroconv/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py", line 429, in configure_and_write_nwbfile
    with IO(nwbfile_path, mode="w") as io:
         ~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/repack_env/lib/python3.13/site-packages/hdmf/utils.py", line 578, in func_call
    return func(args[0], **pargs)
  File "/opt/anaconda3/envs/repack_env/lib/python3.13/site-packages/pynwb/__init__.py", line 454, in __init__
    super().__init__(path, manager=manager, mode=mode, file=file_obj, comm=comm,
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                     driver=driver, aws_region=aws_region, herd_path=herd_path)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/repack_env/lib/python3.13/site-packages/hdmf/utils.py", line 578, in func_call
    return func(args[0], **pargs)
  File "/opt/anaconda3/envs/repack_env/lib/python3.13/site-packages/hdmf/backends/hdf5/h5tools.py", line 104, in __init__
    super().__init__(manager, source=path, herd_path=herd_path)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/repack_env/lib/python3.13/site-packages/hdmf/utils.py", line 578, in func_call
    return func(args[0], **pargs)
  File "/opt/anaconda3/envs/repack_env/lib/python3.13/site-packages/hdmf/backends/io.py", line 41, in __init__
    self.open()
    ~~~~~~~~~^^
  File "/opt/anaconda3/envs/repack_env/lib/python3.13/site-packages/hdmf/backends/hdf5/h5tools.py", line 737, in open
    self.__file = File(self.source, open_flag, **kwargs)
                  ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/repack_env/lib/python3.13/site-packages/h5py/_hl/files.py", line 566, in __init__
    fid = make_fid(name, mode, userblock_size, fapl, fcpl, swmr=swmr)
  File "/opt/anaconda3/envs/repack_env/lib/python3.13/site-packages/h5py/_hl/files.py", line 247, in make_fid
    fid = h5f.create(name, h5f.ACC_TRUNC, fapl=fapl, fcpl=fcpl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5f.pyx", line 124, in h5py.h5f.create
OSError: Unable to synchronously create file (unable to truncate a file which is already open)
```
</p>
</details>

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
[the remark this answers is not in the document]
h-mayorquin: You might be right that I am relying too much on my own sense of what feels "natural" to me. Let me drop that angle and emphasize the one I think matters the most. I believe we should implement two separate functions. Using a boolean flag to control the function's behavior goes against the single responsibility principle and is widely recognized as a code smell (which I agree with):


https://alexkondov.com/should-you-pass-boolean-to-functions/

Furthermore, I do believe that there is space for a function that takes a configuration and fills non-configured datasets with defaults to further modularize the code.

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
h-mayorquin: Great.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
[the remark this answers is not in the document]
pauladkisson: Sorry, I missed this request earlier. Working on it now...

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_backend.py
```diff
@@ -57,6 +60,21 @@ def model_json_schema(cls, **kwargs) -> dict[str, Any]:
 
     @classmethod
     def from_nwbfile(cls, nwbfile: NWBFile) -> Self:
+        """
```
h-mayorquin: Ok.

src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
```diff
@@ -108,7 +108,7 @@ def get_default_dataset_io_configurations(
 
     known_dataset_fields = ("data", "timestamps")
     manager = get_manager()
-    builder = manager.build(nwbfile)
+    builder = manager.build(nwbfile, export=True)
```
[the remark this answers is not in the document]
h-mayorquin: Could we add a comment here with exactly that information?

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_existing_backend_configuration.py
```diff
@@ -0,0 +1,317 @@
...
+
+@pytest.fixture(scope="session")
+def zarr_nwbfile_path(tmpdir_factory):
+    compressor = Blosc(cname="lz4", clevel=5, shuffle=Blosc.SHUFFLE, blocksize=0)
+    filter1 = Blosc(cname="zstd", clevel=1, shuffle=Blosc.SHUFFLE)
+    filter2 = Blosc(cname="zstd", clevel=2, shuffle=Blosc.SHUFFLE)
+    filters = [filter1, filter2]
+
+    nwbfile_path = tmpdir_factory.mktemp("data").join("test_default_backend_configuration_hdf5_nwbfile.nwb.zarr")
+    if not Path(nwbfile_path).exists():
```
h-mayorquin: Same question here about the session scope.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_existing_backend_configuration.py
h-mayorquin: This looks good. I don't like the testing pattern of testing against the strings and I think we should test against the actual object but this pattern is already in the library so it is out of scope to change it here.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
h-mayorquin: This comment lead me nowhere. Can you double check?

h-mayorquin (commented):
I did a more detailed reading of this. Some request to add comments and some other questions about API design.

The rest looks good.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
h-mayorquin: Ok, let me read `_infer_dtype` again to think about this.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
pauladkisson: How about this [one](https://github.com/catalystneuro/neuroconv/pull/1003/files#r2516903381)?

src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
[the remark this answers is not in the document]
pauladkisson: Done!

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -421,4 +423,72 @@ def configure_and_write_nwbfile(
...
+            io.export(nwbfile=nwbfile, src_io=nwbfile.read_io, write_args=dict(link_data=False))
+        else:
+            io.write(nwbfile)
+
+
+def repack_nwbfile(
+    *,
+    nwbfile_path: Path,
+    export_nwbfile_path: Path,
+    backend: Literal["hdf5", "zarr"] = "hdf5",
```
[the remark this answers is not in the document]
pauladkisson: Yes, we need to specify the backend because it can be changed from HDF5 to Zarr (or vice versa). So you can't just infer it from the existing file.

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -421,4 +423,72 @@ def configure_and_write_nwbfile(
...
+    export_backend: Literal["hdf5", "zarr", None] = None,
+    use_default_backend_configuration: bool = True,
+    backend_configuration_changes: dict[str, dict] = None,
```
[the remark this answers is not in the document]
pauladkisson: Yeah, I think my default approach is to just make functions as complete and powerful as possible, but I think it's totally warranted in this case to reduce the scope of repack_nwbfile to a simpler high-level convenience function. I think it's a good suggestion. 

Will work on implementing...

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_existing_backend_configuration.py
```diff
@@ -0,0 +1,317 @@
...
+    processed_time_series = mock_TimeSeries(name="ProcessedTimeSeries", data=processed_array)
+    ecephys_module.add(processed_time_series)
+
+    return nwbfile
+
+
+@pytest.fixture(scope="session")
+def hdf5_nwbfile_path(tmpdir_factory):
+    nwbfile_path = tmpdir_factory.mktemp("data").join("test_existing_backend_configuration_hdf5_nwbfile.nwb.h5")
+    if not Path(nwbfile_path).exists():
```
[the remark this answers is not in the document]
pauladkisson: We don't. It's also using the old tmpdir_factory instead of the newer tmp_path_factory. This was some excessively defensive AI code that I must have missed. Thank you for the catch!

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_get_existing_dataset_io_configurations.py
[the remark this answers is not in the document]
pauladkisson: Yeah, doing it this way with the parameterization saved some effort in the writing of the tests, because I could do it once conceptually and then just adjust the assertions for zarr, but I could definitely see the argument for splitting them up.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_helpers/test_configure_backend_equivalency.py
```diff
@@ -85,7 +85,7 @@ def test_configure_backend_equivalency(
         elif backend == "zarr":
             assert written_data.compressor == numcodecs.GZip(level=2)
 
-        assert_array_equal(nwbfile_1.acquisition["TestTimeSeries"].data[:], written_data[:])
```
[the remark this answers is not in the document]
pauladkisson: It's the updates to configure_backend that broke this assertion. The `nwbfile_1.acquisition["TestTimeSeries"].data` gets wrapped in a data chunk iterator by configure_backend, so this simple indexing doesn't work (TypeError: 'DataChunkIterator' object is not subscriptable).

Asserting equality to the original array_1 is a simple and straightforward fix and keeps the intention of this test.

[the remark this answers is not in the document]
pauladkisson: With the latest changes, we could revert this now, and it wouldn't cause errors, but honestly, I think that it's better using array_1 and less convoluted.

src/neuroconv/tools/nwb_helpers/_configure_backend.py
```diff
@@ -43,14 +50,31 @@ def configure_backend(
 
         neurodata_object = nwbfile.objects[object_id]
         is_dataset_linked = isinstance(neurodata_object.fields.get(dataset_name), TimeSeries)
+        location_in_file = _find_location_in_memory_nwbfile(neurodata_object=neurodata_object, field_name=dataset_name)
```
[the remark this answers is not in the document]
h-mayorquin: Yes, makes sense. Thanks.

PR #1005, opened 2024-08-13
Deprecate H5DataIO from videointerface

h-mayorquin:
Draft

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
+++ b/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
@@ -440,7 +440,7 @@ class VideoInterface(BaseDataInterface):
             if compression is not None or compression_options is not None:
                 warnings.warn(
                     message=(
-                        "Specifying compression methods and their options at the level of tool functions has been deprecated. "
+                        "Specifying compression methods and their options for this interface has been deprecated. "
                         "Please use the `configure_backend` tool function for this purpose."
                     ),
                     category=DeprecationWarning,
--- a/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
+++ b/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
@@ -1,11 +1,10 @@
+import warnings
 from copy import deepcopy
 from pathlib import Path
 from typing import List, Literal, Optional
-from warnings import warn
 
 import numpy as np
 import psutil
-from hdmf.backends.hdf5.h5_utils import H5DataIO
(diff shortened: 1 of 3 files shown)


src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
```diff
@@ -437,14 +436,18 @@ def add_to_nwbfile(
...
-                wrapped_io_data = H5DataIO(
-                    iterable,
-                    compression=compression,
-                    compression_opts=compression_options,
-                    chunks=chunks,
+            # TODO: remove completely after 03/1/2024
+            if compression is not None or compression_options is not None:
+                warnings.warn(
+                    message=(
+                        "Specifying compression methods and their options at the level of tool functions has been deprecated. "
```
CodyCBakerPhD, line 443:
```suggestion
                        "Specifying compression methods and their options for this interface has been deprecated. "
```

Copied and pasted from elsewhere? This isn't a tool function, but an interface

h-mayorquin: Yes, I did, thanks for the catch!

CodyCBakerPhD:
One comment on warning

Can you also remove the docstring for these arguments? https://github.com/catalystneuro/neuroconv/blob/cade1204bb50d08cd03dd887d38ea61eb3f6e1ad/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py#L323-L329

What do you think about the tests in https://github.com/catalystneuro/neuroconv/blob/b9507bf58a977f90f03aed743c272c059cc21fbf/tests/test_behavior/test_video_utils.py#L154 - necessary to keep or can be removed?

h-mayorquin:
I removed the docstrings. 

Concerning the tests I am leaning to remove them just to make the testing slightly simpler but I really don't feel strongly either way.

CodyCBakerPhD:
I agree with removing; it isolates the responsibility of that support to the main configuration tool testing

h-mayorquin:
OK, removed.

CodyCBakerPhD (approved):

CodyCBakerPhD:
Cool!

Next step is to finally unlock the Zarr option in places like https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/nwbconverter.py#L175 (by including it in the literal)

Together with enhancing testing to ensure all interfaces can write+read back Zarr format

Did you want to handle that too?

h-mayorquin:
Yep. Let's do it.

PR #967, opened 2024-07-22
Some fixes for the DLC module

vigji:
This PR addresses a serious time bottleneck produced by the fact that currently the DLC interface always read through all the linked videos to create the timestamps arrays (line 166: `timestamps = _get_movie_timestamps(video[0], infer_timestamps=infer_timestamps)`), even when external timestamps are supplied. 

This PR simply adds an argument to the `_get_pes_args` function with the option to skip this search, and speeds up conversion from minutes to seconds in case there's both DLC videos and annotations. 

It actually also fixes an old bug of dlc2nwb that was supposed to be fixed, when `_get_movie_timestamps` ask for the Len of a `cv2` reader.

Along the way, it also purges `os` in favour of `pathlib`. Overall, those utility functions could really help some further cleanup, would be happy to do more but testing is currently a bit lacking for all the edge cases and I probably don't won't to risk too much refactoring...

Let me know how this looks!

PR #1006, opened 2024-08-13
Add timestamps optimization warning in dlc

h-mayorquin:
Agreed follow up on https://github.com/catalystneuro/neuroconv/pull/967 

Check it out @vigji  

image

Note that a half an hour video took less than a minute in my computer. Your performance was way worse @vigji ?


vigji:
Sorry, don’t have my computer and can’t these it this week, but from memory it was approx 5-10 mins to load 2 h movies at 100 Hz from an SSD using a MacBook Pro 2021

h-mayorquin:
Ok, that makes sense, it is on the ballpark. I just wanted to know if performance was more variable than that.

What do you think of the warning message on the screenshot?

vigji:
It looks reasonable to me! Would definitively address my concern.

h-mayorquin:
Good, this is ready then @CodyCBakerPhD .

CodyCBakerPhD (approved):

PR #1016, opened 2024-08-19
[Pydantic I] Refactor schema inference

CodyCBakerPhD:
@h-mayorquin Here is WIP for the pydantic refactor of the `get_schema_from_method_signature` function

Vastly reduces the complexity of the code we need to support there, but there's still a few 'extra' bits that we need to tack on to maintain back-compatability

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/lightningpose/lightningposeconverter.py
+++ b/src/neuroconv/datainterfaces/behavior/lightningpose/lightningposeconverter.py
@@ -1,5 +1,5 @@
 from copy import deepcopy
-from typing import Optional
+from typing import List, Optional
 
 from pynwb import NWBFile
 
@@ -70,20 +70,11 @@ class LightningPoseConverter(NWBConverter):
             self.data_interface_objects.update(dict(LabeledVideo=VideoInterface(file_paths=[labeled_video_file_path])))
 
     def get_conversion_options_schema(self) -> dict:
-        conversion_options = self.data_interface_objects["PoseEstimation"].get_conversion_options_schema()
-        conversion_options = dict_deep_update(
-            conversion_options, self.data_interface_objects["OriginalVideo"].get_conversion_options_schema()
+        conversion_options_schema = get_schema_from_method_signature(
+            method=self.add_to_nwbfile, exclude=["nwbfile", "metadata"]
         )
 
-        starting_frames = conversion_options["properties"].pop("starting_frames")
-        conversion_options["properties"].update(
-            dict(
-                starting_frames_original_videos=starting_frames,
-                starting_frames_labeled_videos=starting_frames,
-            )
-        )
-
-        return conversion_options
+        return conversion_options_schema
 
     def get_metadata(self) -> DeepDict:
         metadata = self.data_interface_objects["PoseEstimation"].get_metadata()
@@ -115,8 +106,8 @@ class LightningPoseConverter(NWBConverter):
         reference_frame: Optional[str] = None,
         confidence_definition: Optional[str] = None,
         external_mode: bool = True,
-        starting_frames_original_videos: Optional[list] = None,
-        starting_frames_labeled_videos: Optional[list] = None,
+        starting_frames_original_videos: Optional[List[int]] = None,
+        starting_frames_labeled_videos: Optional[List[int]] = None,
         stub_test: bool = False,
     ):
         original_video_interface = self.data_interface_objects["OriginalVideo"]
--- a/src/neuroconv/datainterfaces/ecephys/alphaomega/alphaomegadatainterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/alphaomega/alphaomegadatainterface.py
@@ -1,5 +1,6 @@
+from pydantic import DirectoryPath
+
 from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
-from ....utils.types import FolderPathType
 
 
 class AlphaOmegaRecordingInterface(BaseRecordingExtractorInterface):
@@ -19,7 +20,7 @@ class AlphaOmegaRecordingInterface(BaseRecordingExtractorInterface):
         source_schema["properties"]["folder_path"]["description"] = "Path to the folder of .mpx files."
         return source_schema
 
-    def __init__(self, folder_path: FolderPathType, verbose: bool = True, es_key: str = "ElectricalSeries"):
+    def __init__(self, folder_path: DirectoryPath, verbose: bool = True, es_key: str = "ElectricalSeries"):
         """
         Load and prepare data for AlphaOmega.
 
--- a/src/neuroconv/utils/json_schema.py
+++ b/src/neuroconv/utils/json_schema.py
@@ -69,7 +69,7 @@ def get_base_schema(
     return base_schema
 
 
-def get_schema_from_method_signature(method: Callable, exclude: list[str] | None = None) -> dict:
+def get_schema_from_method_signature(method: Callable, exclude: Optional[List[str]] = None) -> dict:
     """Deprecated version of `get_json_schema_from_method_signature`."""
     message = (
         "The method `get_schema_from_method_signature` is now named `get_json_schema_from_method_signature`."
@@ -80,7 +80,7 @@ def get_schema_from_method_signature(method: Callable, exclude: list[str] | None
     return get_json_schema_from_method_signature(method=method, exclude=exclude)
 
 
-def get_json_schema_from_method_signature(method: Callable, exclude: list[str] | None = None) -> dict:
+def get_json_schema_from_method_signature(method: Callable, exclude: Optional[List[str]] = None) -> dict:
     """
     Get the equivalent JSON schema for a signature of a method.
 
--- a/src/neuroconv/utils/json_schema.py
+++ b/src/neuroconv/utils/json_schema.py
@@ -100,10 +100,11 @@ def get_json_schema_from_method_signature(method: Callable, exclude: list[str] |
         The JSON schema corresponding to the method signature.
     """
     exclude = exclude or []
-    exclude += ["self"]
+    exclude += ["self", "cls"]
 
     signature = inspect.signature(obj=method)
     parameters = signature.parameters
+    additional_properties = False
     arguments_to_annotations = {}
     for argument_name in parameters:
         if argument_name in exclude:
@@ -112,6 +113,7 @@ def get_json_schema_from_method_signature(method: Callable, exclude: list[str] |
         parameter = parameters[argument_name]
 
         if parameter.kind == inspect.Parameter.VAR_KEYWORD:  # Skip all **{...} usage
+            additional_properties = True
             continue
 
         annotation = parameter.annotation
@@ -131,8 +133,8 @@ def get_json_schema_from_method_signature(method: Callable, exclude: list[str] |
     # But Pydantic does automatically
     json_schema = _copy_without_title_keys(temp_json_schema)
 
-    # We also freeze additional properties, but Pydantic does not
-    json_schema["additionalProperties"] = False
+    # Pydantic does not make determinations on additionalProperties
+    json_schema["additionalProperties"] = additional_properties
 
     # Attempt to find descriptions within the docstring of the method
     parsed_docstring = docstring_parser.parse(method.__doc__)
--- a/src/neuroconv/utils/json_schema.py
+++ b/src/neuroconv/utils/json_schema.py
@@ -6,6 +6,7 @@ from datetime import datetime
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional
 
+import docstring_parser
 import hdmf.data_utils
 import numpy as np
 import pydantic
@@ -83,6 +84,8 @@ def get_json_schema_from_method_signature(method: Callable, exclude: list[str] |
     """
     Get the equivalent JSON schema for a signature of a method.
 
+    Also uses `docstring_parser` (NumPy style) to attempt to find descriptions for the arguments.
+
     Parameters
     ----------
     method : callable
@@ -128,6 +131,18 @@ def get_json_schema_from_method_signature(method: Callable, exclude: list[str] |
 
     # We also freeze additional properties, but Pydantic does not
     json_schema["additionalProperties"] = False
+
+    # Attempt to find descriptions within the docstring of the method
+    parsed_docstring = docstring_parser.parse(method.__doc__)
+    for parameter_in_docstring in parsed_docstring.params:
+        if (
+            parameter_in_docstring.arg_name in json_schema["properties"]
+            and parameter_in_docstring.description is not None
+        ):
+            json_schema["properties"][parameter_in_docstring.arg_name].update(
+                description=parameter_in_docstring.description
+            )
+    # TODO: could also add Field support for more direct control over docstrings (and enhanced validation conditions)
 
     return json_schema
 
--- a/src/neuroconv/utils/json_schema.py
+++ b/src/neuroconv/utils/json_schema.py
@@ -1,20 +1,18 @@
 import collections.abc
 import inspect
 import json
+import warnings
 from datetime import datetime
 from pathlib import Path
-from typing import Callable, Dict, List, Literal, Optional
+from typing import Callable, Dict, List, Optional
 
-import docstring_parser
 import hdmf.data_utils
 import numpy as np
+import pydantic
 import pynwb
 from jsonschema import validate
 from pynwb.device import Device
 from pynwb.icephys import IntracellularElectrode
-
-from .dict import dict_deep_update
-from .types import FilePathType, FolderPathType
 
 
 class NWBMetaDataEncoder(json.JSONEncoder):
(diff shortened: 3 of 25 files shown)


src/neuroconv/datainterfaces/ecephys/alphaomega/alphaomegadatainterface.py
```diff
@@ -1,5 +1,6 @@
+from pydantic import DirectoryPath
```
CodyCBakerPhD, line 1: I tested this out here as a showcase of how the rest of the interfaces would be refactored, but left that for a followup since it's a lot of changes all at once

There will technically be a small back-compatibility break until that follow-up is merged, in the 'format' of the path type in each schema. However, this field is not used by JSON validation at all so it's not expected to break anything in the interim

src/neuroconv/utils/json_schema.py
```diff
@@ -70,104 +69,93 @@ def get_base_schema(
     return base_schema
 
 
-def get_schema_from_method_signature(method: Callable, exclude: list = None) -> dict:
+def get_schema_from_method_signature(method: Callable, exclude: list[str] | None = None) -> dict:
```
CodyCBakerPhD, line 72: Much in the way Pydantic decided to deprecate their generic `.schema` method in favor of explicit/verbose pattern, I think we should too since we deal with many other schemas nearby to this utility and elsewhere in our ecosystem (HDMF, NWB, LinkML, etc. etc.)

src/neuroconv/utils/json_schema.py
```diff
@@ -70,104 +69,93 @@ def get_base_schema(
...
+
+        annotation = parameter.annotation
+
+        # Pydantic uses ellipsis for required
+        pydantic_default = ... if parameter.default is inspect._empty else parameter.default
+
+        arguments_to_annotations.update({argument_name: (annotation, pydantic_default)})
+
+    model = pydantic.create_model(
+        "_TempModel", __config__=pydantic.ConfigDict(arbitrary_types_allowed=True), **arguments_to_annotations
```
CodyCBakerPhD, line 127: If you're wondering about the config here, it is a requirement to support any numpy types (which, how we do it here, could also be improved such as in https://github.com/pydantic/pydantic/issues/7017)

For example, https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/ecephys/spikegadgets/spikegadgetsdatainterface.py#L27

src/neuroconv/utils/json_schema.py
```diff
@@ -70,104 +69,93 @@ def get_base_schema(
...
-                    input_schema["properties"].update({param_name: dict(format="directory")})
-        else:
-            raise NotImplementedError(
-                f"The annotation type of '{param}' in function '{method}' is not implemented! "
-                "Please request it to be added at github.com/catalystneuro/nwb-conversion-tools/issues "
-                "or create the json-schema for this method manually."
+    for parameter_in_docstring in parsed_docstring.params:
+        if (
+            parameter_in_docstring.arg_name in json_schema["properties"]
+            and parameter_in_docstring.description is not None
```
bendichter, line 144: consider throwing a warning

tests/test_minimal/test_utils/test_json_schema_utils.py
```diff
@@ -39,35 +35,6 @@ def sort_item(item):
...
-
-    correct_schema = dict(
-        additionalProperties=False,
-        properties=dict(
-            a=dict(type="number"),
-            b=dict(type="number"),
-            c=dict(type="string"),
-            d=dict(type="boolean"),
-            e=dict(default="hi", type="string"),
-            f=dict(type="object", additionalProperties={"^.*$": dict(type="string")}),
```
h-mayorquin, line 57: I was looking for a reference to the regex-like expression here in f in the new tests. Is this something we are changing?

h-mayorquin (commented):
Did a first reading. Some questions/comments. I will read the other PR that depends on this and come back to this one. 

Thanks for making the changes small and readable!

src/neuroconv/utils/json_schema.py
```diff
@@ -70,104 +69,93 @@ def get_base_schema(
...
 
     Parameters
     ----------
-    method: function
-    exclude: list, optional
+    method : callable
+        The method to generate the JSON schema from.
+    exclude : list of str, optional
+        List of arguments to exclude from the schema generation.
+        Always includes 'self' and 'kwargs'.
```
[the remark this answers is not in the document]
bendichter: Yes, in fact this is changed in the code on line 115 to work for *any* var that is unpacked with `**` regardless of its name.

bendichter, line 95:
```suggestion
        Always includes 'self' and 'cls'.
```

src/neuroconv/utils/json_schema.py
```diff
@@ -70,104 +69,93 @@ def get_base_schema(
...
-        tuple="array",
-        FilePathType="string",
-        FolderPathType="string",
+    exclude = exclude or []
+    exclude += ["self", "cls"]
+
+    signature = inspect.signature(obj=method)
+    parameters = signature.parameters
+    additional_properties = False
+    arguments_to_annotations = {}
```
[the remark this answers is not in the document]
bendichter: `annotation` is a term used by the `inspect` library. We don't get pydantic field definitions until line 126.

src/neuroconv/utils/json_schema.py
```diff
@@ -70,104 +69,93 @@ def get_base_schema(
...
+        annotation = parameter.annotation
+
+        # Pydantic uses ellipsis for required
+        pydantic_default = ... if parameter.default is inspect._empty else parameter.default
+
+        arguments_to_annotations.update({argument_name: (annotation, pydantic_default)})
+
+    model = pydantic.create_model(
+        "_TempModel", __config__=pydantic.ConfigDict(arbitrary_types_allowed=True), **arguments_to_annotations
     )
```
bendichter, line 128:
```suggestion
    # The ConfigDict is required to support any numpy types
    model = pydantic.create_model(
        "_TempModel", __config__=pydantic.ConfigDict(arbitrary_types_allowed=True), **arguments_to_annotations
    )
```

bendichter: this seems to me important enough to add as a comment

src/neuroconv/datainterfaces/behavior/lightningpose/lightningposeconverter.py
```diff
@@ -70,20 +70,11 @@ def __init__(
             self.data_interface_objects.update(dict(LabeledVideo=VideoInterface(file_paths=[labeled_video_file_path])))
 
     def get_conversion_options_schema(self) -> dict:
-        conversion_options = self.data_interface_objects["PoseEstimation"].get_conversion_options_schema()
-        conversion_options = dict_deep_update(
-            conversion_options, self.data_interface_objects["OriginalVideo"].get_conversion_options_schema()
+        conversion_options_schema = get_schema_from_method_signature(
+            method=self.add_to_nwbfile, exclude=["nwbfile", "metadata"]
```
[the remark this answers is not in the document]
h-mayorquin: I thought that your config would allow those ones to work by just checking instances.

src/neuroconv/utils/json_schema.py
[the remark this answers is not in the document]
h-mayorquin: Yes, I am aware that that is the term used but the dictionary we are building does not have annotations as items. Plus, the dictionary that we are building is called field_definitions in the pydantic function spec as per the link I showed.

src/neuroconv/utils/json_schema.py
h-mayorquin: I think it is a good idea to motivate the need to pass a config but this is not only support np.ndarray but also things like nwbfile and instance checking for general objects.

src/neuroconv/utils/json_schema.py
CodyCBakerPhD: Good catch - this was copied from the old docstring

src/neuroconv/datainterfaces/behavior/lightningpose/lightningposeconverter.py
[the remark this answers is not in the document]
CodyCBakerPhD: Good question - the original reasoning is because that's how it is now: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/basedatainterface.py#L47

But taking a closer look on this PR...

The issue is that `pynwb.NWBFile` has no valid JSON type - setting `arbitrary_types_allowed=True` in the config allows any field of any type in the Pydantic model to be set, but then JSON schema generation fails for such a model without custom decoding, and I'm not really sure how I would describe that in-memory object in JSON for that matter

As such, the `nwbfile` argument really should be skipped

But I will see about relaxing the `metadata` restriction - previously that would have been because we couldn't support anything too complicated, like `Dict[str, Any]` but maybe now we can

src/neuroconv/utils/json_schema.py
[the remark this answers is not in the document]
CodyCBakerPhD: Yes, we used to skip the name `kwargs` but that of course is just convention not necessity

The true skip is with respect to the check `inspect.Parameter.VAR_KEYWORD`, which is unique to the usage of `**` to dynamically pass additional keyword arguments that aren't annotated

src/neuroconv/utils/json_schema.py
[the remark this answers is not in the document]
CodyCBakerPhD: > Yes, I am aware that that is the term used but the dictionary we are building does not have annotations as items.

[Line 124](arguments_to_annotations.update({argument_name: (annotation, pydantic_default)})): `arguments_to_annotations.update({argument_name: (annotation, pydantic_default)})`

where `annotation` is `parameter.annotation`

So 'annotation' is the classic Pythonic sense of associating a type with an argument via `:`, and 'annotation' is the built-in reference via the `inspect.signature`

Pydantic models have 'fields' which are themselves annotations and can be thought of as extensions beyond conventional Python annotations, such as what is possible via the `pydantic.Field` class

It is, however, a custom class - and our goal here is to coerce annotations to become simple pydantic fields

Some grey area could occur if/when full `pydantic.Field` support is added/used in the library here, for example, as a way of asserting integer ranges such as `number_of_cpus = pydantic.Field(int, ge=1, le=os.cpu_count(), default=1, description="The number of CPUs to use for the process."). But I'd argue you still don't know ahead of time whether or not any given annotation is going to be a 'Field'

src/neuroconv/utils/json_schema.py
CodyCBakerPhD: Note we aren't using any Pydantic-based validation here (saving that for 3rd PR of this chain) but when we do add that - yes, those do get skipped - however, they are not skipped when generating a JSON schema for the class, which then fails if it can't figure out an appropriate type to associate

src/neuroconv/utils/json_schema.py
```diff
@@ -70,104 +69,93 @@ def get_base_schema(
...
-    args_spec = dict()
+
+    temp_json_schema = model.model_json_schema()
+
+    # We never used to include titles in the lower schema layers
```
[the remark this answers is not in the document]
CodyCBakerPhD: > Does it matter if we do from now on? To avoid adding extra code.

In this PR I am merely trying to maintain as much similarity as possible with the current state

I think we can safely include titles since we artificially generate them on the GUIDE anyway

But maybe this can be step 4 of the chain of PRs

unroot does quite a bit more than just that, so I would rather keep the targetted title cleaning. It's pretty minimal for now

src/neuroconv/utils/json_schema.py
```diff
@@ -70,104 +69,93 @@ def get_base_schema(
...
-            args_spec[param_name].update(default=param.default)
-        input_schema["properties"] = dict_deep_update(input_schema["properties"], args_spec)
-        input_schema["additionalProperties"] = param.kind == inspect.Parameter.VAR_KEYWORD
-    return input_schema
+    # TODO: could also add Field support for more direct control over docstrings (and enhanced validation conditions)
+
+    return json_schema
+
+
+def _copy_without_title_keys(d: Any, /) -> Optional[dict]:
```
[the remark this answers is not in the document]
CodyCBakerPhD: Lol, I noticed it on Pydantic (they seem to love it) so figured I'd try it out for a small private helper function

In general you know much much I love keyword-only for everything, including single-argument functions

tests/test_minimal/test_utils/test_json_schema_utils.py
CodyCBakerPhD: If you read the regex carefully it's not saying anything different from the current form (both say 'a string, including the empty string'). According to tests does not alter any validation conditions either

Though if anything the regex ought to be `*.+$` to indicate that each key in the dictionary must have at least one character

tests/test_minimal/test_utils/test_get_schema_from_method_signature.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    These are annotations newly supported by the Pydantic-based inference.
+
+    They should also be compatible with __future__.annotations for SpikeInterface.
+    """
+
+    # TODO: enable | usage cases when 3.11 is minimal
+    def advanced_method(
+        old_list_of_strings: List[str],
+        # new_list_of_strings: list[str],
```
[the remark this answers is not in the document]
CodyCBakerPhD: Right you are; it's only the `|` as `Union` that awaits in the future

Now that we support `list` instead of `List` I'll add a subsequent PR updating to the more modern annotation everywhere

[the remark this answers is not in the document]
h-mayorquin: Yei!

src/neuroconv/utils/json_schema.py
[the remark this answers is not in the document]
h-mayorquin: OK, thanks for explaining. This makes sense.

src/neuroconv/utils/json_schema.py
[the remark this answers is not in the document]
h-mayorquin: I am failing to convey what I mean.

> arguments_to_annotations.update({argument_name: (annotation, pydantic_default)}))
Yes, I understand and I saw it. What I said above is that:
1) "the dictionary does not have annotations as its items". Its items are a tuple.
2) The object that we are building is called is called `field_definitions`. Check my highlight from the link that I shared:

image

This is a minimal thing, feel free to ignore it if you guys like `arguments_to_annotations` as a variable name. I am just focusing on what we are building instead of how we are building it.

tests/test_minimal/test_utils/test_json_schema_utils.py
h-mayorquin: Good. To be honest, I don't know regex enougn and I always have to read it. I was pointing it out in case it was a change by omission instead of by intention. If the code does the same and we don't need to put regex on it I think that's a win.

src/neuroconv/utils/json_schema.py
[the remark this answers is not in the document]
h-mayorquin: Right.

src/neuroconv/datainterfaces/behavior/lightningpose/lightningposeconverter.py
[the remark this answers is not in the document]
h-mayorquin: Yes. When I wrote this I did not relize that under the hood we will still be using the json schema to do the validation.

passing the config like this allows you to do pydantic validation by instance but leaves you with no json schema.

src/neuroconv/utils/json_schema.py
CodyCBakerPhD: The deprecation here in favor of `..._json_schema_...` to be explicit about what type of schema we are inferring

src/neuroconv/utils/json_schema.py
CodyCBakerPhD: Added and tested

src/neuroconv/utils/json_schema.py
h-mayorquin: Ah, got you, yeah I agree it makes sense!

src/neuroconv/utils/json_schema.py
h-mayorquin: Yep. This is what I failed to realize. We still need json.  

Also, the suggestion disapleyd wrong. What I was suggesting was to pass the model_name as keyword: `model_name="_TempModel"` instead of positionally as we are doing here.

CodyCBakerPhD: @h-mayorquin You know I would love nothing more than that, and it's a rare day when you catch me doing anything but

However, as I mentioned elsewhere, Pydantic loves their first positional arguments 😂 So it is actually not allowed at all in this case, much to my dismay

CodyCBakerPhD:
@h-mayorquin BTW I will save the relaxation of `metadata` in exclusion at various places in a follow-up to to a targetted assessment of what effect it will have

src/neuroconv/utils/json_schema.py
h-mayorquin: Mmm I don't really like this. I think that warnings should be actionable. If someone pushed a function to the repo with a typo  and then someone else uses the functionthey will be getting warnings that they can't do nothing about. What use is this information to them? If we want the creator of the function to really add a correct docstring then maybe it should be an assertion or handled somewhere else I feel.

h-mayorquin:
>BTW I will save the relaxation of metadata in exclusion at various places in a follow-up to to a targetted assessment of what effect it will have

OK

The thing that I don't see tested is the use of np.ndarray. Given that you added logic for that here I think it would be good to create that test. I am asking because I expet

src/neuroconv/utils/json_schema.py
h-mayorquin: I don't like to put that burden on the user and I don't see why we need to do it if we can inform the developer.

tests/test_minimal/test_utils/test_get_json_schema_from_method_signature.py
```diff
@@ -0,0 +1,314 @@
...
+    """
+
+    # TODO: enable | instead of union when 3.11 is minimal
+    def advanced_method(
+        old_list_of_strings: List[str],
+        new_list_of_strings: list[str],
+        old_dict_of_ints: Dict[str, int],
+        new_dict_of_ints: dict[str, int],
+        nested_list_of_strings: List[List[str]],
+        array_type: ArrayType,
```
CodyCBakerPhD, line 62: @h-mayorquin Here is an explicit test of array

Previously I was simply relying on their presence in various data interfaces, which failed their respective data interface tests without the config fix

CodyCBakerPhD: Note however that it is (must be) via our usage of the `ArrayType` (which itself could be improved no doubt); if it is simply a `np.array` only then it must be excluded in order to pass this inference

tests/test_minimal/test_utils/test_get_json_schema_from_method_signature.py
```diff
@@ -0,0 +1,325 @@
...
+            "nested_list_of_strings",
+        ],
+        "type": "object",
+    }
+
+    assert test_json_schema == expected_json_schema
+
+
+def test_get_json_schema_from_method_signature_exclude():
+    def basic_method(
```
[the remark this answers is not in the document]
CodyCBakerPhD: Done

src/neuroconv/utils/json_schema.py
CodyCBakerPhD: Another thought

In theory, this method is only currently applied to core supported data interfaces in NeuroConv - do we trust ourselves enough to catch any such typos in `__init__` and `add_to_nwbfile` signatures+docstrings? If so then no worry

But actually then, when a **user** creates their own custom data interface with their own custom signatures and their own docstrings, then this becomes a useful warning in that case that they wouldn't have noticed without their own linting system

tests/test_minimal/test_utils/test_get_json_schema_from_method_signature.py
h-mayorquin: OK, I just tested using the official numpy typing in your test:

```python
    from numpy.typing import ArrayLike
    # TODO: enable | instead of union when 3.11 is minimal
    def advanced_method(
        old_list_of_strings: List[str],
        new_list_of_strings: list[str],
        old_dict_of_ints: Dict[str, int],
        new_dict_of_ints: dict[str, int],
        nested_list_of_strings: List[List[str]],
        array_type: ArrayLike,
        # more_nested_list_of_strings: list[list[list[str]]],
        # pathalogical_case: list[dict[str | int | None, list[Optional[dict[str, list[Literal["a", "b"] | None]]]]],],
    ):
        pass

    test_json_schema = get_json_schema_from_method_signature(method=advanced_method)
```

The schema returns
```python
test_json_schema["properties"]["array_type"]
{'anyOf': [{'type': 'boolean'}, {'type': 'integer'}, {'type': 'number'}, {'type': 'string'}, {'format': 'binary', 'type': 'string'}]}
```

But this is not yet supported:
https://github.com/pydantic/pydantic-extra-types/issues/31

Something to keep an eye on.

h-mayorquin (approved):
LGTM.

I don't like the warning thing but I guess this is a minimal disagreement.

src/neuroconv/utils/json_schema.py
h-mayorquin: Yeah, for that reason I would rather made an assertion. That way we tell the developer of the problem **but also** we don't allow him to be lazy and pass this standard error pollution to the end user.

CodyCBakerPhD: Can you open a subsequent PR elevating it to an error then? For discussion

tests/test_minimal/test_utils/test_get_json_schema_from_method_signature.py
CodyCBakerPhD: Yes, exactly. People have been working on this for a long time and I haven't seen a single perfect accepted solution yet

src/neuroconv/utils/json_schema.py
h-mayorquin: If you really prefer the warning then I think that I would rather have @bendichter break the vote here for assertion vs warning.

If he does not vote I will assume that he prefers a warning and we can let this aside.

I don't think it makes sense to open and issue, everything has been said here.

PR #1017, opened 2024-08-19
[Pydantic II] Change all paths to pydantic types

CodyCBakerPhD:
@h-mayorquin The follow-up replacing all usage with the Pydantic ones

Still need to make my way through the ecephys ones


src/neuroconv/utils/types.py
```diff
@@ -1,12 +1,7 @@
-from pathlib import Path
-from typing import Optional, TypeVar, Union
+from typing import Optional, Union
 
 import numpy as np
 
-FilePathType = TypeVar("FilePathType", str, Path)
```
CodyCBakerPhD, line 6: Yes, this is a hard deprecation

I could not think of a good way to do a soft one without utilizing module-level `__getattr__` magic to inject a deprecation warning

CodyCBakerPhD:
@h-mayorquin All tests should be green now

h-mayorquin (approved):
Unrelated but I would like to eliminate the OptionalArray as well

CodyCBakerPhD:
> Unrelated but I would like to eliminate the OptionalArray as well

One step at a time I say 😉

CodyCBakerPhD:
But yes, happy to do that in follow-up

PR #1019, opened 2024-08-20
Enable backend test for zarr

h-mayorquin:
Now that all the wrappers for data have been removed we are enablign both typing and testing for the zarr backend at the run_conversion level.

### DIFF
--- a/src/neuroconv/basedatainterface.py
+++ b/src/neuroconv/basedatainterface.py
@@ -125,11 +125,8 @@ class BaseDataInterface(ABC):
         nwbfile: Optional[NWBFile] = None,
         metadata: Optional[dict] = None,
         overwrite: bool = False,
-        # TODO: when all H5DataIO prewraps are gone, introduce Zarr safely
-        # backend: Union[Literal["hdf5", "zarr"]],
-        # backend_configuration: Optional[Union[HDF5BackendConfiguration, ZarrBackendConfiguration]] = None,
-        backend: Optional[Literal["hdf5"]] = None,
-        backend_configuration: Optional[HDF5BackendConfiguration] = None,
+        backend: Literal["hdf5", "zarr"] = "hdf5",
+        backend_configuration: Optional[Union[HDF5BackendConfiguration, ZarrBackendConfiguration]] = None,
         **conversion_options,
     ):
         """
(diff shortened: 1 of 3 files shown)


src/neuroconv/basedatainterface.py
```diff
@@ -146,15 +143,15 @@ def run_conversion(
...
-            BackendConfiguration object, and pass that instead.
-            Otherwise, all datasets will use default configuration settings.
+        backend : {'hdf5', 'zarr'}, default: 'hdf5'
+            The storage backend for the NWB file.
+            If `backend_configuration` is provided, this value is overridden by the configuration's backend type.
+        backend_configuration : HDF5BackendConfiguration or ZarrBackendConfiguration, optional
+            Configuration settings for the chosen backend. If not provided, default settings are used.
+            To customize, obtain the default configuration using `.get_default_backend_configuration(...)`,
+            modify it, and pass the modified configuration here.
+        conversion_options : dict, optional
```
CodyCBakerPhD, line 153: This is not true though; the above passes it dynamically

CodyCBakerPhD:
Thanks for opening

Hope tests go OK

src/neuroconv/basedatainterface.py
```diff
@@ -125,11 +125,8 @@ def run_conversion(
...
         metadata: Optional[dict] = None,
         overwrite: bool = False,
-        # TODO: when all H5DataIO prewraps are gone, introduce Zarr safely
-        # backend: Union[Literal["hdf5", "zarr"]],
-        # backend_configuration: Optional[Union[HDF5BackendConfiguration, ZarrBackendConfiguration]] = None,
-        backend: Optional[Literal["hdf5"]] = None,
-        backend_configuration: Optional[HDF5BackendConfiguration] = None,
+        backend: Literal["hdf5", "zarr"] = "hdf5",
+        backend_configuration: Optional[Union[HDF5BackendConfiguration, ZarrBackendConfiguration]] = None,
         **conversion_options,
```
CodyCBakerPhD, line 130: @h-mayorquin There is no keyword argument for `conversion_options` here

src/neuroconv/basedatainterface.py
h-mayorquin, line 153: Got you, thanks.
```suggestion
        **conversion_options : dict, optional
```
So you mean it should be like this? 

Not sure about the typing.

CodyCBakerPhD: IDK that we need to type it in the base class

It's mostly to allow pass-through from child classes that would define and annotate specific arguments

h-mayorquin: OK!

h-mayorquin:
Closing in favor of #1056

PR #1020, opened 2024-08-20
Add source data checks to interface

h-mayorquin:
Draft. As in the title.


CodyCBakerPhD:
Why not just use `pydantic.validate_call` and branch from the pydantic PR

h-mayorquin:
> pydantic.validate_call

I want things to be as similar as possible between interfaces and converters.

Do you think the pydantic mehthod would be better? I will take a look. If so, I might change that in a follow up PR.

CodyCBakerPhD:
I was going to add in a follow-up PR, as I mentioned on Slack

CodyCBakerPhD:
For purposes of initialization of an interface, the pydantic validation would be superior to JSON

It would also be triggered once per interface by an `NWBConverter` during its initialization without extra effort

I would not recommend doing JSON validation on all interfaces like you have it here because because the NWBConverter initialization already validations the entire JSON schema as constructed by that converter; validation the sublevels of each interface by their respective subschemas is quite wasteful, possibly even slow

h-mayorquin:
OK, we discussed this in person. Makes sense to handle with pydantic.

PR #1022, opened 2024-08-21
[Pydantic IVa] Add validation to most interfaces and converters

CodyCBakerPhD:
We'll see what the tests say


src/neuroconv/datainterfaces/ecephys/spikegadgets/spikegadgetsdatainterface.py
```diff
@@ -22,6 +22,7 @@ def get_source_schema(cls) -> dict:
         source_schema["properties"]["file_path"].update(description="Path to SpikeGadgets (.rec) file.")
         return source_schema
 
+    # @validate_call
```
CodyCBakerPhD, line 25: Some particular lines have an issue that might be related to the way the current ArrayType is defined; I'm leaving them as commented out and will debug in the next followup

h-mayorquin:
Can we use the same mechanism in converters?

h-mayorquin:
Or asked differently, do we still need the mechanism in converters if this is doing the job? Should we get rid of those checks?

CodyCBakerPhD:
> Can we use the same mechanism in converters?

I changed the PR title accordingly (the CHANGLOG was always correct)

I applied it where I could and commented out where I couldn't for the next followup

https://github.com/catalystneuro/neuroconv/pull/1022/files#diff-2a5b50c18c9b282ee03800d3fbaf47c5cbb4b300248c49cc2f6b96841c91759dR76

Notably the converters still also do JSON validation because there isn't a super easy way to ensure the inputs to an NWBConverter (the `source_data` dict of dicts) are valid being the basic typing

h-mayorquin:
But now the interfaces are validate when they are intialized. Does not make this the validation at the converter level redundant?

CodyCBakerPhD:
> But now the interfaces are validate when they are intialized. Does not make this the validation at the converter level redundant?

The `NWBConverter` Pydantically validates that it received a `dict` of some strings mapping onto other `dict` objects but nothing further

The interfaces when initialized will be Pydantically validated that their inputs are appropriate paths, they exist, whatnot

But before that, the `NWBConverter` input `source_data` dictionary is JSON validated that the specific keys and all their subdicts are organized as expected by the converter

So it's subtle, but could actually catch different inconsistencies

If we made `source_data` an explicit `BaseModel` with explicit keys for what is expected, it would then be equivalent. But that would be tough to do without at least some `__new__` injection magic

h-mayorquin:
> But before that, the NWBConverter input source_data dictionary is JSON validated that the specific keys and all their subdicts are organized as expected by the converter

> So it's subtle, but could actually catch different inconsistencies

Can you be more specific and point one for me to understand. I am looking at the code here:

https://github.com/catalystneuro/neuroconv/blob/8e20a25ad30609aefffc5ad6f9b56354ee5e6d10/src/neuroconv/nwbconverter.py#L59-L84

CodyCBakerPhD:
1. Very simple Pydantic validation of input to `NWBConverter`

https://github.com/catalystneuro/neuroconv/blob/add_source_validation/src/neuroconv/nwbconverter.py#L76-L77

2. Detailed validation of `source_data` dictionary

https://github.com/catalystneuro/neuroconv/blob/add_source_validation/src/neuroconv/nwbconverter.py#L72

relative to full schema expected specific to the `NWBConverter` from

https://github.com/catalystneuro/neuroconv/blob/add_source_validation/src/neuroconv/nwbconverter.py#L46-L57

(assembled via calls to each interface, yes, but some extra stuff at top level)

3. Pydantic validation of each interface as it gets initialized on path inputs etc.

h-mayorquin:
OK, so after discussing more with Cody this is one concrete example that pydantic won't catch:

```python
from datetime import datetime
from zoneinfo import ZoneInfo
from pathlib import Path
from neuroconv import NWBConverter
from neuroconv.datainterfaces import SpikeGLXRecordingInterface, PhySortingInterface

# For this interface we need to pass the location of the ``.bin`` file. Change the file_path to the location in your system
file_path = f"{ECEPHY_DATA_PATH}/spikeglx/Noise4Sam_g0/Noise4Sam_g0_imec0/Noise4Sam_g0_t0.imec0.ap.bin"
interface_spikeglx = SpikeGLXRecordingInterface(file_path=file_path, verbose=False)

folder_path = f"{ECEPHY_DATA_PATH}/phy/phy_example_0"  # Change the folder_path to the location of the data in your system
interface_phy = PhySortingInterface(folder_path=folder_path, verbose=False)

 # Now that we have defined the two interfaces we pass them to the ConverterPipe which will coordinate the
 # concurrent conversion of the data

class TestConverter(NWBConverter):
    data_interface_classes = {
        "A": SpikeGLXRecordingInterface,
        "B": PhySortingInterface
    }
    
source_data = {"A": dict(file_path=file_path), "WRONG": dict(folder_path=folder_path)}

converter = TestConverter(source_data=source_data, verbose=True)
ValidationError: Additional properties are not allowed ('WRONG' was unexpected)

Failed validating 'additionalProperties' in schema:
```

But I think that can be easily fixed for the `NWBConverter` by just checking that the keys of source_data are in the keys of the data_interfaces attributes.

I will give it another thought but I wanted to leave the concrete example here.

CodyCBakerPhD:
@h-mayorquin I had a better idea for how to do this; will see tomorrow if it works

CodyCBakerPhD:
@h-mayorquin OK after trying and thinking on several other strategies, I still think that having the two different 'attempts' at validation could be better than attempting to consolidate to one or the other

Reasoning: the act of wrapping an interface with `validate_call` is not (and cannot easily be) required - we are doing it here for all core interfaces, but not requiring any custom interfaces to be wrapped in the same way

Whereas the `NWBConverter.get_source_schema` + JSON validation applies without any intentional wrapping and relies only on the act of using annotation typing on interface signatures.

Technically, it too is 'optional' in the sense that it can be overridden by any custom definition of a `NWBConverter` child class that does not itself make a call to `super().__init__(...)`, but that at least takes more work than forgetting to apply the wrapper

So I view the additional validation provided by `validate_call` to be a 'nicety' or 'extra feature' to provide better feedback earlier when using the most popular interfaces, but not as a complete replacement to our existing workflow

How does that sound?

I guess one question that remains is if you think it's worth it then to also apply the JSON schema logic at the base interface level as well

CodyCBakerPhD:
FYI @pauladkisson failing credentials on live service tests: https://github.com/catalystneuro/neuroconv/actions/runs/10685941506/job/29620008386?pr=1022

Make sure you updated with your staging key (not main archive)

CodyCBakerPhD:
FYI @pauladkisson, codecov token failing too: https://github.com/catalystneuro/neuroconv/actions/runs/10685941506/job/29620006161?pr=1022

pauladkisson:
> FYI @pauladkisson, codecov token failing too: https://github.com/catalystneuro/neuroconv/actions/runs/10685941506/job/29620006161?pr=1022

I only just updated it 23mins ago and these tests triggered 53mins ago, so maybe it was using the old one.

h-mayorquin:
@CodyCBakerPhD 
> How does that sound?

Ok, thanks for giving it a thought though.

> I guess one question that remains is if you think it's worth it then to also apply the JSON schema logic at the base interface level as well

Yes, I would like that. This is in the context of https://github.com/catalystneuro/neuroconv/issues/42 and following the philsophy that things should fail early.

CodyCBakerPhD:
> Yes, I would like that. This is in the context of https://github.com/catalystneuro/neuroconv/issues/42 and following the philsophy that things should fail early.

OK, perhaps in a follow-up?

CodyCBakerPhD:
Or you could even restore that previous PR of yours implementing that

h-mayorquin:
Yes, I can take care of the other side. This is ready. I think you might have to modify the changelog (not sure, just double check).  Will you take a look at the array issue that you mentioned?

Also the run-live-service-test should be fixed in a bit, right? Maybe let's wait for that.

CodyCBakerPhD:
> Will you take a look at the array issue that you mentioned?

As mentioned the array issue will be resolved in a follow-up (if it actually can be without too much effort) - worst case scenario I just remove the commented blocks if Pydantic validation can't be easily made to apply. More evidence in favor of it being a non-holistic 'additional' type checker

h-mayorquin:
Yeah, that's my understanding that it would be done in a follow up, I was not implying or asking otherwise. Can you fix the changelog?

h-mayorquin (approved):
Oh, you did change the Chanelog already. Great.

Issue #465, opened 2023-06-01
refactor: change DataInterfaceTestMixin to only handle one input at a time


bendichter:
### What would you like to see added to NeuroConv?

Currently, there is a 1:1 relationship between a DataInterface class and a DataInterfaceTextMixin class. The DataInterfaceTextMixin class can handle multiple kwargs_inputs representing different test inputs to the same data interferface. This was originally chosen to be analogous to the way neo builds tests [example](https://github.com/NeuralEnsemble/python-neo/blob/29d73f68070fbd4388daec7bdbfc6a666fd53080/neo/test/rawiotest/test_blackrockrawio.py#L21-L30).

This decision has made the code more complex than it needs to be for several reasons:
* we often need to check metadata for different inputs, which requires some sort of if/else logic
* when a test fails, it is not always clear which input is causing the failure
* the DataInterfaceTextMixin class needs to handle different data types for kwargs_inputs, which is not ideal design and leads to unnecessary complexity

Proposed solution:
A DataInterfaceTextMixin.kwargs_inputs should only be a dictionary. If you want to test multiple different inputs, create a distinct class for each input.

I wrote the original test structure and I support this change.

### Do you have any interest in helping implement the feature?

No.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

PR #1024, opened 2024-08-22
[testing] Re-organize data tests showcase

h-mayorquin:
Currently for tests that are identical but tests different datasets (e.g video) we need to create an internal list and loop over it:

https://github.com/catalystneuro/neuroconv/blob/8e20a25ad30609aefffc5ad6f9b56354ee5e6d10/tests/test_on_data/test_behavior_interfaces.py#L306-L316

https://github.com/catalystneuro/neuroconv/blob/8e20a25ad30609aefffc5ad6f9b56354ee5e6d10/src/neuroconv/tools/testing/data_interface_mixins.py#L216-L220

This makes dififcult to time tests and figure out which specific dataset is problematic with common debuggin tools and the pytest framework.

This PR handles this with [pytest fixture parameterization](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#fixture-parametrize) and now tests are separated in pytest per case like this:

image

This PR limits itself to changing this for the VideoInterface and I will start propagating this to other modalities. At the end, the number of lines will be greatly reduced as that would eliminate a lot of duplicated logic.

### DIFF
--- a/tests/test_on_data/test_behavior_interfaces.py
+++ b/tests/test_on_data/test_behavior_interfaces.py
@@ -303,16 +303,34 @@ class TestFicTracDataInterfaceTiming(TemporalAlignmentMixin, unittest.TestCase):
     save_directory = OUTPUT_PATH
 
 
-class TestVideoInterface(VideoInterfaceMixin, unittest.TestCase):
+import pytest
+
+
+class TestVideoInterface(VideoInterfaceMixin):
     data_interface_cls = VideoInterface
-    interface_kwargs = [
-        dict(file_paths=[str(BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_avi.avi")]),
-        dict(file_paths=[str(BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_flv.flv")]),
-        dict(file_paths=[str(BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_mov.mov")]),
-        dict(file_paths=[str(BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_mp4.mp4")]),
-        dict(file_paths=[str(BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_wmv.wmv")]),
-    ]
     save_directory = OUTPUT_PATH
+
+    @pytest.fixture(
+        params=[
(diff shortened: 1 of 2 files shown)


CodyCBakerPhD:
fix #https://github.com/catalystneuro/neuroconv/issues/465

tests/test_on_data/test_behavior_interfaces.py
```diff
@@ -303,17 +303,35 @@ class TestFicTracDataInterfaceTiming(TemporalAlignmentMixin, unittest.TestCase):
...
+        params=[
+            (dict(file_paths=[str(BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_avi.avi")])),
+            (dict(file_paths=[str(BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_flv.flv")])),
+            (dict(file_paths=[str(BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_mov.mov")])),
+            (dict(file_paths=[str(BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_mp4.mp4")])),
+            (dict(file_paths=[str(BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_wmv.wmv")])),
+        ],
+        ids=["avi", "flv", "mov", "mp4", "wmv"],
+    )
+    def setup_interface(self, request):
```
CodyCBakerPhD, line 323: I think our eventual preferred syntax was actually separate classes altogether, like

```python
class TestVideoInterfaceAVI(VideoInterfaceMixin):
    interface_kwargs = {"file_path": str(BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_avi.avi")}
```

That way you don't need the extra `setup_interfaces` step, and you can attach separate docstrings to each class (I personally find that very useful when adding new test cases for bug fixes since I can link to the original issue in the docstring as a reminder)

Technically a little bit more writing, but better verbosity and less magic overall

I find pytest params are great for expanding matrices of inputs and input types in simple functions (like testing both `str` and `path` inputs in cases like this) but I think it's overkill here

h-mayorquin: I am fine with that too. It is also necessary when the test of the interface will check for metadata.

But I will just do this refactor so the diff is minimal and then we can break classes with parameterized in multiple classes after that.

PR #1026, opened 2024-08-23
[testing] Re-organize data tests around the one dataset per test

### LINKED ISSUE #465: refactor: change DataInterfaceTestMixin to only handle one input at a time

h-mayorquin:
The complete version of #1024.

Should close #465.

### DIFF
--- a/src/neuroconv/tools/testing/data_interface_mixins.py
+++ b/src/neuroconv/tools/testing/data_interface_mixins.py
@@ -8,7 +8,7 @@ from pathlib import Path
 from typing import Literal, Optional, Type, Union
 
 import numpy as np
-from hdmf.testing import TestCase as HDMFTestCase
+import pytest
 from hdmf_zarr import NWBZarrIO
 from jsonschema.validators import Draft7Validator, validate
 from numpy.testing import assert_array_equal
@@ -35,14 +35,11 @@ from neuroconv.tools.nwb_helpers import (
 )
 from neuroconv.utils import NWBMetaDataEncoder
 
-from .mock_probes import generate_mock_probe
-
 
 class DataInterfaceTestMixin:
     """
     Generic class for testing DataInterfaces.
 
-    This mixin must be paired with unittest.TestCase.
 
     Several of these tests are required to be run in a specific order. In this case,
     there is a `test_conversion_as_lone_interface` that calls the `check` functions in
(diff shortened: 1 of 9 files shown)


src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -213,59 +215,52 @@ def run_custom_checks(self):
...
     Generic class for testing temporal alignment methods.
-
-    This mixin must be paired with a unittest.TestCase class.
     """
 
     data_interface_cls: Type[BaseDataInterface]
-    interface_kwargs: Union[dict, list[dict]]
+    interface_kwargs: dict
+    save_directory: Path = Path(tempfile.mkdtemp())
+    conversion_options: dict = dict()
```
CodyCBakerPhD, line 252: Minor: should this class set a mutable default?

h-mayorquin: Yes, you are right!

CodyCBakerPhD (approved):

CodyCBakerPhD:
OK - followup will enhance readability / per case customizability?

h-mayorquin:
> OK - followup will enhance readability / per case customizability?

Yes, I broke some of the classes already. I also want to make some of the checks test on themselves so the failures are pint-pointed more precisely.

CodyCBakerPhD:
Sounds good 👍

Issue #1030, opened 2024-08-26
Extract more information from the `config.yaml`  file with `DeepLabCutInterface` 


h-mayorquin:
Currently, we are only extracting the task as a session description and experimenter:

https://github.com/catalystneuro/neuroconv/blob/727228c8bd871d0ffdcca48ac76a2010f293652d/src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py#L55-L66

We will make this optional.

We are on a crunch now for a project but in a future project that uses DeepLabCut's  `config.yaml` we should figure out what more metadata could be extracted from those configuration files. From the test data it seems that at least the names of the body parts could be incluced as metadata somewhere.

This is the `config.yaml` file contents from the test data that we have on gin:

```yaml
    # Project definitions (do not edit)
Task: openfield
scorer: Pranav
date: Aug20
multianimalproject: false
identity:

    # Project path (change when moving around)
project_path: /Users/sakshamsharda/Documents/NWB/GIN/behavior_testing_data/DLC

    # Annotation data set configuration (and individual video cropping parameters)
video_sets:
  /Data/openfield-Pranav-2018-08-20/videos/m1s1.mp4:
    crop: 0, 640, 0, 480
  /Data/openfield-Pranav-2018-08-20/videos/m1s2.mp4:
    crop: 0, 640, 0, 480
  /Data/openfield-Pranav-2018-08-20/videos/m2s1.mp4:
    crop: 0, 640, 0, 480
  /Data/openfield-Pranav-2018-08-20/videos/m3s1.mp4:
    crop: 0, 640, 0, 480
  /Data/openfield-Pranav-2018-08-20/videos/m3s2.mp4:
    crop: 0, 640, 0, 480
  /Data/openfield-Pranav-2018-08-20/videos/m4s1.mp4:
    crop: 0, 640, 0, 480
  /Data/openfield-Pranav-2018-08-20/videos/m5s1.mp4:
    crop: 0, 800, 0, 800
  /Data/openfield-Pranav-2018-08-20/videos/m6s1.mp4:
    crop: 0, 800, 0, 800
  /Data/openfield-Pranav-2018-08-20/videos/m6s2.mp4:
    crop: 0, 800, 0, 800
  /Data/openfield-Pranav-2018-08-20/videos/m7s1.mp4:
    crop: 0, 800, 0, 800
  /Data/openfield-Pranav-2018-08-20/videos/m7s2.mp4:
[...]

PR #1031, opened 2024-08-26
Make `config_file_path` optional on `DeepLabCutInterface`

h-mayorquin:
Needed for the Cohen conversion where the files are (hopefully temporarily) missing. Plus, as described in https://github.com/catalystneuro/neuroconv/issues/1030 we don't really extract any critical information from them.


CodyCBakerPhD:
Can you add a small test case with the existing example data really quick?

h-mayorquin:
> Can you add a small test case with the existing example data really quick?

What kind of test are you envisioning? That it does not fail when you pass None instead of a file_path?

CodyCBakerPhD:
A showcase of w/e new behavior you are adding in this PR insofar as it is different from the current test cases

h-mayorquin:
The new behavior is that you don't require to pass a configuration file path. Would the test I mentioned cover it?

CodyCBakerPhD:
Possibly - unless there is logic in DLC interface for 'searching' for a nearby config, in which case you'd have to making a tmpdir copy of example data w/o the config file

h-mayorquin:
> Possibly - unless there is logic in DLC interface for 'searching' for a nearby config, in which case you'd have to making a tmpdir copy of example data w/o the config file

There is no such a thing.

h-mayorquin:
This was a good suggestion @CodyCBakerPhD. I found out another nested use of the configuration file deep inside the call stack of the utils.

h-mayorquin:
Will need this:
#1047

CodyCBakerPhD (approved):

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
```diff
@@ -220,7 +238,7 @@ def _get_video_info_from_config_file(config_file_path: Path, vidname: str):
             break
 
     if video is None:
-        warnings.warn(f"The corresponding video file could not be found...")
+        warnings.warn(f"The corresponding video file could not be foundd in the config file")
```
h-mayorquin, line 241: Typo here!

h-mayorquin, line 241:
```suggestion
        warnings.warn(f"The corresponding video file could not be found in the config file")
```

tests/test_on_data/behavior/test_behavior_interfaces.py
```diff
@@ -365,6 +365,32 @@ def check_read_nwb(self, nwbfile_path: str):
...
+            assert "PoseEstimation" in processing_module_interfaces
+
+            pose_estimation_series_in_nwb = processing_module_interfaces["PoseEstimation"].pose_estimation_series
+            expected_pose_estimation_series = ["ind1_leftear", "ind1_rightear", "ind1_snout", "ind1_tailbase"]
+
+            expected_pose_estimation_series_are_in_nwb_file = [
+                pose_estimation in pose_estimation_series_in_nwb for pose_estimation in expected_pose_estimation_series
+            ]
+
+            assert all(expected_pose_estimation_series_are_in_nwb_file)
```
bendichter, line 391:
```suggestion
            assert all(
                pose_estimation in pose_estimation_series_in_nwb for pose_estimation in ["ind1_leftear", "ind1_rightear", "ind1_snout", "ind1_tailbase"]
            )
```

this would be clearer to me

h-mayorquin: This is a pet peeve of mine that I think makes debuggin easier so I would like to explain myself.

The reason that I like avoid definitions inside of other scopes is that it gets on the way of the standard debuggin workflow.  With the current code If the test fails, I just set a debuggin point in line 390 (`pdb.set_trace()`) and then I can check immediatley what is the value of `expected_pose_estimation_series_are_in_nwb_file` and think about what failed. With the proposed change I would need to evaluate the expression again to inspect it.

Moreover, [before python 3.12 list comprehension is debugged as a separate frame](https://docs.python.org/3/whatsnew/3.12.html#pep-709-comprehension-inlining) so if I use the automated tools of my IDE I can't be in a frame that has access to both values of interest: the values inside the list comprehension and the context from which is called.

All of that said I think the first definition in line 385 can be eliminated to make this more concise but I like having all the defintions before the assertions/checks so I can debug everything in the same frame.

bendichter: ok, I rarely use pdb.set_trace, so I mostly go by readability, but I can see how it could get in the way for you. Feel free to ignore

PR #1033, opened 2024-08-27
[Pydantic IV] Restore soft deprecation of old types

CodyCBakerPhD:
As per meeting today

@h-mayorquin after this you are probably good for a fresh release


src/neuroconv/utils/types.py
CodyCBakerPhD, line 1: listed in both locations since you can do (and have always been able to do) either `from neuroconv.utils` or `from neuroconv.utils.types`

h-mayorquin (approved):
Thanks for the quick fix:

image

PR #1042, opened 2024-08-29
Added chunking/compression for string-only objects

### LINKED ISSUE #1041: [Feature]: Add Compression support for compound objects

`DatasetIOConfiguration.from_neurodata_object` currently raises a `NotImplementedError` for compound objects with dtype 'object'. 

This makes it very difficult to use the `KiloSortSortingInterface`, which includes a compound object for unit quality ('good' or 'noise').

_No response_

Yes.

### Code of Conduct

pauladkisson:
Addresses #1041


h-mayorquin:
OK, this should not be throwing an error. Can you just pass a configuration with no chukning or compression for those objects?

I also need to take a look on the gin dataset from phy.

pauladkisson:
> Can you just pass a configuration with no chukning or compression for those objects?

Passed in no compression (chunk_shape is required to pass validation, so I kept that) and removed the warning.

h-mayorquin:
Yeah, I remember that I discussed that issue about the chunk_shape and full_shape being mandatory. I think that the argument was that we should not being creating configurations for those cases at all. Does that ring a bell, @CodyCBakerPhD ?

I think his idea was it should be skipped at a higher level like `get_default_dataset_io_configurations` where  `from_neurodata_object` is called.

pauladkisson:
> I think that the argument was that we should not being creating configurations for those cases at all.

I really don't see why we should disallow this, esp since the docs seem to imply that setting these properties to None is the way to disable chunking/compression while still providing a backendConfiguration: https://neuroconv.readthedocs.io/en/main/user_guide/backend_configuration.html

h-mayorquin:
OK, I am weary of this when if the array is too large because we will be creating chunks that are too large but I think that this will be an edge case. In a future PR let's think about how to differentiate objects that are strings (the motivation of this PR) from things that are real objects (I am not even sure we can write the latter to NWB)

Can you add a test for this for this behavior?

pauladkisson:
> OK, I am *wary of this when if the array is too large because we will be creating chunks that are too large but I think that this will be an edge case.

Maybe we should keep the warning then?

pauladkisson:
> Can you add a test for this for this behavior?

Done!

h-mayorquin:
This looks good. 

Here you are testing for arrays that have object type but are in fact string arrays. I would like to hear your thoughts about this comment: https://github.com/catalystneuro/neuroconv/issues/1041#issuecomment-2332668040. I want to move the code in that direction. Would you like to help? It can be here or in another PR but your test will need to be modified a bit. Otherwise, I can take care of it in another PR.


Also, can you unlink the issue so it is not closed when we merge?

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -264,6 +265,21 @@ def from_neurodata_object(cls, neurodata_object: Container, dataset_name: Litera
         full_shape = get_data_shape(data=candidate_dataset)
         dtype = _infer_dtype(dataset=candidate_dataset)
 
+        if dtype == np.dtype(
```
h-mayorquin, line 268: I would add this logic inside the last if in the branch below so it is a sub-case of object

h-mayorquin: I mean here:
https://github.com/catalystneuro/neuroconv/blob/562bf7ff5222ea6d90f8ba6b4f440b3afcd64078/src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py#L295-L302

h-mayorquin (approved):
LGTM

h-mayorquin:
Cody brought a new case to me that I want to consider before merging this. I will do some experimentation.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -277,10 +278,33 @@ def from_neurodata_object(cls, neurodata_object: Container, dataset_name: Litera
...
+                )
+                compression_method = "gzip"
+            else:
+                chunk_shape = full_shape  # validate_all_shapes fails if chunk_shape or buffer_shape is None
+                buffer_shape = full_shape
+                compression_method = None
+                warnings.warn(
+                    f"Default chunking and compression options for compound objects are not optimized. "
+                    f"Consider manually specifying DatasetIOConfiguration for dataset at '{location_in_file}'."
+                )
```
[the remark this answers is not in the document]
h-mayorquin: It is the same issue. If you use the latest version of spikeinterface (to be released soon) that problem should be solved.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -277,10 +278,33 @@ def from_neurodata_object(cls, neurodata_object: Container, dataset_name: Litera
...
             compression_method = "gzip"
         elif dtype == np.dtype("object"):  # Unclear what default chunking/compression should be for compound objects
-            raise NotImplementedError(
-                f"Unable to create a `DatasetIOConfiguration` for the dataset at '{location_in_file}'"
-                f"for neurodata object '{neurodata_object}' of type '{type(neurodata_object)}'!"
-            )
+            # pandas reads in strings as objects by default: https://pandas.pydata.org/docs/user_guide/text.html
+            max_string_length = 1
+            all_elements_are_strings = True
+            for element in candidate_dataset[:].flat:
```
h-mayorquin, line 284: You can do this kind of thing with list comprehension which are optimized and I also think is more readable.

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -277,10 +278,33 @@ def from_neurodata_object(cls, neurodata_object: Container, dataset_name: Litera
...
+                if not isinstance(element, str):
+                    all_elements_are_strings = False
+                    break
+                string_length = len(element)
+                if string_length > max_string_length:
```
h-mayorquin, line 289: Instead of estimating yourself the dtype,  I would get do numpy do the casting `np.array(data).dtype` and get the type from them. It would be less brittle I feel.

h-mayorquin (changes requested):
OK, I changed my mind about the compound objects for three reasons:
1)  I don't have time to test the troubling example this week and I need this https://github.com/catalystneuro/neuroconv/pull/1067 first.
2) I would do another patch release before the minor release. 

For the patch release I would like to include the "compound objects that are actually strings" changes of this PR.

I would suggest do the following:

Let's keep the code for objects that are real strings but still keep the previous behavior (throw an error) for compound types. 

This could be achieved by changing the warning to an error and temporarily commenting out the corresponding tests

I can take care of these modifications @pauladkisson but if you want to do it that's fine as well.

pauladkisson:
> For the patch release I would like to include the "compound objects that are actually strings" changes of this PR.
> I would suggest do the following:
> Let's keep the code for objects that are real strings but still keep the previous behavior (throw an error) for compound types.
> This could be achieved by changing the warning to an error and temporarily commenting out the corresponding tests

tbh I'd rather just wait and take care of all compound objects (including the problematic mixed str-NaN case) in this PR.
Is there a reason you need it in the upcoming patch?

pauladkisson:
> In short: I want to have the bug solved and I want to think about the complicated case more. They are two different issues so I would prefer them to address them piece-wise.

Ok, I am conviced. 👍

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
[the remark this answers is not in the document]
h-mayorquin: Ah, of course because I am using the latest spikeinterface version. If you don't have it, your array will be loaded as an object that will have Nan.

pauladkisson:
> > > In short: I want to have the bug solved and I want to think about the complicated case more. They are two different issues so I would prefer them to address them piece-wise.
> > 
> > 
> > Ok, I am conviced. 👍
> 
> Thanks but man I feel bad about my lack of clear thinking on this issue. I have made too many requests as I have changed my mind and learned thinking about this. I can take care of rebasing and organizing this PR so you don't have to if you want.

No worries.

h-mayorquin (approved):
Thanks again for all the work.

PR #1050, opened 2024-09-01
Make some methods private

h-mayorquin:
@CodyCBakerPhD 

These are methods that we don't want to make guarantee and we also don't believe should be on the API docs. I am making them private. We can change them if the needs come to be.


CodyCBakerPhD (approved):

CodyCBakerPhD:
Note that though this marks them as private, they are all still publically exposed to the submodules

PR #1053, opened 2024-09-02
unit table descriptions for phy and kilosort

pauladkisson:
This PR adds description fields for the unit table from phy and kilosort.


zm711:
Had this open on my computer and bumped into this PR. Just one question if I may. How will you deal with the fact that users can add columns to the TSV and they will be displayed? In this case you'll just ignore them right. Is there some way you could take any additional headers and I guess the description for any extra columns if added would be something like `user defined column unspecified` or something of that sort.

pauladkisson:
If users add columns to the .tsv, they should propagate into the units table, with description 'No description.'

Then, if they want to add a description, they can simply add an entry to the metadata yaml like so:
```yaml
Ecephys:
  Device:
    ...
  ElectrodeGroup:
    ...
  ElectricalSeries:
    ...
  UnitProperties:
    - name: <user_defined_column>
      description: <custom_description>
```

or update the metadata dictionary directly like so:
```python
metadata["Ecephys"]["UnitProperties"].append(dict(name=<user_defined_column>, description=<custom_description>))
```

zm711:
Perfect. Just wanted to doublecheck since we sometimes have issues on phy with people adding stuff and then wanting to understand how the information propagates. So just wanted to make sure it was covered her since they don't always exist :)

h-mayorquin (commented):
These things are almost the same extractor in another PR maybe we can think of doing some deduplication.

Can you add a test for this?

pauladkisson:
> These things are almost the same extractor in another PR maybe we can think of doing some deduplication.

👍

pauladkisson:
> Can you add a test for this?

Done.

src/neuroconv/datainterfaces/ecephys/kilosort/kilosortdatainterface.py
```diff
@@ -36,3 +36,31 @@ def __init__(
         verbose: bool, default: True
         """
         super().__init__(folder_path=folder_path, keep_good_only=keep_good_only, verbose=verbose)
+
+    def get_metadata(self):
+        metadata = super().get_metadata()
+        # See Kilosort save_to_phy() docstring for more info on these fields: https://github.com/MouseLand/Kilosort/blob/main/kilosort/io.py
+        # Or see phy documentation: https://github.com/cortex-lab/phy/blob/master/phy/apps/base.py
+        metadata["Ecephys"]["UnitProperties"] = [
+            dict(name="units", description="Autogenerated by neuroconv."),
```
h-mayorquin, line 45: Wait, what is this one? Is it for the name of the units table?

h-mayorquin:
This looks good. Three more requests:
1) An explanation of the "units" description
2) Changelog.
3) Can you add a test that checks that the `UnitDescriptions` actually end up as a column descriptions of the units table? Something like this:

```
class TestPhySortingInterface(SortingExtractorInterfaceTestMixin):
    data_interface_cls = PhySortingInterface
    interface_kwargs = dict(folder_path=str(DATA_PATH / "phy" / "phy_example_0"))
    save_directory = OUTPUT_PATH

    
    def check_units_table_description_propagation():

        interface = self.data_interface_cls(**self.interface_kwargs)
        unit_descriptions = metadata["Ecephys"]["UnitDescriptions"]
        nwbfile = interface.create_nwbfile()
        
        units_table = nwbfile.units       
        for property in ...  
```

In another PR we can propagate this as a general test in the sorting mixing.

src/neuroconv/datainterfaces/ecephys/kilosort/kilosortdatainterface.py
pauladkisson: For some reason, I thought I needed to include this, but it was an artifact from how I was reading the UnitProperties for Schneider Conversion.

pauladkisson:
> This looks good. Three more requests:

Done!

h-mayorquin (approved):
LGTM

I will do a small PR after this to have discussion about testing in general where I will like your feedback.

PR #1055, opened 2024-09-04
[Pydantic IVb] Pydantic validation on arrays

CodyCBakerPhD:
Follow up to add remaining validation calls for interfaces using arrays


CodyCBakerPhD:
@h-mayorquin I think this is resolved now

h-mayorquin (approved):
OK, this makes sense. Maybe we should add these to all the validate calls so no one needs to encounter this error if they decide to use a non pydantic type in the future.

I wish there was some kind of global configuration on this case or that we could call the validate decorator in the BaseDataInterface

PR #1056, opened 2024-09-06
Enable zarr backend testing in data tests [1]

h-mayorquin:
This enables zarr testing for running conversions with backend specified as a string (still missing the building a backend configuration part)

It also makes some metadata checks proper tests on their own which has two advantages:
* They will fail on their own and be identified as such by pytest
* Can be parallelized with xdist.

### DIFF
--- a/src/neuroconv/tools/testing/data_interface_mixins.py
+++ b/src/neuroconv/tools/testing/data_interface_mixins.py
@@ -220,7 +220,7 @@ class DataInterfaceTestMixin:
     @pytest.mark.parametrize("backend", ["hdf5", "zarr"])
     def test_run_conversion_with_backend(self, setup_interface, tmp_path, backend):
 
-        nwbfile_path = str(tmp_path / f"conversion_with_backend{backend}_{self.test_name}.nwb")
+        nwbfile_path = str(tmp_path / f"conversion_with_backend{backend}-{self.test_name}.nwb")
 
         metadata = self.interface.get_metadata()
         if "session_start_time" not in metadata["NWBFile"]:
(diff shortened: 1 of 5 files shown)


src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -223,6 +217,26 @@ def run_custom_checks(self):
         """Override this in child classes to inject additional custom checks."""
         pass
 
+    @pytest.mark.parametrize("backend", ["hdf5", "zarr"])
+    def test_run_conversion_with_backend(self, setup_interface, tmp_path, backend):
+
+        nwbfile_path = str(tmp_path / f"conversion_with_backend{backend}_{self.test_name}.nwb")
```
CodyCBakerPhD, line 223:
```suggestion
        nwbfile_path = str(tmp_path / f"conversion_with_backend-{backend}_{self.test_name}.nwb")
```

h-mayorquin: you prefer `-` over `_`?

CodyCBakerPhD: whichever you prefer

I kind of tend to mirror what DANDI does and `_` separate thematically similar blocks, with `-` separation within a block

h-mayorquin: I really don't have a strong preference, I will add the commit, just curios.

h-mayorquin: Changed

h-mayorquin:
@CodyCBakerPhD OK, we got the point were the failures are actually related to zarr, check out the video test failure:

https://github.com/catalystneuro/neuroconv/actions/runs/10744684959/job/29802201096?pr=1056

Does it ring any bell? I will dig into it.

h-mayorquin:
Note for myself to also add the documentation:

#1019

CodyCBakerPhD:
Don't believe I've seen this one myself, but from the message the path to the Zarr array (file): https://github.com/catalystneuro/neuroconv/actions/runs/10744684959/job/29802201096?pr=1056#step:31:3426

is not a valid Windows path (should have replaced the `/` with `\\`?)

h-mayorquin:
@CodyCBakerPhD 
OK, this should be ready but it requires the fix on #1070 before (as does everything else).

The cause of the Windows error is here:

https://github.com/hdmf-dev/hdmf-zarr/issues/219

h-mayorquin:
@CodyCBakerPhD this is ready to go.

CodyCBakerPhD (approved):

CodyCBakerPhD:
Coolio

PR #1057, opened 2024-09-06
Fix a bug in `add_sorting_to_nwbfile` where `unit_electrode_indices` was only propagated if `waveform_means` was passed

h-mayorquin:
Fixed this bug and added a test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Upcoming
 
+### Bug fixes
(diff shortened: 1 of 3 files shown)


alejoe91:
@h-mayorquin I think that was done on purpose. The `electrode_indices` column is a map for the waveforms field. No?

h-mayorquin:
> @h-mayorquin I think that was done on purpose. The `electrode_indices` column is a map for the waveforms field. No?

The documentation indicates it should map to the electrodes (in this channels) that the unit came from:

https://pynwb.readthedocs.io/en/stable/pynwb.file.html#pynwb.file.NWBFile.add_unit

Am I misunderstanding something?

alejoe91:
Makes sense :) My interpretation was too strict then!

alejoe91 (approved):

PR #1061, opened 2024-09-06
Enable testing arm64 Mac architecture in the CI.

### LINKED ISSUE #1054: [Feature]: Add arm64 Mac testing

Just noticed that the testing matrix does not include `macos-latest` which uses arm64 (not intel) runners: https://github.com/catalystneuro/neuroconv/blob/main/.github/workflows/testing.yml#L23

_No response_

Yes.

### Code of Conduct

h-mayorquin:
As in the title. Shoul fix #1054

Let's get the ball rolling on this.

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
+++ b/src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
@@ -379,7 +379,9 @@ def add_subject_to_nwbfile(
     video_name, scorer = h5file.stem.split("DLC")
     scorer = "DLC" + scorer
 
-    df = _ensure_individuals_in_header(pd.read_hdf(h5file), individual_name)
+    data_frame_from_hdf5 = pd.read_hdf(h5file)
+
+    df = _ensure_individuals_in_header(data_frame_from_hdf5, individual_name)
 
     # Note the video here is a tuple of the video path and the image shape
     if config_file is not None:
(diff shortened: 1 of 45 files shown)


h-mayorquin:
OK, this is working, we need somehow to skip the doc test for DeepLabCut with M1s and python=3.9. On the other hand, if we follow the numpy convention we should stop supporting python 3.9 now so maybe we don't need to do that.

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
```diff
@@ -379,7 +379,9 @@ def add_subject_to_nwbfile(
     video_name, scorer = h5file.stem.split("DLC")
     scorer = "DLC" + scorer
 
-    df = _ensure_individuals_in_header(pd.read_hdf(h5file), individual_name)
+    data_frame_from_hdf5 = pd.read_hdf(h5file)
```
h-mayorquin, line 382:
```suggestion
```

pauladkisson:
> But this also has the support for the intel mac which I think was only discontinued from pros two years ago or something? To be honest, I would prefer to test only on latest mac, the CI is already super bloated on my view but I think there might be a case here for the two.
> 
> How do you feel about it?

I am still using an intel mac, so I would like the tests to include it. Plus, I think they're still relatively common. I agree that the CI is a bit bloated and takes too long to pick up new jobs, but I don't think this is the place to make cuts.

h-mayorquin:
>  but I don't think this is the place to make cuts.
This make sense. 

Let's brainstorm how to improve the CI at some point.

bendichter (approved):

PR #1063, opened 2024-09-07
Add more docstrings to public methods

h-mayorquin:
Should come after #1062

This was the most annoying one.


src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
```diff
@@ -511,6 +511,33 @@ def __init__(self, file_path: FilePath, verbose: bool = True):
...
+        - The method reads the `.session.mat` file using `pymatreader` and extracts `extracellular` data.
+        - It creates a dummy recording extractor using `spikeinterface.core.numpyextractors.NumpyRecording`.
+        - The generated extractor includes channel IDs and other relevant metadata such as number of channels,
+        number of samples, and sampling frequency.
+        - Channel metadata is added to the dummy extractor using the `add_channel_metadata_to_recoder` function.
+        - If the `.session.mat` file is not found, no extractor is returned.
+
+        Warnings
+        --------
+        Ensure that the `.session.mat` file is correctly located in the expected session path, or the method will not generate
```
pauladkisson, line 536: I think it is worthwhile to explicitly specify the expected path: self.session_path / f"{self.session_id}.session.mat"

src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py
```diff
@@ -127,14 +148,15 @@ def get_metadata(
...
+    def get_original_timestamps(self) -> np.ndarray:  # noqa: D102
         reinitialized_extractor = self.get_extractor()(**self.source_data)
         return reinitialized_extractor.frame_to_time(frames=np.arange(stop=reinitialized_extractor.get_num_frames()))
 
-    def get_timestamps(self) -> np.ndarray:
+    def get_timestamps(self) -> np.ndarray:  # noqa: D102
         return self.imaging_extractor.frame_to_time(frames=np.arange(stop=self.imaging_extractor.get_num_frames()))
 
     def set_aligned_timestamps(self, aligned_timestamps: np.ndarray):
+        """Replace all timestamps for this interface with those aligned to the common session start time."""
```
pauladkisson, line 159: This needs a Parameters section for aligned_timestamps

src/neuroconv/datainterfaces/ophys/basesegmentationextractorinterface.py
```diff
@@ -89,23 +110,24 @@ def get_metadata_schema(self) -> dict:
...
         return reinitialized_extractor.frame_to_time(frames=np.arange(stop=reinitialized_extractor.get_num_frames()))
 
-    def get_timestamps(self) -> np.ndarray:
+    def get_timestamps(self) -> np.ndarray:  # noqa: D102
         return self.segmentation_extractor.frame_to_time(
             frames=np.arange(stop=self.segmentation_extractor.get_num_frames())
         )
 
     def set_aligned_timestamps(self, aligned_timestamps: np.ndarray):
+        """set the aligned timestamps for the segmentation extractor."""
```
pauladkisson, line 130: Again here, needs Parameters section for aligned_timestamps

src/neuroconv/datainterfaces/ophys/brukertiff/brukertiffconverter.py
```diff
@@ -187,6 +223,22 @@ def add_to_nwbfile(
...
+        nwbfile : NWBFile
+            The NWBFile object to which the data will be added.
+        metadata : dict
+            Metadata dictionary containing information about the data to be added.
+        stub_test : bool, optional
+            If True, only a subset of the data (defined by `stub_frames`) will be added for testing purposes,
+            by default False.
+        stub_frames : int, optional
+            The number of frames to include in the subset if `stub_test` is True, by default 100.
+
```
pauladkisson, line 240: remove empty line

src/neuroconv/datainterfaces/ophys/brukertiff/brukertiffdatainterface.py
```diff
@@ -28,6 +29,7 @@ def get_streams(
         folder_path: DirectoryPath,
         plane_separation_type: Literal["contiguous", "disjoint"] = None,
     ) -> dict:
+        """get streams for the Bruker TIFF imaging data."""
```
pauladkisson, line 32: Needs Parameters section for folder_path and plane_separation_type

src/neuroconv/datainterfaces/ophys/brukertiff/brukertiffdatainterface.py
```diff
@@ -191,6 +195,7 @@ def get_source_schema(cls) -> dict:
 
     @classmethod
     def get_streams(cls, folder_path: DirectoryPath) -> dict:
+        """get streams for the Bruker TIFF imaging data."""
```
pauladkisson, line 198: Needs Parameters section

src/neuroconv/datainterfaces/ophys/miniscope/miniscopeimagingdatainterface.py
```diff
@@ -74,11 +76,13 @@ def get_metadata(self) -> DeepDict:
         return metadata
 
     def get_metadata_schema(self) -> dict:
+        """Get the metadata schema for the Miniscope imaging data."""
         metadata_schema = super().get_metadata_schema()
         metadata_schema["properties"]["Ophys"]["definitions"]["Device"]["additionalProperties"] = True
         return metadata_schema
 
     def get_original_timestamps(self) -> np.ndarray:
+        """Get the original timestamps from the Miniscope data."""
```
pauladkisson, line 85: Needs a Returns section

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterfaces.py
```diff
@@ -87,6 +88,7 @@ class ScanImageLegacyImagingInterface(BaseImagingExtractorInterface):
 
     @classmethod
     def get_source_schema(cls) -> dict:
+        """ " "Get the source schema for the ScanImage legacy imaging interface."""
```
pauladkisson, line 91: Looks like a typo with the extra "s

src/neuroconv/datainterfaces/ophys/suite2p/suite2pdatainterface.py
```diff
@@ -140,6 +142,31 @@ def add_to_nwbfile(
...
+        stub_test : bool, optional
+            If True, only a subset of the data (defined by `stub_frames`) will be added for testing purposes,
+            by default False.
+        stub_frames : int, optional
+            The number of frames to include in the subset if `stub_test` is True, by default 100.
+        include_roi_centroids : bool, optional
+            Whether to include the centroids of regions of interest (ROIs) in the data, by default True.
+        include_roi_acceptance : bool, optional
+            Whether to include acceptance status of ROIs, by default True.
+        mask_type : str, optional
```
pauladkisson, line 163: mask_type should include the full info from basesegmentationinterface:
```python
mask_type : str, default: 'image'
            There are three types of ROI masks in NWB, 'image', 'pixel', and 'voxel'.

            * 'image' masks have the same shape as the reference images the segmentation was applied to, and weight each pixel
              by its contribution to the ROI (typically boolean, with 0 meaning 'not in the ROI').
            * 'pixel' masks are instead indexed by ROI, with the data at each index being the shape of the image by the number
              of pixels in each ROI.
            * 'voxel' masks are instead indexed by ROI, with the data at each index being the shape of the volume by the number
              of voxels in each ROI.

            Specify your choice between these two as mask_type='image', 'pixel', 'voxel', or None.
            If None, the mask information is not written to the NWB file.
```

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
```diff
@@ -24,21 +24,33 @@ def __init__(
...
-
         self._read_kwargs = read_kwargs
         self.dataframe = self._read_file(file_path, **read_kwargs)
         self.time_intervals = None
 
     def get_metadata(self) -> dict:
+        """
+        Get metadata for the time intervals.
+
+        Returns
```
pauladkisson, line 49: This get_metadata fn has a Returns section, but all the others do not.

We should probably pick one style and stick with it.

src/neuroconv/nwbconverter.py
```diff
@@ -166,13 +166,34 @@ def create_nwbfile(self, metadata: Optional[dict] = None, conversion_options: Op
...
+        nwbfile : NWBFile
+            The NWB file object with the data from the data interfaces added to it.
+        """
         conversion_options = conversion_options or dict()
         for interface_name, data_interface in self.data_interface_objects.items():
             data_interface.add_to_nwbfile(
                 nwbfile=nwbfile, metadata=metadata, **conversion_options.get(interface_name, dict())
             )
 
+        return nwbfile
```
pauladkisson, line 195: I'm not sure why you added a return value to add_to_nwbfile (it may need one, idk), but it definitely should not be in this PR.

src/neuroconv/tools/hdmf.py
```diff
@@ -50,6 +50,8 @@ def estimate_default_chunk_shape(chunk_mb: float, maxshape: tuple[int, ...], dty
     def estimate_default_buffer_shape(
         buffer_gb: float, chunk_shape: tuple[int, ...], maxshape: tuple[int, ...], dtype: np.dtype
     ) -> tuple[int, ...]:
+        """ "Add docstring to this"""
```
pauladkisson, line 53: lets do it then

pauladkisson (changes requested):
Thank you @h-mayorquin for working through all of this!

One point of discussion is should we include Returns sections for standard methods like `get_metadata`, `get_metadata_schema`, etc.? Or should we just have them be 1-liners.

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
h-mayorquin: That's a good point.

src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py
h-mayorquin: Thanks.

src/neuroconv/tools/hdmf.py
h-mayorquin: x D

Yep!

src/neuroconv/nwbconverter.py
h-mayorquin: Agree.

src/neuroconv/datainterfaces/ecephys/cellexplorer/cellexplorerdatainterface.py
h-mayorquin: Done, will be up in the next commit.

src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py
h-mayorquin: Changed to noqa, this docstring is identical and should be inherited from the base class.

src/neuroconv/datainterfaces/ophys/basesegmentationextractorinterface.py
h-mayorquin: Changed to noqa, this docstring is identical and should be inherited from the base class:

https://github.com/catalystneuro/neuroconv/blob/0d5d6ccb430e545b4b899e861876d600e04c5fbf/src/neuroconv/basetemporalalignmentinterface.py#L42-L56

src/neuroconv/datainterfaces/ophys/brukertiff/brukertiffconverter.py
h-mayorquin: Done.

src/neuroconv/datainterfaces/ophys/brukertiff/brukertiffdatainterface.py
h-mayorquin: Done.

src/neuroconv/datainterfaces/ophys/brukertiff/brukertiffdatainterface.py
h-mayorquin: Done

src/neuroconv/datainterfaces/ophys/miniscope/miniscopeimagingdatainterface.py
h-mayorquin: Same, noqa and should inherit from base class.

src/neuroconv/datainterfaces/ophys/scanimage/scanimageimaginginterfaces.py
h-mayorquin: Done

src/neuroconv/datainterfaces/ophys/suite2p/suite2pdatainterface.py
h-mayorquin: Thanks, should be done in the next commit.

src/neuroconv/datainterfaces/text/timeintervalsinterface.py
h-mayorquin: Also changing the noqa, would be easier to unify at the base class and then propagate speicifities if needed.

src/neuroconv/nwbconverter.py
h-mayorquin: Done

h-mayorquin:
> ne point of discussion is should we include Returns sections for standard methods like get_metadata, get_metadata_schema, etc.? Or should we just have them be 1-liners.

I am fine with adding it, but this is just a minimal effort to get the ruff passing and then we can improve piecewise and unify.

bendichter:
When a method in a child class is overriding a parent class and does not have a docstring, it takes on the docstring of the parent. I think this works fine and in many cases makes it unnecessary to create a new docstring. However, it looks like this ruff inspector does not respect this and requires docstrings to be defined even on child classes where the parent has an appropriate docstring defined. Is that correct? If that's the case, I think we should be able to write a custom check that will work better for us.

bendichter:
e.g. 

```python
import importlib
import inspect
import pkgutil

def check_package_docstrings(package_name):
    """
    Check if all public methods of public classes in public modules of a package have docstrings.

    Args:
    package_name (str): The name of the package to check.

    Returns:
    dict: A dictionary containing information about missing docstrings.
    """
    missing_docstrings = {}
    package = importlib.import_module(package_name)
    
    for _, module_name, _ in pkgutil.walk_packages(package.__path__, package.__name__ + '.'):
        try:
            module = importlib.import_module(module_name)
            
            for name, obj in inspect.getmembers(module):
                if inspect.isclass(obj) and not name.startswith('_'):
                    class_missing = check_class_docstrings(obj)
                    if class_missing:
                        missing_docstrings[f"{module_name}.{name}"] = class_missing
        
        except ImportError as e:
            print(f"Error importing module {module_name}: {e}")
    
    return missing_docstrings

def check_class_docstrings(cls):
    """
    Check if all public methods of a class have docstrings.

    Args:
    cls (type): The class to check.

    Returns:
    list: A list of method names without docstrings.
    """
    missing = []
    for name, method in inspect.getmembers(cls, inspect.isfunction):
        if not name.startswith('_') and not method.__doc__:
            missing.append(name)
    return missing

# Example usage
if __name__ == "__main__":
    package_name = "your_package_name"
    result = check_package_docstrings(package_name)
    
    if result:
        print("Missing docstrings found:")
        for class_name, methods in result.items():
            print(f"  {class_name}: {', '.join(methods)}")
    else:
        print("All public methods have docstrings.")
```

h-mayorquin:
> When a method in a child class is overriding a parent class and does not have a docstring, it takes on the docstring of the parent. I think this works fine and in many cases makes it unnecessary to create a new docstring. However, it looks like this ruff inspector does not respect this and requires docstrings to be defined even on child classes where the parent has an appropriate docstring defined. Is that correct? If that's the case, I think we should be able to write a custom check that will work better for us.

Yeah, that's a downside. The approach that I am taking here is just annotating it like this so it gets ignored:

image

This is good because it forces one to be explicit about it (you add it when you know the docstrings should be inherited) and is less complex than implementing something on our own.

bendichter:
I tend to side more on not adding chores for devs. They tend to pile up in a way that creates a lot of busy work and makes it difficult to onboard people and is unwelcoming to new contributors. In this particular case I suppose I could go either way

pyproject.toml
```diff
@@ -131,6 +131,7 @@ select = [
     "F401",  # Unused import
     "I",  # All isort rules
     "D101",  # Missing docstring in public class
+    "D102",  # Missing docstring in public method
```
h-mayorquin, line 134:
```suggestion
```
Removing this line
@pauladkisson this eliminates the check

We still have the problem that I added many #noqa in this PR.

pauladkisson: Done! Now this is just docstring improvements.

h-mayorquin:
@pauladkisson this is ready to go. It is only docstring improvements now and I changed the title accordingly.

pauladkisson (approved):
Looks good!

PR #1070, opened 2024-09-10
Fix paths

h-mayorquin:
Granular version without the verbose.


pauladkisson (changes requested):
Needs changelog entry but otherwise looks good.

h-mayorquin:
Done.

pauladkisson (approved):

PR #1076, opened 2024-09-11
Make `get_extractor` work with `MockImagingInterface`

h-mayorquin:
These changes are necessary so the `get_extractor` method works with ` MockImagingInterface`

https://github.com/catalystneuro/neuroconv/blob/9a886587aece41b3be0c5951b2d6ed54ece83304/src/neuroconv/baseextractorinterface.py#L21-L31

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # Upcoming
 
 ## Bug Fixes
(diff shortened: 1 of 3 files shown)


pauladkisson (changes requested):
Needs changelog but otherwise looks good.  Makes much more sense to me in a separate PR too, thx.

h-mayorquin:
Added changelog.

pauladkisson (approved):

PR #1078, opened 2024-09-11
Only run changelog request when status is not draft

h-mayorquin:
As in the title. 

@pauladkisson Here the idea is that the changelog check is only enforced when someone marks the PR as ready for review and from then onwards. This achieves the following which I find desirable:

* The PR should not be marked immediately as a failure as it is now when is on draft.
* PRs also change when they are in draft and the right point to write a changelog is when the PR is ready for review.
* It will save runners on the CI and currently, we are getting too little on them and waiting too long on our workflows.

### DIFF
--- a/.github/workflows/deploy-tests.yml
+++ b/.github/workflows/deploy-tests.yml
@@ -17,7 +17,7 @@ jobs:
 
   detect-changelog-updates:
     needs: assess-file-changes
-    if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' && github.event.pull_request.draft == 'false' }}
+    if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' && github.event.pull_request.draft == false }}
     name: Auto-detecting CHANGELOG.md updates
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/deploy-tests.yml
+++ b/.github/workflows/deploy-tests.yml
@@ -2,6 +2,7 @@ name: Deploy tests
 
 on:
   pull_request:
+    types: [synchronize, opened, reopened, ready_for_review]
   merge_group:
   workflow_dispatch:
 
@@ -16,7 +17,7 @@ jobs:
 
   detect-changelog-updates:
     needs: assess-file-changes
-    if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
+    if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' && github.event.pull_request.draft == 'false' }}
     name: Auto-detecting CHANGELOG.md updates
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/deploy-tests.yml
+++ b/.github/workflows/deploy-tests.yml
@@ -2,7 +2,7 @@ name: Deploy tests
 
 on:
   pull_request:
-    types: [synchronize, opened, reopened, ready_for_review]
+    types: [synchronize, opened, reopened, ready_for_review] # defaults + ready_for_review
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-tests.yml
```diff
@@ -2,6 +2,7 @@ name: Deploy tests
 
 on:
   pull_request:
+    types: [synchronize, opened, reopened, ready_for_review]
```
pauladkisson, line 5: What type(s) does this exclude?

pauladkisson (approved):
Agree on all counts, seems like a good idea.

.github/workflows/deploy-tests.yml
h-mayorquin: Yes, I should have added a comment.  
The first three types are the defaults and  `ready_for_the_review` is added on top of that so that the changelog test happens when the button "ready for review" is clicked.

Here is the full list:
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request

h-mayorquin:
OK, tests successfully skipped on draft:

image

.github/workflows/deploy-tests.yml
pauladkisson, line 5:
```suggestion
    types: [synchronize, opened, reopened, ready_for_review] # defaults + ready_for_review
```

.github/workflows/deploy-tests.yml
```diff
@@ -16,7 +17,7 @@ jobs:
 
   detect-changelog-updates:
     needs: assess-file-changes
-    if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
+    if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' && github.event.pull_request.draft == 'false' }}
```
h-mayorquin, line 20:
```suggestion
    if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' && github.event.pull_request.draft == false }}
```

.github/workflows/deploy-tests.yml
```diff
@@ -16,7 +17,7 @@ jobs:
 
   detect-changelog-updates:
     needs: assess-file-changes
-    if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' }}
+    if: ${{ needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true' && github.event.pull_request.draft == false }}
```
h-mayorquin, line 20: @pauladkisson I think this is something good to be aware of. Before, in the second condition, I had the boolean with quotes:

`github.event.pull_request.draft == 'false'`

As we have in the first condition here (`needs.assess-file-changes.outputs.SOURCE_CHANGED == 'true'`) and in many other places. However, for some reason that did not work. I am wondering why those booleans behave differently. Could it be because it is ourselves who are setting them? Do you have any idea?

Now I changed the second condition boolean to false (without the quotes) and it works as intended. The changelog failed when I changed from draft to ready to review:

image

pauladkisson: I think it's because the inputs/outputs to/from reusable workflows and composite actions must be strings whereas the native github.event.pull_request.draft field is a proper boolean.

See https://github.com/actions/runner/issues/2238.

h-mayorquin: Thanks for the pointer. I encountered this problem before in the cache action and I thought it would be good for you to be aware of it.

h-mayorquin: This comment there in the thread you shared is also useful:

https://github.com/actions/runner/issues/2238#issuecomment-1686609260

PR #1080, opened 2024-09-13
add macos-latest to testing matrix

pauladkisson:
Fixes #1061


h-mayorquin:
I already did this here:

 #1061

h-mayorquin:
But this also has the support for the intel mac which I think was only discontinued from pros two years ago or something? To be honest, I would prefer to test only on latest mac, the CI is already super bloated on my view but I think there might be a case here for the two. 

How do you feel about it?

pauladkisson:
> I already did this here:
> 
> #1061

Not sure how I missed that, sorry.

PR #1085, opened 2024-09-16
[Cloud Deployment IVb] Rclone in AWS on EFS

CodyCBakerPhD:
@pauladkisson Can you setup your Google Drive in accordance with [these instructions](https://github.com/catalystneuro/neuroconv/pull/1085/files#diff-0e57bdea7148b9f2f27d41aca0b7f1c71fdf9dc09a4d0887abb2e4e7da7ba1f6R303-R311)? (just copy over https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/spikeglx/Noise4Sam_g0 as shown) let me know when that's ready so I can try to deploy the CI for this


pauladkisson:
> @pauladkisson Can you setup your Google Drive in accordance with [these instructions](https://github.com/catalystneuro/neuroconv/pull/1085/files#diff-0e57bdea7148b9f2f27d41aca0b7f1c71fdf9dc09a4d0887abb2e4e7da7ba1f6R303-R311)? (just copy over https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/src/master/spikeglx/Noise4Sam_g0 as shown) let me know when that's ready so I can try to deploy the CI for this

Done!

pauladkisson:
> @pauladkisson And while you're at it, could you include the `phy` folder in anticipating of using the test specification files? https://github.com/catalystneuro/neuroconv/blob/main/tests/test_on_data/test_yaml/conversion_specifications/GIN_conversion_specification.yml#L53

Where exactly does the phy folder go? Under testing_rclone_spikeglx/ci_tests? Or somewhere else?

CodyCBakerPhD:
> Where exactly does the phy folder go? Under testing_rclone_spikeglx/ci_tests? Or somewhere else?

Good question - actually, in light of that, I've modified the structure shown in the docstring

Sorry about that, could you change it when you get a chance?

tests/test_minimal/test_tools/aws_tools.py
```diff
@@ -297,3 +297,150 @@ def test_submit_aws_batch_job_with_efs_mount():
...
+    """
+    To allow this test to work, the developer must create a folder on the outer level of their personal Google Drive
+    called 'testing_rclone_spikegl_and_phy' with the following structure:
+
+    testing_rclone_spikeglx_and_phy
+    ├── ci_tests
+    ├──── spikeglx
+    ├────── Noise4Sam_g0
+    ├──── phy
+    ├────── phy_example_0
```
CodyCBakerPhD, line 312: @pauladkisson Here is the new structure that mirrors that GIN specification test

pauladkisson: Ok, the files are up on my google drive [here](https://drive.google.com/drive/folders/1c9uYyhwQzOyQxYKvgr99jrlON8rDR2dO?usp=drive_link). Should be matching the structure you outlined, but if there are any problems just lmk.

CodyCBakerPhD:
@h-mayorquin This is ready now

Proof of tests passing: https://github.com/catalystneuro/neuroconv/actions/runs/11085279132/job/30801565771

src/neuroconv/tools/aws/_submit_aws_batch_job.py
```diff
@@ -171,7 +171,9 @@ def submit_aws_batch_job(
     job_dependencies = job_dependencies or []
     container_overrides = dict()
     if environment_variables is not None:
-        container_overrides["environment"] = [{key: value} for key, value in environment_variables.items()]
+        container_overrides["environment"] = [
+            {"name": key, "value": value} for key, value in environment_variables.items()
+        ]
```
CodyCBakerPhD, line 176: Had to fix this, was not tested before now

src/neuroconv/tools/aws/_submit_aws_batch_job.py
```diff
@@ -530,7 +532,11 @@ def _generate_job_definition_name(
     """
     docker_tags = docker_image.split(":")[1:]
     docker_tag = docker_tags[0] if len(docker_tags) > 1 else None
-    parsed_docker_image_name = docker_image.replace(":", "-")  # AWS Batch does not allow colons in job definition names
+
+    # AWS Batch does not allow colons, slashes, or periods in job definition names
+    parsed_docker_image_name = str(docker_image)
+    for disallowed_character in [":", r"/", "."]:
+        parsed_docker_image_name = parsed_docker_image_name.replace(disallowed_character, "-")
```
CodyCBakerPhD, line 539: Turns out these are really restrictive, had a fix this using a GHCR image source

src/neuroconv/tools/aws/_submit_aws_batch_job.py
```diff
@@ -540,7 +546,6 @@ def _generate_job_definition_name(
         job_definition_name += f"_{efs_id}"
     if docker_tag is None or docker_tag == "latest":
         date = datetime.now().strftime("%Y-%m-%d")
-        job_definition_name += f"_created-on-{date}"
```
CodyCBakerPhD, line 543: After internal debate, I decided to roll this back since even latest images will be pulled, not at time of job definition declaration, but rather at time of instance run (which could be well after)

Should reduce the amount of spam for the normal AWS tests

CodyCBakerPhD:
@h-mayorquin Also leaving this as a note:

You may want to include job definition cleanup in the `tearDown` step. I'm not for the time being because there is no cost and it contains provenance with very useful debug info (and the basic non-EFS test should only create one for all time, should even theoretically re-use without making new revisions). But EFS mounting is a very important association with it so there is technically one job definition per EFS related test

Or you can just clean them up in a separate task periodically (monthly?), it's only a couple lines of boto3 to find the names that match some pattern and deregister them

Also, there is a rare condition where if the test is interrupted via GitHub (therefore not triggering the natural `tearDown`) the EFS might not get cleaned up - I'm not worried about cost since even if you forget about it for these tests for weeks it'll still only be pennies (though you should check it from time to time just to be sure) - but instead the test assertion errors out due to multiple copies of EFS of the same name, so if this becomes a common annoyance I recommend adding a pre-step in the test case setup to clean any existing EFS of the expected name

h-mayorquin (approved):

PR #1086, opened 2024-09-16
[Cloud Deployment IVc] Deploy NeuroConv in AWS with EFS

CodyCBakerPhD:
The most important step in the process, combining each of the previous developments


src/neuroconv/tools/aws/_submit_aws_batch_job.py
```diff
@@ -544,8 +541,6 @@ def _generate_job_definition_name(
     job_definition_name += f"_{minimum_worker_cpus}-CPU"
     if efs_id is not None:
         job_definition_name += f"_{efs_id}"
-    if docker_tag is None or docker_tag == "latest":
-        date = datetime.now().strftime("%Y-%m-%d")
```
CodyCBakerPhD, line 548: Minor cleanups here caused by the previous removal of the creation date tag

CodyCBakerPhD:
@h-mayorquin This one is ready too!

Proof of key tests passing: https://github.com/catalystneuro/neuroconv/actions/runs/11089506338/job/30810706117?pr=1086

src/neuroconv/tools/aws/_deploy_neuroconv_batch_job.py
```diff
@@ -0,0 +1,241 @@
...
+    )
+
+    available_efs_volumes = efs_client.describe_file_systems()
+    matching_efs_volumes = [
+        file_system
+        for file_system in available_efs_volumes["FileSystems"]
+        for tag in file_system["Tags"]
+        if tag["Key"] == "Name" and tag["Value"] == efs_volume_name
+    ]
+    max_iterations = 10
```
CodyCBakerPhD, line 130: This middle-man EFS management is rather annoying but had bugs with duplicates otherwise. Maybe a better way in the future would be direct EFS ID passing that overrides the 'search' based creation

src/neuroconv/tools/aws/_deploy_neuroconv_batch_job.py
```diff
@@ -0,0 +1,241 @@
...
+        )
+        warnings.warn(message=message, stacklevel=2)
+
+        return info
+
+    # Cleanup EFS after job is complete - must clear mount targets first, then wait before deleting the volume
+    efs_volumes = efs_client.describe_file_systems()
+    matching_efs_volumes = [
+        file_system
+        for file_system in efs_volumes["FileSystems"]
```
CodyCBakerPhD, line 218: @h-mayorquin Also, in a follow-up, you may wish to add logic for detecting if the EFS volume exists before this method is called, in which case we might not *want* to clean it up automatically (might be shared across jobs or something)

And/or always control the behavior with an extra argument `cleanup: bool` perhaps. Given the billing consequences I tend to err on the side of always perform cleanup, hence the current logic

h-mayorquin (approved):
Merging this to have this done.

PR #1081, opened 2024-09-13
Consolidate daily workflows into one workflow and add email notifs

pauladkisson:
This PR consolidates the daily workflows into dailies.yml and adds an email notif for each to make it easier to keep track of them all.

See https://github.com/catalystneuro/neuroconv/actions/runs/10889811236

PR #1088, opened 2024-09-16
reorganized weeklies into dedicated workflow like dailies

pauladkisson:
Continuation of #1081 for weekly workflows.

### DIFF
--- a/.github/workflows/build_and_upload_docker_image_rclone_with_config.yml
+++ b/.github/workflows/build_and_upload_docker_image_rclone_with_config.yml
@@ -1,9 +1,13 @@
 name: Build and Upload Docker Image of Rclone With Config to GHCR
 
 on:
-  schedule:
-    - cron: "0 16 * * 1"  # Weekly at noon EST on Monday
   workflow_dispatch:
+  workflow_call:
+    secrets:
+      DOCKER_UPLOADER_USERNAME:
+        required: true
+      DOCKER_UPLOADER_PASSWORD:
+        required: true
 
 concurrency:  # Cancel previous workflows on the same pull request
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


pauladkisson:
@CodyCBakerPhD, for my own understanding, what is the reason that `Build and Upload Docker Image of Rclone With Config to GHCR` only runs weekly while all the rest of the active workflows run daily?

CodyCBakerPhD:
> for my own understanding, what is the reason that Build and Upload Docker Image of Rclone With Config to GHCR only runs weekly while all the rest of the active workflows run daily?

The base image for rclone is not assumed to update that regularly or that its updates are not that important

Could maybe even go down to monthly

Main edge case benefit is that you gain access to new service providers (but most of our expected usage is drive/dropbox anyway)

pauladkisson:
> The base image for rclone is not assumed to update that regularly or that its updates are not that important

Ok, sounds good.

h-mayorquin (commented):
Should we put the formatwise gallyer tests here as well?

https://github.com/catalystneuro/neuroconv/blob/fddd4d28cb7839920479d8154f49417f78a391af/.github/workflows/formatwise-installation-testing.yml#L1-L73

Also, maybe the dev tests? They are on the dailies but we won't be able to fix them on a daily basis so once you get the notification once you will keep getting the notification every day. I guess that's your call.

pauladkisson:
> Should we put the formatwise gallyer tests here as well?

Those are disabled, so I didn't include them.

pauladkisson:
> Also, maybe the dev tests?

No, I definitely want to know about the dev tests on a daily basis.

h-mayorquin:
> > Should we put the formatwise gallyer tests here as well?
> 
> Those are disabled, so I didn't include them.

OK, I think we should activate them. But that's outside of the scope of this PR.

h-mayorquin (approved):

PR #1090, opened 2024-09-17
Add json schema validation for source data at the interface level

### LINKED ISSUE #42: Discussion: Unify `run_conversion` interfaces between base classes and `nwbconverter`

h-mayorquin:
Should close #42.


h-mayorquin:
@CodyCBakerPhD 

This is ready. Three things:
1) Some of the typing was missing. There were some things marked as `value: float = None` where I had to mark them as optional because the JSON schema would fail to validate. I am surprised that Pydantic missed that. This adds to the point that you made before that both have different strengths, maybe?
2) I created a way to clarify which source arguments are used to initialize an extractor in the extractor interfaces. The problem is that we pass everything to the super init where it is validated but sometimes we modify the arguments names within the interface and the JSON validation fails. I think it is better to have this explicit anyway (explicit is better than implicit).
3) Some of the typing was infraspecific, for example in the MockRecording the typing for durations was tuple(int) but apparently it needs to be `tuple(int, ...)` to indicate there is more than one.

docs/conversion_examples_gallery/behavior/lightningpose.rst
```diff
@@ -23,7 +23,7 @@ Convert LightningPose pose estimation data to NWB using :py:class:`~neuroconv.da
     >>> labeled_video_file_path = str(folder_path / "labeled_videos/test_vid_labeled.mp4")
 
     >>> converter = LightningPoseConverter(file_path=file_path, original_video_file_path=original_video_file_path, labeled_video_file_path=labeled_video_file_path, verbose=False)
-
+    Source data is valid!
```
CodyCBakerPhD, line 26: Here I was thinking you despised verbosity in the testing suite 😉

h-mayorquin: I do, but I did not want to change there because I saw that @pauladkisson explicitly set `verbose=True` so I went with what I assumed was the intent.

CodyCBakerPhD:
> 1. Some of the typing was missing. There were some things marked as value: float = None where I had to mark them as optional because the JSON schema would fail to validate. I am surprised that Pydantic missed that. This adds to the point that you made before that both have different strengths, maybe?

I too am surprised by this, but maybe not so much - `docval` in HDMF behaves similarly, where you don't need to indicate `None` is a possible value to be passed to an argument even if that is exactly its default value. I think Pydantic does not consider the typing vs. default a fundamental incompatability since the validation call only checks if the user specified (not default) input matches annotation type. If user specifies same as default value, I also don't know what would happen but suppose it would ignore

> 2. I created a way to clarify which source arguments are used to initialize an extractor in the extractor interfaces. The problem is that we pass everything to the super init where it is validated but sometimes we modify the arguments names within the interface and the JSON validation fails. I think it is better to have this explicit anyway (explicit is better than implicit).

Nice - yeah that has always proved a subtle pain point, since we do utilize additional `kwargs` in the interface `__init__` that are not a part of the extractor `kwargs`, usually to perform 'extra' metadata or neo related operations

> Some of the typing was infraspecific, for example in the MockRecording the typing for durations was tuple(int) but apparently it needs to be tuple(int, ...) to indicate there is more than one.

Good catches - accurate validation must rely on accurate annotations, which have only ever been used so far as a form of developer communication

CodyCBakerPhD (approved):

CodyCBakerPhD:
Just beware, this (like the Pydantic one before it) could potentially break custom conversion pipelines using latest/dev NeuroConv by requiring them to use accurate annotations on all interface `__init__`

h-mayorquin:
> Just beware, this (like the Pydantic one before it) could potentially break custom conversion pipelines using latest/dev NeuroConv by requiring them to use accurate annotations on all interface `__init__`

This is a very important point.

Just to be clear:  if they were using `NWBConverter` before this PR should not change things, right? Validation was already done there:

https://github.com/catalystneuro/neuroconv/blob/f28410598cfbf830dd7f25ff774c5cf6553273f9/src/neuroconv/nwbconverter.py#L45-L85

CodyCBakerPhD:
> Just to be clear: if they were using NWBConverter before this PR should not change things, right? Validation was already done there:

I suppose that's true, but I've been guilty myself of using ConverterPipes sometimes

What is mildly annoying now is that JSON validation occurs in both NWBConverter initialization and individual interface initialization, but overall benefit is probably worth it in the end

h-mayorquin:
> What is mildly annoying now is that JSON validation occurs in both NWBConverter initialization and individual interface initialization, but overall benefit is probably worth it in the end

Agree with it, I would think long term on how to avoid the overlap.

PR #1013, opened 2024-08-15
Fix stream in Plexon 1 interface

h-mayorquin:
Neo released, this is ready.

Should fix https://github.com/catalystneuro/neuroconv/issues/1000.

Related neo PR:
https://github.com/NeuralEnsemble/python-neo/pull/1524

Issue #993, opened 2024-08-07
[DevOps]: Separate data caching from other workflows


CodyCBakerPhD:
### What would you like to see added to NeuroConv?

Both doctests and the later parts of our main test suite here on NeuroConv rely on the example data cache

The GUIDE also has multiple parts that each rely on such a mechanism, but we did something there that made it nicer - we made the caching it's own job and made the downstream runs dependent on its success: https://github.com/NeurodataWithoutBorders/nwb-guide/blob/main/.github/workflows/example_data_cache.yml

I suggest following that strategy @pauladkisson @h-mayorquin as it would alleviate the failures that started to be seen today, such as in https://github.com/catalystneuro/neuroconv/pull/992

It also simplifies things by keeping the actions more modular rather than having cache handling steps integrated within the main CI tests

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

h-mayorquin:
This is a good idea. There are new developments in cross oss caching that might simplify this even further:

https://github.com/actions/cache/blob/main/tips-and-workarounds.md#cross-os-cache

h-mayorquin:
@pauladkisson do you want to optimize this further or are you pleased with the improvements on #1095?

pauladkisson:
Yeah, I think #1095 should close this issue.

PR #1095, opened 2024-09-18
Use a github action to centralize gin data preparation

h-mayorquin:
While working on  #1013 I was having trouble getting the data right from the caches (see the diff in that PR). Centralizing the logic should make this kind of problem easier to deal with.

This is a partial move in the direction of #993.

### DIFF
--- a/.github/actions/load-data/action.yml
+++ b/.github/actions/load-data/action.yml
@@ -1,5 +1,5 @@
 name: 'Prepare Datasets'
-description: 'Restores data form caches or downloads it from S3.'
+description: 'Restores data from caches or downloads it from S3.'
 inputs:
   aws-access-key-id:
     description: 'AWS Access Key ID'
@@ -60,6 +60,7 @@ runs:
 
     - name: Determine if downloads are required
       id: download-check
+      shell: bash  # Added shell property
       run: |
         if [[ "${{ steps.cache-ephys-datasets.outputs.cache-hit }}" != 'true' || \
               "${{ steps.cache-ophys-datasets.outputs.cache-hit }}" != 'true' || \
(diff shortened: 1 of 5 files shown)


pauladkisson (changes requested):
Looks good, this PR should probably delete update-s3-testing-data.yml to avoid confusion.

h-mayorquin:
> Looks good, this PR should probably delete update-s3-testing-data.yml to avoid confusion.

What possible confusion do you in mind?

pauladkisson:
> > Looks good, this PR should probably delete update-s3-testing-data.yml to avoid confusion.
> 
> What possible confusion do you in mind?

update-s3-testing-data.yml is disabled, and actions/load-data seems to be a replacement (is it not?), so we should delete update-s3-testing-data.yml to avoid maintaining vestigial workflows that aren't in use.

h-mayorquin:
> > > Looks good, this PR should probably delete update-s3-testing-data.yml to avoid confusion.
> > 
> > 
> > What possible confusion do you in mind?
> 
> update-s3-testing-data.yml is disabled, and actions/load-data seems to be a replacement (is it not?), so we should delete update-s3-testing-data.yml to avoid maintaining vestigial workflows that aren't in use.

No, it is not meant to replace it. This assumes the data is already on the s3 and loads it either from the cache or from the s3 if the former fails. At the moment we don't have a mechanism to automatically transfer the data from gin to s3. We should meet at brainstorm that.

pauladkisson:
> No, it is not meant to replace it. This assumes the data is already on the s3 and loads it either from the cache or from the s3 if the former fails. 

I see. update-testing-data.yml on roiextractors is more like load-data/action.yml than update-s3-testing-data.yml, so that's why I misunderstood.

> At the moment we don't have a mechanism to automatically transfer the data from gin to s3. We should meet at brainstorm that.

gin is very finnicky, so we might just be better off keeping our current (manual) system.

pauladkisson (approved):

pauladkisson:
> What troubles have you experienced?

Failed downloads mostly.

PR #1104, opened 2024-09-29
[Cloud Deployment IVd] AWS usage docs

CodyCBakerPhD:
Does not technically depend on the other PRs, but of course would not be usable without them

Feel free to trim/reword/modify as needed

It's pretty hefty on the amount of setup required, but I don't see any way to reduce that. It is what it is I suppose...


oruebel:
@h-mayorquin is this PR ready to merge? It would be nice if I could point to the AWS demo on the main docs.

h-mayorquin (approved):

Issue #1106, opened 2024-10-03
[Feature]: EDF with multiple offsets


bendichter:
### What would you like to see added to NeuroConv?

I was just in a meeting with Hezekiah Branch, who wants to convert EDF files to NWB and share them on DANDI. When following the docs from the conversion gallery for EDF, we got an error that we can't handle data that has different channel offsets. apparently this file specifies different channel offsets for different channels. I know we've seen different conversion factors, but I can't recall different offsets. Indeed, looking at [the docs for ElectricalSeries in PyNWB](https://pynwb.readthedocs.io/en/stable/pynwb.ecephys.html#pynwb.ecephys.ElectricalSeries), it looks like we allow for multiple conversion factors but not multiple offsets.  Have we dealt with this kind of thing before? Maybe there's a way around this. I have asked Hezekiah to share one of these files with us so we can dig in and see what is going on. 

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

PR #1110, opened 2024-10-10
Add skip channels to EDF interface

h-mayorquin:
Related to #1106

This adds the user to the possibility of skipping non-neural channels on the EDF interface.

Some relevant comments:
* Can't add testing right now. The EDF by neo is blocked when reading the format. That means that it does not work well with the parallel machinery that we have in pytest. I have been meaning to fix this on the neo level or add a hack to our pytest machinery but other priorities come first (see https://github.com/NeuralEnsemble/python-neo/issues/1557).
* Up for debate on how to expose the available channels so the user can pick for them. Ideally, this should be a static or class method but then because of the blocking problem described in the previous point this will not work in a script (getting the channel names requires opening the file). Maybe this is simpler to fix at the neo level by providing a method to closing the objects but I did not want that to get in the way.
* I will open another PR to improve the friendliness of the error for multiple offsets. (Done here https://github.com/catalystneuro/neuroconv/pull/1111)

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/edf/edfdatainterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/edf/edfdatainterface.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from pydantic import FilePath
 
 from ..baserecordingextractorinterface import BaseRecordingExtractorInterface
@@ -23,7 +25,22 @@ class EDFRecordingInterface(BaseRecordingExtractorInterface):
         source_schema["properties"]["file_path"]["description"] = "Path to the .edf file."
         return source_schema
 
-    def __init__(self, file_path: FilePath, verbose: bool = True, es_key: str = "ElectricalSeries"):
+    def _source_data_to_extractor_kwargs(self, source_data: dict) -> dict:
+
+        extractor_kwargs = source_data.copy()
+        extractor_kwargs.pop("channels_to_skip")
+        extractor_kwargs["all_annotations"] = True
+        extractor_kwargs["use_names_as_ids"] = True
+
+        return extractor_kwargs
+
+    def __init__(
+        self,
+        file_path: FilePath,
+        verbose: bool = True,
+        es_key: str = "ElectricalSeries",
+        channels_to_skip: Optional[list] = None,
+    ):
         """
         Load and prepare data for EDF.
         Currently, only continuous EDF+ files (EDF+C) and original EDF files (EDF) are supported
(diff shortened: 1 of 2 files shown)


src/neuroconv/datainterfaces/ecephys/edf/edfdatainterface.py
```diff
@@ -23,7 +25,22 @@ def get_source_schema(cls) -> dict:
         source_schema["properties"]["file_path"]["description"] = "Path to the .edf file."
         return source_schema
 
-    def __init__(self, file_path: FilePath, verbose: bool = True, es_key: str = "ElectricalSeries"):
+    def _source_data_to_extractor_kwargs(self, source_data: dict) -> dict:
+
+        extractor_kwargs = source_data.copy()
+        extractor_kwargs.pop("channels_to_skip")
+        extractor_kwargs["all_annotations"] = True
+        extractor_kwargs["use_names_as_ids"] = True
```
h-mayorquin, line 33: @bendichter 
This is necessary to display meaningful names that the user can suppress with the newly added parameter. Otherwise, the channel ids are "1", "2", "3", etc.

This is technically a breaking change because we are changing the way we add the data to the NWBFile but I also consider this an improvement:

image

src/neuroconv/datainterfaces/ecephys/edf/edfdatainterface.py
```diff
@@ -36,15 +53,24 @@ def __init__(self, file_path: FilePath, verbose: bool = True, es_key: str = "Ele
...
+            Key for the ElectricalSeries metadata
+        channels_to_skip : list, default: None
+            Channels to skip when adding the data to the nwbfile. These parameter can be used to skip non-neural
+            channels that are present in the EDF file.
+
         """
         get_package(
             package_name="pyedflib",
-            excluded_platforms_and_python_versions=dict(darwin=dict(arm=["3.8", "3.9"])),
+            excluded_platforms_and_python_versions=dict(darwin=dict(arm=["3.9"])),
```
bendichter, line 64: why are you changing this?

bendichter:
@h-mayorquin this looks good! Can you confirm that this is able to read and convert the sample EDF file?

src/neuroconv/datainterfaces/ecephys/edf/edfdatainterface.py
h-mayorquin: We no longer support python 3.8, unecessary change though, reverting.

h-mayorquin:
> @h-mayorquin this looks good! Can you confirm that this is able to read and convert the sample EDF file?

Yes, can't attach the stub here but I will send it on slack.

src/neuroconv/datainterfaces/ecephys/edf/edfdatainterface.py
bendichter: oh, right, I forgot about that. It's fine to keep it, I was just wondering.

h-mayorquin: Ok.

bendichter (approved):

PR #1113, opened 2024-10-14
Fix Failing Dailies

pauladkisson:
Dailies were failing bc os-versions and python-versions were not being propagated properly.

Now fixed -- see here: https://github.com/catalystneuro/neuroconv/actions/runs/11463322663

Daily neuroconv docker tests are failing (see #1121), but since that is a separate problem, we can solve it separately.


.github/workflows/neuroconv_docker_testing.yml
```diff
@@ -81,7 +81,7 @@ jobs:
 
       - name: Pull docker image
         run: |
-          docker pull ghcr.io/catalystneuro/neuroconv:latest
+          docker pull ghcr.io/catalystneuro/neuroconv:v0.6.4
```
h-mayorquin, line 84: Can you add more context about this change?

h-mayorquin (commented):
What do you think about changing the following lines to make the argument non-required for this and the other workflows:

https://github.com/catalystneuro/neuroconv/blob/977144edbd2d0412b7a47fbc52e916376311d579/.github/workflows/testing.yml#L8-L11

That way we only need to keep those default versions in one place.

pauladkisson:
> That way we only need to keep those default versions in one place.

Yeah, I think that's a good idea.

h-mayorquin:
> > That way we only need to keep those default versions in one place.
> 
> Yeah, I think that's a good idea.

See here for more related discussion:

https://github.com/catalystneuro/neuroconv/pull/1117#issuecomment-2427842819

Tag me when this is ready to take a look.

pauladkisson:
@h-mayorquin, this is ready to go.

h-mayorquin:
Where are you setting `vars.ALL_PYTHON_VERSIONS`? I would like to have the ability to open a PR when adding support for new versions as I would like to run the testing suit before merging.

h-mayorquin:
I would prefer to have them in code so we can have a workflow where changing them requires and triggers a PR.

bendichter (approved):

bendichter:
Oh yeah, that's a good point. Is there a way to put this as a variable in code?

pauladkisson:
> Oh yeah, that's a good point. Is there a way to put this as a variable in code?

As far as I could tell, the only way to include this in the code is to define it separately for each workflow. If we want all the workflows to reference a single variable, I think it has to be a github repository variable: https://github.com/catalystneuro/neuroconv/settings/variables/actions

h-mayorquin:
We can either use the defaults to pass to reduce the duplication (simpler and probably the way to go) or you can write the variable to a file and use bash or an action to read them.

pauladkisson:
@h-mayorquin how about this solution?

h-mayorquin:
Yes, I am glad that the reading from a file option worked. This LGTM.

PR #1117, opened 2024-10-21
Add python 3.13 to CI

bendichter:


h-mayorquin:
You need to modify the Python version here:

https://github.com/catalystneuro/neuroconv/blob/52cd6aa0ea46c0d143739767da4951ade17c3b74/.github/workflows/deploy-tests.yml#L42-L44

Maybe it would be a good idea to manage that with the defaults or remove the defaults as in the current state is confusing, what do you think? they are required but they have a default (?!).

h-mayorquin:
I am as confused as you are but I can take a look. Let me know if this becomes a blocker.

pauladkisson:
After #1113, adding python 3.13 shouldn't require a PR, so we can close this.

h-mayorquin:
I think that adding support for new versions and old versions should still require a PR independent of how we define the versions. Both might require changes on requirements and/or code and it feels to me that a PR is a good place to discuss/highlight this.

pauladkisson:
> I think that adding support for new versions and old versions should still require a PR independent of how we define the versions. Both might require changes on requirements and/or code and it feels to me that a PR is a good place to discuss/highlight this.

Well, the pyproject.toml would still need to add/remove the changed versions, so that would require a PR.

Unfortunately, I don't think there's any good way to require a PR to actually change the configuration variable for the repo.

pauladkisson:
> Unfortunately, I don't think there's any good way to require a PR to actually change the configuration variable for the repo.

Well I was dead wrong there lol.

h-mayorquin:
On the ephys side I am following here:
https://github.com/SpikeInterface/spikeinterface/pull/348
We should do the same for the ophys side

Then, what would remain at our side is some behavior I think.

pauladkisson:
> On the ephys side I am following here: [SpikeInterface/spikeinterface#348](https://github.com/SpikeInterface/spikeinterface/issues/348) We should do the same for the ophys side
> 
> Then, what would remain at our side is some behavior I think.

I'm sorry, what are you talking about?

h-mayorquin:
> > On the ephys side I am following here: [SpikeInterface/spikeinterface#348](https://github.com/SpikeInterface/spikeinterface/issues/348) We should do the same for the ophys side
> > Then, what would remain at our side is some behavior I think.
> 
> I'm sorry, what are you talking about?

Adding Python 3.13 support? : /

pauladkisson:
Based on install errors, looks like we are waiting on numba: https://github.com/numba/numba/pull/9682

h-mayorquin:
I removed the numpy ceiling. After #1204 I think this should work, fingers crossed. If it does not, then I will remove the numpy ceiling in a separate PR.

h-mayorquin:
OK, tests failing. Let's remove the numpy ceiling on another PR:

https://github.com/catalystneuro/neuroconv/pull/1206

CHANGELOG.md
```diff
@@ -8,9 +8,10 @@
 ## Features
 
 ## Improvements
+* Support python 3.12 [PR #1117](https://github.com/catalystneuro/neuroconv/pull/1117)
```
bendichter, line 11:
```suggestion
* Support python 3.13 [PR #1117](https://github.com/catalystneuro/neuroconv/pull/1117)
```

h-mayorquin:
I am hoping that #1235 might be the final thing we need to add this to the CI.

h-mayorquin:
We will need a release of roiexctractors. We already fixed the current error there


https://github.com/catalystneuro/roiextractors/pull/397/files

pauladkisson:
> We will need a release of roiexctractors.

I'll cut a new one after github.com/catalystneuro/roiextractors#395 and github.com/catalystneuro/roiextractors#396 get merged.

CHANGELOG.md
```diff
@@ -51,8 +52,7 @@
 * Simple writing no longer uses a context manager [PR #1180](https://github.com/catalystneuro/neuroconv/pull/1180)
 * Added Returns section to all getter docstrings [PR #1185](https://github.com/catalystneuro/neuroconv/pull/1185)
 * ElectricalSeries have better chunking defaults when data is passed as plain array [PR #1184](https://github.com/catalystneuro/neuroconv/pull/1184)
-* Ophys interfaces now call `get_metadata` by default when no metadata is passed [PR #1200](https://github.com/catalystneuro/neuroconv/pull/1200) and [PR #1232](https://github.com/catalystneuro/neuroconv/pull/1232)
-
+* Ophys interfaces now call `get_metadata` by default when no metadata is passed [PR #1200](https://github.com/catalystneuro/neuroconv/pull/1200)
```
h-mayorquin, line 55:
```suggestion
* Ophys interfaces now call `get_metadata` by default when no metadata is passed [PR #1200](https://github.com/catalystneuro/neuroconv/pull/1200) and [PR #1232](https://github.com/catalystneuro/neuroconv/pull/1232)
```

h-mayorquin:
There are still some errors related to the change on how numpy 2.0 represents scalar values as strings. We can take those piecewise but I guess this is not urgent.

h-mayorquin:
OK, it seems that everything is fixed but we will need to wait for a release of SpikeInterface.

h-mayorquin:
Mmm even with the latest spikeinterface release another deep problem with numpy scalar value representations still managed to leak all the way here:

https://github.com/NeuralEnsemble/python-neo/issues/1678

I think this can be fixed here by using ids I hope.

h-mayorquin (approved):
Ok, with the latest release of neo this should be ready. Yei.

PR #1119, opened 2024-10-21
Live service test to only use oldest and latest for PR CI

h-mayorquin:
Follow up on this:
https://github.com/catalystneuro/neuroconv/pull/1082#discussion_r1759632013

I think there are still situations where we want to test live srivces on the PR (something changing something related to them) so I am just reducing here here to oldest and latest.


pauladkisson (commented):
I think if we need live service tests for PRs (which seems reasonable to me), we should just run it for all python versions (like we do currently).

Issue #1112, opened 2024-10-11
[Bug]: Namespace collision with two or more SpikeGLX interfaces


simon-ball:
### What happened?

A large proportion of the experimental data recorded within our lab uses two or more Neuropixels probes (recorded with SpikeGLX) simultaneously. Each probe recording is then clustered independently via KiloSort. I expect to then generate a single NWB file, containing these multiple sets of SpikeGLX data, correctly linked to the correct member of multiple sets of Kilosort data.

However, I have not yet found a way to handle this case via Neuroconv. Possibly I am missing something obvious, possibly it was not envisaged during development. Nowhere in the documentation nor CatalystNeuro examples can I find something similar. 

Data from a typical recording session looks something like this (irrelevant files excluded for clarity)
```bash
> data
   > subject_1
      > session_1
         > probe_1
            > ks2.5_01
            npx.imec.ap.bin
            npx.imec.ap.meta
         > probe_2
            > ks2.5_01
            npx.imec.ap.bin
            npx.imec.ap.meta
```

---

I have tried the following approaches, with the following failure modes:

**Attempt 1: `NWBConverter`**, following the documentation [here ](https://neuroconv.readthedocs.io/en/main/user_guide/nwbconverter.html)

< See code `attempt_1` >

This runs into several problems
* Since source_data must be provided as a dictionary, not a list, it does not scale to arbitrary _n_ probes. I would need to duplicate the code for new, separate classes `ThreeProbeConverter`, `FourProbeConverter`, etc
* There is no obvious way to specify that `ks_1` is associated with `probe_1` and not with `probe_2` 
* Most crucially, the `SpikeGLXRecordingInterface` does not name its component parts uniquely (despite having access to the unique serial number encoded in the metadata file used to extract the model type). Therefore, for example, the `Device` entity for each `SpikeGLXRecordingInterface` is identical as far as name, description etc, as are the various ElectrodeGroups, and so when the metadata dictionary is repeatedly updated within `md = converter.get_metadata()`, the outcome only lists a single device, and as many electrode groups as a single probe

I did not get as far as figuring out how to crosslink `ks_1` with `probe_1`, as the device issue proved terminal.

---

Attempt 2: `ConverterPipe`, similar to the documentation [here](https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/combinations/spikeglx_and_phy.html), overwriting the metadata as needed to provide unique device IDs

< see code `attempt_2`>

[...]

PR #1124, opened 2024-11-05
Propagate `unit_electrode_indices` to `SortingInterface`

h-mayorquin:
This will allow the user to pass a mapping between units and the electrodes on the electrode table. Eventually I will use this to add a more robust solution like here:

https://github.com/catalystneuro/neuroconv/issues/1112#issuecomment-2422864725


src/neuroconv/nwbconverter.py
```diff
@@ -144,7 +144,7 @@ def validate_conversion_options(self, conversion_options: dict[str, dict]):
 
     def create_nwbfile(self, metadata: Optional[dict] = None, conversion_options: Optional[dict] = None) -> NWBFile:
         """
-        Create and return an in-memory pynwb.NWBFile object with this interface's data added to it.
+        Create and return an in-memory pynwb.NWBFile object with the conversion data added to it.
```
h-mayorquin, line 147:
```suggestion
        Create and return an in-memory pynwb.NWBFile object with this interface's data added to it.
```

src/neuroconv/datainterfaces/ecephys/basesortingextractorinterface.py
```diff
@@ -312,9 +313,15 @@ def add_to_nwbfile(
         units_name : str, default: 'units'
             The name of the units table. If write_as=='units', then units_name must also be 'units'.
         units_description : str, default: 'Autogenerated by neuroconv.'
+        unit_electrode_indices : list of lists of int, optional
+            A list of lists of integers indicating the indices of the electrodes that each unit is associated with.
```
pauladkisson, line 317: This corresponds to the sparsity in spikeinterface right?

pauladkisson (commented):
This looks good. 

What do you think about adding some kind of how-to explaining what to do if you have 2 probes and run kilosort like in #1112? 

I think it could be an effective way to communicate to users about this new feature.

pauladkisson:
> What do you think about adding some kind of how-to explaining what to do if you have 2 probes and run kilosort like in #1112?

Actually, if this is supposed to be used in `SortedRecordingConverter` specifically, maybe just a conversion gallery entry for that one would be sufficient...

h-mayorquin:
> > What do you think about adding some kind of how-to explaining what to do if you have 2 probes and run kilosort like in #1112?
> 
> Actually, if this is supposed to be used in `SortedRecordingConverter` specifically, maybe just a conversion gallery entry for that one would be sufficient...

Yeah, I would be adding a tutorial there.

pauladkisson (approved):

PR #1130, opened 2024-11-06
Change deprecated `get_schema_from_method_signature` to `get_json_schema_from_method_signature`

h-mayorquin:
The method was deprecated but we are still using it in many places. This PR corrects that oversight.


weiglszonja:
Looks good to me, did you want to include this in the changlog as well?

h-mayorquin:
@weiglszonja  . Thatnks for the review. I think this does not require a changelog. The function was already deprecated and no functionality is changed. It is more of an internal change. That said, I don't feel that strong about it and can add it to the changelog if you want.

weiglszonja (approved):

PR #961, opened 2024-07-14
add max_channel column for phy interface

bendichter:
Some older Phy-based NWB files, e.g. https://neurosift.app/?p=/nwb&url=https://api.dandiarchive.org/api/assets/b4a15b14-3a7b-49ae-b9ba-fa2c4d8f6c29/download/&dandisetId=000053&dandisetVersion=0.210819.0345 do not have any information that could associate a each unit to an electrode (or physical location). Without this information it is not possible to map a unit to 3D coordinates nor brain area, which has limited the utility of the data for secondary analysis. This is an attempt to remedy this by getting the templates from the Phy output files, calculating the peak channel of each template, and adding that to as a property of the units table.

PR #1132, opened 2024-11-07
`SortedRecordingConverter` for solving unit electrode provenance in sorted data

h-mayorquin:
This has #1124 as a base and should be done after that.

This converter can be used to write sorted data and keep the correct electrode provenance of the units (display electrodes in the units table). The converted takes care of handling the correct map between units and electrodes which is important for multiple recording/probes such as the one discussed in #1112.

Also related to this:
https://github.com/catalystneuro/neuroconv/pull/961


h-mayorquin:
OK, so we can have this as a base for the manual case. In another PR coming after this we can use sortinganalyzer tools to automatically build the mapping on the fly. After that, let's think on an interface that will generalize this to converters with multiple interfaces.

@alejoe91 can you take a look at this? I will add the sorting_analyzer automatic detection in a future PR later.

alejoe91 (approved):
Looks good to me! Thanks @h-mayorquin

PR #1136, opened 2024-11-12
Use mixing tests for mocks

h-mayorquin:
We have specific tests for the mock recording interface that duplicates the tests on the mixing. This PR solves this by using the mixing tests for the mock (We already doing that in the ophys side). 

This PR also separates some changes that I did on https://github.com/catalystneuro/neuroconv/pull/1124 and should come before it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Imaging interfaces have a new conversion option `always_write_timestamps` that can be used to force writing timestamps even if neuroconv's heuristics indicates regular sampling rate [PR #1125](https://github.com/catalystneuro/neuroconv/pull/1125)
 
 ## Improvements
+* Use mixing tests for ecephy's mocks [PR #1136](https://github.com/catalystneuro/neuroconv/pull/1136)
 
 # v0.6.5 (November 1, 2024)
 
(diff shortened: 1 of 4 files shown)


pauladkisson (changes requested):
Looks good to me, but we should remove the 'old' Test classes before we merge right?

h-mayorquin:
> Looks good to me, but we should remove the 'old' Test classes before we merge right?

I remove the ones from the Old Recording Extractor. I don't really like the stub tests were done in the old tests. Since those were written I have written better machinery on the SpikeInterface side. I will update those in a separate PR though and then remove the old ones.

pauladkisson:
I see, so you'll remove those old tests after #1124 is ready to go?

pauladkisson:
Also looks like test_no_slash_in_name is failing...

h-mayorquin:
> I see, so you'll remove those old tests after #1124 is ready to go?

Yes, if not before.

h-mayorquin:
> Also looks like test_no_slash_in_name is failing...

Should be fixed now.

pauladkisson:
> > I see, so you'll remove those old tests after #1124 is ready to go?
> 
> Yes, if not before. I just don't want this to get in the way of #1124 as it would just be a cleaning/refactor PR and #1124 is building towards and enhancement coming from a user request.

Sounds good.  As long as we have a clear plan to remove it, so it doesn't just end up hanging around with a comment saying it should be removed.

pauladkisson (approved):
Tests are failing, but should pass after #1140 gets merged.

PR #1137, opened 2024-11-12
[Draft] Make dynamic tables expandable

h-mayorquin:
Sup, @bendichter I wanted to write this here so we have an anchor point for discussion next time we meet. Taking the electrodes table as a starting point there are two objects that we are not catching for chunkg at the moment:
* ElectrodeGroups column 
* Ids.

Check the code below for more comments but the gist of where I got stuck is here:

```

<HDF5 dataset "id": shape (2,), type "<i8">
data.chunks
(2,)
data.maxshape
(2,)
data.resize(shape)
  File "<string>", line 1, in <module>
  File "/home/heberto/miniconda3/envs/work/lib/python3.11/site-packages/h5py/_hl/dataset.py", line 679, in resize
    self.id.set_extent(size)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5d.pyx", line 335, in h5py.h5d.DatasetID.set_extent
RuntimeError: Unable to synchronously change a dataset's dimensions (dimension cannot e
```

For the ids, the data was chunk but the maxshape was also so it becomes non-expandable.

### DIFF
--- a/src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
+++ b/src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
@@ -5,7 +5,6 @@ from typing import Generator, Literal, Union
 import h5py
 import numpy as np
 import zarr
-from hdmf import Container
 from hdmf.data_utils import DataIO
 from hdmf.utils import get_data_shape
 from hdmf_zarr import NWBZarrIO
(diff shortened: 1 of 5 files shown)


src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
```diff
@@ -104,10 +103,11 @@ def get_default_dataset_io_configurations(
     known_dataset_fields = ("data", "timestamps")
     for neurodata_object in nwbfile.objects.values():
         if isinstance(neurodata_object, DynamicTable):
-            dynamic_table = neurodata_object  # For readability
 
-            for column in dynamic_table.columns:
-                candidate_dataset = column.data  # VectorData object
+            dynamic_table = neurodata_object  # For readability
+            for column in (*dynamic_table.columns, dynamic_table.id):
```
h-mayorquin, line 108: Here, we add the ids of the tables to the set that we catch for setting the DataIO.

PR #1138, opened 2024-11-12
Fix make_or_load_nwbfile

### LINKED ISSUE #1114: [Bug]: Issue with DLC interface not loading
### What happened?

I am replicating the funny behavior I already described in #967: when running DLCInterface conversions, it fails silently not adding the correct fields in the nwb.

This time it was more painful as I forgot about the issue previously raised, and here it was 'in production'. I was running a batch conversion of multiple experiments:

```python
for exp in experiments:
    load_stuff()
    interfaces_pipe = make_interfaces()
    interfaces_pipe.run_conversion()
```

This resulted in *only the first experiment* missing all the DLC related stuff - which was driving me completely nuts for a good day of debugging, being those also quite slow operations, until I realized it sounded like the previous issue - and I fixed it adding a random 
```
from ndx_pose import PoseEstimation, PoseEstimationSeries
```

At the beginning of the script (alternatively, running twice the conversion of the first experiment :D)

This was massively annoying, I strongly suggest looking deeper into this or change the DLC example to document it.

This time it was happening from a different machine and different OS! (same interface though - always VSCode, running a script)

### Steps to Reproduce

The following code results in the bug:

```python
from datetime import datetime
from zoneinfo import ZoneInfo
from pathlib import Path
import numpy as np

file_path = ".../....h5"
config_file_path = ".../config.yaml"
path_to_save_nwbfile = ".../test_nwb.nwb"

timestamps = np.array([1,2,3])  # stupid, speeds up processing if tstamps provided

interface = DeepLabCutInterface(file_path=file_path, config_file_path=config_file_path, 
                                subject_name="ind1", verbose=False)
interface.set_aligned_timestamps(timestamps)

session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=ZoneInfo("US/Pacific"))

interface.run_conversion(nwbfile_path=path_to_save_nwbfile, metadata=metadata)

from pynwb import NWBHDF5IO

# confirm the file contains the new TimeSeries in acquisition
with NWBHDF5IO(path_to_save_nwbfile, "r") as io:
    read_nwbfile = io.read()
    print(read_nwbfile.processing)
```

The print states:
```
Fields:
  data_interfaces: {
    PoseEstimation <class 'pynwb.core.NWBDataInterface'>
  }
  description: processed behavioral data
}
```
And there is no data inside, unless I run the code twice or I add the initial import, in which case I get
```
Fields:
  data_interfaces: {
    PoseEstimation <class 'ndx_pose.pose.PoseEstimation'>
  }
  description: processed behavioral data
}
```
and data can be accessed.

### Traceback

_No response_

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.10

### Package Versions

`neuroconv == 0.6.4`

### Code of Conduct

pauladkisson:
This PR fixes a bug with `make_or_load_nwbfile` that pops up whenever extensions are lazily imported.

Fixes #1114 

See https://github.com/rly/ndx-pose/issues/36 for detailed example.

### DIFF
--- a/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
+++ b/src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
@@ -238,7 +238,7 @@ def make_or_load_nwbfile(
         else:
             load_kwargs.update(mode="w")
 
-        io = backend_io_class(**load_kwargs)
+        read_io = backend_io_class(**load_kwargs)
 
     read_nwbfile = nwbfile_path_is_provided and append_mode
     create_nwbfile = not read_nwbfile and not nwbfile_is_provided
@@ -249,7 +249,7 @@ def make_or_load_nwbfile(
         if nwbfile_is_provided:
             nwbfile = nwbfile_in
         elif read_nwbfile:
-            nwbfile = io.read()
+            nwbfile = read_io.read()
         elif create_nwbfile:
             if metadata is None:
                 error_msg = "Metadata is required for creating an nwbfile "
@@ -266,7 +266,13 @@ def make_or_load_nwbfile(
     finally:
         if nwbfile_path_is_provided and nwbfile_loaded_succesfully:
             try:
-                io.write(nwbfile)
+                read_io.close()
+                if read_nwbfile:
+                    warn("Appending to an existing on-disk NWB file may be slow for large files.")
+                    load_kwargs.update(mode="w")
+                # read_io CANNOT be used for writing bc it may not have access to all of the namespaces (extensions) added to the nwbfile -- see: https://github.com/rly/ndx-pose/issues/36
+                with backend_io_class(**load_kwargs) as write_io:
+                    write_io.write(nwbfile)
 
                 if verbose:
                     print(f"NWB file saved at {nwbfile_path_in}!")
@@ -274,8 +280,7 @@ def make_or_load_nwbfile(
                 nwbfile_written_succesfully = False
                 raise write_error
             finally:
-                io.close()
-                del io
+                del read_io
 
                 if not nwbfile_written_succesfully:
                     _attempt_cleanup_of_existing_nwbfile(nwbfile_path=nwbfile_path_in)


pauladkisson:
@h-mayorquin, @rly,
 This fix works, but it rewrites any NWB files that are supposed to be appended, which could be slow for large NWB files.  I couldn't come up with anything better but interested to know what you both think.

pauladkisson:
Ok, based on @rly's comments, I think we should wait on this until we have a function in pynwb that can update an io with the current global namespace. Then, we can keep append-mode support and just update the existing io.

h-mayorquin:
@pauladkisson thanks for doing the detective work of tracking this bug and building an example. I tried this before and I failed.

This context manager has been a source of headaches and on top of that, I find it really difficult to read and understand. Cody and I wanted to break it into different pieces but never found the time and resources to do it.

If we could peel away the append mode functionality from the rest then we can create the io once all the calls to `add_to_nwbfile` are done and solve the problem for the 95 % of use cases.  We can then fix the append mode once the necessary machinery on pynwb is added. 

I think this is the way to move forward, let me know if there is something I am missing.

rly:
> This context manager has been a source of headaches and on top of that, I find it really difficult to read and understand.

I agree. I know it is a Python feature, but I have never seen this syntax before and I find it unintuitive and confusing.

pauladkisson:
> If we could peel away the append mode functionality from the rest then we can create the io once all the calls to add_to_nwbfile are done and solve the problem for the 95 % of use cases. We can then fix the append mode once the necessary machinery on pynwb is added.

Yeah, I think this is reasonable. What should we do with append in the mean time?

h-mayorquin:
We can discuss a couple of general solutions but, as I mentioned in #1114  I think the necessary imports for `add_to_nwbfile` to work should be called at the `__init__` anyway. This will avoid this problem because all the uses of the context manager are inside `run_conversion` (which requires the interfaces to be already initialized and therefore the extension imports are already in place). Plus, I think it is good practice in general: Things should fail early if the extension library is missing in the executing environment.

PR #1139, opened 2024-11-13
Support datetime in conversion options

h-mayorquin:
Hi, @alessandratrapani .

I need to add a test but this is the reason you are seeing this error. As you might know we use json schema to do validation of the conversion options, metadata and interface arguments (source data).  The json schema validation requires arguments to be string and we use encoders to accomplish this. Check out the code. I still need to add a test for this but I wanted to share this with you as quick as possible.

A temporal solution in your code is to transform the datetime object to string (see the diff here for how) and then transform it back to datetime in `add_to_nwb`. But we can merge this quickly as well once I add a test.

### DIFF
--- a/src/neuroconv/utils/json_schema.py
+++ b/src/neuroconv/utils/json_schema.py
@@ -60,6 +60,23 @@ class _NWBSourceDataEncoder(_NWBMetaDataEncoder):
         return super().default(obj)
 
 
+class _NWBConversionOptionsEncoder(_NWBMetaDataEncoder):
+    """
+    Custom JSON encoder for conversion options of the data interfaces and converters (i.e. kwargs).
+
+    This encoder extends the default JSONEncoder class and provides custom serialization
+    for certain data types commonly used in interface source data.
+    """
+
+    def default(self, obj):
+
+        # Over-write behaviors for Paths
+        if isinstance(obj, Path):
+            return str(obj)
+
+        return super().default(obj)
+
+
 def get_base_schema(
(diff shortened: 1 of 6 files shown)


alessandratrapani (approved):
THank you for taking care of this. PR looks good to me

src/neuroconv/utils/json_schema.py
```diff
@@ -60,6 +60,23 @@ def default(self, obj):
...
+    for certain data types commonly used in interface source data.
+    """
+
+    def default(self, obj):
+
+        # Over-write behaviors for Paths
+        if isinstance(obj, Path):
+            return str(obj)
+
+        return super().default(obj)
```
bendichter, line 77: Why not add this to `_NWBMetaDataEncoder` instead of making a new class? I'm kind of surprised this isn't the default behavior for Path objs. What is the default behavior?

h-mayorquin: The default behavior is to throw an error. 

Yeah, maybe we can use the same encoder for everything. Note that even before this PR we already had a `_NWBSourceDataEncoder`:

https://github.com/catalystneuro/neuroconv/blob/dcd248b3faec7f9b6f003784dbd899945b565df8/src/neuroconv/utils/json_schema.py#L46-L60

I think that the original idea was that metadata, source data and the conversion options might require different behavior so it was created that way. Check out the inheritance pattern here that kind of implements this simplification:

https://github.com/catalystneuro/neuroconv/pull/1142/files#r1856981408

PR #1140, opened 2024-11-13
Added CSV support to DeepLabCutInterface

### LINKED ISSUE #1131: [Feature]: Add support for .csv output files from DeepLabCut

DeepLabCut optionally exports to .csv instead of .h5 (https://deeplabcut.github.io/DeepLabCut/docs/standardDeepLabCut_UserGuide.html), so DeepLabCutInterface should be able to support both types of outputs.

_No response_

Yes.

### Code of Conduct

pauladkisson:
Fixes #1131


pauladkisson:
@h-mayorquin, I know you are working on some improvements for DeepLabCut.  Lmk what you think the merge order should be.

pauladkisson:
TODO: Add note to conversion gallery about updated support

h-mayorquin:
I think your PR should have precedence as it is related to a conversion project. I will handle the conflicts later when sleap-io is ready.

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -50,7 +51,8 @@ def __init__(
         from ._dlc_utils import _read_config
 
         file_path = Path(file_path)
-        if "DLC" not in file_path.stem or ".h5" not in file_path.suffixes:
+        suffix_is_valid = ".h5" in file_path.suffixes or ".csv" in file_path.suffixes
+        if not "DLC" in file_path.stem or not suffix_is_valid:
             raise IOError("The file passed in is not a DeepLabCut h5 data file.")
```
h-mayorquin, line 56: Remove the h5 reference in the IOError.

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
```diff
@@ -341,21 +326,21 @@ def _write_pes_to_nwbfile(
 
 def add_subject_to_nwbfile(
```
h-mayorquin, line 327: We should make these functions private I think. We are using them to convert and the thing that should be exposed is our interface. 

I am saying this because technically we are broken backwards comp by changing the name of the file, the right way would be to introduce a soft deprecation, create a new argument, etc.  But I don't think we should be that careful with these ones ... to avoid this problem in the future let's make this private now.

What do you think?

tests/test_on_data/behavior/test_behavior_interfaces.py
```diff
@@ -454,6 +454,40 @@ def check_read_nwb(self, nwbfile_path: str):
         pass
 
 
+@pytest.mark.skipif(
+    platform == "darwin" and python_version < version.parse("3.10"),
+    reason="interface not supported on macOS with Python < 3.10",
+)
+class TestDeepLabCutInterfaceFromCSV(DataInterfaceTestMixin):
```
h-mayorquin, line 461: I was wondering why not to use the `DeepLabCutInterfaceMixin` but then I realize it is empty.

h-mayorquin (commented):
Is the data on gin already? 

I wanted to discuss this policy on the meeting with Ben this week but there was no time. You and our other colleagues seem to have a preference for not using gin which makes sense to me but we all should agree on this before changing the practice.

pauladkisson:
> Is the data on gin already?

Here's the PR: https://gin.g-node.org/CatalystNeuro/behavior_testing_data/pulls/14

And I uploaded to our AWS folder as well.

tests/test_on_data/behavior/test_behavior_interfaces.py
pauladkisson: Yeah, I was confused about that as well. Should we delete it? One of the testing classes uses it even though it does nothing...

h-mayorquin: Yeah, I think we should delete it. Could you do the honors?

h-mayorquin (approved):
LGTM. Feel free to erase the mixing that we are not using if you have the time.

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
pauladkisson: To be honest, I would get rid of `add_subject_to_nwbfile` entirely and unpack its contents into the interface (which would still not break backwards compatibility of the interface). It adds another redirection that seems completely unnecessary.

What do you think?

tests/test_on_data/behavior/test_behavior_interfaces.py
```diff
@@ -454,6 +454,40 @@ def check_read_nwb(self, nwbfile_path: str):
...
+    data_interface_cls = DeepLabCutInterface
+    interface_kwargs = dict(
+        file_path=str(
+            BEHAVIOR_DATA_PATH
+            / "DLC"
+            / "SL18_D19_S01_F01_BOX_SLP_20230503_112642.1DLC_resnet50_SubLearnSleepBoxRedLightJun26shuffle1_100000_stubbed.csv"
```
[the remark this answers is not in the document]
pauladkisson: ok! What to call the existing .h5 and config.yaml files? `open_field_without_video`?

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
pauladkisson: Nm, too much work for just an aesthetic improvement (although I do dislike excessive nesting/redirection).

tests/test_on_data/behavior/test_behavior_interfaces.py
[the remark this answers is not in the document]
h-mayorquin: Yes, that's how they call it in the tutorial. Yours can be called `csv_only` or something like that. 

I am thinking that the folder name should be useful for us so when we see the file path we know what it is. The provenance and details can then be on the readme.

h-mayorquin:
I think the cache has not been updated?

pauladkisson:
> I think the cache has not been updated?

Forgot to update the paths on the conversion gallery -- should be fixed now.

PR #1142, opened 2024-11-14
restore public status of `NWBMetaDataEncoder` encoder

### LINKED ISSUE #1141: [Bug]: No ImportError when calling `from neuroconv.utils import NWBMetaDataEncoder`
### What happened?

In https://github.com/catalystneuro/neuroconv/pull/1050, some classes such as `NWBMetaDataEncoder` were renamed `_NWBMetaDataEncoder`. This breaks code in NWB GUIDE that calls `from neuroconv.utils import NWBMetaDataEncoder`. Because of the override to `__getitem__` in `neuroconv/utils/__init__.py` (https://github.com/catalystneuro/neuroconv/pull/1033), the line `from neuroconv.utils import NWBMetaDataEncoder` now fails silently. It should return something like `ImportError: cannot import name 'AA' from 'collections' (/Users/rly/mambaforge/envs/test/lib/python3.12/collections/__init__.py)` but `NWBMetaDataEncoder` is set to None without an error. This was confusing to debug.

I suggest changing `__getitem__` to end with something like:
```python
if name in globals():
    return globals()[name]
raise ImportError(f"cannot import name '{name}' from 'neuroconv.utils'")
```
ref: https://peps.python.org/pep-0562/

I am also in favor of making `NWBMetaDataEncoder` public again since it is used by NWB GUIDE. But if not, I can just copy the code over to NeuroConv...

### Steps to Reproduce

```python
from neuroconv.utils import serjsklfjskl
```

### Traceback

_No response_

### Operating System

macOS

### Python Executable

Python

### Python Version

3.12

### Package Versions

neuroconv==0.6.5

### Code of Conduct

h-mayorquin:
This should fix #1141. 

I want to do this after #1139 to minimize conflict errors. This PR has #1139 as base.


src/neuroconv/utils/json_schema.py
```diff
@@ -16,13 +16,8 @@
...
 
 
-class _NWBMetaDataEncoder(json.JSONEncoder):
-    """
-    Custom JSON encoder for NWB metadata.
-
-    This encoder extends the default JSONEncoder class and provides custom serialization
-    for certain data types commonly used in NWB metadata.
-    """
+class _GenericNeuroconvEncoder(json.JSONEncoder):
```
h-mayorquin, line 19: @bendichter here.

src/neuroconv/utils/json_schema.py
```diff
@@ -36,45 +31,38 @@ def default(self, obj):
...
-        return super().default(obj)
+class _NWBSourceDataEncoder(_GenericNeuroconvEncoder):
+    """
+    Custom JSON encoder for data interface source data (i.e. kwargs).
+    """
 
 
-class _NWBConversionOptionsEncoder(_NWBMetaDataEncoder):
+class _NWBConversionOptionsEncoder(_GenericNeuroconvEncoder):
     """
```
bendichter, line 59: I think this is a strange pattern. Why define child classes that are identical to the parent class? Why not just use the parent class?

h-mayorquin: They might diverge and we might not want the same behavior for each of those cases.

h-mayorquin: Given that they are private though, I don't mind having the simple case as you want it and then refactoring if and when we need it.

bendichter (approved):

Issue #1143, opened 2024-11-14
Peel appart `make_or_load_nwbfile ` context manager.


h-mayorquin:
The `make_or_load_nwbfile` has led to many errors and difficulties #1138 #1114 #911 #910 #912
Plus, even [experienced programmers agree that this pattern is difficult to undersand](https://github.com/catalystneuro/neuroconv/pull/1138#issuecomment-2475096400) a view which I strongly agree with.

After providing interfaces and converters with `create_nwbfile` we don't need this complication for anything other than append mode (opening a file from disk and adding data to it). We should use a normal write without context manager for the other cases.

h-mayorquin:
Closed by [#1180](https://github.com/catalystneuro/neuroconv/pull/1180)

PR #1144, opened 2024-11-15
Move imports of extensions to the init of the interfaces

### LINKED ISSUE #1114: [Bug]: Issue with DLC interface not loading

h-mayorquin:
Fix #1114.

As discussed in the general meeting. This is the simplest solution. Later, when time allows, we will do  https://github.com/catalystneuro/neuroconv/issues/1143 which should be a more robust solution. 

I added a test to assure myself that this closes #1114

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
+++ b/src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
@@ -46,6 +46,7 @@ class AudioInterface(BaseTemporalAlignmentInterface):
 
         verbose : bool, default: False
         """
+
         suffixes = [suffix for file_path in file_paths for suffix in Path(file_path).suffixes]
         format_is_not_supported = [
             suffix for suffix in suffixes if suffix not in [".wav"]
--- a/src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
+++ b/src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
@@ -46,6 +46,8 @@ class AudioInterface(BaseTemporalAlignmentInterface):
 
         verbose : bool, default: False
         """
+        # This import is to assure that the ndx_pose is in the global namespace when an pynwb.io object is created
+        import ndx_sound  # noqa: F401
 
         suffixes = [suffix for file_path in file_paths for suffix in Path(file_path).suffixes]
         format_is_not_supported = [
@@ -167,7 +169,6 @@ class AudioInterface(BaseTemporalAlignmentInterface):
         stub_frames: int = 1000,
         write_as: Literal["stimulus", "acquisition"] = "stimulus",
         iterator_options: Optional[dict] = None,
-        compression_options: Optional[dict] = None,  # TODO: remove completely after 10/1/2024
         overwrite: bool = False,
         verbose: bool = True,
     ):
@@ -225,7 +226,6 @@ class AudioInterface(BaseTemporalAlignmentInterface):
                 write_as=write_as,
                 starting_time=starting_times[file_index],
                 iterator_options=iterator_options,
-                compression_options=compression_options,  # TODO: remove completely after 10/1/2024; still passing for deprecation warning
             )
 
         return nwbfile
--- a/src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
+++ b/src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
@@ -279,13 +279,14 @@ def _write_pes_to_nwbfile(
         else:
             timestamps_cleaned = timestamps
 
+        timestamps = np.asarray(timestamps).astype("float64", copy=False)
         pes = PoseEstimationSeries(
             name=f"{animal}_{keypoint}" if animal else keypoint,
             description=f"Keypoint {keypoint} from individual {animal}.",
             data=data[:, :2],
             unit="pixels",
             reference_frame="(0,0) corresponds to the bottom left corner of the video.",
-            timestamps=timestamps_cleaned,
+            timestamps=timestamps,
             confidence=data[:, 2],
             confidence_definition="Softmax output of the deep neural network.",
         )
@@ -298,6 +299,7 @@ def _write_pes_to_nwbfile(
 
     # TODO, taken from the original implementation, improve it if the video is passed
     dimensions = [list(map(int, image_shape.split(",")))[1::2]]
+    dimensions = np.array(dimensions, dtype="uint32")
     pose_estimation_default_kwargs = dict(
         pose_estimation_series=pose_estimation_series,
         description="2D keypoint coordinates estimated using DeepLabCut.",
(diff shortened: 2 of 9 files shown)


src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
```diff
@@ -46,6 +46,9 @@ def __init__(self, file_paths: list[FilePath], verbose: bool = False):
 
         verbose : bool, default: False
         """
+        # This import is to assure that the ndx_pose is in the global namespace when an pynwb.io object is created
```
pauladkisson, line 49:
```suggestion
        # This import is to assure that ndx_sound is in the global namespace when an pynwb.io object is created.
        # For more detail, see https://github.com/rly/ndx-pose/issues/36
```

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
```diff
@@ -279,13 +279,14 @@ def _write_pes_to_nwbfile(
...
             timestamps_cleaned = timestamps
 
+        timestamps = np.asarray(timestamps).astype("float64", copy=False)
         pes = PoseEstimationSeries(
             name=f"{animal}_{keypoint}" if animal else keypoint,
             description=f"Keypoint {keypoint} from individual {animal}.",
             data=data[:, :2],
             unit="pixels",
             reference_frame="(0,0) corresponds to the bottom left corner of the video.",
-            timestamps=timestamps_cleaned,
```
pauladkisson, line 288: What is going on here?

src/neuroconv/datainterfaces/behavior/lightningpose/lightningposedatainterface.py
```diff
@@ -80,6 +80,9 @@ def __init__(
         verbose : bool, default: True
             controls verbosity. ``True`` by default.
         """
+        # This import is to assure that the ndx_pose is in the global namespace when an pynwb.io object is created
```
pauladkisson, line 83:
```suggestion
        # This import is to assure that the ndx_pose is in the global namespace when an pynwb.io object is created
        # For more detail, see https://github.com/rly/ndx-pose/issues/36
```

src/neuroconv/datainterfaces/behavior/medpc/medpcdatainterface.py
```diff
@@ -73,6 +73,9 @@ def __init__(
         verbose : bool, optional
             Whether to print verbose output, by default True
         """
+        # This import is to assure that the ndx_events is in the global namespace when an pynwb.io object is created
```
pauladkisson, line 76:
```suggestion
        # This import is to assure that the ndx_events is in the global namespace when an pynwb.io object is created
        # For more detail, see https://github.com/rly/ndx-pose/issues/36
```

src/neuroconv/tools/audio/audio.py
```diff
@@ -53,17 +52,6 @@ def add_acoustic_waveform_series(
         "acquisition",
     ], "Acoustic series can be written either as 'stimulus' or 'acquisition'."
 
-    # TODO: remove completely after 10/1/2024
```
pauladkisson, line 56: I guess this is fine, but kinda off-topic for this PR.

pauladkisson (changes requested):
Caught a copy-paste error, and added some suggestions but otherwise looks good.

I also noticed a few off topic changes, would you mind explaining what's going on?

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
```diff
@@ -279,13 +279,14 @@ def _write_pes_to_nwbfile(
         else:
             timestamps_cleaned = timestamps
```
pauladkisson, line 280: This block is now unnecessary/unused with the timestamps fix

h-mayorquin: Can you elaborate?

h-mayorquin: Ah, I see, I think the fix is wrong. Let me check.

h-mayorquin: Should be fixed now.

pauladkisson (approved):
LGTM

h-mayorquin:
Thanks for the review @pauladkisson .

PR #1146, opened 2024-11-22
Fix chunking bug with compound dtypes

### LINKED ISSUE #1122: [Bug]: Recommended Chunk Shape doesn't take into account compound dtypes
### What happened?

Dev tests are failing: https://github.com/catalystneuro/neuroconv/actions/runs/11567195585/job/32197117750

Tracked this down to an update with hdmf that uncovered these lines: https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/backends/hdf5/h5tools.py#L1427-L1435

And our chunking recommendation based on the data shape here: https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py#L263

Notice how in hdmf, if the data has a compound dtype, the shape is `(len(data),)`, but in neuroconv the shape is always `get_data_shape(data)`.

This throws an error when they mismatch in the case of a Caiman pixel_mask, which is a compound dtype in NWB.

The initial solution that I came up with was to load the NWB schema to figure out if a dataset is compound or not, but I was having some trouble finding the right code to load in the schema...

### Steps to Reproduce

```python
n/a
```

### Traceback

_No response_

### Operating System

Linux

### Python Executable

Conda

### Python Version

3.9

### Package Versions

_No response_

### Code of Conduct

pauladkisson:
Fixes #1122 

Comapare Dev Tests after the fix: https://github.com/catalystneuro/neuroconv/actions/runs/12960177532/job/36153641201 (2 failed -- see #1188)

To Dev Tests before: https://github.com/catalystneuro/neuroconv/actions/runs/12995279932/job/36241639533 (8 failed)


pauladkisson:
@rly, lmk what you think

pauladkisson:
@h-mayorquin, this is ready for review. Basically I use the hdmf.build.builders.BaseBuilder to check if a neurodata object would have a compound dtype. Most of the complexity is introduced by the need to find a match between the neurodata object and its location in the builder, which is outlined in the docstrings. Lmk what you think!

h-mayorquin:
I did a first reading. Two things:
* I think, we should update hdmf on the pyproject to the latest version or greater.
* Looking at the tests that were failing before they seem related to the pixel mask. Can we create a more direct test of this in the dataset configuration tests? I think it would be better to have something more unit-test-like that would fail quicker if we break this (or if we can to refactor). Can we build something simpler with pixel-mask so we don't rely on the full segmentation conversion test?

pauladkisson:
> I think, we should update hdmf on the pyproject to the latest version or greater.

I updated to include everything <4, which zarr-related issues. Should be able to add hdmf 4.0 soon -- see: #1191

pauladkisson:
> Looking at the tests that were failing before they seem related to the pixel mask. Can we create a more direct test of this in the dataset configuration tests? I think it would be better to have something more unit-test-like that would fail quicker if we break this (or if we can to refactor). Can we build something simpler with pixel-mask so we don't rely on the full segmentation conversion test?

Definitely needs some unit tests. I'll put together some.

pauladkisson:
From Meeting: move has_compound_dtype inside get_data_shape

pauladkisson:
> From Meeting: move has_compound_dtype inside get_data_shape

Actually, `get_data_shape` comes from hdmf.utils, so no way to move that in this PR. Also, if I remember correctly, they use `get_data_shape` in ways that would make incorporating this compound_dtype fix difficult.

pauladkisson:
@h-mayorquin, i added tests, so this should be good to go!

src/neuroconv/tools/nwb_helpers/_configuration_models/_base_dataset_io.py
```diff
@@ -245,7 +249,9 @@ def model_json_schema(cls, **kwargs) -> dict[str, Any]:
         return super().model_json_schema(mode="validation", schema_generator=PureJSONSchemaGenerator, **kwargs)
 
     @classmethod
-    def from_neurodata_object(cls, neurodata_object: Container, dataset_name: Literal["data", "timestamps"]) -> Self:
+    def from_neurodata_object(
```
h-mayorquin, line 252: I would like to make the builder argument optional. If the builder is not passed, then the check for a compound object is not performed.

Reasons:
1. We might find a better way to handle this in the future, and I want to avoid breaking backwards compatibility if that happens.
2. I want a quick way to test the effect of compound types on the full shape.
3. We don't need to complicate tests for cases without compound data types by requiring the building of a manager, passing it, or finding the builder.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_models/test_dataset_io_configuration_helpers.py
h-mayorquin, line 1: Most of those tests are for the infrastructure that sustains the desired behavior and is not clear to me how they used to fail before this PR. Can you do something like:

```
from neuroconv.tools.nwb_helpers import HDF5DatasetIOConfiguration

nwbfile = mock_NWBFile()
imaging_plane = mock_ImagingPlane(nwbfile=nwbfile)
plane_segmentation = PlaneSegmentation(
    name="PlaneSegmentation",
    description="description",
    imaging_plane=imaging_plane,
)
pixel_mask = [[0, 0, 1]]
plane_segmentation.add_roi(pixel_mask=pixel_mask)
nwbfile.processing["ophys"].add(plane_segmentation)

neurodata_object = plane_segmentation.columns[1]


manager = get_manager()
builder = manager.build(nwbfile)
dst_config = HDF5DatasetIOConfiguration.from_neurodata_object(
    neurodata_object=neurodata_object,
    dataset_name="data",
   builder=builder,
)
assert dst_config.full_shape == (1, )
``


or these the `get_full_shape_function`  instead of the `HDF5DatasetIOConfiguration.from_neurodata_object` (see main comment)

**Review by h-mayorquin — CHANGES_REQUESTED** [review]
My apologies regarding the `get_data_shape` function.

I propose we create a new function called `get_full_data_shape` in `src/neuroconv/tools/hdmf.py` and consolidate all related functionality there. It seems that this should be something provided by HDMF, and isolating this complexity from the rest of the code would be ideal.

There is also a coupling issue with the calculation of the manager for the entire file. I would prefer not to require `from_neurodata_object` to accept an additional argument that is build related. However, calculating the manager for every dataset is probably too expensive, and using the manager to build only for the dataset doesn't work (based on my initial attempt).

**Requests:**

1. **Centralize Logic:**  
   Move the code to HDMF in the `tools` directory and create a new function, `get_full_shape`, that takes the builder as an optional argument so that all the logic is centralized. In the docstring, document that `get_data_shape` fails for compound objects and that this behavior is desired for building a dataset I/O configuration object.

2. **Add Tests:**  
   Add a test that asserts the correct behavior of either the new function `get_dataset_full_shape` or the dataset IO configuration produced by the `from_neurodata_object` method. See the more concrete request on the review.

File: tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_models/test_dataset_io_configuration_helpers.py [hunk 2]
```diff
@@ -0,0 +1,190 @@
...
+from neuroconv.tools.nwb_helpers._configuration_models._base_dataset_io import (
+    _find_location_in_memory_nwbfile,
+    _find_sub_builder,
+    _infer_dtype,
+    get_dataset_builder,
+    has_compound_dtype,
+)
+
+
+def test_find_location_in_memory_nwbfile():
```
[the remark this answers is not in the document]
pauladkisson: > we don't have them somewhere else?

Nope, no dedicated unit tests, so I figured I would take the opportunity to make some.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_models/test_dataset_io_configuration_helpers.py
pauladkisson: tests/test_minimal/test_tools_hdmf.py: test_get_full_data_shape_compound should accomplish what you're asking for here.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_models/test_dataset_io_configuration_model.py
```diff
@@ -89,8 +90,12 @@ def get_data_io_kwargs(self):
     data = np.array(["test", "string", "abc"], dtype=object)
     nwbfile.add_trial_column(name="test", description="test column with object dtype but all strings", data=data)
     neurodata_object = nwbfile.trials.columns[2]
+    manager = get_manager()
```
h-mayorquin, line 93: Do we still need this here after making the builder optional?

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_models/test_dataset_io_configuration_helpers.py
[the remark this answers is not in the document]
h-mayorquin: OK, makes sense. 
Thanks.

tests/test_minimal/test_tools/test_backend_and_dataset_configuration/test_models/test_dataset_io_configuration_model.py
pauladkisson: Not strictly speaking, so I made one version of the test with a builder and one without.

h-mayorquin: How are you thinking about it? 
If you are testing consistency between passing builder and not builder shouldn't they be the same test to see that the output is the same?

To be honest I was just thinking in not having a test for the builder at this level. This behavior does not dependent on it (it is only relevant for compound data types) but maybe I am missing something, how are you thinking?

pauladkisson: I added it for completeness (testing all the paths that the code could travel through), but honestly I agree that not having a test for builder behavior at this level is fine since it will be tested at the `get_default_dataset_io_configurations` level and at the `get_full_data_shape` level.

h-mayorquin (approved):
LGTM

Should we, in another PR, increase the floor of the hdmf version? Remove the ceiling?

pauladkisson:
> Should we, in another PR, increase the floor of the hdmf version? Remove the ceiling?

Sure, can you take care of that?

h-mayorquin:
@pauladkisson 
I can take care of the ceiling but I am not certain if we should increase the floor. Any idea?

pauladkisson:
> I can take care of the ceiling but I am not certain if we should increase the floor. Any idea?

I don't think we need to as long as it remains compatible with spikeinterface, etc.

PR #1151, opened 2024-12-04
Fix live-services tests (dandi) on windows

h-mayorquin:
We are getting failure on dandi tests on windows for well known reasons:

https://github.com/catalystneuro/neuroconv/pull/1086

The teardown mechanism of the unittest structure removes the directory which fails sometimes on windows because of permissions. This PR changes the testing from unittest to pytest so instead of using the ad-hoc teardown it uses pytest tmp_path methods which should be safer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,4 @@
-# Upcoming
-
-## Features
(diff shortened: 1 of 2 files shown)


pauladkisson (approved):
Looks good!

PR #1150, opened 2024-12-04
Support arbitrary multi probe and multi trigger structures in spikeglx

h-mayorquin:
This should fix #1149 and #978

Issue #629, opened 2023-11-06
[Feature]: solve naming collisions in metadata


bendichter:
### What would you like to see added to NeuroConv?

`NWBConverter`s currently have a usability problem due to naming collisions. For example, let's define a converter with two RecordingInterface classes, both of which define `es_key = "ElectricalSeries"` by default.

```python
class NWBConverterChild(NWBConverter):
    data_interface_classes = dict(Interface1=MockRecordingInterface, Interface2=MockRecordingInterface)

converter = NWBConverterChild(dict(Interface1=dict(), Interface2=dict()))
converter.get_metadata_schema()["properties"]["Ecephys"]["properties"].keys()
```

```python
dict_keys(['Device', 'ElectrodeGroup', 'Electrodes', 'definitions', 'ElectricalSeries'])
```

There are two interfaces, yet we end up with a single field ElectricalSeries, because their is a naming collision: both interfaces default to `es_key="ElectricalSeries"`. So a user needs to know to override this with `es_key="something_unique"` in the source input for each interface, which is unintuitive and not a great user experience.

**Solution**
In a `NWBConverter`, we know that the interface names are unique, since they are all keys in the same `dict`, so if we prepend with those names, all of the "ElectricalSeries" fields in the metadata struct should also be unique.

Basically, I want:

```python
>>> converter.get_metadata_schema()["properties"]["Ecephys"]["properties"].keys()
dict_keys(['Device', 'ElectrodeGroup', 'Electrodes', 'definitions', 'Interface1_ElectricalSeries', 'Interface2_ElectricalSeries'])
```

This comes up most often with ecephys, though ideally we would find a solution that would work for all modalities.

Thoughts, @CodyCBakerPhD ?

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

CodyCBakerPhD:
A name generator like in PyNWB (https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/testing/mock/utils.py) or like pytest factories use is usually the way to go however, some annoyances that can result from that are

[...]

PR #1152, opened 2024-12-10
Support digital channels on NIDQ interface and use TimeSeries instead of ElecricalSeries for analog channels

### LINKED ISSUE #923: [Feature]: Events extraction from Neuropixels data

How can we best get support for this https://github.com/NeuralEnsemble/python-neo/pull/1383 included in neuroconv?

@CodyCBakerPhD
@manimoh

Converting neuropixels data

### Code of Conduct

### LINKED ISSUE #994: [Discussion] Should NIDQ channels be treated like ElectricalSeries?
From a discussion with @bendichter about examples where we want to extend a column in the electrodes table.

While working on https://github.com/catalystneuro/neuroconv/pull/985 I realize the that we are writing the nidq as an ElectricalSeries:

```python
from neuroconv.converters import SpikeGLXConverterPipe

converter = SpikeGLXConverterPipe(folder_path=SPIKEGLX_PATH / "Noise4Sam_g0")
nwbfile = converter.create_nwbfile()


electrodes_df = nwbfile.electrodes.to_dataframe()
electrodes_df[electrodes_df["channel_name"].str[0] == "X"]
```

image

Reading the docs in the schema it seems that NIDQ does not really fit there:

https://github.com/NeurodataWithoutBorders/nwb-schema/blob/63ac845b2e1afdf6fc79466c1feafdd34538bb10/core/nwb.ecephys.yaml#L1-L7

Quoting:

> A time series of acquired voltage data from **extracellular recordings** ....

(bold mine) 

Maybe it is OK to write NIDQ as ElectricalSeries because the units are volts but what seems odd to me is that we are also writing the channels in the electrodes table.

I feel this should not be the case?
Thoughts?

h-mayorquin:
This should come after https://github.com/catalystneuro/neuroconv/pull/1150 and has it as a base.

This should close #994 and also should close #923


src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -56,12 +52,18 @@ def __init__(
             Path to .nidq.bin file.
         verbose : bool, default: True
             Whether to output verbose text.
-        load_sync_channel : bool, default: False
-            Whether to load the last channel in the stream, which is typically used for synchronization.
-            If True, then the probe is not loaded.
         es_key : str, default: "ElectricalSeriesNIDQ"
         """
 
+        if load_sync_channel:
```
bendichter, line 58: What is the behavior change? What should be done instead?

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -56,12 +52,18 @@ def __init__(
...
+            import warnings
+
+            warnings.warn(
+                "The load_sync_channel parameter is deprecated and will be removed in June 2025.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
bendichter, line 65: What is the behavior change? What should be done instead?

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -72,18 +74,36 @@ def __init__(
...
+        if self.has_digital_channels:
+            import ndx_events  # noqa: F401
+            from spikeinterface.extractors import SpikeGLXEventExtractor
+
+            self.event_extractor = SpikeGLXEventExtractor(folder_path=self.folder_path)
+
         super().__init__(
             verbose=verbose,
             load_sync_channel=load_sync_channel,
             es_key=es_key,
```
bendichter, line 99: should es_key still be here? "es" stands for ElectricalSeries

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -101,24 +121,199 @@ def get_metadata(self) -> dict:
...
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
+        if write_electrical_series is not True:
+            warnings.warn(
+                "The 'write_electrical_series' parameter is deprecated and will be removed in June 2025.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
bendichter, line 189: For each of these, it would be helpful to provide a bit more explanation about why it is being deprecated/what the new behavior is/what to do to get the same result instead of using this arg

bendichter:
This looks great! Thanks for putting this together. Just a minor request about the warn messages

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
h-mayorquin: The last channel of the ap and lf bands of SpikeGLX binaries are used to store synch data. This command does not make sense with the NIDQ interface because there is no such channel. To be honest, I don't know what role was playing here before.

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
h-mayorquin: Yes, I think you are right. 

At the moment, it is not clear how to pass metadata to this object but also there is not clear metadata to be passed. I discussed yesterday with a user that has this type of data about what could be useful for their use case and the only thing that came up is a description of the channels. 

I will open another issue to discuss this.

Related, I really like the proposal that you had here:
https://github.com/catalystneuro/neuroconv/issues/629

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
bendichter: yeah, that makes sense. I just think we need a sentence or two alluding to this for users who are less familiar and may be copying/pasting old code

h-mayorquin: Yes, I updated all the warning strings to add context and information. Take a look and let me know what you think.

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
h-mayorquin: OK, I opened an issue here:
https://github.com/catalystneuro/neuroconv/issues/1154

I would like to hear your thoughts on it when you have some time.

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
h-mayorquin: Improved the warning to give more context.

bendichter (approved):
Perfect!

PR #1153, opened 2024-12-10
Set `verbose = False` as default on all the interfaces

h-mayorquin:
We have discussed this many times but never did it.


src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
```diff
@@ -171,7 +171,7 @@ def add_to_nwbfile(
         write_as: Literal["stimulus", "acquisition"] = "stimulus",
         iterator_options: Optional[dict] = None,
         overwrite: bool = False,
-        verbose: bool = True,
```
pauladkisson, line 174: This verbose is actually unused (and redundant with the one in the init of the same interface), so it should be removed.

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
```diff
@@ -20,11 +20,11 @@ class BaseRecordingExtractorInterface(BaseExtractorInterface):
 
     ExtractorModuleName = "spikeinterface.extractors"
 
-    def __init__(self, verbose: bool = True, es_key: str = "ElectricalSeries", **source_data):
+    def __init__(self, verbose: bool = False, es_key: str = "ElectricalSeries", **source_data):
         """
         Parameters
         ----------
-        verbose : bool, default: True
+        verbose : bool, default: Falsee
```
pauladkisson, line 27:
```suggestion
        verbose : bool, default: False
```

pauladkisson (commented):
Looks good overall. 

Do you think we should propagate verbose from NWBConverter to the datainterfaces (if they aren't already specified in `source_data`)? I usually want either all the interfaces verbose or none, but I don't know about other people's usage.

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
```diff
@@ -171,7 +171,7 @@ def add_to_nwbfile(
...
+        verbose: bool = False,
```
pauladkisson, line 174: This verbose is actually unused (and redundant with the one in the init of the same interface), so it should be removed.

src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
h-mayorquin: Thanks for catching this.

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
h-mayorquin: You are right. Not sure where it is best. I think that I prefer the __init__ one.

h-mayorquin:
> Do you think we should propagate verbose from NWBConverter to the datainterfaces (if they aren't already specified in source_data)? I usually want either all the interfaces verbose or none, but I don't know about other people's usage.

Yes, I think this is a good idea.

src/neuroconv/datainterfaces/behavior/audio/audiointerface.py
pauladkisson: init is definitely best.  It's not even used in the add_to_nwbfile method at all. Plus all other interfaces define verbose in the init (I checked).

h-mayorquin: Makes sense, removed. 

I also removed the overwrite here. I think this was copied verbatim from some other place or we forgot to update it . 

Thanks for noticing.

bendichter (approved):

PR #1155, opened 2024-12-11
Soft deprecate `file_path` for `SpikeGLXRecordingInterface` and `SpikeGLXNIDQInterface`

### LINKED ISSUE #978: Match signature of `SpikeGLXRecordingInterface` interface to `SpikeGLXRecordingExtractor` extractor
`SpikeGLXRecordingInterface` takes a file_path as an input and uses it to determine the stream whereas `SpikeGLXRecordingExtractor` takes a folder as input and uses the stream to determine the file.

I think we are kind of duplicating efforts here and it feels to me that we should strieve to keep the interafce between the two libraries consistent. Any downsides of moving in this direction that I am not seeing?

h-mayorquin:
As this is an important change, we are giving it  a year instead of our typical six months. 

Should close #978

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
@@ -36,7 +36,7 @@ class SpikeGLXNIDQInterface(BaseDataInterface):
         self,
         file_path: Optional[FilePath] = None,
         verbose: bool = True,
-        load_sync_channel: Optional[bool] = False,
+        load_sync_channel: Optional[bool] = None,
         es_key: str = "ElectricalSeriesNIDQ",
         folder_path: Optional[DirectoryPath] = None,
     ):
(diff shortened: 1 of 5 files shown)


src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
```diff
@@ -36,7 +36,7 @@ def __init__(
         self,
         file_path: Optional[FilePath] = None,
         verbose: bool = True,
-        load_sync_channel: bool = False,
+        load_sync_channel: Optional[bool] = False,
```
bendichter, line 39:
```suggestion
        load_sync_channel: Optional[bool] = None,
```

bendichter: this is supposed to be `None`, right?

bendichter:
lgtm

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglxnidqinterface.py
h-mayorquin: Yes.

bendichter (approved):

PR #1158, opened 2024-12-13
Ecephys warnings removal and clean up 

h-mayorquin:
I have been working on ecephys pipeline this week and these are some warnings that made it more difficult for me to find the errors. Besides, I am doing some other minor clean up.

* @pauladkisson, I am removing the stub tests that [we discussed before](https://github.com/catalystneuro/neuroconv/pull/1136#issuecomment-2475129398) as those are now tested with the mocking interface
* I am deprecating the `starting_time` of `add_electrical_series` as we have the time alignment methods now. We did the same in #1152

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 * Removed use of `jsonschema.RefResolver` as it will be deprecated from the jsonschema library [PR #1133](https://github.com/catalystneuro/neuroconv/pull/1133)
 * Completely removed compression settings from most places [PR #1126](https://github.com/catalystneuro/neuroconv/pull/1126)
 * Soft deprecation for `file_path` as an argument of  `SpikeGLXNIDQInterface` and `SpikeGLXRecordingInterface` [PR #1155](https://github.com/catalystneuro/neuroconv/pull/1155)
+* `starting_time` in RecordingInterfaces has given a soft deprecation in favor of time alignment methods [PR #1158](https://github.com/catalystneuro/neuroconv/pull/1158)
+
 
 ## Bug Fixes
 * datetime objects now can be validated as conversion options [#1139](https://github.com/catalystneuro/neuroconv/pull/1126)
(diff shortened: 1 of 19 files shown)


pauladkisson:
> * @pauladkisson, I am removing the stub tests that [we discussed before](https://github.com/catalystneuro/neuroconv/pull/1136#issuecomment-2475129398) as those are now tested with the mocking interface

Looks good.

> * I am deprecating the `starting_time` of `add_electrical_series` as we have the time alignment methods now. We did the same in [Support digital channels on NIDQ interface and use TimeSeries instead of ElecricalSeries for analog channels #1152](https://github.com/catalystneuro/neuroconv/pull/1152)

If we are removing the ability to set the starting times directly, we should probably include a way to set the starting_time and rate (currently, the temporal alignment methods require that aligned starting times use full timestamps.

pauladkisson:
> We have methods for this:
> ...
> And the function to add ElectricalSeries decides whether to write starting_time + rate or timestamps.

Ok, so to set the starting time, you would call `set_aligned_starting_time`, which would 
1. use `recording_extractor.get_times()` to automatically construct a full timestamps vector from starting_time and rate
2. add `aligned_starting_time` to every timestamp in that vector
3. when the `add_electrical_series_to_nwbfile` sees these aligned timestamps, it would calculate that they are regular, and extract the starting time and rate to put in the nwb file.

I guess this works, it's a little convoluted/inefficient, but simple for the user, which is more important...

I think this is a case (temporal alignment in general really) where it would be really nice to have a how-to in the docs to make it easier to use.  But...that's definitely outside the scope of this PR and would require some discussion.

h-mayorquin:
Correct.

Yes, it is inefficient as of now. I will make some improvements from SpikeInterface. The advantage of this is just to have this code in only one place instead of having two ways of doing the same spread across the codebase.


> I think this is a case (temporal alignment in general really) where it would be really nice to have a how-to in the docs to make it easier to use. But...that's definitely outside the scope of this PR and would require some discussion.

Yes, that would be nice.

pauladkisson (approved):
Looks good!

PR #1160, opened 2024-12-16
Add missing typing to `NWBConverter` backend and backend options 

h-mayorquin:
Running local tests for the next release I noticed and changed three things:

1) The converter was missing the zarr option in the signature for typing. I added the typing and modified the docstring accordingly.
2) The miniscope tests were comparing session start time  with time zone included. This only works on the CI but fails locally if someone is not in the time zone of the CI server. I made the comparison on the naive datetime object only.
3) The TDT interface for fiber photometry is outputting files on the repo directory. I changed the conversion gallery so it uses the same machinery as all the other interfaces (where the output of the gallery is written to a temporary directory).

### DIFF
--- a/src/neuroconv/nwbconverter.py
+++ b/src/neuroconv/nwbconverter.py
@@ -231,7 +231,7 @@ class NWBConverter:
             The configuration model to use when configuring the datasets for this backend.
             To customize, call the `.get_default_backend_configuration(...)` method, modify the returned
             BackendConfiguration object, and pass that instead.
-            Otherwise, all datasets will use default configuration settings..
+            Otherwise, all datasets will use default configuration settings.
         conversion_options : dict, optional
             Similar to source_data, a dictionary containing keywords for each interface for which non-default
             conversion specification is requested.
--- a/src/neuroconv/nwbconverter.py
+++ b/src/neuroconv/nwbconverter.py
@@ -204,11 +204,8 @@ class NWBConverter:
         nwbfile: Optional[NWBFile] = None,
         metadata: Optional[dict] = None,
         overwrite: bool = False,
-        # TODO: when all H5DataIO prewraps are gone, introduce Zarr safely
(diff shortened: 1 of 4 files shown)


src/neuroconv/nwbconverter.py
```diff
@@ -226,15 +223,15 @@ def run_conversion(
...
             The type of backend to use when writing the file.
             If a `backend_configuration` is not specified, the default type will be "hdf5".
             If a `backend_configuration` is specified, then the type will be auto-detected.
-        backend_configuration : HDF5BackendConfiguration, optional
+        backend_configuration : HDF5BackendConfiguration or ZarrBackendConfiguration, optional
             The configuration model to use when configuring the datasets for this backend.
             To customize, call the `.get_default_backend_configuration(...)` method, modify the returned
             BackendConfiguration object, and pass that instead.
-            Otherwise, all datasets will use default configuration settings.
+            Otherwise, all datasets will use default configuration settings..
```
weiglszonja, line 234:
```suggestion
            Otherwise, all datasets will use default configuration settings.
```

tests/test_on_data/ophys/test_miniscope_converter.py
```diff
@@ -159,8 +159,8 @@ def assertNWBFileStructure(self, nwbfile_path: str):
             nwbfile = io.read()
 
             self.assertEqual(
-                nwbfile.session_start_time,
-                datetime(2021, 10, 7, 15, 3, 28, 635).astimezone(),
+                nwbfile.session_start_time.replace(tzinfo=None),
```
weiglszonja, line 162: Great catch, I didn't think about that this can cause issues. Thanks!

weiglszonja (approved):
Looks good, one small typo. Thanks!

src/neuroconv/nwbconverter.py
h-mayorquin: Thanks!

PR #1162, opened 2024-12-17
Improve temporally_align_data_interfaces

### LINKED ISSUE #1161: [Feature]: NWBConverter.temporally_align_data_interfaces() should take metadata and conversion_options as kwargs

I often need access to info in the metadata and/or conversion_options for temporal alignment, but in order to do so, I have to redefine `run_conversion` with this info passed on to `temporally_align_data_interfaces`. By providing this info in our default implementation it would reduce duplicate code.

_No response_

Yes.

### Code of Conduct

pauladkisson:
Fixes #1161 

Unfortunately, this will break back-compatibility for anyone who implemented their own version of `temporally_align_data_interfaces` w/o re-implementing `run_conversion`. So, probably needs to be slated for 0.7.0.

@h-mayorquin, do you have a list of TODOs for 0.7.0 or a branch or something so that it doesn't get lost?


h-mayorquin:
> @h-mayorquin, do you have a list of TODOs for 0.7.0 or a branch or something so that it doesn't get lost?

No, but after the next release, we will make breaking changes so the next should be 0.7.0 anyway.

h-mayorquin:
I slept on this, and here is my take:

* I can't find any good argument against having this, and it seems that this should have been the case from the beginning.
* I am not a heavy user of this function; I usually work directly at the interface level. I think you should bring it up in one of the team meetings to get feedback—especially from Ben, who, if I remember correctly, is a user of this and likely wants more input on high-level API decisions.
* It is worth considering whether some judicious use of kwargs and/or positional-only arguments could avoid the backward compatibility break—not necessarily here, but at least in other situations where we might need to change the function signature in the future.
* I still think the accompanying issue requires you to write a concrete case so better reasons for the change are documented somewhere.

So, I am a +1 on this. It is unclear to me what kind of test, if any, this might require.

h-mayorquin:
I think no one had arguments against this. Let's add it to the next major version. The new current changelog has a section for the 0.7.0. Can you add the changelog there:

https://github.com/catalystneuro/neuroconv/blob/cd7764e699ea77bb279ba7f04c974d5e27f12393/CHANGELOG.md#L1-L9

Issue #932, opened 2024-06-27
Decide and test priority of `group_name` and `group` in `/tools/spikeinterface/spikeinterface.py` functions.


h-mayorquin:
From here:
#634

The relevant comments are here:

https://github.com/catalystneuro/neuroconv/issues/634#issuecomment-1805902718

And more specifically: 
@alejoe91 
> When loading and setting a probe on the SI side, the group property is set. However, the add_electrode_groups has a precedence on the group_name over the group property to define electrode groups (see [here](https://github.com/catalystneuro/neuroconv/blob/283c37423d0e6184c57cd645eb0164013fcf4854/src/neuroconv/tools/spikeinterface/spikeinterface.py#L131-L134)). Therefore, the probe group is ignored.

So if the recorder of the user has a `group_name` our functions will write that one instead of the `group` from the probe which is probably what the user intended. 

I need to check a couple of cases on the spikeinterface side. I will come back to this but I wanted to document this so we don't forget.

PR #1164, opened 2024-12-18
Fix extra electrode groups ephys

### LINKED ISSUE #1163: Ecephys pipeline writing extra device and extra electrode groups when setting `group_names`
This issue arose during the DiCarlo data upload. Since my last conversion, the names of the electrode groups (in this case, the Intan ports) have been added to the metadata. This addition has caused the NWB file to contain more devices and electrode groups than expected, and the extra ones are incorrect.

Upon investigation, I found this to be a general problem. I will first address the general issue and then implement Intan-specific fixes. Here is a minimal example illustrating the general issue:

```python
from neuroconv.tools.testing.mock_interfaces import MockRecordingInterface


interface = MockRecordingInterface(num_channels=4)

recording_extractor = interface.recording_extractor
recording_extractor.set_channel_groups(groups=[0, 1, 2, 3], channel_ids=[0, 1, 2, 3])   
recording_extractor.set_property(key="group_name", values=["group1", "group2", "group3", "group4"])
nwbfile = interface.create_nwbfile()

assert len(nwbfile.devices) == 1
assert len( len(nwbfile.ec_electrode_groups) == 4
``` 

Both fail on main (the results being 2 and 8).\

I will push both fixes soon.

h-mayorquin:
Should fix #1163. 

Also related to https://github.com/catalystneuro/neuroconv/issues/932 as that should be more or less the last place where we did not priority set.


pauladkisson (approved):
Looks good!

h-mayorquin:
Thanks for the review @pauladkisson

PR #1165, opened 2024-12-18
Detect mismatch errors between group and group names when writing `ElectrodeGroups`

### LINKED ISSUE #932: Decide and test priority of `group_name` and `group` in `/tools/spikeinterface/spikeinterface.py` functions.

h-mayorquin:
This should come after  #1164 and has it as its base.

This should close  #932 as we have now a function that decides the right priority between "group_name" and "group". This PR adds some errors for two different types of mismatch:

1) The group_names have different number of unique entries than groups.
2) The group_names and the groups are consistent across the array

@alejoe91 could you take a look?


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -259,6 +267,23 @@ def _get_group_name(recording: BaseRecording) -> np.ndarray:
     # If for any reason the group names are empty, fill them with the default
     group_names[group_names == ""] = default_value
 
+    # Validate group names against groups
+    if groups is not None:
+        unique_groups = np.unique(groups)
```
h-mayorquin, line 272: Mmm I will change this to sets as I remember np.unique being buggy for strings.

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -245,11 +245,19 @@ def _get_group_name(recording: BaseRecording) -> np.ndarray:
...
+
+    Raises
+    ------
+    ValueError
+        If the number of unique group names doesn't match the number of unique groups,
+        or if the mapping between group names and group numbers is inconsistent.
     """
     default_value = "ElectrodeGroup"
     group_names = recording.get_property("group_name")
+    groups = recording.get_channel_groups()
```
alejoe91, line 257: I think this us never None, so L262 will never be reached

alejoe91 (approved):

src/neuroconv/tools/spikeinterface/spikeinterface.py
h-mayorquin: Made the correction.

PR #1167, opened 2024-12-19
Avoid writing extra devices when using `IntanRecordingInterface` with multiple electrode groups

h-mayorquin:
As in the title. 

This is after #1164 as that laid the ground-work for this.

The problem is the last time I hardcoded that we only add the IntanDevice to the first group. The other groups have a default Device and that's where the extra ones come from. 

I added a test and fixed this.


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Small fixes should be here.
 ## Features
 
 ## Improvements
+* Fix metadata bug in `IntanRecordingInterface` where incorrected devices were added if the recording contained multiple electrode groups or names [#1166](https://github.com/catalystneuro/neuroconv/pull/1166)
```
weiglszonja, line 21:
```suggestion
* Fix metadata bug in `IntanRecordingInterface` where incorrect devices were added if the recording contained multiple electrode groups or names [#1166](https://github.com/catalystneuro/neuroconv/pull/1166)
```

tests/test_on_data/ecephys/test_recording_interfaces.py
```diff
@@ -268,6 +267,29 @@ def test_devices_written_correctly(self, setup_interface):
...
+    def test_not_adding_extra_devices_when_recording_has_groups(self, setup_interface):
+        # Test that no extra-devices are added when the recording has groups
+
+        nwbfile = mock_NWBFile()
+        self.interface.add_to_nwbfile(nwbfile=nwbfile)
+        recording = self.interface.recording_extractor
+        num_channels = recording.get_num_channels()
+        channel_groups = np.full(shape=num_channels, fill_value=0, dtype=int)
+        channel_groups[::2] = 1
+        recording.set_channel_groups(groups=channel_groups)
```
weiglszonja, line 279: Just curious, why is `self.interface.add_to_nwbfile(nwbfile=nwbfile)` not called after `recording.set_channel_groups(groups=channel_groups)`? Does the mock extractor already have channel groups when you add to the file?

CHANGELOG.md
h-mayorquin: Thanks, this is better!

tests/test_on_data/ecephys/test_recording_interfaces.py
h-mayorquin: Thanks, the tests were wrong. Adding the interfaces' data to the nwbfile should be done after the setup. Should be corrected now.

h-mayorquin:
Also blocked by #1175:

See the error here:
https://github.com/catalystneuro/neuroconv/actions/runs/12753555210/job/35545502312?pr=1167

h-mayorquin:
All tests are passing now that we have merged #1175 @weiglszonja

weiglszonja (approved):
Lgtm, thanks!

PR #1170, opened 2024-12-20
Update ndx-pose for Lightning pose

pauladkisson:
Update ndx-pose to >= 0.2 for Lightning pose.


pauladkisson:
@h-mayorquin, the tests are only failing due to SLEAP and DLC, so we can merge this into your PR if you think it looks good.

PR #1175, opened 2025-01-08
Set ceiling for HDMF to avoid failures with the latest release

h-mayorquin:
Currently, there are tests failing due to chunking errors with the latest release of hdmf:

https://github.com/catalystneuro/neuroconv/pull/1168

 I am putting a ceiling on the version right now until we fix that. If I remember well this requires or would be helped by first fixing the PoseEstimation interfaces in other PRs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,13 @@ __pycache__/
 # C extensions
 *.so
 
+
+# UV
+#   Similar to Pipfile.lock, it is generally recommended to include uv.lock in version control.
+#   This is especially recommended for binary packages to ensure reproducibility, and is more
+#   commonly ignored for libraries.
+uv.lock
+
 # Distribution / packaging
 .Python
 build/
@@ -127,3 +134,4 @@ venv.bak/
 
 # Spyder
 .spyproject
+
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -6,6 +6,13 @@ __pycache__/
 # C extensions
 *.so
 
+
```
h-mayorquin, line 9: I have been using uv on my mac and I would like to add uv to our gitignore.

This is in line with what github is doing now, see:
https://github.com/github/gitignore/pull/4545

For the same change "upstream".

tests/test_on_data/setup_paths.py
h-mayorquin, line 1: This was failing to create the file for me on mac when using uv. I guess the problem is that uv uses a different base directory.

h-mayorquin: I also tested this on a conda base installation on my linux system and it correctly moves the file.

h-mayorquin:
https://github.com/catalystneuro/neuroconv/pull/1167 also needs this:

This is the error:
image

tests/test_on_data/setup_paths.py
h-mayorquin: I can't remember right now. Either source or destination was incorrect during the copy.

weiglszonja (approved):

Issue #1147, opened 2024-11-22
Fixes for upcoming hdmf version


pauladkisson:
Dev tests are failing for several reasons due to changes on the dev branch of hdmf.

- [x] Compound dtypes should be fixed by #1146

The deprecated `call_docval_func` function has been removed, but still lingering on some extensions:
- [x] ndx-pose should be fixed by #1128 
- [x] ndx-events should be fixed by #1176

h-mayorquin:
Should this be closed?

PR #1176, opened 2025-01-13
updated ndx-events to 0.2.1

pauladkisson:
The latest release of ndx-events (0.2.1) fixes compatibility issues with dev version of hdmf.

This is a partial fix for #1147

See https://github.com/rly/ndx-events/issues/20


h-mayorquin (approved):
I don't think this needs a changelog.

PR #1177, opened 2025-01-14
SpikeGLX: Add inter-sample shift description and fix ElectrodeGroup naming

h-mayorquin:
From the IBL project. The inspector run reveals that the description of inter-sample-shift is missing. As we always have this for SpikeGLX it is a good idea to add it.

Plus, the current ElectrodeGroup naming is confusing when some of the probes have shanks. Here, the diff is illustrative:

Version on main:
image

Note that the shank number is confusing as they are concatenated with the device number of the probe making it hard to read and identify.

Version on this branch:

image

In this case, the second and third probes have shanks and the ElectrodeGroups indicate the situation more clearly.


alejoe91:
@h-mayorquin this looks great! One comment: I think the name should be Neuropixels instead of Neuropixel, since that's the correct probe name (https://www.neuropixels.org/)

h-mayorquin:
Ah thanks for the naming thing. It was inconsistent between devices and ElectrodeGroups and I opted for the singular. Thanks for the correction.

alejoe91 (approved):

Issue #1023, opened 2024-08-22
[Feature]: Selecting channels from OpenEphys data to convert


rly:
### What would you like to see added to NeuroConv?

I have an OpenEphys binary data file `continuous.dat` that has shape (num_samples = 157733308, num_channels = 264). The first 256 channels are neural data. The last 8 channels are analog data (the first one contains received TTL pulses). I want to convert the 256 channels into an ElectricalSeries. Is there a way to configure NeuroConv to do this? Similarly, I would like to convert just channel index 256 to a TimeSeries.  

At the JHU workshop, we ran into a related issue where someone was using their last recording channel to record something non-neural. I forget which system. I wasn't sure how to use NeuroConv for this.

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

h-mayorquin:
Hi, @rly . Can you tell me more about your case?  I would like to fix this at the root and it would be useful to document cases when spikeinterface is loading more channels that it should (see [here](https://github.com/SpikeInterface/spikeinterface/issues/3197))

Meanwhile, yes, there is a way. Here is example code of how would you do that:

```python
from neuroconv.tools.testing.mock_interfaces import MockRecordingInterface
from neuroconv import ConverterPipe


interface = MockRecordingInterface(num_channels=4)
# Suppose you only want to convert the first two channels

recording = interface.recording_extractor
channel_ids_to_convert = [0, 1]
interface.recording_extractor = recording.select_channels(channel_ids=channel_ids_to_convert)

data_interfaces = [interface]

# data_interfaces.append(other_data_interfaces_of_the_conversion)
 
converter_pipe = ConverterPipe(data_interfaces=data_interfaces)

converter_pipe.run_conversion(**run_conversion_kwargs)
```

Let me know if it makes sense.

rly:
@h-mayorquin thanks! that's useful. The OpenEphys system was set up with 266 channels. In `settings.xml`, you can kind of see how different channels are grouped by inspecting the channel info:

```xml
<?xml version="1.0" encoding="UTF-8"?>

<SETTINGS>
  <INFO>
    <VERSION>0.4.5</VERSION>
[...]

PR #1179, opened 2025-01-16
Do not convert non-neural channels to ElectricalSeries on `OpenEphysBinaryRecordingInterface`

h-mayorquin:
Should close https://github.com/catalystneuro/neuroconv/issues/1023

Part of an effort like the one in:

https://github.com/catalystneuro/neuroconv/pull/1152

Will add another PR at some point to transform the non-neural channels to a TimeSeries automatically.

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/openephys/openephysbinarydatainterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/openephys/openephysbinarydatainterface.py
@@ -86,6 +86,13 @@ class OpenEphysBinaryRecordingInterface(BaseRecordingExtractorInterface):
         if stub_test:
             self.subset_channels = [0, 1]
 
+        # Check if the recording has ADC channels
+        recording = self.recording_extractor
+        channel_ids = recording.get_channel_ids()
+        neural_channels = [id for id in channel_ids if "ADC" not in id]
+        if len(neural_channels) < len(channel_ids):
+            self.recording_extractor = recording.select_channels(channel_ids=neural_channels)
+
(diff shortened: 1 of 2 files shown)


src/neuroconv/datainterfaces/ecephys/openephys/openephysbinarydatainterface.py
```diff
@@ -108,6 +108,13 @@ def __init__(
         if stub_test:
             self.subset_channels = [0, 1]
 
+        # Check if the recording has ADC channels
```
alessandratrapani, line 111: In the data from Kind Lab, they have AUX (AUX1, AUX2, AUX3), not ADC, for the accelerometer signals.
Should we go the other way around? --> checking that the channel_id contains CH, and only if it contains CH we add that as ElectricalSeries.

h-mayorquin: That's a great suggestion. Let me debug against that data. Could you talk to them to see if we can see use some of their data in our gin tests? I can then stub some of that .

h-mayorquin: Ah, but I just checked and your data is legacy, right? This PR is for the binary format where the channels are only ADC:

https://open-ephys.github.io/gui-docs/User-Manual/Recording-data/Open-Ephys-format.html
https://open-ephys.github.io/gui-docs/User-Manual/Recording-data/Binary-format.html

Your data might need another PR altogether.

bendichter (approved):

PR #1190, opened 2025-02-04
Add `ImageInterface` 

### LINKED ISSUE #1186: Add `ImageInterface`
The goal of interfaces is to make conversions easier and we can make the creation of [Images containers](https://pynwb.readthedocs.io/en/stable/pynwb.base.html#pynwb.base.Images) more straightforward in several ways:

* Reducing boilerplate. The user provides the file_paths or a folder_path with images and we can load the images, extract resolution and orientation if they are available in the metadata and write to disk.
* Automatically choose the correct data type based on the image metadata. We currently have three data types in pynwb, [GrayscaleImage](https://pynwb.readthedocs.io/en/stable/pynwb.image.html#pynwb.image.GrayscaleImage), [RGBImage](https://pynwb.readthedocs.io/en/stable/pynwb.image.html#pynwb.image.RGBImage) and [RGBAImage](https://pynwb.readthedocs.io/en/stable/pynwb.image.html#pynwb.image.RGBAImage). The interface will select the correct data type automatically based on the image metadata making life easier for the user.
* We can set good chunking (probably None in most cases?) and compression defaults.
* If there are too many images (say they will require more than 1GiB) we can write them with an iterator to enable the workflow to work in most computers.

h-mayorquin:
Should close #1186.


src/neuroconv/datainterfaces/image/imageinterface.py
```diff
@@ -0,0 +1,269 @@
...
+            for path in self.file_paths:
+                if not Path(path).exists():
+                    raise ValueError(f"File {path} does not exist")
+
+    def get_metadata(self) -> DeepDict:
+        """Get metadata for the images."""
+        metadata = super().get_metadata()
+
+        # Add basic metadata about the images
+        metadata["Images"] = dict(description="Images loaded through ImageInterface", num_images=len(self.file_paths))
```
bendichter, line 216: why would you need to indicate the number of images?

h-mayorquin: We don't as far as I see. I uncritically accepted this from cline I think.

h-mayorquin: Changed, I also added the correct headers on the `rst` that you mentioned.

docs/user_guide/image.rst
h-mayorquin, line 1: Ben mentioned that a better place for this might be the conversion gallery and I think he is right.

h-mayorquin: I moved it.

h-mayorquin:
@bendichter this is ready to go.

bendichter (approved):

PR #1193, opened 2025-02-06
Fix plexon 2 tests

h-mayorquin:
Mysterious failure related to the [zugbruecke](https://zugbruecke.readthedocs.io/en/latest/) library on intel mac. I am not sure if the mac os runner configuration changed as this was working well before and there is **no** [recent release](https://pypi.org/project/zugbruecke/
) of the library.

Skipping tests as this is sort of an edge case.


h-mayorquin:
@pauladkisson This should get the other PRs working. I don't think this needs a changelog.

pauladkisson (approved):
Thanks for tracking this down. I take it that test_raw_recordings.py is redundant with test_recording_interfaces.py?

h-mayorquin:
> Thanks for tracking this down. I take it that test_raw_recordings.py is redundant with test_recording_interfaces.py?

Yes, sorry for not explaining that.

The only test that was remaining in `test_raw_recordings` was the Plexon2 test. I moved it to the new format which is what is available on `test_recordings_interfaces.py`

PR #1192, opened 2025-02-06
updated secrets for live-service-testing

pauladkisson:
Fixes failing dailies that were broken by #1089

See https://github.com/catalystneuro/neuroconv/actions/runs/13170714295

PR #1195, opened 2025-02-07
Reactivate Plexon 2 tests on Mac arm.

h-mayorquin:
As discussed on #1192.

Let's monitor the test for a couple of merges.


h-mayorquin:
@pauladkisson seems to be working here : /

Let's look at this across a couple of updates to see if the error re-appears.

pauladkisson:
> seems to be working here : /

What do you mean? This one is failing: macos-latest Python 3.10

h-mayorquin:
> > seems to be working here : /
> 
> What do you mean? This one is failing: macos-latest Python 3.10

My bad, I spoke too soon. Ok. So, at least we can track it and fix it here.

h-mayorquin:
@pauladkisson this is finally working again. I changed the way that wine is installed on the action.

PR #1198, opened 2025-02-11
Fixed docker tests

pauladkisson:
See https://github.com/catalystneuro/neuroconv/actions/runs/13254425466


h-mayorquin (approved):
Great, now we need to fix that live test that is failing everywhere.

PR #1199, opened 2025-02-11
Update dandi version floor

h-mayorquin:
Checking if the test in MacOS latest still fails with this.


pauladkisson (approved):
Looks good!

PR #1200, opened 2025-02-11
Call get_metadata when no metadata is passed on ophys interfaces

h-mayorquin:
For ecephys interfaces, we call `get_metadata` when `metadata = None` in `add_to_nwbfile`. This is convenient because the default metadata is always added ensuring that the file is as complete as it can be with what can be extracted automatically.

This PR unifies the behavior between ophys and ecephys by adding this to ophys.


alessandratrapani (approved):
Makes sense to me!

PR #1205, opened 2025-02-18
Fixed daily docker dev image workflow

pauladkisson:
The daily docker dev image workflow (build_and_upload_docker_image_dev.yml) kept getting cancelled, so I removed the concurrency, which should prevent it from getting canceled. Since it's not a PR test, not having concurrency enabled shouldn't be a big deal.

See dailies run from this branch here: https://github.com/catalystneuro/neuroconv/actions/runs/13383889282


pauladkisson:
> Can you add more on why it was being canceled? was it taking too long? What is the frequency of the action? If the frequency is once per day how come concurrency is getting on the way?

Honestly, I'm not sure. I tried to look into it, but I wasn't able to quickly figure out what went wrong, so I figured this was a quick fix.

h-mayorquin:
OK, if this is working I guess we can think on this as a technical debt and flag it as "something weird might be going on there but is not a priority" until it breaks. Fine by me.

h-mayorquin:
Ah, just one thing, can you quickly check if the action is not running for a ridiculous amount of time? I think github has a timeout so probably not problem but it they get stuck and we get a github maybe github actions will penalize us for having a lot of actions running.

h-mayorquin (approved):

PR #1222, opened 2025-02-22
Add MP3 support to AudioInterface and update tests

### LINKED ISSUE #1219: [New Format]: support for mp3 audio

 . 

_No response_

### Existing APIs for format

_No response_

_No response_

Yes.

### Code of Conduct

bendichter:
fix #1219


h-mayorquin:
I hadn’t seen this before opening the other ones—my mistake.

I’ve been thinking about this, and the main challenge is that many of these libraries rely on FFmpeg for reading audio files. However, packaging FFmpeg in a way that works reliably across the three major operating systems within a pip-installable framework is difficult.

To test this, [I’ve created a small set of stub files ](https://gist.github.com/h-mayorquin/6ed9ad9038e5643977aee189ae63d946)for the most common archival formats (WAV, FLAC, AIFF), along with MP3 and OGG, which are also widely used. We can them put them in gin. I believe this is necessary to establish a testing framework where these formats can be accessed without requiring FFmpeg or any other dependency that isn’t pip-installable. This will also allow us to enable CI testing in an environment similar to what our users experience.

A preliminary review suggests that torchaudio could be a straightforward solution, even though it’s quite heavy. Once these test files are available on Gin, we can explore lighter alternatives if we want since PyTorch itself is very large (5GiB as a dependency as they vendorize things inside of the package).

That said, I might be overlooking a better approach. What do you think, Ben?

bendichter:
I see. I'm not crazy about a 5GB dependency just to read MP3s. If we use ffmpeg, the downsides are that
1. Users will need to install ffmpeg if they have not already
2. This will require us to either not include this capability in GUIDE or to put some work into including ffmpeg, which will be a bit tricky because it is OS-dependent.

This may have been why we previously stopped at wav files. I think I would prefer going with librosa even given those down-sides though I agree it's not ideal

PR #1224, opened 2025-02-24
Add mp3 support by Claude Code with Sonet 3.7

h-mayorquin:
Good attempt, it would be nice to play to see if it can make the reading lazy.

#1225

PR #1225, opened 2025-02-24
Add mp3 support by Cline with Sonet 3.7

h-mayorquin:
#1224


h-mayorquin:
OK, after thinking about this I look more the solution provided by Claude Code. Closing this.

PR #1226, opened 2025-02-25
Add support for 24-bit WAV files in AudioInterface and utility functions

### LINKED ISSUE #1218: [Bug]: AudioInterface not working for wav files that use 3 bytes
### What happened?

mmap functionality not working for data that uses 3 byte precision

### Steps to Reproduce

```python
.
```

### Traceback

```python

```

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.8

### Package Versions

_No response_

### Code of Conduct

bendichter:
fix #1218


tests/test_behavior/test_audio_interface.py
```diff
@@ -208,3 +210,40 @@ def test_run_conversion(self):
...
+            # Verify the file was created and can be read
+            with NWBHDF5IO(path=nwbfile_path, mode="r") as io:
+                nwbfile = io.read()
+
+                # Check that the acoustic waveform series exists
+                audio_name = metadata["Behavior"]["Audio"][0]["name"]
+                assert audio_name in nwbfile.stimulus
+
+                # Try to read the data
+                acoustic_series = nwbfile.stimulus[audio_name].data[:]
```
bendichter, line 248:
```suggestion
                acoustic_series = nwbfile.stimulus[audio_name].data
```

h-mayorquin (approved):
Discussion:

I don’t like the utils pattern, as I prefer to keep all the code focused on the interfaces. I would rather keep the `get_wave_bit_depth` code as a method within the interface file (either stand alone or static). If we find more use cases for it, we can consider generalizing it into a utility, but I’d prefer to do that when the need arises. What do you think?

If you still prefer the utils pattern, could we make it private to minimize the API surface? Keeping it private by default, unless we have a strong reason to expose it, would reduce the risk of breaking changes and make the code easier to modify in the future.

This LGTM to me otherwise but I would like to hear what you think.

bendichter:
How about we make it a private function within the audiointerface file until we have a use for it

h-mayorquin:
Yeah, that looks great to me!

Issue #1156, opened 2024-12-12
[Bug]: DANDI Upload with session ID that contains "/" fails


rly:
### What happened?

See https://github.com/dandi/helpdesk/issues/169#issuecomment-2537137478

The user has an NWB file with `session_id` = `"BLA-projecting NBM/SI neuron from animal M24 on 26/07/2023"`. It looks like NeuroConv or dandi-cli is replacing the "/" in the session ID with "\\" when renaming the file before upload, and therefore the file cannot be found. 

DANDI uses the following format for file names: `{dandiset_id}/sub-{subject_id}/sub-{subject_id}_ses-{session_id}.nwb` so a "/" in the session ID (or subject ID) would cause problems unless handled in a special way. 

The error is reported in NWB GUIDE, but from what I can trace, the error comes from calling neuroconv. See traceback below. The inputs to `automatic_dandi_upload` is a folder `/Users/rly/NWB_GUIDE/.temp/upload/temp_20241211-150938` with contents:
```
ls -al ~/NWB_GUIDE/.temp/upload/temp_20241211-150938 
total 0
drwxr-xr-x  3 rly  staff   96 Dec 11 15:09 .
drwxr-xr-x  6 rly  staff  192 Dec 11 15:09 ..
lrwxr-xr-x  1 rly  staff   60 Dec 11 15:09 MACAQUE_CHOL_NBM_FL_20230726T173713.nwb -> /Users/rly/Downloads/MACAQUE_CHOL_NBM_FL_20230726T173713.nwb
```

So the file is renamed with the session ID by neuroconv or dandi organize.

Related: 
- https://github.com/hdmf-dev/hdmf/pull/1219
- https://github.com/NeurodataWithoutBorders/nwbinspector/issues/543

### Steps to Reproduce

```python
See linked issue
```

### Traceback

```python
[2024-12-11 15:09:41,427] ERROR in app: Exception on /neuroconv/upload [POST]
Traceback (most recent call last):
  File "/Users/rly/miniconda3/envs/nwb-guide/lib/python3.9/site-packages/flask/app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/rly/miniconda3/envs/nwb-guide/lib/python3.9/site-packages/flask/app.py", line 1469, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
[...]

PR #1227, opened 2025-02-25
corrected session_id by replacing invalid characters with underscores…

anoushkajain:
solves [issue  #1156](https://github.com/catalystneuro/neuroconv/issues/1156)


h-mayorquin:
Thanks for the draft @anoushkajain. Don't forget to add a test for it and ping me when you think is ready or you need any help with it.

tests/test_minimal/test_tools/dandi_transfer_tools.py
h-mayorquin, line 1: What do you think of creating more than one file given that dandi organize is meant to work across a set of many files. Let's do 4-5-6 or something like that.

Also, for the last bit of the test I think it is better to just be explicit instead of relying on a loop to make the error obvious and transparent. 

What do you think? @alessandratrapani @anoushkajain

alessandratrapani:
This PR was created by mistake. The issue was on nwbinspector (https://github.com/NeurodataWithoutBorders/nwbinspector/issues/543) and fixed here: https://github.com/NeurodataWithoutBorders/nwbinspector/pull/570

h-mayorquin:
I thought it was solving #1156

PR #1228, opened 2025-02-25
Add warning for non-.nwb file extensions in NWB conversion for Converters and Interfaces

### LINKED ISSUE #1212: [Feature]: warning if the output nwbfile_path does not end in ".nwb"

Generally, we advise that NWB files should end in ".nwb". If the user tries to use a path other that .nwb, we should warn them

_No response_

Yes.

### Code of Conduct

bendichter:
fix #1212

### DIFF
--- a/src/neuroconv/basedatainterface.py
+++ b/src/neuroconv/basedatainterface.py
@@ -1,6 +1,7 @@
 import importlib
 import json
 import uuid
+import warnings
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Literal, Optional, Union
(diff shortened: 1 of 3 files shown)


h-mayorquin:
What do you think of throwing an error instead of a warning? 

Reason being that nwb is already throwing a warning [here](https://github.com/NeurodataWithoutBorders/pynwb/pull/1978) and the library should follow good practices, right?

bendichter:
Well I could imagine situations where someone might want to use another suffix. For example, they might want to have x.nwb.h5 and x.nwb.zarr to examine different compression strategies.

But maybe this would break `dandi organize`? I don't know.

There is not a huge advantage of also having the warning here. One small advantage is instead of on `NWBHDF5IO` is that the warning will occur before the conversion begins, which may take a long time.

h-mayorquin:
Check this out:
https://github.com/NeurodataWithoutBorders/nwbinspector/issues/567

If the convention of having `nwb.zarr` is embraced we will need extra logic for this.

Taking into account all of this discussion I am leaning to just outsource this to upstream libraries. They already have a warning on place and that way we don't need to add extra logic here. 

Maybe we should update our dependencies to be at least the ones that have the warnings though.

bendichter:
yeah that warning is recent and I think was just released today. I'd be fine with relying on that

h-mayorquin:
OK.

PR #1229, opened 2025-02-25
Add first draft of .clinerules file

bendichter:


.clinerules
```diff
@@ -0,0 +1,5 @@
+* When developing a new feature or fixing a bug, always write corresponding tests and run the tests to ensure the feature worked
+* When adding tests, first read through the existing tests and try to fit new tests in the existing testing infrastructure
+* If a new feature requires a new dependency, make sure to add it in the pyproject.toml
+* Always make a contribution to the CHANGELOG file summarizing any changes
+* Use typehints and numpy-style docstrings
```
luiztauffer, line 5:
```suggestion
* Use typehints and numpy-style docstrings
* always apply Black formatting
* avoid making any modifications that are not strictly necessary to the task you've been requested to do
* if you spot any possible errors or have suggestions for improvement outside of the scope of your requested task, you can add them to your Task Completed message
```

luiztauffer:
we could also include the [cline memory bank instructions](https://docs.cline.bot/improving-your-prompting-skills/custom-instructions-library/cline-memory-bank) in the `.clinerules` file

h-mayorquin (commented):
I like the changelog suggestion!

Some suggestions that come to mind:
* Maybe give instructions on how to run the tests and where the test are located?
* I don't think black formatting is necessary that is done very cheaply and quickly by the formatter and the pre-commit and might increase the number of iterations that cline does (which reduces feedback time and increases costs). 
* Maybe we should add a request to not modify docstrings outside of the scope of the feature being added to reduce iterations.

luiztauffer:
> I don't think black formatting is necessary that is done very cheaply and quickly by the formatter and the pre-commit and might increase the number of iterations that cline does (which reduces feedback time and increases costs).

yeah, that's a good point. It should probably be a general rule to avoid asking the LLMs to do things that we can automate otherwise

.clinerules
```diff
@@ -0,0 +1,8 @@
...
+* always apply Black formatting
+* avoid making any modifications that are not strictly necessary to the task you've been requested to do
```
bendichter, line 7:
```suggestion
* avoid making any modifications that are not strictly necessary to the task you've been requested to do (including dostrings)
```

.clinerules
```diff
@@ -0,0 +1,8 @@
+* When developing a new feature or fixing a bug, always write corresponding tests and run the tests to ensure the feature worked
+* When adding tests, first read through the existing tests and try to fit new tests in the existing testing infrastructure
+* If a new feature requires a new dependency, make sure to add it in the pyproject.toml
+* Always make a contribution to the CHANGELOG file summarizing any changes
+* Use typehints and numpy-style docstrings
+* always apply Black formatting
+* avoid making any modifications that are not strictly necessary to the task you've been requested to do (including dostrings)
```
bendichter, line 7:
```suggestion
* avoid making any modifications that are not strictly necessary to the task you've been requested to do (including docstrings)
```

bendichter:
Thanks for the thoughtful feedback!

My experience is that our test launching infrastructure for this repo is pretty standard, and as a result cline is pretty good at running tests. It will often run targeted runs through pytest - just a single relevant test or just a single test file. I really like that, so I am inclined to let it do its thing there without additional guidance.

I agree black formatting is not necessary, but it's nice to have it in place so you don't have to constantly re-pull after the formatting is applied, so let's try leaving it. We can remove it later if it is causing problems.

Let's test the memory bank approach ourselves a bit before we impose that as standard practice on a core repo

bendichter:
Might be also worth checking out @magland 's .clinerules file here: https://github.com/flatironinstitute/neurosift/blob/main-v2/.clinerules

h-mayorquin (approved):
OK for me.

PR #1232, opened 2025-02-27
Call `get_metadata` when no metadata is passed on segmentation interfaces 

h-mayorquin:
Follow up on #1200. Explanation can be found there.


alessandratrapani (approved):
LGTM

PR #1234, opened 2025-03-04
Fix parsing of group names with numpy 2.0

alejoe91:
This is a weird behavior with numpy 2.0 in [this function](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/spikeinterface/spikeinterface.py#L1815).

In the current implementation, applying `list` to `np.unique` [here](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/spikeinterface/spikeinterface.py#L1820) results in:
```
print(group_names)

>>> [np.str_('ProbeA')]
```
This messes up the [pandas query](`https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/spikeinterface/spikeinterface.py#L1827`), which doesn't recognize the `np.` symbol.

This is a simple fix to make sure all elements of the `group_names` are parsed to python `str`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Deprecations and Changes
 
 ## Bug Fixes
+* Fix parsing of group_names in `tools.spikeinterface`
 
 ## Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Deprecations and Changes
 
 ## Bug Fixes
-* Fix parsing of group_names in `tools.spikeinterface`
+* Fix parsing of group_names in `tools.spikeinterface` [PR #1234](https://github.com/catalystneuro/neuroconv/pull/1234/files)
 
 ## Features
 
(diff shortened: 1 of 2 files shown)


h-mayorquin:
> This messes up the pandas query, which doesn't recognize the np. symbol.

Which pandas query?

h-mayorquin:
Ah, this query:

https://github.com/catalystneuro/neuroconv/blob/fe9badc6c42e0fdada321487087f7da4a276175b/src/neuroconv/tools/spikeinterface/spikeinterface.py#L1827

h-mayorquin:
LGTM to me, is there a test we can add?

h-mayorquin:
Oh, this is surfacing here:
https://github.com/catalystneuro/neuroconv/actions/runs/13640334933/job/38128592505?pr=1117

So it seems this is either a python 3.13 or newer numpy behavior?

alejoe91:
> Oh, this is surfacing here: https://github.com/catalystneuro/neuroconv/actions/runs/13640334933/job/38128592505?pr=1117
> 
> So it seems this is either a python 3.13 or newer numpy behavior?

Exactly! I think it's numpy 2.0 behavior, since this popped up on my end with 3.12

h-mayorquin (approved):
LGTM.

PR #1240, opened 2025-03-06
Filter out warning for missing timezone in the CI

h-mayorquin:
Having many warnings on the CI makes navigating the CI log harder. For many tests, we artificially add a timezone because one isn't available, making the warning pointless in the CI context. I'd like to filter these warnings out of the CI. If there are cases where the timezone is critical, I believe we should have explicit test failures rather than warnings.

What do you think?


pauladkisson (approved):
Makes sense to me.

PR #1242, opened 2025-03-06
Added stub_test option to TDTFiberPhotometryInterface

### LINKED ISSUE #1241: [Feature]: TDTFiberPhotometryInterface needs a stub_test option

`stub_test` is a common conversion option to allow users to quickly and easily stub their data. Although TDTFiberPhotometryInterface exposes t1 and t2, which could be used for this purpose, it doesn't provide the handy `stub_test` option.

_No response_

Yes.

### Code of Conduct

pauladkisson:
Fixes #1241

### DIFF
--- a/src/neuroconv/datainterfaces/ophys/tdt_fp/tdtfiberphotometrydatainterface.py
+++ b/src/neuroconv/datainterfaces/ophys/tdt_fp/tdtfiberphotometrydatainterface.py
@@ -273,6 +273,7 @@ class TDTFiberPhotometryInterface(BaseTemporalAlignmentInterface):
         self,
         nwbfile: NWBFile,
         metadata: dict,
+        *,
         stub_test: bool = False,
         t1: float = 0.0,
         t2: float = 0.0,
--- a/src/neuroconv/datainterfaces/ophys/tdt_fp/tdtfiberphotometrydatainterface.py
+++ b/src/neuroconv/datainterfaces/ophys/tdt_fp/tdtfiberphotometrydatainterface.py
@@ -273,6 +273,7 @@ class TDTFiberPhotometryInterface(BaseTemporalAlignmentInterface):
         self,
         nwbfile: NWBFile,
         metadata: dict,
+        stub_test: bool = False,
         t1: float = 0.0,
         t2: float = 0.0,
         timing_source: Literal["original", "aligned_timestamps", "aligned_starting_time_and_rate"] = "original",
@@ -286,6 +287,8 @@ class TDTFiberPhotometryInterface(BaseTemporalAlignmentInterface):
             The in-memory object to add the data to.
         metadata : dict
             Metadata dictionary with information used to create the NWBFile.
+        stub_test : bool, optional
+            If True, only add a subset of the data (1s) to the NWBFile for testing purposes, default = False.
         t1 : float, optional
             Retrieve data starting at t1 (in seconds), default = 0 for start of recording.
         t2 : float, optional
@@ -306,6 +309,12 @@ class TDTFiberPhotometryInterface(BaseTemporalAlignmentInterface):
         )
 
         # Load Data
+        if stub_test:
+            assert (
+                t2 == 0.0
+            ), f"stub_test cannot be used with a specified t2 ({t2}). Use t2=0.0 for stub_test or set stub_test=False."
+            t2 = t1 + 1.0
+
         tdt_photometry = self.load(t1=t1, t2=t2)
 
         # timing_source is used to avoid loading the data twice if alignment is NOT used.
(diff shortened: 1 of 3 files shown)


src/neuroconv/datainterfaces/ophys/tdt_fp/tdtfiberphotometrydatainterface.py
```diff
@@ -273,6 +273,7 @@ def add_to_nwbfile(
         self,
         nwbfile: NWBFile,
         metadata: dict,
+        stub_test: bool = False,
```
h-mayorquin, line 276: Note that this is a breaking change for anyone that was relying on the order of arguments in the conversion options.

pauladkisson: Good catch, another point in favor of kw-only args I would say.

tests/test_on_data/ophys/test_fiber_photometry_interfaces.py
```diff
@@ -311,6 +313,34 @@ def test_all_conversion_checks(self, _, timing_source):
...
+
+    def test_all_conversion_checks_stub_test_invalid(self):
+        metadata_file_path = Path(__file__).parent / "fiber_photometry_metadata.yaml"
+        editable_metadata = load_dict_from_file(metadata_file_path)
+        metadata = self.data_interface_cls(**self.interface_kwargs).get_metadata()
+        metadata = dict_deep_update(metadata, editable_metadata)
+
+        self.conversion_options["stub_test"] = True
+        self.conversion_options["t2"] = 1.0
+        error_message = re.escape(
```
h-mayorquin, line 336: Out of curiosity, why do we need the re.escape here for matching the error.

pauladkisson: Honestly, not sure, but the test was failing and this was the suggested fix in the error message.

h-mayorquin: I see. Those texts for matching strings are really annoying. I am on the fence about whether they are worth the effort. Anyway, this is out of topic.

h-mayorquin (approved):

tests/test_on_data/ophys/test_fiber_photometry_interfaces.py
pauladkisson: I think it is worth it. It's important the error that gets raises is exactly the one I expect otherwise the test/code isn't working properly, and matching the message is a way to ensure uniqueness.

h-mayorquin: I mean, it is good, but there is a trade-off between effort, debuggability, and thoroughness. Some points for consideration:

* Here is a case where a good improvement is creating unnecessary work: https://github.com/catalystneuro/neuroconv/pull/1246#discussion_r1986124766. What is the test really buying us? immutability of the original message?

* Here is a case where it might be beneficial. I made changes on [these](https://github.com/catalystneuro/neuroconv/pull/1235) [PRs](https://github.com/SpikeInterface/spikeinterface/pull/3751) because some error tests are failing due to changes in how NumPy represents scalar values. It's taking additional work not just here in neuroconv but also in Spikeinterface but arguably the test is valuable because the text with the representation is not as clear as the old one: . 

* In this PR, we introduced code that neither of us fully understands, solely to pass the test. You spent more time just to add match the message exactly because of oddities of string matching. Cumulatively, I've spent several developer hours simply handling escape sequences in strings (e.g., \n, slashes, etc.) to get certain tests to work with exact matches. It seems to me that those hours would be better spend doing something else.

I feel that most of the value comes from confirming that the test catches the correct type of error and perhaps specific keywords within the error message. It is not clear to me that going for the exact match brings extra benefits worth the cost. You might be right. Maybe I am overcounting the costs and undercounting the benefits? I am overall on the fence.

pauladkisson: 1. ScanImageImagingInterface: For this case, I would say that matching the error message is even more important here than usual. In this PR we're improving an error message, and the tests clearly show that the error message has changed and now appropriately shows the available channels (or available planes).
2. Not sure where the matching strings are for those PRs
3. I mean, is `re.escape` really so opaque that it's a problem for the codebase? Sure, I don't know all of the special character combinations in regex but all I really need to know is that `re.escape` ignores them. It doesn't seem like much dev effort to me with the clear benefit of being thorough in testing.

Overall, though, I think whether we opt for exact string matching or not it won't make _much_ of a difference in our lives as developers, so our energy is better spent on more important issues.

h-mayorquin: > Overall, though, I think whether we opt for exact string matching or not it won't make much of a difference in our lives as developers, so our energy is better spent on more important issues.

Makes sense. We disagree on costs but I see no way way of settling that and there is no a large benefit to actually settle it.

PR #1244, opened 2025-03-06
Added support for changing names in VideoInterface

### LINKED ISSUE #1243: [Feature]: Add support for changing video name in VideoInterface

Currently VideoInterface automatically constructs metadata for each video with `name=f"Video {Path(file_path).stem}"`. However in a lot of cases, the names in the metadata are awkward and long.

To get around this problem, in the past I have overwritten the video metadata with custom metadata, but to my knowledge there is no way to simply specify the names of the videos as I would like them in a metadata.yaml.

To solve this problem, I propose adding a `video_names` argument in the `__init__` of VideoInterface, which would then propagate to the `get_metadata` function. 

_No response_

Yes.

### Code of Conduct

pauladkisson:
Fixes #1243 

Cline (Claude 3.7 Sonnet) re-initialized memory bank and wrote the new test for $0.37
<details>
Prompt:

```
Follow your custom instructions.

Here is a description of the task we are working on:
I just added the optional video_names argument to VideoInterface and now I need to add a test to make sure that it works. Please add a test to test_video_interface.py following the examples already present in that file.

Initialize the memory bank accordingly
```

</details>

### DIFF
--- a/tests/test_behavior/test_video_interface.py
+++ b/tests/test_behavior/test_video_interface.py
@@ -98,6 +98,32 @@ class TestVideoInterface(TestCase):
         )
         return VideoTestNWBConverter(source_data=source_data)
 
+    def test_video_names_parameter(self):
+        """Test that the video_names parameter works correctly."""
+        # Test with custom video_names
+        custom_names = ["Custom Video 1", "Custom Video 2", "Custom Video 3"]
+        interface_with_names = VideoInterface(
+            file_paths=self.video_files, video_names=custom_names, metadata_key_name="CustomVideos"
+        )
+        metadata_with_names = interface_with_names.get_metadata()
+
+        # Verify custom names are used in metadata
+        video_metadata = metadata_with_names["Behavior"]["CustomVideos"]
+        self.assertEqual(len(video_metadata), len(custom_names))
+        for i, name in enumerate(custom_names):
+            self.assertEqual(video_metadata[i]["name"], name)
+
+        # Test without video_names (default behavior)
+        interface_without_names = VideoInterface(file_paths=self.video_files, metadata_key_name="DefaultVideos")
+        metadata_without_names = interface_without_names.get_metadata()
+
+        # Verify default names are used in metadata
+        video_metadata = metadata_without_names["Behavior"]["DefaultVideos"]
+        self.assertEqual(len(video_metadata), len(self.video_files))
+        for i, file_path in enumerate(self.video_files):
+            expected_name = f"Video {Path(file_path).stem}"
+            self.assertEqual(video_metadata[i]["name"], expected_name)
+
 
 @unittest.skipIf(skip_test, "cv2 not installed")
 class TestExternalVideoInterface(TestVideoInterface):
--- a/tests/test_behavior/test_video_interface.py
+++ b/tests/test_behavior/test_video_interface.py
@@ -110,8 +110,8 @@ class TestVideoInterface(TestCase):
         # Verify custom names are used in metadata
         video_metadata = metadata_with_names["Behavior"]["CustomVideos"]
         self.assertEqual(len(video_metadata), len(custom_names))
-        for i, name in enumerate(custom_names):
-            self.assertEqual(video_metadata[i]["name"], name)
+        for meta, name in zip(video_metadata, custom_names):
+            self.assertEqual(meta["name"], name)
 
         # Test without video_names (default behavior)
         interface_without_names = VideoInterface(file_paths=self.video_files, metadata_key_name="DefaultVideos")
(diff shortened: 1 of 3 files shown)


tests/test_behavior/test_video_interface.py
```diff
@@ -98,6 +98,32 @@ class VideoTestNWBConverter(NWBConverter):
...
+        interface_with_names = VideoInterface(
+            file_paths=self.video_files, video_names=custom_names, metadata_key_name="CustomVideos"
+        )
+        metadata_with_names = interface_with_names.get_metadata()
+
+        # Verify custom names are used in metadata
+        video_metadata = metadata_with_names["Behavior"]["CustomVideos"]
+        self.assertEqual(len(video_metadata), len(custom_names))
+        for i, name in enumerate(custom_names):
+            self.assertEqual(video_metadata[i]["name"], name)
```
bendichter, line 114: Not a big deal, but I would use `zip` here

h-mayorquin:
How different is this mechanism for how we provide names for Electrical and Imaging interfaces? I need to review that I think we should do the same.

pauladkisson:
> How different is this mechanism for how we provide names for Electrical and Imaging interfaces? I need to review that I think we should do the same.

For electrical this is already covered by `es_key`, but I don't think there is a way to do it for Imaging

h-mayorquin:
We have `metadata_key_name` which is like `es_key` or am I wrong? What are we missing?

pauladkisson:
> We have `metadata_key_name` which is like `es_key` or am I wrong? What are we missing?

`metadata_key_name` only determines where in the metadata videointerface looks for the video-related metadata, but (prior to this PR) the actual name of the video that gets propagated to NWB is always determined by the file path.

`es_key` determines both where to look in the metadata _and_ the name of the ElectricalSeries as it appears in NWB.

h-mayorquin:
| es_key determines both where to look in the metadata and the name of the ElectricalSeries as it appears in NWB.

Right. 

so my point is, can/should we do the same here for the `VideoIinterface`?

The mechanisms should be similar across interfaces.

bendichter:
Maybe it makes more sense to call it `metadata_key` and have that as an arg for every interface?

pauladkisson:
> Maybe it makes more sense to call it metadata_key and have that as an arg for every interface?

Yeah, I think that is reasonable, but not sure it is worth the effort of a big refactor. Maybe just a standard we can adopt for new interfaces going forward.

> so my point is, can/should we do the same here for the VideoIinterface?

So, initially I was thinking that it would be a little different since VideoInterface currently takes a list of videos rather than just one, so each would need separate `metadata_key` or a list of `metadata_keys` or something like that. But, after looking into this a bit more, I realized that VideoInterface doesn't actually use any of the metadata entries for 2nd video, 3rd video, etc. See https://github.com/catalystneuro/neuroconv/blob/25b3e65aa926aa8607305d400d9cf8d3090f0565/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py#L356

This makes sense bc if `external_mode` is True then all of the videos are written to the same `ImageSeries` object, so there's only 1 description for all of them. And if `external_mode` is False then multiple file paths aren't supported so there has to just be one video anyway.

Ofc this is very confusing for the user especially since VideoInterface complains if the number of metadata entries doesn't match the number of file paths even though that metadata info isn't used at all!
See https://github.com/catalystneuro/neuroconv/blob/25b3e65aa926aa8607305d400d9cf8d3090f0565/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py#L341

To resolve this confusion I would propose a flatter metadata structure in which each metadata entry corresponds 1:1 with a VideoInterface and an ImageSeries ex.
```yaml
Behavior:
  Video:
  - name: Video1
    description: description of video 1
  - name: Video2
    description: description of video 2
```

Lmk what you think @h-mayorquin @bendichter

h-mayorquin:
Couple of ideas:
1) Maybe it is time to create a separate interface for ExternalVideo. I think this will simplify the coding, the metadata handling and the user interface. We have discussed this many times but never did it.
2) It would be good to take this into account as I think it is a great idea that we have not adopted that much because is breaking: https://github.com/catalystneuro/neuroconv/issues/269

pauladkisson:
> Two ideas:
> 
> 1. Maybe it is time to create a separate interface for ExternalVideo. I think this will simplify the coding, the metadata handling and the user interface. We have discussed this many times but never did it.

Love it, happy to take this one.

> 2. It would be good to take this into account as I think it is a great idea that we have not adopted it that much because is breaking: [[Feature]: change metadata lists to dictionaries #269](https://github.com/catalystneuro/neuroconv/issues/269)

I agree with this in principle, but it seems like it might be a big lift. Like, would we have to change the way [dict_deep_update](https://github.com/catalystneuro/neuroconv/blob/135a28de3ca99452195581d1fcbc817033ea1689/src/neuroconv/utils/dict.py#L109) works? So many things depend on the way metadata is currently structured with lists of dicts with name keys. It seems like it would be a big refactor, and hard to move incrementally.

h-mayorquin:
> Like, would we have to change the way [dict_deep_update](https://github.com/catalystneuro/neuroconv/blob/135a28de3ca99452195581d1fcbc817033ea1689/src/neuroconv/utils/dict.py#L109) works?

How are you thinking? This did not occur to me. I was thinking that what you need to change is the metadata schema which is breaking but maybe you are right about the update function.

pauladkisson:
> How are you thinking?

Actually, after thinking about it a bit more carefully, `dict_deep_update` would be fine, since the special "name" key only becomes relevant for lists of dicts anyway. So if we changed the lists of dicts to nested dicts, it should still update properly.

h-mayorquin:
Should we close this now that #1251 is ongoing?

pauladkisson:
Replaced by #1251

PR #1249, opened 2025-03-07
CI: test only for oldest and latest version of (supported) python 

h-mayorquin:
Hey, @bendichter. I think we are getting throttled by the CI, and I don't really see a big advantage in testing all possible versions of Python. Usually, we just want to confirm that we aren't using language constructs unavailable in older versions, and also make sure we're forward-compatible so things don't break with the latest version. Everything in between isn't as critical.

I think this would lighten our CI load and get us quicker feedback. I am actually curios on this one, maybe I am missing something. What do you think?


bendichter (approved):

bendichter:
Yes, this seems like a good idea

PR #587, opened 2023-10-05
Add support for adding `OnePhotonSeries`/`TwoPhotonSeries` to processing module

weiglszonja:
Resolve #581
Add `parent_container` argument to `add_photon_series` that defines where to add the photon series (`nwbfile.acquisition` or `nwbfile.processing["ophys"]`).

PR #1251, opened 2025-03-11
Split VideoInterface into ExternalVideoInterface and InternalVideoInterface

pauladkisson:
This PR splits VideoInterface into 2: ExternalVideointerface for external files and InternalVideoInterface for internal ImageSeries. VideoInterface acts as a router between these two options.

Closes #1244


pauladkisson:
Also, VideoInterface currently has an option for its own chunking, but chunking is now automatically applied at the neuroconv level. Maybe we should remove the option to disable chunking?

src/neuroconv/datainterfaces/behavior/video/internalvideointerface.py
```diff
@@ -0,0 +1,345 @@
...
+        verbose: bool = False,
+        *,
+        video_name: str = "Video1",
+    ):
+        """
+        Create the interface for writing videos as ImageSeries.
+
+        Parameters
+        ----------
+        file_paths : list of FilePaths
```
h-mayorquin, line 44: This docstring is wrong, it should be only file_path

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -0,0 +1,330 @@
...
+    # Other suffixes, while they can be opened by OpenCV, are not supported by DANDI so should probably not list here
+    info = "Interface for handling standard video file formats."
+
+    @validate_call
+    def __init__(
+        self,
+        file_paths: list[FilePath],
+        verbose: bool = False,
+        *,
+        video_name: str = "Video1",
```
h-mayorquin, line 31: I think this name is a bit confusing when passing the file paths of multiple videos.

I don't have a good idea now on how to improve it but I am flagging this for myself and to hear your thoughts.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -0,0 +1,330 @@
...
+from ....tools import get_package
+from ....tools.nwb_helpers import get_module
+from ....utils import dict_deep_update, get_base_schema, get_schema_from_hdmf_class
+
+
+class ExternalVideoInterface(BaseDataInterface):
+    """Data interface for writing videos as ImageSeries."""
+
+    display_name = "Video"
+    keywords = ("movie", "natural behavior", "tracking")
```
h-mayorquin, line 20: should be video instead of movie, right?

Not so sure on the other keywords as this interface can be used for many many things...

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -0,0 +1,330 @@
...
+
+from .video_utils import VideoCaptureContext
+from ....basedatainterface import BaseDataInterface
+from ....tools import get_package
+from ....tools.nwb_helpers import get_module
+from ....utils import dict_deep_update, get_base_schema, get_schema_from_hdmf_class
+
+
+class ExternalVideoInterface(BaseDataInterface):
+    """Data interface for writing videos as ImageSeries."""
```
h-mayorquin, line 17: Something something external series. 

The docstring should reference the other interface and where do they stand with respect to each other I think.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -0,0 +1,330 @@
...
+        self._number_of_files = len(file_paths)
+        self._timestamps = None
+        self._segment_starting_times = None
+        self.video_name = video_name
+        super().__init__(file_paths=file_paths)
+
+    def get_metadata_schema(self):
+        metadata_schema = super().get_metadata_schema()
+        image_series_metadata_schema = get_schema_from_hdmf_class(ImageSeries)
+        # TODO: in future PR, add 'exclude' option to get_schema_from_hdmf_class to bypass this popping
```
h-mayorquin, line 70: This is available at the moment, right?

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -0,0 +1,330 @@
...
+
+    def align_by_interpolation(self, unaligned_timestamps: np.ndarray, aligned_timestamps: np.ndarray):
+        raise NotImplementedError("The `align_by_interpolation` method has not been developed for this interface yet.")
+
+    def add_to_nwbfile(
+        self,
+        nwbfile: NWBFile,
+        metadata: Optional[dict] = None,
+        starting_frames: Optional[list[int]] = None,
+        module_name: Optional[str] = None,
```
h-mayorquin, line 249: What about the "parent container" name for this that you proposed for other pipelines? Does it fit here?

src/neuroconv/datainterfaces/behavior/video/internalvideointerface.py
```diff
@@ -0,0 +1,345 @@
...
+        metadata_key_name : str, optional
+            The key used to identify this video data within the overall experiment metadata.
+            Defaults to "Videos".
+
+            This key is essential when multiple video streams are present in a single experiment.
+            The associated metadata should be a list of dictionaries, with each dictionary
+            containing metadata for a specific video segment:
+
+            ```
+            metadata["Behavior"][metadata_key_name] = [
```
h-mayorquin, line 56: I think the docstring does not represent the new naming that you have.

So, if we follow https://github.com/catalystneuro/neuroconv/issues/269 this would change to a dict of vide metadata instead of a list, right?

src/neuroconv/datainterfaces/__init__.py
h-mayorquin, line 1: Concerning the old interface we should for route it to the new ones backwards compatibility, testing and simplification purposes  (probably with new).

Less sure about deprecating but I guess we don't need to make a decision in this PR.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin, line 1: Conversion gallery for the interfaces? what do you think?

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: Can we call it just `ExternalVideo` (without the number 1)?

pauladkisson:
> > Also, VideoInterface currently has an option for its own chunking, but chunking is now automatically applied at the neuroconv level. Maybe we should remove the option to disable chunking?
> 
> I think this is a data chunk iterator for iterative buffered writing, the parameter has an unfortunate name though, in the other interfaces it is called something like iterator_type I think.

Ok, I added a deprecation warning for this option.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: > Concerning the old interface we should for route it to the new ones backwards compatibility, testing and simplification purposes (probably with new).

I feel that these two requests are in tension with each other a bit. Either we should recommend that users use the router VideoInterface and not provide documentation about the private sub-interfaces or we document them and recommend their usage instead.

What do you think?

h-mayorquin: Yes, this is a good point, and I don’t think I was very clear.

To state my preference more clearly:

* I would like the new interfaces to become the official ones and eventually deprecate the old interface. I believe the new interfaces are clearer in terms of their purpose and usage.
* From that perspective, the new interfaces should be featured in the gallery, and we can eventually remove the old one.
* While we are deprecating the old interface, I think it’s useful to turn it into a router. This allows us to port over fixes and catch errors that can be resolved in the new interfaces, since I expect some people will continue using the old one for a while. Additionally, this approach will simplify the code of the old interface.

Do you still feel that those are in tension?

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: Only problem with `parent_container` is it constrains the options for processing module, which breaks back-compatibility with videointerface

pauladkisson:
@h-mayorquin this is ready for another round of review.

tests/test_behavior/video_metadata.yaml
h-mayorquin, line 1: Remove till we agree on having this type of example on the gallery.

tests/test_behavior/test_external_video_interface.py
```diff
@@ -0,0 +1,306 @@
...
+        self.metadata = self.nwb_converter.get_metadata()
+        self.metadata["NWBFile"].update(session_start_time=datetime.now(tz=gettz(name="US/Pacific")))
+        self.nwbfile_path = self.test_dir / "video_test.nwb"
+        self.aligned_segment_starting_times = [0.0, 50.0]
+
+    def tearDown(self) -> None:
+        shutil.rmtree(self.test_dir)
+        del self.nwb_converter
+
+    def create_videos(self):
```
h-mayorquin, line 42: This is one of the slowest part of our testing suit. 

I feel that it does not add anything more than the tests on data that we have and if that is true I would prefer to have that one only.

If it does add something that I am not aware off we should make it a fixture for the three interfaces so it is reused.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -0,0 +1,345 @@
...
+
+
+class ExternalVideoInterface(BaseDataInterface):
+    """Data interface for writing videos as external_file ImageSeries."""
+
+    display_name = "Video"
+    keywords = ("video", "behavior")
+    associated_suffixes = (".mp4", ".avi", ".wmv", ".mov", ".flx", ".mkv")
+    # Other suffixes, while they can be opened by OpenCV, are not supported by DANDI so should probably not list here
+    info = "Interface for handling standard video file formats."
```
h-mayorquin, line 23:
```suggestion
    info = "Interface for handling standard video file formats and write them as ImagingSeries with external_files."
```
or something like that.

I think this is displayed on the guide

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -0,0 +1,345 @@
...
+        For writing videos internally with just an ImageSeries object, use the InternalVideoInterface.
+
+        Parameters
+        ----------
+        file_paths : list of FilePaths
+            Many video storage formats segment a sequence of videos over the course of the experiment.
+            Pass the file paths for this videos as a list in sorted, consecutive order.
+        verbose : bool, optional
+            If True, display verbose output. Defaults to False.
+        video_name : str, optional
```
h-mayorquin, line 46: It always confuses me that optional in the typing system means that something can be None but in numpy style docstrings it means that it is not necessary to specifiy.

There is also default instead of optional which I think is more explicit. 

I know that Ben has a preference but I always forget which one is it.

https://numpydoc.readthedocs.io/en/latest/format.html#parameters

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -0,0 +1,345 @@
...
+            The name of this video as it will appear in the ImageSeries.
+            Defaults to "ExternalVideo".
+
+            This key is essential when multiple video streams are present in a single experiment.
+            The associated metadata should be a list of dictionaries, with each dictionary
+            containing metadata for a single video stream comprised of potentially multiple segments:
+
+            ```
+            metadata["Behavior"]["Video"] = [
```
h-mayorquin, line 55: Probably I am confused but shouldn't this be:

```python
metadata["Behavior"]["Video"] = {
    "ExternalVideo1": {
        "description": "description 1.",
        "unit": "Frames",
        **video1_metadata,
    },
    "ExternalVideo2": {
        "description": "description 2.",
        "unit": "Frames",
        **video2_metadata,
    },
    "ExternalVideo3": {
        "description": "description 3.",
        "unit": "Frames",
        **video3_metadata,
    },
}
```

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -0,0 +1,345 @@
...
+        file_paths = self.source_data["file_paths"]
+
+        # Be sure to copy metadata at this step to avoid mutating in-place
+        videos_metadata = deepcopy(metadata).get("Behavior", dict()).get("Video", None)
+        if videos_metadata is None:
+            videos_metadata = deepcopy(self.get_metadata()["Behavior"]["Video"])
+        image_series_kwargs = metadata["Behavior"]["Video"][self.video_name]
+        image_series_kwargs["name"] = self.video_name
+
+        timing_type = self.get_timing_type()
```
h-mayorquin, line 311: Probably not for this PR, but we should unify this behavior with the other interfaces. The default should be to check whether the timestamps are uniform and, if so, write starting_time and rate. This behavior can then be overridden with the always_write_timestamps option.

References for current implementations:

Ecephys:
https://github.com/catalystneuro/neuroconv/blob/2f9661ff4c3920b39c5eef2bce8063751390784f/src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py#L323-L335

Ophys:
https://github.com/catalystneuro/neuroconv/blob/2f9661ff4c3920b39c5eef2bce8063751390784f/src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py#L140-L150

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -0,0 +1,345 @@
...
+
+        # Attach image series
+        image_series = ImageSeries(**image_series_kwargs)
+        if parent_container == "acquisition":
+            nwbfile.add_acquisition(image_series)
+        elif parent_container == "processing/behavior":
+            get_module(nwbfile=nwbfile, name="behavior", description=module_description).add(image_series)
+        else:
+            raise ValueError(
+                f"parent_container must be either 'acquisition' or 'processing/behavior', not {parent_container}."
```
h-mayorquin, line 342: As per your comment in other PRs validation logic should be at the start of the function, what do you think?

src/neuroconv/datainterfaces/behavior/video/internalvideointerface.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def align_by_interpolation(self, unaligned_timestamps: np.ndarray, aligned_timestamps: np.ndarray):
+        raise NotImplementedError("The `align_by_interpolation` method has not been developed for this interface yet.")
+
+    def add_to_nwbfile(
+        self,
+        nwbfile: NWBFile,
+        metadata: Optional[dict] = None,
+        stub_test: bool = False,
+        chunk_data: bool = True,
```
h-mayorquin, line 217: should we choose a better name for this one so buffered writing is not confused with chunking?

src/neuroconv/datainterfaces/behavior/video/internalvideointerface.py
```diff
@@ -0,0 +1,361 @@
...
+            frames, the equivalent uncompressed RAM usage is around 2GB per minute of video. The default is True.
+        parent_container: {'acquisition', 'processing/behavior'}
+            The container where the ImageSeries is added, default is nwbfile.acquisition.
+            When 'processing/behavior' is chosen, the ImageSeries is added to nwbfile.processing['behavior'].
+        module_description: str, optional
+            If parent_container is 'processing/behavior', and the module does not exist,
+            it will be created with this description. The default description is the same as used by the
+            conversion_tools.get_module function.
+        """
+        if not chunk_data:
```
h-mayorquin, line 268: Ah, I see, yeah, I think it should be like the other interfaces. There is a way to not use iterators.

h-mayorquin (commented):
OK, I did another reading today. This is looking good.

docs/conversion_examples_gallery/behavior/video.rst
```diff
@@ -7,10 +7,98 @@ Install NeuroConv with the additional dependencies necessary for reading multime
 
     pip install "neuroconv[video]"
 
-This interface can handle conversions from avi, mov, mp4, wmv, flv and most FFmpeg_ supported formats to NWB using the
-:py:class:`~neuroconv.datainterfaces.behavior.video.videodatainterface.VideoInterface` class.
+This interface can handle conversions from avi, mov, mp4, wmv, flv and most `FFmpeg <https://ffmpeg.org/>_` supported formats to NWB.
 
-.. _FFmpeg: https://ffmpeg.org/
+If linking to a list of external files (recommended), use the
```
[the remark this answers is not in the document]
h-mayorquin: https://nwbinspector.readthedocs.io/en/dev/best_practices/image_series.html#storage-of-imageseries

tests/test_behavior/test_mixed_video_interfaces.py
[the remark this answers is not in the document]
h-mayorquin: Ah, the shared metadata mmm.

tests/test_behavior/test_external_video_interface.py
h-mayorquin: To be more clear, I claim that everything on this tests is or should be covered here:
https://github.com/catalystneuro/neuroconv/blob/2f9661ff4c3920b39c5eef2bce8063751390784f/tests/test_on_data/behavior/test_behavior_interfaces.py#L259-L272

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: Can you elaborate?

docs/conversion_examples_gallery/behavior/video.rst
```diff
@@ -7,10 +7,98 @@ Install NeuroConv with the additional dependencies necessary for reading multime
...
+:py:class:`~neuroconv.datainterfaces.behavior.video.externalvideodatainterface.ExternalVideoInterface` class.
+
+Specify the metadata (optional)
```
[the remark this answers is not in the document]
h-mayorquin: Yes, please separate, we can discuss that as a separate issue.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: Previously, a user could specify and module_name (ex. 'my_arbitrarily_named_module'), but with parent_container we only have 'acquisition' or 'processing/behavior'.

tests/test_behavior/test_mixed_video_interfaces.py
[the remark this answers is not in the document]
pauladkisson: > Ah, the shared metadata mmm.

Yep

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: Is that a limitation of the other streams?
If so, I would prefer to unify and then expand. 

But if you want to keep it as it is currently I am fine with delaying this and address this separately.

tests/test_behavior/test_external_video_interface.py
```diff
@@ -0,0 +1,306 @@
...
+
+try:
+    import cv2
+
+    skip_test = False
+except ImportError:
+    skip_test = True
+
+
+@unittest.skipIf(skip_test, "cv2 not installed")
```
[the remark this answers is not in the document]
pauladkisson: I'll add a note in the .clinerules to that effect

tests/test_behavior/test_external_video_interface.py
```diff
@@ -0,0 +1,306 @@
...
+    def setUp(self) -> None:
+        self.test_dir = Path(tempfile.mkdtemp())
+        self.video_files = self.create_videos()
+        self.nwb_converter = self.create_video_converter()
+        self.metadata = self.nwb_converter.get_metadata()
+        self.metadata["NWBFile"].update(session_start_time=datetime.now(tz=gettz(name="US/Pacific")))
+        self.nwbfile_path = self.test_dir / "video_test.nwb"
+        self.aligned_segment_starting_times = [0.0, 50.0]
+
+    def tearDown(self) -> None:
```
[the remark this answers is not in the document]
h-mayorquin: That's great!
Did it generate that diagram by itself? : O

tests/test_behavior/test_mixed_video_interfaces.py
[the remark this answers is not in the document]
pauladkisson: I think the only downside would be the metadata dictionary is slightly more complex for users, but that shouldn't really be much of an issue.

tests/test_behavior/test_external_video_interface.py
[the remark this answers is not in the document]
pauladkisson: Yeah, in Plan mode, I usually ask it to express its plan as a mermaid diagram. I find it's much quicker for me to check the logic and I think it might help with the eventual output too (based on this [blog post](https://cline.bot/blog/memory-bank-how-to-make-cline-an-ai-agent-that-never-forgets))

> Here's why this matters: Mermaid diagrams aren't just prettier than text – they're a formal language for describing workflows. Each node, connection, and decision point is explicitly defined. For an AI, this structured format might actually be easier to understand than paragraphs of natural language trying to describe the same process.
>
> When we write text instructions, there's room for ambiguity. But a flowchart in Mermaid code is unambiguous – each state and transition is clearly defined. It's like giving the AI a precise map instead of vague directions.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: > Is that a limitation of the other streams as well (ephys and ophys)?

It is in #587 at least.

> If so, I would prefer to unify and then expand.
> But if you want to keep it as it is currently I am fine with delaying this and address this separately.

I would prefer to keep it limited (we don't really recommend making arbitrary processing modules anyway), but it depends on how important true back-compatibility is for this PR.

tests/test_behavior/test_mixed_video_interfaces.py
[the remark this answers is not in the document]
pauladkisson: Updated!

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: Done!

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: Yeah, makes sense to me, but I think beyond the scope of this PR.

tests/test_behavior/test_external_video_interface.py
pauladkisson: I'm confused, why would you ask to refactor the existing tests to pytest style if you think they should just be completely removed in favor of these test_on_data tests?

pauladkisson: And more to the point, only test_video_interface.py tests error handling, irregular timestamps, and multiple video files in the same interface, so it's not really redundant.

pauladkisson: Plus, this is stretching the scope of this PR. Updating the testing structure to focus more on test_on_data should be it's own PR.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: This should be resolved now.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: I feel like optional is fine, and it's everywhere in our codebase.

pauladkisson: > There is also default instead of optional which I think is more explicit.

This is the kind of change that I think has such low marginal value it's really not worth the time and effort discussing/implementing it.

tests/test_behavior/test_external_video_interface.py
h-mayorquin: > I feel that it does not add anything more than the tests on data that we have and **if that is true** I would prefer to have that one only.

> **If it does add something that I am not aware off** we should make it a fixture for the three interfaces so it is reused.

It seems that the tests that create data on the fly do cover some aspects that are not covered on the tests with data from gin. If that is the case, then I am fine for adding new tests. 

But given that we are adding new tests, do it with a modular fixture so the computation is not repeated.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: Yes,I was just checking with you if there would not be major roadblocks as you are reading this on detail.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: Agree, it is a preference of Ben though so was just letting you know.

tests/test_behavior/test_video_interface.py
```diff
@@ -1,290 +1,314 @@
...
+    return pytest.importorskip("cv2")
+
+
+@pytest.fixture
+def test_dir(tmp_path):
+    """Create and return a test directory using pytest's tmp_path fixture."""
+    return tmp_path
+
+
+@pytest.fixture
```
h-mayorquin, line 25: What matter is that this is well scoped so the computation is not repeated:

https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session

tests/test_behavior/test_internal_video_interface.py
```diff
@@ -0,0 +1,293 @@
...
+
+from neuroconv import NWBConverter
+from neuroconv.datainterfaces.behavior.video.internalvideointerface import (
+    InternalVideoInterface,
+)
+from neuroconv.utils import dict_deep_update
+
+
+@pytest.fixture
+def cv2():
```
h-mayorquin, line 17: This is not how this is used:
See the documentation that I shared and other usees:

https://github.com/search?q=org%3Acatalystneuro%20%20importorskip&type=code

tests/test_behavior/test_internal_video_interface.py
```diff
@@ -0,0 +1,293 @@
...
+    """Import and return cv2 or skip the test if not installed."""
+    return pytest.importorskip("cv2")
+
+
+@pytest.fixture
+def test_dir(tmp_path):
```
h-mayorquin, line 23: this is also unnecessary:

https://docs.pytest.org/en/stable/how-to/tmp_path.html

docs/conversion_examples_gallery/behavior/video.rst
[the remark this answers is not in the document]
h-mayorquin: Thanks.

tests/test_behavior/test_external_video_interface.py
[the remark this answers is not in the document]
h-mayorquin: OK, I will count his as a failure of cline. The  class structure is fine, see the testa at `src/neuroconv/tools/testing/data_interface_mixins.py`.

I just requested to use the tmp_path fixture from pytest to handle file removal (to avoid windows errors) and also module or session scope fixtures for the video creation to avoid repeated computation on the new tests.

docs/conversion_examples_gallery/behavior/video.rst
```diff
@@ -7,10 +7,71 @@ Install NeuroConv with the additional dependencies necessary for reading multime
...
+    >>> from datetime import datetime
+    >>> from zoneinfo import ZoneInfo
+    >>> from pathlib import Path
+
+    >>> from neuroconv.datainterfaces import InternalVideoInterface
+    >>> from neuroconv.utils import dict_deep_update, load_dict_from_file
+
+    >>> LOCAL_PATH = Path(".") # Path to neuroconv
+    >>> video_file_path = BEHAVIOR_DATA_PATH / "videos" / "CFR" / "video_avi.avi"
+    >>> editable_metadata_path = LOCAL_PATH / "tests" / "test_behavior" / "video_metadata.yaml"
```
h-mayorquin, line 60: I guess this will fail.

tests/test_behavior/test_external_video_interface.py
pauladkisson: Ok, it is a reused fixture now.  But more importantly, I reduced the video size from (30, 640, 480) to (30, 64, 48), which speeds things up dramatically.

tests/test_behavior/test_external_video_interface.py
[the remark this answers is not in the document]
pauladkisson: I don't think this is a failure per se. Even if the existing class structure is compatible with pytest, the function-oriented style is the more standard way to write tests with pytest, and makes e.g. parameterization easier to manage.

But, as a general observation, I have found it difficult to prevent Cline from going "above and beyond" the original request.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: Yep, thanks for the catch.

tests/test_behavior/test_external_video_interface.py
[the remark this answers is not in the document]
h-mayorquin: I am talking about the failure to use fixtures on the intended way which is a central feature of pytest. 

The changing to fully functional came because your prompt is too open ended.

tests/test_behavior/test_external_video_interface.py
h-mayorquin: Even better. Thanks.

src/neuroconv/datainterfaces/behavior/video/internalvideointerface.py
pauladkisson: Done!

src/neuroconv/datainterfaces/behavior/video/internalvideointerface.py
pauladkisson: Ok, I updated accordingly.

tests/test_behavior/test_internal_video_interface.py
pauladkisson: My bad, should be fixed now.

tests/test_behavior/test_internal_video_interface.py
pauladkisson: noticed that one earlier, should be fixed already.

pauladkisson:
@h-mayorquin, this should be ready for another round.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: I agree with your reasoning and we are on the same page here. We can generalize later if we need to but we a global solution instead.

src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
```diff
@@ -341,111 +346,72 @@ def add_to_nwbfile(
...
+                starting_frames=starting_frames,
+                parent_container=parent_container,
+                module_description=module_description,
+            )
+
+        # Use InternalVideoInterface for external_mode=False
+        # Validate that we only have one file (required by InternalVideoInterface)
+        if self._number_of_files > 1:
+            raise NotImplementedError(
+                "Multiple file_paths with external_mode=False is not yet supported! "
```
h-mayorquin, line 392:
```suggestion
                "Multiple file_paths with external_mode=False is not supported! "
```

h-mayorquin (changes requested):
I see no big upside for breaking backwards compatibility with the `chunk_data` to `buffer_data` on the old `videointerface` 

Just that and this should be ready to go.

src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
```diff
@@ -266,7 +269,7 @@ def add_to_nwbfile(
         stub_test: bool = False,
         external_mode: bool = True,
         starting_frames: Optional[list[int]] = None,
-        chunk_data: bool = True,
+        buffer_data: bool = True,
```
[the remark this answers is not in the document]
h-mayorquin: The other interfaces are new so we can keep the `buffer_data` that you have chosen.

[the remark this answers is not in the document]
pauladkisson: Done!

h-mayorquin (approved):
Thanks for pushing this forward.

PR #1252, opened 2025-03-11
Add concurrency at the job level for live service tests

h-mayorquin:
They have been failing for a while. We believe this is a consequence of concurrent write access to the DANDI database. If this is correct, this should eliminate the problem.

That said, this solution may be too strong. A more precise alternative would be to have DANDI files that are OS- and Python-version-dependent, preventing overwriting of the same object. However, this approach is easier to implement, and the live service tests are not that long, so perhaps we should opt for the lower-effort solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 * Change `np.NAN` to `np.nan` to support numpy 2.0 [PR #1245](https://github.com/catalystneuro/neuroconv/pull/1245)
 * Testing: only run tests for oldest and newest versions of python [#1249](https://github.com/catalystneuro/neuroconv/pull/1249)
 * Improve error display on scan image interfaces [PR #1246](https://github.com/catalystneuro/neuroconv/pull/1246)
-* Added concurrency to live-service-testing GitHub Actions workflow to prevent simultaneous write to the dandiset.
+* Added concurrency to live-service-testing GitHub Actions workflow to prevent simultaneous write to the dandiset. [#1252](https://github.com/catalystneuro/neuroconv/pull/1252)g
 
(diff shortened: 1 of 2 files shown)


h-mayorquin:
Ah, is more or less the time that it takes the full tests. I think it is fine to test the flakiness of the live service tests with this for a while. What do you think, @pauladkisson ?

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 * Change `np.NAN` to `np.nan` to support numpy 2.0 [PR #1245](https://github.com/catalystneuro/neuroconv/pull/1245)
 * Testing: only run tests for oldest and newest versions of python [#1249](https://github.com/catalystneuro/neuroconv/pull/1249)
 * Improve error display on scan image interfaces [PR #1246](https://github.com/catalystneuro/neuroconv/pull/1246)
+* Added concurrency to live-service-testing GitHub Actions workflow to prevent simultaneous write to the dandiset. [#1252](https://github.com/catalystneuro/neuroconv/pull/1252)g
```
pauladkisson, line 19:
```suggestion
* Added concurrency to live-service-testing GitHub Actions workflow to prevent simultaneous write to the dandiset. [#1252](https://github.com/catalystneuro/neuroconv/pull/1252)
```

pauladkisson:
> Ah, is more or less the time that it takes the full tests. I think it is fine to test the flakiness of the live service tests with this for a while. What do you think, @pauladkisson ?

Yeah, looks good. Just caught a typo in the changelog.

h-mayorquin:
Thanks.

pauladkisson (approved):

PR #1253, opened 2025-03-13
Propagate verbose from converter to interfaces

### LINKED ISSUE #1178: [Enhancement] Propagate verbose from Converters to Interfaces
As suggested by @pauladkisson here:

https://github.com/catalystneuro/neuroconv/pull/1153#pullrequestreview-2516337643

I think this is a good idea. We should do it after #1153 I think.

h-mayorquin:
Fixes #1178

### DIFF
--- a/src/neuroconv/nwbconverter.py
+++ b/src/neuroconv/nwbconverter.py
@@ -4,7 +4,7 @@ import inspect
 import json
 from collections import Counter
 from pathlib import Path
-from typing import Literal, Optional, Union
+from typing import Literal, Optional, Type, Union
 
 from jsonschema import validate
 from pydantic import FilePath, validate_call
@@ -45,7 +45,7 @@ class NWBConverter:
     associated_suffixes: tuple[str] = tuple()
     info: Union[str, None] = None
 
-    data_interface_classes = None
+    data_interface_classes: dict[str, Type[BaseDataInterface]] = {}
 
     @classmethod
     def get_source_schema(cls) -> dict:
--- a/src/neuroconv/nwbconverter.py
+++ b/src/neuroconv/nwbconverter.py
@@ -1,5 +1,6 @@
 """Contains core class definitions for the NWBConverter and ConverterPipe."""
 
+import inspect
 import json
 from collections import Counter
 from pathlib import Path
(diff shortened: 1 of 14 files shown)


src/neuroconv/nwbconverter.py
```diff
@@ -87,11 +88,19 @@ def _validate_source_data(self, source_data: dict[str, dict], verbose: bool = Fa
...
         self.verbose = verbose
-        self.data_interface_objects = {
-            name: data_interface(**source_data[name])
-            for name, data_interface in self.data_interface_classes.items()
-            if name in source_data
-        }
+        self.data_interface_objects = dict()
+        for interface_name, data_interface in self.data_interface_classes.items():
+            # If the name of an interface is not passed in the source_data, skip it
+            # TODO: shouldn't this throw an error? keeping the current behavior now
```
pauladkisson, line 94: I think this is intended, or at least I make use of this functionality fairly often. For some conversions, there are sessions that only use a subset of the interfaces. Rather than having to define multiple NWBConverters for each subset, I can just define one NWBConverter class and modify the source_data dictionary to only include the interfaces that I want.

pauladkisson (commented):
Looks good, but I think we should remove the TODO.

src/neuroconv/nwbconverter.py
h-mayorquin: I see. That makes sense to me.

h-mayorquin: I removed the TODO and added an explanation on how this is used.

pauladkisson (approved):
I'm a little confused why the live-service-tests got cancelled when `cancel-in-progress` is false, but that's not really a problem for this PR.

h-mayorquin:
> I'm a little confused why the live-service-tests got cancelled when `cancel-in-progress` is false, but that's not really a problem for this PR.

I noticed that. I did not set something right when I modified them.

pauladkisson:
> > I'm a little confused why the live-service-tests got cancelled when `cancel-in-progress` is false, but that's not really a problem for this PR.
> 
> I noticed that. I did not set something right when I modified them.

I think it might be somehow interacting with the `Deploy Tests` concurrency that gets triggered when the PR updates. It was working properly when I re-ran them until the latest merge commit.

PR #1258, opened 2025-03-14
Add cline rule about inlining imports.

h-mayorquin:
I think adding this is also a good idea. @luiztauffer @bendichter


.clinerules
```diff
@@ -1,6 +1,7 @@
 * When developing a new feature or fixing a bug, always write corresponding tests and run the tests to ensure the feature worked
 * When adding tests, first read through the existing tests and try to fit new tests in the existing testing infrastructure
 * If a new feature requires a new dependency, make sure to add it in the pyproject.toml
+* Avoid making imports at the top of the file for dependencies that are not pyproject dependencies. Inline the imports for extra dependencies.
```
bendichter, line 4:
```suggestion
* To enable dependencies that are specific to certain Interface or Converter classes, import these additional dependencies within the class using `tools.get_package` instead of importing them at the top of the file.
```

PR #1264, opened 2025-03-19
Fix live service tests

pauladkisson:
Removing concurrency check for live service tests so that they don't interfere in multiple commits on the same PR.

### DIFF
--- a/.github/workflows/live-service-testing.yml
+++ b/.github/workflows/live-service-testing.yml
@@ -25,11 +25,6 @@ on:
 env:
   DANDI_API_KEY: ${{ secrets.DANDI_API_KEY }}
 
-# We are adding concurrency here to avoid concurrent writes to the dandi API.
-concurrency:
-  group: live-service-testing
-  cancel-in-progress: false
-
 jobs:
   run:
     name: ${{ matrix.os }} Python ${{ matrix.python-version }}


.github/workflows/live-service-testing.yml
```diff
@@ -25,11 +25,6 @@ on:
 env:
   DANDI_API_KEY: ${{ secrets.DANDI_API_KEY }}
 
-# We are adding concurrency here to avoid concurrent writes to the dandi API.
```
h-mayorquin, line 28: I thought that we needed to cancel this within the same PR:

```
concurrency:  # Cancel previous workflows on the same pull request
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

pauladkisson: No, bc it should be handled by the deploy-tests's concurrency settings

h-mayorquin: See, that's my confusing. I thought that that was for each of the jobs on the matrix and that this top level thing would contro inter-PR.

pauladkisson: I think this section prevents jobs within the matrix from running at the same time (macos-13 python 3.9, windows  python 3.12, etc.), which is doing most of the work to prevent a race condition.
```yaml
strategy:
      # This avoids two tests within the matrix running at the same time
      max-parallel: 1
```

pauladkisson: The concurrency was trying to prevent different PRs from racing, but it was interfering with the concurrency defined at the deploy-tests.yml level.

pauladkisson: That's my theory at any rate.

h-mayorquin (commented):
Question: if you updated the same PR. Does it cancel the previous tests?

pauladkisson:
> Question: if you updated the same PR. Does it cancel the previous tests?

Yes, you can see this behavior from the rapid commits.

h-mayorquin (approved):
LGTM

PR #1270, opened 2025-03-26
Fix use of `nwbfile_path`  in documentation of `configure_and_write_nwbfile`

h-mayorquin:
In the documentation sometimes `save_path` is used and other times the correct `output_filepath`. I am fixing this and also changing the argument to `nwbfile_path` to unify this with other functions in the library.


pauladkisson (approved):
LGTM!

PR #1256, opened 2025-03-14
Add append mode to `run_conversion`

h-mayorquin:
Should fix #1213.

PR #1272, opened 2025-03-27
Fix conversion gallery docstring video interface

h-mayorquin:
@pauladkisson 
Small fixes:
1) The internal video interface, as per the best practices suggestion, should not contain beahvior.
2) The nwbfile_path was not included in the conversion gallery
3) Because the temporal path is only unique per rst doc and until we add something like https://github.com/catalystneuro/neuroconv/pull/1256 I would like to keep the writing safe so I added `overwrite=True` on the conversion gallery.


pauladkisson:
looks good, while we're at it, let's add back the metadata.yaml in this PR as well.

h-mayorquin:
I will leave that you guys.

pyproject.toml
```diff
@@ -42,7 +42,7 @@ dependencies = [
     "hdmf>=3.13.0",
     "hdmf_zarr>=0.7.0",
     "pynwb>=2.7.0",
-    "pydantic>=2.0.0",
+    "pydantic>=2.0.0,<2.11",  # Pydantic 2.11 changed how the schema works for generic dict types
```
h-mayorquin, line 45:
```suggestion
    "pydantic>=2.0.0",
```

PR #1278, opened 2025-04-01
Change video interface metadata access path

h-mayorquin:
Hey, @pauladkisson 

Before the release, what do you think of making the dict of the external and internal videos plural to indicate there might be more than one.

This is aligned with the current use in the old behavior interface:

https://github.com/catalystneuro/neuroconv/blob/002bcae7e3e84580e6779f8033e222963be736a1/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py#L295-L303

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
+++ b/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
@@ -352,9 +352,9 @@ class VideoInterface(BaseDataInterface):
 
         # Use appropriate metadata key based on external_mode
         if external_mode:
-            metadata_reformatted["Behavior"]["ExternalVideo"] = {video_name: videos_metadata[0]}
+            metadata_reformatted["Behavior"]["ExternalVideos"] = {video_name: videos_metadata[0]}
         else:
-            metadata_reformatted["Behavior"]["InternalVideo"] = {video_name: videos_metadata[0]}
+            metadata_reformatted["Behavior"]["InternalVideos"] = {video_name: videos_metadata[0]}
 
         parent_container = "acquisition" if module_name is None else "processing/behavior"
         if parent_container == "processing/behavior":


pauladkisson:
> Before the release, what do you think of making the dict of the external and internal videos plural to indicate there might be more than one.

Yeah, sounds like a good idea. I'd like to get consensus on this idea since existing interfaces use singular (ex. all the RecordingInterfaces use 'Device' and 'ElectrodeGroup' even though there could be many devices and ElectrodeGroups).

pauladkisson:
See #1280

h-mayorquin:
> Yeah, sounds like a good idea. I'd like to get consensus on this idea since existing interfaces use singular (ex. all the RecordingInterfaces use 'Device' and 'ElectrodeGroup' even though there could be many devices and ElectrodeGroups).

I think it might be difficult to get a big push to change this generally and I’m planning to make a release this week. Once it’s released, I’d prefer not to change it unless there’s a lot of push for a breaking change in your newly created issue (note it will break yaml file reading).

My argument is that this follows the precedent set by the previous `VideoInterface`, where the metadata entry was `Videos`. That said, since you did the work on this interface, I think the decision should be yours.

pauladkisson:
> I think it might be difficult to get a big push to change this generally and I’m planning to make a release this week. Once it’s released, I’d prefer not to change it

Yeah, this would only apply to new interfaces (like ExternalVideo), so it shouldn't be much marginal work.

h-mayorquin:
> Yeah, this would only apply to new interfaces (like ExternalVideo), so it shouldn't be much marginal work.

What I mean is that once it's released, any changes would be breaking, so the barrier to making them will be higher. It’s not about the amount of work involved.

pauladkisson:
> What I mean is that once it's released, any changes would be breaking, so the barrier to making them will be higher. It’s not about the amount of work involved.

Sure, we can get this done before the release.

Actually though, each interface only has 1 video, so one would only get multiple entries under `ExternalVideo` from multiple `ExternalVideoInterface`s. Do you still think we should use plural?

h-mayorquin:
Even with the internal interface, users might add more than one video (although they’ll need two interfaces to do so). As a result, the metadata entry could end up containing metadata from multiple videos.

pauladkisson:
> Even with the internal interface, users might add more than one video (although they’ll need two interfaces to do so). As a result, the metadata entry could end up containing metadata from multiple videos.

Yeah, ok. I just wanted to make sure I was understanding under what conditions these metadata categories should be plural. Seems like anytime it's reasonable to have multiple entries _in the metadata_, they should be plural to convey that.

PR #1279, opened 2025-04-01
Use `always_write_timestmaps` on the new video interfaces

h-mayorquin:
@pauladkisson as discussed during https://github.com/catalystneuro/neuroconv/pull/1251

This will unify video with ecehpys and ophys.

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
+++ b/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
@@ -173,7 +173,7 @@ class ExternalVideoInterface(BaseDataInterface):
         """
         Set the aligned starting time for the ImageSeries in this interface.
 
-        This method will not do anything if the timestamps have already been set.
+        If the timestamps have already been set, each segment will be shifted by aligned_starting_time.
 
         Must be in units seconds relative to the common 'session_start_time'.
 
@@ -312,7 +312,7 @@ class ExternalVideoInterface(BaseDataInterface):
             image_series_kwargs["device"] = device
 
         if always_write_timestamps:
-            timestamps = self._timestamps if self._timestamps is not None else self.get_timestamps()
+            timestamps = self.get_timestamps()
             image_series_kwargs.update(timestamps=np.concatenate(timestamps))
         elif self._timestamps is not None:
             # Check if timestamps are regular
--- a/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
+++ b/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
@@ -10,7 +10,12 @@ from .video_utils import VideoCaptureContext
 from ....basedatainterface import BaseDataInterface
 from ....tools import get_package
 from ....tools.nwb_helpers import get_module
-from ....utils import dict_deep_update, get_base_schema, get_schema_from_hdmf_class
+from ....utils import (
+    calculate_regular_series_rate,
+    dict_deep_update,
+    get_base_schema,
+    get_schema_from_hdmf_class,
+)
 
 
 class ExternalVideoInterface(BaseDataInterface):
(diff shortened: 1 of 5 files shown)


src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -122,37 +127,6 @@ def get_original_timestamps(self, stub_test: bool = False) -> list[np.ndarray]:
                 timestamps.append(video.get_video_timestamps(max_frames=max_frames))
         return timestamps
 
-    def get_timing_type(self) -> Literal["starting_time and rate", "timestamps"]:
```
pauladkisson, line 125: This should also be removed for InternalVideoInterface right?

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -314,7 +294,29 @@ def add_to_nwbfile(
...
+                image_series_kwargs.update(starting_time=starting_time, rate=rate)
+            else:
+                image_series_kwargs.update(timestamps=timestamps)
+        else:
+            if self._number_of_files > 1 and self._segment_starting_times is None:
+                raise ValueError(
+                    f"No timing information is specified and there are {self._number_of_files} total video files! "
+                    "Please specify the temporal alignment of each video."
+                )
+            starting_time = self._segment_starting_times[0] if self._segment_starting_times is not None else 0.0
```
pauladkisson, line 315: I realize this is from the old implementation, but what's the point of setting a list of `self._segment_starting_times` if only the first one ever gets used...?

pauladkisson: Specifically, I think it would be confusing for a user to specify a list of segment starting times, and then check the ImageSeries only to find it has just a single `starting_time` and `rate`.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: Yes.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: I find the current timing functions a bit confusing and would like to improve them but I think we can't get to the situation you propose because of this assert.

https://github.com/catalystneuro/neuroconv/blob/dc3b4a63d29fd5a8095c40754b3d2c94428b5e45/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py#L209-L213

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: Should be removed.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: I have an idea: What if instead of raising a ValueError in the else here: https://github.com/catalystneuro/neuroconv/blob/dc3b4a63d29fd5a8095c40754b3d2c94428b5e45/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py#L163-L190

We set 
```python
self._segment_starting_times[0] = aligned_starting_time
```

And then we raise a ValueError in the else here: https://github.com/catalystneuro/neuroconv/blob/dc3b4a63d29fd5a8095c40754b3d2c94428b5e45/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py#L192-L224

pauladkisson: Then a user would only use `set_aligned_segment_starting_times` if they are aligning all the timestamps, and if they are using a starting_time and rate representation, they would use `set_aligned_starting_time`.

pauladkisson: > Hahaha maybe we should separate the ExternalFile into two lol

Yeah, that's not a bad idea, but I don't think we _really_ need to.

pauladkisson: > Would you like to create a PR with this as a base for that improvement?

Sure

h-mayorquin: Ok, thanks a lot for that, @pauladkisson 

Anything else you think we should do or modify for this PR?

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -193,46 +167,36 @@ def set_aligned_timestamps(self, aligned_timestamps: list[np.ndarray]):
...
-        ), "If setting both timestamps and starting times, please set the timestamps first so they can be shifted by the starting times."
         self._timestamps = aligned_timestamps
 
-    def set_aligned_starting_time(self, aligned_starting_time: float, stub_test: bool = False):
+    def set_aligned_starting_time(self, aligned_starting_time: float):
         """
-        Align all starting times for all videos in this interface relative to the common session start time.
+        Set the aligned starting time for the ImageSeries in this interface.
+
+        This method will not do anything if the timestamps have already been set.
```
pauladkisson, line 176:
```suggestion
        If the timestamps have already been set, each segment will be shifted by aligned_starting_time.
```

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -343,7 +311,29 @@ def add_to_nwbfile(
                 nwbfile.add_device(device)
             image_series_kwargs["device"] = device
 
-        timing_type = self.get_timing_type()
+        if always_write_timestamps:
+            timestamps = self._timestamps if self._timestamps is not None else self.get_timestamps()
```
pauladkisson, line 315:
```suggestion
            timestamps = self.get_timestamps()
```

self.get_timestamps already checks if self._timestamps is None, so this should be the same but simpler.

src/neuroconv/datainterfaces/behavior/video/internalvideointerface.py
```diff
@@ -364,13 +344,24 @@ def add_to_nwbfile(
 
         image_series_kwargs.update(data=iterable)
 
-        if timing_type == "starting_time and rate":
+        from ....utils import calculate_regular_series_rate
+
+        if always_write_timestamps:
+            timestamps = self._timestamps if self._timestamps is not None else self.get_timestamps()
```
pauladkisson, line 350:
```suggestion
            timestamps = self.get_timestamps()
```

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: Yes, thanks.

pauladkisson (approved):
Looks good!

PR #1281, opened 2025-04-01
Update .clinerules

pauladkisson:
This PR removes a line in clinerules that encourages Cline to update the Changelog. Since there is usually only 1 entry in the changelog / PR, and many sub-tasks for cline / PR, this request doesn't really make sense, and is annoying to correct.

### DIFF
--- a/.clinerules
+++ b/.clinerules
@@ -3,7 +3,6 @@
 * When working with tests, use pytest if possible instead of unittest or hdmf.testing.TestCase
 * If a new feature requires a new dependency, make sure to add it in the pyproject.toml
 * To enable dependencies that are specific to certain Interface or Converter classes, import these additional dependencies within the class using `tools.get_package` instead of importing them at the top of the file.
-* Always make a contribution to the CHANGELOG file summarizing any changes
 * Use typehints and numpy-style docstrings
 * always apply Black formatting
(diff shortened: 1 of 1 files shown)


pauladkisson:
Alternatively, maybe this kind of thing should just be in the .gitignore so that everyone can develop however they want?

luiztauffer (approved):
yeah, I can see cline overdoing it on changelog, I agree with the change.

alessandratrapani (approved):
Maybe instead of "...summarizing any changes" --> "...summarizing the main changes" or similar?

pauladkisson:
> Agree! Maybe instead of "...summarizing any changes" --> "...summarizing the main changes" or similar?

The issue is I really don't want/need Cline messing with the changelog unless I specifically ask it to. Summarizing the steps that it took should be in a separate document that we likely don't want to actually include in the repo ex. progress.md.

h-mayorquin (changes requested):
Let's wait for Ben on this one.

h-mayorquin:
Hey, @bendichter, I agree with these changes but you are a user of this, what do you think?

bendichter:
I can see how that line would get annoying. I'm fine with removing it

h-mayorquin (approved):

PR #1276, opened 2025-03-27
Added metadata section to video conversion gallery

pauladkisson:
See #1269

PR #1282, opened 2025-04-01
Added device to video interface

### LINKED ISSUE #1277: Video Interface should link to a device
pynwb.image.ImageSeries has an option for `device` which links the `ImageSeries` to the pynwb.device.Device that was used to capture the images/video.

As it currently stands, none of the video interfaces support this linkage.

VideoInterface should optionally create a device for a camera and link it to the ImageSeries.

pauladkisson:
Fixes #1277

Will update conversion gallery in #1276

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
+++ b/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
@@ -4,6 +4,7 @@ from typing import Literal, Optional
 import numpy as np
 from pydantic import FilePath, validate_call
 from pynwb import NWBFile
+from pynwb.device import Device
 from pynwb.image import ImageSeries
 
 from .video_utils import VideoCaptureContext
@@ -29,6 +30,7 @@ class ExternalVideoInterface(BaseDataInterface):
         verbose: bool = False,
         *,
         video_name: str = "ExternalVideo",
+        device_name: str = "ExternalVideoCamera",
     ):
         """
         Initialize the interface.
@@ -68,6 +70,7 @@ class ExternalVideoInterface(BaseDataInterface):
         self._timestamps = None
         self._segment_starting_times = None
         self.video_name = video_name
+        self.device_name = device_name
         super().__init__(file_paths=file_paths)
 
     def get_metadata_schema(self):
@@ -79,6 +82,8 @@ class ExternalVideoInterface(BaseDataInterface):
             image_series_metadata_schema["properties"].pop(key)
             if key in image_series_metadata_schema["required"]:
                 image_series_metadata_schema["required"].remove(key)
+        device_metadata_schema = get_schema_from_hdmf_class(Device)
+        device_metadata_schema["required"].remove("name")
         metadata_schema["properties"]["Behavior"] = get_base_schema(tag="Behavior")
         metadata_schema["properties"]["Behavior"]["required"].append("ExternalVideo")
         metadata_schema["properties"]["Behavior"]["properties"]["ExternalVideo"] = {
@@ -87,13 +92,21 @@ class ExternalVideoInterface(BaseDataInterface):
             "required": [self.video_name],
             "additionalProperties": True,
         }
+        metadata_schema["properties"]["Behavior"]["required"].append("ExternalVideoDevices")
+        metadata_schema["properties"]["Behavior"]["properties"]["ExternalVideoDevices"] = {
+            "type": "object",
+            "properties": {self.device_name: device_metadata_schema},
+            "required": [self.device_name],
+            "additionalProperties": True,
+        }
         return metadata_schema
 
     def get_metadata(self):
         metadata = super().get_metadata()
         video_metadata = {
             "Behavior": {
-                "ExternalVideo": {self.video_name: dict(description="Video recorded by camera.", unit="Frames")}
+                "ExternalVideo": {self.video_name: dict(description="Video recorded by camera.", unit="Frames")},
+                "ExternalVideoDevices": {self.device_name: dict(description="Video camera used for recording.")},
             }
         }
         return dict_deep_update(metadata, video_metadata)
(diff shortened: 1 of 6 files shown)


src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -29,6 +30,7 @@ def __init__(
         verbose: bool = False,
         *,
         video_name: str = "ExternalVideo",
+        device_name: str = "ExternalVideoCamera",
```
h-mayorquin, line 33: I am not sure about this. We don't do this for any other device in the other interfaces, we just make the metadata match.

Can the same workflow be achieved here?

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -311,9 +350,21 @@ def add_to_nwbfile(
         videos_metadata = deepcopy(metadata).get("Behavior", dict()).get("ExternalVideo", None)
         if videos_metadata is None:
             videos_metadata = deepcopy(self.get_metadata()["Behavior"]["ExternalVideo"])
-        image_series_kwargs = metadata["Behavior"]["ExternalVideo"][self.video_name]
+        image_series_kwargs = videos_metadata[self.video_name]
```
h-mayorquin, line 353: I saw this bug. Let's fix it first actually. I will take care.

h-mayorquin (commented):
I took a first look.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: https://github.com/catalystneuro/neuroconv/pull/1283

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: Not if we want to be able to easily rename devices. In other interfaces I have to manually remove the default device fields from the metadata.  It would be much simpler to specify the device name at the init.

See https://github.com/catalystneuro/schneider-lab-to-nwb/blob/f1954ffd57857cc6309e6c75a00a6e24d332f0c3/src/schneider_lab_to_nwb/zempolich_2024/zempolich_2024_open_ephys_recording_interface.py#L17

h-mayorquin: That's a bug that we should solve:

https://github.com/catalystneuro/neuroconv/pull/1059

pauladkisson: Exactly, and I think this is the simplest, most explicit way to solve it.

pauladkisson: What if you had two Intan interfaces with different devices? Would #1059 still work? Probably an uncommon situation for Intan but very common for video.

h-mayorquin: I mean for OpenEphys we should fix the bug. Passing a device in the metadata should lead to no devices being added in the ecephys interfaces.

> Exactly, and I think this is the simplest, most explicit way to solve it.

I disagree.

In this case, I believe disambiguation is better achieved by nesting the device metadata under the corresponding ExternalVideo section in the metadata. This way, someone specifying metadata via YAML doesn't need to know or match the device name argument from the code in advance. In other words, we avoid coupling the YAML metadata structure to internal code details.

By nesting, the user simply adds the device metadata under the relevant ExternalVideo entry. There’s a single, unambiguous way to specify the device.

This approach:
*Avoids collisions when dealing with multiple videos (each device’s metadata is scoped to its video).
* Keeps the interface signature simpler.
* Avoids coupling across unrelated parts of the YAML schema, making `video_name` the only dynamic key in the specification.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -87,13 +104,21 @@ def get_metadata_schema(self):
...
+        }
         return metadata_schema
 
     def get_metadata(self):
         metadata = super().get_metadata()
         video_metadata = {
             "Behavior": {
-                "ExternalVideos": {self.video_name: dict(description="Video recorded by camera.", unit="Frames")}
+                "ExternalVideos": {self.video_name: dict(description="Video recorded by camera.", unit="Frames")},
+                "ExternalVideoDevices": {self.device_name: dict(description="Video camera used for recording.")},
```
h-mayorquin, line 121: I think this should be nested inside the metadata of the corresponding video to avoid collisions and have one less dynamic element on the metadata.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: I take your point. How about the current implementation?

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -87,13 +91,26 @@ def get_metadata_schema(self):
...
         metadata = super().get_metadata()
         video_metadata = {
             "Behavior": {
-                "ExternalVideos": {self.video_name: dict(description="Video recorded by camera.", unit="Frames")}
+                "ExternalVideos": {
+                    self.video_name: dict(
+                        description="Video recorded by camera.", unit="Frames", device=self._default_device_name
+                    )
+                },
+                "ExternalVideoDevices": {
```
h-mayorquin, line 111: Do you disagree with having the device metadata nested inside the corresponding video?

I was thinking on something like this:

```python
video_metadata = {
    "Behavior": {
        "ExternalVideos": {
            self.video_name: dict(
                description="Video recorded by camera.",
                unit="Frames",
                device={
                    "name": self._default_device_name,
                    "description": "Video camera used for recording."
                },
            )
        }
    }
}
```

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: Looking at it, I added a comment below related to follow on this.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: Actually, that looks very nice, I wasn't understanding what you were asking for earlier. Done!

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -315,6 +333,15 @@ def add_to_nwbfile(
             videos_metadata = deepcopy(self.get_metadata()["Behavior"]["ExternalVideos"])
         image_series_kwargs = videos_metadata[self.video_name]
         image_series_kwargs["name"] = self.video_name
+        device_kwargs = image_series_kwargs.pop("device", None)
```
h-mayorquin, line 336: OK, one last question. If the user passes metadata but doesn't include the nested `device` dictionary, then no device will be created, right? Is that the intended behavior?

There are also a couple of edge cases, for example, if the user provides the `device` dictionary but not the `name`. I think this will be caught by the doc_val or the `validate_metadata` at write time. What do you think?

h-mayorquin (commented):
This looks good. One more question on my side and this should be ready to go.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
pauladkisson: > There are also a couple of edge cases, for example, if the user provides the device dictionary but not the name. I think this will be caught by the doc_val or the validate_metadata at write time. What do you think?

Yep, this gets caught. I just added a test to ensure this behavior.

h-mayorquin: Good. It makes sense to have a way of doing that.

h-mayorquin (approved):
LGTM

Thanks for adding another test. That was above and beyond the call of duty : )

PR #1283, opened 2025-04-01
Fix default metadata bug for ExternalVideoInterface

h-mayorquin:
This escaped us in the recent merge. This does not need a changelog because is a fix for a bug that has not been released.

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
+++ b/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
@@ -308,7 +308,8 @@ class ExternalVideoInterface(BaseDataInterface):
 
         # Be sure to copy metadata at this step to avoid mutating in-place
         videos_metadata = deepcopy(metadata).get("Behavior", dict()).get("ExternalVideos", None)
-        if videos_metadata is None:
+        # If no metadata is provided use the default metadata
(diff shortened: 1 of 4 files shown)


pauladkisson (approved):

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -308,7 +308,8 @@ def add_to_nwbfile(
 
         # Be sure to copy metadata at this step to avoid mutating in-place
         videos_metadata = deepcopy(metadata).get("Behavior", dict()).get("ExternalVideos", None)
-        if videos_metadata is None:
+        # If no metadata is provided use the default metadata
+        if videos_metadata is None or self.video_name not in videos_metadata:
```
pauladkisson, line 312: When is this going to happen?

h-mayorquin: Metadata with entries for `ExternalVideos` is received as an input but there is no entry for this specific video.

h-mayorquin: Wait that sounds wrong.

h-mayorquin: Forget it, that seems correct to me.

PR #1284, opened 2025-04-02
Fix regression. Restore default naming behavior for VideoInterfaces

h-mayorquin:
This should come after #1283 and has it as its base.

One recurring issue we've encountered is that users often want to add two interfaces of the same type, which leads to naming clashes. We saw this during the previous roadshow with BlackRock, and it's been a consistent pain point that needs to be addressed.

While working on #1243 and specifically on #1251 we updated the default naming to use cleaner, more readable names. However, this change now requires users to manually rename interfaces to avoid conflicts.

This PR restores the previous behavior by assigning a unique default name, allowing users to add multiple interfaces of the same type without running into metadata clashes. The ease of changing the name to something better is kept. 

No changelog necessary as this has not been released.


pauladkisson:
I understand why we need a mechanism to keep the default names unique in the case of multiple interfaces, but I think we can do better than `f"Video {ugly file name}"`.

How about just `"ExternalVideo 1"` for the first video added, `"ExternalVideo 2"` for the second one and so on?

h-mayorquin:
If we can make an improvement in that direction I am fine with it as this is a new interface and a new naming convention would be OK.

Can you create a PR to showcase your idea?

That said, I would like to have this as a default so if we don't manage to agree on time we release functioning (although ugly) code and then we fix improve it later.

pauladkisson:
> Can you create a PR to showcase your idea?

Ok, it actually ended up being a harder problem than I thought. If you use the nwbfile to determine number than the order they get added determines the name (which seems wrong). If you use the order of creation then it breaks the isolation/independence of each interface. So, I think using f"Video {file_path.stem}" is actually fine. I still think it looks ugly and will specify video_name for my conversions, but it unambiguously communicates which video is which without the need for interdependent interfaces.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
```diff
@@ -45,7 +46,7 @@ def __init__(
             If True, display verbose output. Defaults to False.
         video_name : str, optional
             The name of this video as it will appear in the ImageSeries.
-            Defaults to "ExternalVideo".
+            Defaults to the name of the first file (without extension) in the list of file_paths.
```
pauladkisson, line 49:
```suggestion
            Defaults to f"Video {file_paths[0]}".
```

h-mayorquin:
Yes, it is hard problem. We should address it in its generality at some point. Well, at least we now have simpler interfaces and a simpler way of changing the name.

src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
h-mayorquin: Thanks.

src/neuroconv/datainterfaces/behavior/video/internalvideointerface.py
```diff
@@ -50,7 +50,7 @@ def __init__(
             If True, display verbose output. Defaults to False.
         video_name : str, optional
             The name of this video as it will appear in the ImageSeries.
-            Defaults to "InternalVideo".
+            Defaults to the `Video file_path.stem` if not provided.
```
pauladkisson, line 53:
```suggestion
            Defaults to f"Video {file_path.stem}" if not provided.
```

pauladkisson (changes requested):
Looks good, might as well be explicit about the defaults.

h-mayorquin:
OK, did the latest changes.

pauladkisson (approved):
LGTM!

PR #1285, opened 2025-04-02
Fix `VideoInterface` when metadata is not passed

h-mayorquin:
I think this is another regression caused by the recent changes. If metadata is not passed to the VideoInterface, it fails because the dictionary doesn't contain the expected behavior when reformatting the metadata.

No changelog for this either.

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
+++ b/src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
@@ -348,8 +348,10 @@ class VideoInterface(BaseDataInterface):
 
         # Transform metadata from list format to nested dictionary format
         metadata_reformatted = deepcopy(metadata)
-        video_name = videos_metadata[0].pop("name")
+        if "Behavior" not in metadata_reformatted:
+            metadata_reformatted["Behavior"] = dict()
 
+        video_name = videos_metadata[0].pop("name")
         # Use appropriate metadata key based on external_mode
         if external_mode:
             metadata_reformatted["Behavior"]["ExternalVideos"] = {video_name: videos_metadata[0]}
--- a/tests/test_behavior/test_video_interface.py
+++ b/tests/test_behavior/test_video_interface.py
@@ -4,9 +4,19 @@ import numpy as np
 import pytest
 from dateutil.tz import gettz
 from pynwb import NWBHDF5IO
+from pynwb.testing.mock.file import mock_NWBFile
 
 from neuroconv import NWBConverter
 from neuroconv.datainterfaces import VideoInterface
+
+
+def test_video_interface_without_metadata(video_files):
+    """Test that the VideoInterface class can be imported."""
+
+    interface = VideoInterface(file_paths=[video_files[0]])
+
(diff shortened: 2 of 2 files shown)


src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
```diff
@@ -348,8 +348,10 @@ def add_to_nwbfile(
...
         # Transform metadata from list format to nested dictionary format
         metadata_reformatted = deepcopy(metadata)
-        video_name = videos_metadata[0].pop("name")
+        if "Behavior" not in metadata_reformatted:
+            metadata_reformatted["Behavior"] = dict()
 
+        video_name = videos_metadata[0].pop("name")
         # Use appropriate metadata key based on external_mode
         if external_mode:
             metadata_reformatted["Behavior"]["ExternalVideos"] = {video_name: videos_metadata[0]}
```
pauladkisson, line 357: Alternatively, we could make metadata a DeepDict on line 330, and then this assignment would be fine.

pauladkisson: and I think we really should bc in BaseDataInterface, metadata is a DeepDict

h-mayorquin: We are very inconsistent about this. The code that you wrote assumed that it was not a deep dict (hence all the get and default options) so I kept with that.

pauladkisson: > Yes, I thought about this but just did not know how. Can you do `DeepDict(dict)` and that would work?

Yeah `metadata = DeepDict(metadata)` should work just fine.

pauladkisson: > We are very inconsistent about this.

Yeah, that's true, we really should be more consistent about it. It's a pretty easy swap.

h-mayorquin: Good to know.

Feel free to unify this in a PR after this. We should also add return types to get_metadata.

h-mayorquin: this seems like a good job for cline.

pauladkisson: Yeah, maybe we can live-code at the next AI meeting (I'm supposed to present and it fits with my theme).

h-mayorquin: Seems good. I can volunteer to fix the DeepDict approach here after we have merged all the related PRs as well but I would like to fix this regression as soon as possible.

Do you have any other comments regarding the fix of this bug?

pauladkisson (approved):

src/neuroconv/datainterfaces/behavior/video/videodatainterface.py
pauladkisson: Ok, sounds good.

PR #1286, opened 2025-04-02
Update video timestamps

pauladkisson:
slightly different from what I described in #1279.

Basically, `set_aligned_segment_starting_times` assumes you want to save the full timestamps, and sets them automatically if they haven't been set already. Then, `set_aligned_starting_time` can be used if you want to use starting_time and rate.

### DIFF
--- a/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
+++ b/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py
@@ -187,13 +187,17 @@ class ExternalVideoInterface(BaseDataInterface):
                 segment_starting_time + aligned_starting_time for segment_starting_time in self._segment_starting_times
             ]
         else:
-            raise ValueError("There are no timestamps or starting times set to shift by a common value!")
+            self._segment_starting_times = [np.nan] * self._number_of_files
+            self._segment_starting_times[0] = aligned_starting_time
 
     def set_aligned_segment_starting_times(self, aligned_segment_starting_times: list[float], stub_test: bool = False):
         """
         Align the individual starting time for each video (segment) in this interface relative to the common session start time.
 
         Must be in units seconds relative to the common 'session_start_time'.
+
+        If the timestamps have not already been set, this method will set them to the original timestamps and then shift
+        them by the aligned segment starting times.
 
         Parameters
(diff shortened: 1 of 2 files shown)


h-mayorquin:
I like the modification to `set_aligned_segment_starting_times`, since we're not really using `self._segment_starting_times` and those values would always be set based on the timestamps. This is a good improvement.

However, I'm not in favor of removing the shifting behavior from `set_aligned_starting_time`, as I think it's a useful feature overall. I typically set the aligned timestamps first, and then shift them later, at a different point in the code when the shift time becomes available, without needing to reset them again. Even if the timestamps are regular, this section of the code ensures the correct starting time is applied:

https://github.com/catalystneuro/neuroconv/blob/a8ea8a60407ab7c6b66c3ba54293253a27e6ff39/src/neuroconv/datainterfaces/behavior/video/externalvideointerface.py#L294-L296

What do you think about it?

pauladkisson:
> I typically set the aligned timestamps first, and then shift them later, at a different point in the code when the shift time becomes available, without needing to reset them again. Even if the timestamps are regular, this section of the code ensures the correct starting time is applied:

I'm not sure I understand this use case. Is there a reason you would want to use `set_aligned_starting_time` instead of `set_aligned_segment_starting_times`?

h-mayorquin:
I see—I think I may have misunderstood your intention.

The view that I favor is that `set_aligned_starting_time` is intended for the single file path case, while `set_aligned_segment_starting_times` handles the multiple file path case. From that perspective, I expected the function to apply the start time as a shift when timestamps already exist.

However, it sounds like your interpretation may be more consistent with the previous implementation where `set_aligned_starting_time` is used when writing data using a `start_time` and `rate`, and `set_aligned_segment_starting_times` is used when timestamps are present, offering a convenient way to apply a shift in that case.

Does that sound like an accurate reflection of your thinking?

pauladkisson:
> In RecordingInterfaces, the shift is applied in the single file/segment case, so I’d like to keep that behavior here as well.

Ok, `set_aligned_starting_time` will shift pre-existing timestamps now.

h-mayorquin (approved):
Thanks, @pauladkisson

PR #1292, opened 2025-04-07
Improve ElectricalSeries units in recording interfaces

### LINKED ISSUE #1266: add_electrical_series_to_nwbfile's write_scaled doesn't do anything if True
I was chasing through `add_electrical_series_to_nwbfile()` in neuroconv/tools/spikeinterface/spikeinterface.py today for https://github.com/catalystneuro/dan-lab-to-nwb/pull/9, and I noticed that `write_scaled` doesn't actually do anything if it's True. I would have expected it to set `returned_scaled=True` for `_recording_traces_to_hdmf_iterator` so that the data is appropriately converted to uV. Maybe I'm missing something?

h-mayorquin:
We have been discussing with the SpikeInterface team about improving unit handling [here](https://github.com/SpikeInterface/spikeinterface/pull/3844), and this is the first PR moving in that direction. This PR improves the following:

1) It checks whether the recording has the necessary information to scale to microvolts (gains and offsets). If not, it throws a warning and writes the data as it is. Note that, unlike before, it does not allow partially specified offsets and gains. This is in preparation for a future error that SpikeInterface will raise if units are not properly scalable to microvolts. I might make this an error once we make further changes in SpikeInterface as I don't want to break anything at the moment.

2) This should also fix #1266, which is currently not working and is something we want to discourage anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 # v0.7.3 (Upcoming)
 
 ## Deprecations and Changes
-* `write_scaled` behavior on `add_electrical_series_to_nwbfile` is deprecated and will be removed in or after October 2025 [PR #TBD](https://github.com/catalystneuro/neuroconv/pull/TBD)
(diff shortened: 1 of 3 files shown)


pauladkisson (approved):
LGTM!

PR #1293, opened 2025-04-07
Write rate and starting time in  `DeepLabCutInterface` if timestamps are regular

h-mayorquin:
Draft. From the Marie Suver conversion.


luiztauffer (approved):
looks good to me

PR #1295, opened 2025-04-07
Update blackrock.rst

bendichter:


h-mayorquin (approved):
Maybe we should add something to the same effect on the index.

Issue #1280, opened 2025-04-01
[Discussion]: New interfaces should use plural names when categories can contain multiple elements


pauladkisson:
Currently, some interfaces use singular names when categorizing potentially multiple objects. Ex. for RecordingInterfaces, `metadata["Ecephys"]["Device"]` is a list of devices.

In such cases we should use the plural to communicate that it could be more than one object. Ex. `metadata["Ecephys"]["Devices"]`.

h-mayorquin:
I am +1 on this.

This will be a breaking change for yaml files which is costly.

If we decide to move forward we should do it together with other metadata schema breaking changes, maybe https://github.com/catalystneuro/neuroconv/issues/269.

h-mayorquin:
I will handle this for old interface when I do the change on #269.

h-mayorquin:
Done as part of #269, which moved the whole metadata structure to dicts keyed by a name you choose. Every field that can hold more than one entry is plural now, top-level `Devices` and `DeviceModels`, then `ElectrodeGroups` and `ElectricalSeries` under `Ecephys`, `ImagingPlanes` and `PlaneSegmentations` under `Ophys`, `Skeletons` and `PoseEstimations` under `Pose`, and `EventTables` under `Events`.

The rule is written down in the developer guide now, so a new interface has something to follow:
https://github.com/catalystneuro/neuroconv/blob/664c0ade0aa3d7708b2a26964e9d5a62b3c64cb6/docs/developer_guide/metadata_principles.rst#L200-L213

The settled design is written down per modality, #1653 for ophys, #1759 for events and #1778 for fiber photometry, with the cross-modality rules in #1801 and #1822. Ecephys is the one still missing and that is #1729:
https://neuroconv.readthedocs.io/en/main/developer_guide/metadata_principles.html

The yaml breaking change I was worried about here is paid for, it went out with the format switch in #1934 instead of on its own.

PR #1237, opened 2025-03-06
Add analog interface for openephys binary

h-mayorquin:
Follow up on #1179 and similar to #1152

PR #1296, opened 2025-04-08
Add  `add_time_series_to_nwbfile` 

h-mayorquin:
As part of the ongoing effort to write all available data (i.e. not only neural) from a format in an easy way, this PR introduces the `add_time_series_to_nwbfile` function.

This function enables users to add non-neural recording data to NWB as a `TimeSeries`. It performs metadata validation and, when available, adds unit information, gains, and offsets.

This update also simplifies the process of integrating analog interfaces such as the `SpikeGLXNIDQInterface` interface, which I’ve refactored in this PR to demonstrate the new functionality.

Additional context comes from a request by @bendichter in this comment: https://github.com/catalystneuro/neuroconv/pull/1237#discussion_r1985650901

This complies with https://github.com/catalystneuro/neuroconv/issues/269 and #1280

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # v0.7.3 (Upcoming)
 
 ## Deprecations and Changes
+* `write_scaled` behavior on `add_electrical_series_to_nwbfile` is deprecated and will be removed in or after October 2025 [PR #1292](https://github.com/catalystneuro/neuroconv/pull/1292)
+* `add_electrical_series_to_nwbfile` now requires both gain and offsets to write scaling factor for voltage conversion when writing to NWB [PR #1292](https://github.com/catalystneuro/neuroconv/pull/1292)
 
 ## Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ## Bug Fixes
 
 ## Features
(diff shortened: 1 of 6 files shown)


luiztauffer (approved):
That’s nice! But this is specific for writing from spikeinterface recordings, right? I find the name to be too general, and we could perhaps specify the use case in it. `add_time_series_to_nwbfile` sounds like a more general tool, that would cover other modalities as well, behavior for example.

h-mayorquin:
> That’s nice! But this is specific for writing from spikeinterface recordings, right? I find the name to be too general, and we could perhaps specify the use case in it. `add_time_series_to_nwbfile` sounds like a more general tool, that would cover other modalities as well, behavior for example.

Yes, some other names are like this in that module.

* add_electrical_series_to_nwbfile

They are in `src/neuroconv/tools/spikeinterface/spikeinterface.py`

Maybe `add_recording_as_time_series_to_nwbfile` ?

weiglszonja:
+1 for `add_recording_as_time_series_to_nwbfile` or `add_time_series_from_recording_to_nwbfile`, since it expects a `Recoding` object I would add this in the name to make it clear.

h-mayorquin:
@luiztauffer @weiglszonja I think this is a good idea, I changed it.

Maybe, in another PR, I should make `add_electrical_series_to_nwbfile` private and only expose `add_recording_to_nwbfile`

PR #1302, opened 2025-04-17
Fix `OpenEphysRecordingInterface` api docs

### LINKED ISSUE #1045: API Docs for OpenEphys router displays the `__init__` documentation of the base class
You can see here that the documentation is wrong

image

The problem is that this class does not have an`__init__` then the documentation for the `__init__` is inhertied from the BaseClass and is displayed wrongly.

https://neuroconv.readthedocs.io/en/main/api/interfaces.ecephys.html#module-neuroconv.datainterfaces.ecephys.openephys.openephysdatainterface

h-mayorquin:
Should fix #1045

It is fixed now:

image

I also updated the doc dependencies to their latest version, fixed some warnings on the doc building process and did minor corrections. here and there.


CHANGELOG.md
h-mayorquin, line 1: I realize a lot of links to the PRs in the changelog were missing the PR letters. Added them.

weiglszonja (approved):
Thanks! I think the docs look great

PR #1305, opened 2025-04-21
Test that top level modules can be imported with minimal installation

### LINKED ISSUE #1304: [Bug]: neuroconv[suite2p] not installing dependencies
### What happened?

The example provided in [the official doc](https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/segmentation/suite2p.html) does not work. 

After installing `neuroconv` via `pip install "neuroconv[suite2p]"`, running `from neuroconv.datainterfaces import Suite2pSegmentationInterface` would give us following error messages.

> ModuleNotFoundError: No module named 'spikeinterface'

Installing `spikeinterface` separately fixes this issue.

### Steps to Reproduce

```python
pip install "neuroconv[suite2p]"
from neuroconv.datainterfaces import Suite2pSegmentationInterface
```

### Traceback

```python

```

### Operating System

Windows

### Python Executable

Conda

### Python Version

3.9

### Package Versions

_No response_

### Code of Conduct

h-mayorquin:
This should fix #1304. The specific error that trigger the user is now fixed on main but we should have a test for this (which I just added).

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/sortedrecordinginterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/sortedrecordinginterface.py
@@ -7,9 +7,6 @@ from neuroconv.datainterfaces.ecephys.baserecordingextractorinterface import (
 from neuroconv.datainterfaces.ecephys.basesortingextractorinterface import (
     BaseSortingExtractorInterface,
 )
-from neuroconv.tools.spikeinterface.spikeinterface import (
-    _get_electrode_table_indices_for_recording,
-)
 
 
 class SortedRecordingConverter(ConverterPipe):
@@ -90,6 +87,10 @@ class SortedRecordingConverter(ConverterPipe):
             nwbfile=nwbfile,
             metadata=metadata,
             **conversion_options_recording,
+        )
+
+        from ...tools.spikeinterface.spikeinterface import (
+            _get_electrode_table_indices_for_recording,
         )
 
         # This returns the indices in the electrode table that corresponds to the channel ids of the recording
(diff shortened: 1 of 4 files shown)


h-mayorquin:
Great, a similar error surfaced here. I will fix it.

bendichter (approved):

src/neuroconv/datainterfaces/ecephys/sortedrecordinginterface.py
```diff
@@ -92,6 +89,10 @@ def add_to_nwbfile(self, nwbfile, metadata, conversion_options: Optional[dict] =
             **conversion_options_recording,
         )
 
+        from ...tools.spikeinterface.spikeinterface import (
+            _get_electrode_table_indices_for_recording,
+        )
```
bendichter, line 94: Why was this moved? It's not an external import, so it should not depend on optional dependencies

h-mayorquin: The `spikeinterface` module in `tools` imports `BaseRecording` from `spikeinterface`:  
https://github.com/catalystneuro/neuroconv/blob/d09abd9465af00f346b60b11f20af105e35695de/src/neuroconv/tools/spikeinterface/spikeinterface.py#L12-L13

That said, it might be better to make even that module (`neuroconv.tools.spikeinterface`) importable without requiring `spikeinterface`. The tradeoff would be losing type introspection in that module for editors, which might be acceptable.

Note that the equivalent module for interacting with `roiextractors` also imports `roiextractors`:  
https://github.com/catalystneuro/neuroconv/blob/d09abd9465af00f346b60b11f20af105e35695de/src/neuroconv/tools/roiextractors/roiextractors.py#L28-L32

If the goal is to make these modules importable with minimal dependencies, I can handle that in a separate PR after this one. Let me know what you think.

bendichter: I see. Ok I think it's fine the way it is

PR #1307, opened 2025-04-23
Expose iterative write parameters on `BaseImagingExtractor`

h-mayorquin:
Two things:
1) Those parameters were previously not exposed, now they are.
2) Also, now that we are adjusting the semantics of frame-shape-sample I fixed the calculation of the max shape which is buggy at the moment because the previously `get_image_size` did return the size of the volume sometimes.

### DIFF
--- a/src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
+++ b/src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
@@ -114,10 +114,10 @@ class ImagingExtractorDataChunkIterator(GenericDataChunkIterator):
         assert buffer_gb > 0, f"buffer_gb ({buffer_gb}) must be greater than zero!"
         assert all(np.array(chunk_shape) > 0), f"Some dimensions of chunk_shape ({chunk_shape}) are less than zero!"
 
-        sample_shape = self._get_maxshape()[1:]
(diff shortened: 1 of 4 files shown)


src/neuroconv/tools/roiextractors/imagingextractordatachunkiterator.py
```diff
@@ -114,35 +114,45 @@ def _get_scaled_buffer_shape(self, buffer_gb: float, chunk_shape: tuple) -> tupl
...
     def _get_maxshape(self) -> tuple:
-        video_shape = (self.imaging_extractor.get_num_frames(),)
-        image_shape = self.imaging_extractor.get_image_size()
-        width, height = image_shape[1], image_shape[0]  # ROIExtractors convention is flipped
-        video_shape += (width, height)
-        if len(image_shape) == 3:
-            depth = image_shape[2]
-            video_shape += (depth,)
-        return video_shape
+
```
h-mayorquin, line 134: This is fixing a bug because now image_shape always returns the shape of the image while previously it could return the shape of the volume.

Once the next version of roiextractors is released we can make a better fix for this but at the moment the simplest and safest option is just to get a sample and look at the shape.

pauladkisson (approved):
LGTM!

h-mayorquin:
Thanks for the review, @pauladkisson.

Issue #1248, opened 2025-03-07
[Question] How to add location or brain area to the electrodes table in neuroconv


h-mayorquin:
This question came from @TommasoLambresa and @FrancescoNegri from the University of Genoa.

Here is an example with Intan but it would work with any other interface

```python
from neuroconv.datainterfaces import IntanRecordingInterface
from pynwb import NWBFile
from datetime import datetime, timezone

time_now = datetime.now(tz=timezone.utc)

nwbfile = NWBFile(    
    session_description="Session 1",
    identifier="NWB123",
    session_start_time=time_now,
    file_create_date=time_now
)


interface = IntanRecordingInterface(file_path=file_path)

recording_extractor = interface.recording_extractor
# This you will have to fill with the proper location
brain_area = ["brain_location_per_electrode"] * recording_extractor.get_num_channels()
recording_extractor.set_property("brain_area", values=brain_area)

interface.add_to_nwbfile(nwbfile=nwbfile)

```

After that, your nwbfile will have "location" in the electrodes table set to your desired values. 

I will upgrade this with clarifications and then we should add this to the documentation.

FrancescoNegri:
Thank you for your answer @h-mayorquin!

Unluckily, in our case we build the NWB file with unknown location and fix them later, as the location of each electrode is inferred via LFP analyses carried out with SpikeInterface.

Hence, we would like to edit the location column of an already existing NWB file.

According to my understanding, this should be supported and is compliant with the guidelines outlined [here](https://nwbinspector.readthedocs.io/en/dev/best_practices/ecephys.html).

We didn't manage to find a way to edit the location column values for each electrode after it has been set to unknown at the NWB file creation.

h-mayorquin:
By existing you mean that the file is already written to disk?

FrancescoNegri:
Yes, we typically convert our raw data (`.rhs`, `.pl2`, ...) to `.nwb` file with `neuroconv`. Then, we load the `.nwb` file into `SpikeInterface` to perform some analyses and to estimate the location of electrodes from LFP activity. After that, I would like to update the location column in the electrodes table of the `.nwb` file mentioned above.

h-mayorquin:
Ok, then something like this should work:

Once you wrote a file and you want to modify the location in the electrodes table. You open it on read and write mode (r+)

```python
# Read the NWB file
from pynwb import NWBHDF5IO
with NWBHDF5IO(nwbfile_path, "r+") as io:
[...]

PR #1311, opened 2025-04-28
Documentation: Add how to guide for annotating electrode groups and locations in ecephys data

h-mayorquin:
The request for how to annotate electrode metadata came from slack and should also address #1248.


chrishalcrow:
Hey @h-mayorquin, general comment: would be great if guide was clearer about `ElectrodeGroup` being special. I find all this stuff a bit confusing, but am I right that every electrode should belong to a group, and only one group. And is the property key `group` reserved specially for ElectrodeGroups? And this is really part of the nwb specification. Then the second Section is a bit different: you can add any number of arbitrary labels that you make yourself, and this is additional metadata on top of everything else.

docs/how_to/annotate_ecephys_data.rst
```diff
@@ -0,0 +1,115 @@
...
+    metadata = interface.get_metadata()
+
+    # Define multiple electrode groups with different metadata
+    electrode_group_A = {"name": "ElectrodeGroupA", "description": "A description", "location": "Location A"}
+    electrode_group_B = {"name": "ElectrodeGroupB", "description": "B description", "location": "Location B"}
+    electrode_group_C = {"name": "ElectrodeGroupC", "description": "C description", "location": "Location C"}
+
+    # Replace the default ElectrodeGroup with a list of electrode groups
+    electrode_groups = [electrode_group_A, electrode_group_B, electrode_group_C]
+    metadata["Ecephys"]["ElectrodeGroup"] = electrode_groups
```
chrishalcrow, line 34: Is there a danger that the user is overwriting existing, important metadata on this line?

docs/how_to/annotate_ecephys_data.rst
```diff
@@ -0,0 +1,115 @@
...
+    recording = interface.recording_extractor
+    channel_id_to_group_names = {
+        "0": "ElectrodeGroupA",
+        "1": "ElectrodeGroupA",
+        "2": "ElectrodeGroupC",
+        "3": "ElectrodeGroupB",
+        "4": "ElectrodeGroupA",
+    }
+    recording.set_property(
+        key="group",
```
chrishalcrow, line 46: Curious: in spikeinterface we always use group for NP probes to label the shanks. Is that bad?? Are there other groupings people might wanna use to sort into ElectrodeGroups?

docs/how_to/annotate_ecephys_data.rst
```diff
@@ -0,0 +1,115 @@
...
+
+How to Annotate Extracellular Electrophysiology Data
+====================================================
+
+This guide provides instructions on how to annotate extracellular electrophysiology data in NeuroConv.
+
+How to Set ElectrodeGroup Metadata
+----------------------------------
+
+When working with extracellular electrophysiology data, it can be helpful to segment electrodes into different
```
chrishalcrow, line 11: Related to the general comment: to me it seems "helpful" to add metadata like brain area; but it seems really important to group your electrodes into e.g. shanks for NP probes or tetrodes for multi-tetrode data. To me, there are two problems in this guide and it'd be nice to seperate them a bit.

docs/how_to/annotate_ecephys_data.rst
h-mayorquin: It should only change the group metadata and nothing else which is the intended effect.

docs/how_to/annotate_ecephys_data.rst
h-mayorquin: No, it is OK. Groups is just a special property that people might use for partition their electrode table. Intan for example, has different ports, see this PR:

https://github.com/SpikeInterface/spikeinterface/pull/3753

docs/how_to/annotate_ecephys_data.rst
h-mayorquin: Both are very important from the point of view of NWB and both are special columns in the electrodes table. These are the required arguments.

```
def ElectrodeTable(name='electrodes',
                   description='metadata about extracellular electrodes'):
    return _tablefunc(name, description,
                      [('location', 'the location of channel within the subject e.g. brain region'),
                       ('group', 'a reference to the ElectrodeGroup this electrode is a part of'),
                       ('group_name', 'the name of the ElectrodeGroup this electrode is a part of')
                       ]
                      )
```

The reason that both are important is that NWB is an archival format so the brain location matters a lot for future analysis. Electrode group is the basic grouping of your electrodes so if they come from a different probe, shank or port we want to know that.

h-mayorquin:
Hi, @chrishalcrow 

Thanks a lot for taking a look. I will add more clarifying comments in the direction of your questions as I think they will be useful for other users who might be confused in the same way.

>Hey @h-mayorquin, general comment: would be great if guide was clearer about ElectrodeGroup being special.

Yes, as I mentioned above both the location and the electrodes group are special. I will emphasize that.

> I find all this stuff a bit confusing, but am I right that every electrode should belong to a group, and only one group. And is the property key group reserved specially for ElectrodeGroups?

You are correct, every row of the electrode table represents an electrode and every row has an electrode group value.

docs/how_to/annotate_ecephys_data.rst
chrishalcrow: Am I right in thinking that an electrode must be part of a group, but you can upload a file to DANDI without locations?

h-mayorquin: You can't build the electrodes table without locations. If your file requires an electrode table (an a file with an `ElectricalSeries` does) then you need to add locations. That's why you will find a lot of dummy values:

https://github.com/catalystneuro/neuroconv/blob/551fe106033e13764c74797f8cde8e719ee9b99b/src/neuroconv/tools/spikeinterface/spikeinterface.py#L787-L795

But is is very information for future analysis and to have the complete metadata of an experiment so we are adding a guide here. Hopefully, more people will annotate them like this.

docs/how_to/annotate_ecephys_data.rst
```diff
@@ -0,0 +1,140 @@
...
+    )
+
+    # Create the NWBFile with the updated metadata
+    nwbfile = interface.create_nwbfile(metadata=metadata)
+
+    # You can check that your electrodes are correctly representing here
+    nwbfile.electrode_groups
+
+    # And that they are mapped correctly to the channels/electrodes
+    nwbfile.electrodes.to_dataframe()
```
weiglszonja, line 66: I think it would be also useful to show how this table looks like, easier to imagine where the group ends up in the table. but only if you agree as well, this is very minor

docs/how_to/annotate_ecephys_data.rst
```diff
@@ -0,0 +1,140 @@
...
+
+
+This approach allows you to associate different channels with different electrode groups, each with its own metadata such as location.
+
+How to Add Location to the Electrodes Table
+-------------------------------------------
+
+In addition to setting electrode group metadata, you may want to add specific location information for each individual electrode in the electrodes table.
+This is particularly useful when electrodes are located in different brain areas or when you want to provide more detailed anatomical information.
+Use standard atlas names for anatomical regions when possible.
```
weiglszonja, line 80:
```suggestion
Use standard atlas names (e.g. Allen Brain Atlas) for anatomical regions when possible.
```

docs/how_to/annotate_ecephys_data.rst
```diff
@@ -0,0 +1,140 @@
...
+    # You can verify that the brain_area property was added to the electrodes table
+    nwbfile.electrodes.to_dataframe()
+
+    # Write the NWB file to disk
+    nwbfile_path = "your_annotated_file.nwb"
+    configure_and_write_nwbfile(nwbfile=nwbfile, nwbfile_path=nwbfile_path)
+
+This approach allows you to add specific location information for each electrode, which will be included in the NWB file's electrodes table.
+Note that any other property of the electrodes can be added in a similar way such as impedance, filtering, stereotaxic coordinates, etc.
+
```
weiglszonja, line 125: I think it would be worth adding an example for annotating the coordinates of the electrodes here as well, what do you think?

weiglszonja (approved):
LGTM, thanks!

Issue #1310, opened 2025-04-27
[Bug]: Neuroconv cannot be installed with the latest pip 25.1 on Python 3.9


rly:
### What happened?

[pip 25.1 was released today](https://pip.pypa.io/en/stable/news/) and appears to change how dependencies are resolved.

- `dandi` requires `urllib3>=2.0.0` as of 0.64.0 (neuroconv currently requires "dandi>=0.66.5")
- `boto3` requires `botocore` which [currently requires "urllib3>=1.25.4,<1.27"](https://github.com/boto/botocore/blob/174ae89d96676a11589d3e296b8ff5ace3bbc659/setup.cfg#L8) on Python<3.10 (neuroconv currently requires any version of `boto3`)

The previous version of pip, 25.0.1, would spend a LONG time iterating backwards through versions of `boto3` and `botocore` from the current `boto3` 1.38.3 and `botocore` 1.38.3 back to `boto3-1.7.84` and `botocore-1.10.84` which were released 7 years ago when urllib3 was not a requirement (compare https://github.com/boto/botocore/blob/1.11.0/setup.py#L28 and https://github.com/boto/botocore/blob/1.10.84/setup.py#L27)

```
Collecting botocore<1.38.0,>=1.37.37 (from boto3->neuroconv==0.7.3)
  Downloading botocore-1.37.37-py3-none-any.whl.metadata (5.7 kB)
Collecting boto3 (from neuroconv==0.7.3)
  Downloading boto3-1.37.36-py3-none-any.whl.metadata (6.7 kB)

... cut 6300 lines ...

Collecting boto3 (from neuroconv==0.7.3)
  Downloading boto3-1.8.0-py2.py3-none-any.whl.metadata (4.7 kB)
Collecting botocore<1.12.0,>=1.11.0 (from boto3->neuroconv==0.7.3)
  Downloading botocore-1.11.0-py2.py3-none-any.whl.metadata (2.7 kB)
Collecting boto3 (from neuroconv==0.7.3)
  Downloading boto3-1.7.84-py2.py3-none-any.whl.metadata (4.8 kB)
Collecting botocore<1.11.0,>=1.10.84 (from boto3->neuroconv==0.7.3)
  Downloading botocore-1.10.84-py2.py3-none-any.whl.metadata (2.7 kB)
```

[...]

PR #1312, opened 2025-04-28
Python 3.9 dependencies

h-mayorquin:
Testing what might be a simple fix for this #1310


h-mayorquin:
First attempt fail:

```bash
        if server_schema_version != schema_version:
>           raise SchemaVersionError(
                f"Server requires schema version {server_schema_version};"
                f" client only supports {schema_version}.  You may need to"
                " upgrade dandi and/or dandischema."
            )
E           dandi.exceptions.SchemaVersionError: Server requires schema version 0.6.9; client only supports 0.6.8.  You may need to upgrade dandi and/or dandischema.
```

PR #1313, opened 2025-04-28
Drop support for python 3.9

### LINKED ISSUE #1310: [Bug]: Neuroconv cannot be installed with the latest pip 25.1 on Python 3.9

h-mayorquin:
Testing.

Should fix #1310.


h-mayorquin:
Note that this will mean dropping support for Cambridge Electronic Design (CED): 

https://github.com/catalystneuro/neuroconv/blob/cab89fe7094a382746ed94c7b48520717573211c/src/neuroconv/datainterfaces/ecephys/spike2/spike2datainterface.py#L10-L29

luiztauffer (approved):

PR #1318, opened 2025-05-01
Separate subject_id from individual_name in DeepLabCutInterface

pauladkisson:
Fixes #1317

PR #1319, opened 2025-05-06
Add metadata support for `DeepLabCut`

### LINKED ISSUE #1317: [Feature]: Separate subject_id from individual_name in DLC Interface

Currently, the DeepLabCutInterface assumes that the `individual_name` present in the DLC output file (.csv or .h5) matches the `subject_id` for the NWB file.  However, in general, this is not the case. For example, in my DLC output file there are 2 rats labeled 'rat 1' and 'rat 2', but their subject ids are `XFN1` and `XFN3`.

The DLC interface should have a mechanism to specify an individual_name for a local DLC output (ex. 'rat 1') AND a global subject_id for the nwbfile (ex. 'XFN1').

_No response_

Yes.

### Code of Conduct

h-mayorquin:
Should close #1317

This will allow the users to modify all the aspects of metadata of the dlc interface with our common metadata flow.

I also move the container name to the `init` of the file to be in line with other metadata tag identifiers see #269


src/neuroconv/utils/json_schema.py
```diff
@@ -232,6 +233,18 @@ def fill_defaults(schema: dict[str, Any], defaults: dict[str, Any], overwrite: b
     if properties_reference not in schema and "patternProperties" in schema:
         properties_reference = "patternProperties"
 
+    # Handle additionalProperties case
```
pauladkisson, line 236: What's going on here? Why do you need to edit the json_schema?

pauladkisson:
Given that this PR is very metadata-centric, it might be nice to add the metadata structure to the conversion gallery.

pauladkisson:
> Should close https://github.com/catalystneuro/neuroconv/issues/1317

So, to be clear, this replaces #1318 right?

pauladkisson:
> [The] desired workflow can be accomplished by changing the Skeleton subject to the subject id for the desired interface. The rest of the container names and metadata can also be edited in a simpler way.

Using the metadata to change the names instead of the conversion options makes sense to me, but I'm having a bit of trouble understanding how exactly it would work for multi-subject DLC files. Do you think you could add a test equivalent to `class TestDeepLabCutInterfaceSubjectID(DataInterfaceTestMixin):` from #1318?

src/neuroconv/utils/json_schema.py
h-mayorquin: I was debugging something. I removed this now.

h-mayorquin:
> Given that this PR is very metadata-centric, it might be nice to add the metadata structure to the conversion gallery.

Added. let me know what you think.

> So, to be clear, this replaces https://github.com/catalystneuro/neuroconv/pull/1318, right?

Not sure yet. I think this should be done regardless, but I wanted to discuss it with you first.

> Using the metadata to change the names instead of the conversion options makes sense to me, but I'm having a bit of trouble understanding how exactly it would work for multi-subject DLC files. Do you think you could add a test equivalent to `class TestDeepLabCutInterfaceSubjectID(DataInterfaceTestMixin):` from https://github.com/catalystneuro/neuroconv/pull/1318?

Let's discuss this. This PR enables the following workflow:

```python

dlc_interface = DeepLabCutInterface(file_path=file_path, subject_name="rat 1")
metadata = dlc_interface.get_metadata()

subject_id = "XFN1"

# If the skeleton matches the subject_id, it will be added to the metadata
metadata["PoseEstimation"]["Skeletons"]["SkeletonPoseEstimationDeepLabCut_Rat 1"]["subject"] = subject_id

metadata["Subjects"] = [{"subject_id": subject_id}]  # If the subject_id matches the one in the skeleton, it will be linked.
metadata["NWBFile"] = {
    "session_start_time": "2023-07-20T12:00:00",
    "session_id": "log07-20-2023",
    "session_description": "test session",
}

nwbfile = dlc_interface.create_nwbfile(metadata=metadata)

assert nwbfile.processing["behavior"]["Skeletons"]["SkeletonPoseEstimationDeepLabCut_Rat 1"].subject.subject_id == subject_id

If the subject in the Skeleton matches the `subject_id`, the linking happens automatically. Note that for this to work, the subject must be added beforehand (and intentionally). In this case, the subject is added by `create_nwbfile` because it’s included in the metadata.

I'm not a big fan of the automatic subject addition in the current interface. It creates a made-up name when there's only one subject, and when there are two, it just uses the first interface and it sounds bad that it depends on the order of execution.

File: docs/conversion_examples_gallery/behavior/deeplabcut.rst [hunk 2]
```diff
@@ -28,3 +28,100 @@ This interface supports both .h5 and .csv output files from DeepLabCut.
...
+    >>> metadata["PoseEstimation"]["Skeletons"] = skeletons_metadata
+
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> nwbfile_path = f"{path_to_save_nwbfile}"  # This should be something like: "saved_file.nwb"
+    >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata, overwrite=True)
+
+The metadata structure for DeepLabCut includes:
+
+1. **PoseEstimationContainers** - Contains the main metadata for the pose estimation:
+   - ``name``: Name of the pose estimation container
```
pauladkisson, line 113: This list isn't rendering properly.

docs/conversion_examples_gallery/behavior/deeplabcut.rst
```diff
@@ -28,3 +28,100 @@ This interface supports both .h5 and .csv output files from DeepLabCut.
     >>> metadata["NWBFile"].update(session_start_time=session_start_time)
     >>> # Choose a path for saving the nwb file and run the conversion
     >>> interface.run_conversion(nwbfile_path=path_to_save_nwbfile, metadata=metadata)
+
+
+Specifying Metadata
```
pauladkisson, line 33: This is great, thanks.

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
```diff
@@ -334,122 +373,60 @@ def _write_pes_to_nwbfile(
...
+    # Extract device information
+    device_name = container_metadata.get("devices", [f"Camera{container_name}"])[0]
+    device_metadata = pose_estimation_metadata.get("Devices", {}).get(device_name, {})
+
+    # Create or get the device
+    if device_name not in nwbfile.devices:
         camera = nwbfile.create_device(
-            name=camera_name,
-            description="Camera used for behavioral recording and pose estimation.",
+            name=device_name,
```
pauladkisson, line 383: what about the other device kwargs (ex. manufacturer)?

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -30,21 +32,89 @@ def __init__(
...
+
+        Metadata Structure
+        ------------------
+        The metadata is organized in a hierarchical structure:
+
+        metadata["PoseEstimation"]["Skeletons"][skeleton_name] contains:
+        {
+            "name": "SkeletonPoseEstimationDeepLabCut_SubjectName",
+            "nodes": ["bodypart1", "bodypart2", ...],  # List of keypoints/bodyparts
+            "edges": [[0, 1], [1, 2], ...],  # Connections between nodes (optional)
```
pauladkisson, line 73: These seem more like data than metadata to me.

pauladkisson (commented):
Made some comments

pauladkisson:
> If the subject in the Skeleton matches the subject_id, the linking happens automatically. Note that for this to work, the subject must be added beforehand (and intentionally). In this case, the subject is added by create_nwbfile because it’s included in the metadata. Would a test like this works or are you looking for something different?

Yes, this is essentially what I had in mind, and it clarifies significantly what the workflow would look like.

> I'm not a big fan of the automatic subject addition in the current interface. It creates a made-up name when there's only one subject, and when there are two, it just uses the first interface and it sounds bad that it depends on the order of execution.

Agree. It really shouldn't be DLC's job to create a subject if there isn't one already. It's an optional parameter for Skeleton, so I vote we just leave it unspecified if nwbfile.subject is None.

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
```diff
@@ -85,6 +316,122 @@ def get_metadata(self):
...
+                "original_videos": [video_file_path] if video_file_path else None,
+                "PoseEstimationSeries": {},
+            }
+        }
+
+        # Add a series for each bodypart
+        for bodypart in bodyparts:
+            pose_estimation_metadata["PoseEstimationContainers"][container_name]["PoseEstimationSeries"][bodypart] = {
+                "name": f"{self.subject_name}_{bodypart}",
+                "description": f"Keypoint {bodypart} from individual {self.subject_name}.",
```
pauladkisson, line 426: What do you think about leaving the subject_name out of these descriptions by default? Ex.
```python
"name": f"{bodypart}",
"description": "Keypoint for {bodypart}"
```

It would make my life a lot easier with fewer metadata fields to adjust when renaming subjects, and there's only one PoseEstimation container per subject anyway, so it would remain unambiguous.

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
h-mayorquin: I don't want to move too much in this direction because I really prefer to specify those as a device model instead of as a device:

https://github.com/NeurodataWithoutBorders/nwb-schema/issues/607

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
h-mayorquin: This makes sense to me.

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
h-mayorquin: I think that's a good point especially with the edges.

With the nodes, I think it is nice being able to edit them. With the edges, I feel it is OK to leave it as it just as an escape valve for when users want a fix with this.

How do you think about it? What do you think is better?

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
h-mayorquin: I would like to this and the subject modification in a follow up PR after this big change though. Would you be Ok with that?

docs/conversion_examples_gallery/behavior/deeplabcut.rst
h-mayorquin: Thanks. It should be fixed now.

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
pauladkisson: I worry that DeviceModel won't be merged for a while yet. And it should be a really simple fix to just grab the kwargs from the metadata.

h-mayorquin: Let's wait for the next LBNWL or group meeting to settle this then.

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
pauladkisson: If you renamed the nodes in the metadata, would the interface even work? Based on my reading of `_write_pes_to_nwbfile`, it seems like the node names must match the values in the data.

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
pauladkisson: I think maybe you're misunderstanding my request? Let me make it more concrete...

pauladkisson: Instead of
```python
camera = nwbfile.create_device(
    name=device_name,
    description=device_metadata["description"],
)
```

just use 
```python
camera = nwbfile.create_device(**device_metadata)
```

and let the user work out which device metadata fields to include or not.  No need to worry about which ones might get moved to DeviceModel.

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
h-mayorquin: Let me re-read this. Which line is where the clash will occur?

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
h-mayorquin: No, I think I understand. I am thinking on whether we should allow the writing of attributes which will be deprecated.

src/neuroconv/datainterfaces/behavior/deeplabcut/deeplabcutdatainterface.py
pauladkisson: Ah, no, this should be fine actually, my bad. I was conflating keypoints with nodes (by default they are the same), but actually they are distinct.

pauladkisson: Yeah after more consideration, I think this is fine to expose.

src/neuroconv/datainterfaces/behavior/deeplabcut/_dlc_utils.py
pauladkisson: Ah, I see. I still think it shouldn't be a big deal to allow this, but it can wait until we hear from the group.

pauladkisson (approved):
Thanks for all the hard work @h-mayorquin!

PR #1322, opened 2025-05-07
Updated to sleap-io 0.3.0

pauladkisson:
Depends on https://github.com/rly/ndx-pose/pull/41


h-mayorquin (approved):
Great! Thanks a bunch for taking care of this!

h-mayorquin:
I added a changelog to let the users know that now we need ndx-pose larger than 0.2.

h-mayorquin:
Damn, some tests failing : /

pauladkisson:
> Damn, some tests failing : /

The extra processing module that was causing the tests to fail was 'behavior', so I removed those problematic assertions.

PR #1324, opened 2025-05-07
updated sleap version

pauladkisson:
Until https://github.com/rly/ndx-pose/issues/40 gets fixed, sleap must be pinned to the older version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -114,7 +114,7 @@ audio = [
 ]
 sleap = [
     "av>=10.0.0",
-    "sleap-io>=0.0.2",
+    "sleap-io>=0.0.2,<0.3.0",
 ]
 deeplabcut = [
     "ndx-pose>=0.2",


pauladkisson:
> One downside of this is that full (the full dependency in pyproject) would not be able to resolve.

What do you mean? Full resolves no problem. This just prevents sleap-io from installing the latest version that they just released yesterday. Anyway, it's just a stopgap so that the tests pass and don't hold up other projects. I expect  https://github.com/rly/ndx-pose/issues/40 to be fixed in short order.

h-mayorquin:
Ah, you are right, we don't have ndx-pose-estimation specified for sleap-io. I stand corrected.

h-mayorquin (approved):
LGTM.

Thanks, Paul.

talmo:
Agh, sorry guys!! Please let me know if there's anything we can do on the `sleap-io` side!

PR #1327, opened 2025-05-08
Fix AudioInterface for multi-dot filenames

### LINKED ISSUE #1326: [Bug]: AudioInterface doesn't allow multiple dots in filename
### What happened?

When a WAV file has a name like "252_4.1.wav" (where .1 indicates trial number)
the current implementation treats '.1' as an invalid audio format, see:
https://github.com/catalystneuro/neuroconv/blob/8bcb0806ad2767ba6834973cdb15dc064acfdba2/src/neuroconv/datainterfaces/behavior/audio/audiointerface.py#L54-L57

Can we just check the final extension to be wav instead of all suffixes?

### Steps to Reproduce

```python
from neuroconv.datainterfaces import AudioInterface

# some file with extra dots
wav_file = "recording.1.wav"
AudioInterface(wav_file)
```

### Traceback

```python
  File "/Users/weian/catalystneuro/kind-lab-to-nwb/src/kind_lab_to_nwb/rat_behavioural_phenotyping_2025/prey_capture/convert_session.py", line 212, in <module>
    session_to_nwb(
    ~~~~~~~~~~~~~~^
        output_dir_path=output_dir_path,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
        overwrite=overwrite,
        ^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/weian/catalystneuro/kind-lab-to-nwb/src/kind_lab_to_nwb/rat_behavioural_phenotyping_2025/prey_capture/convert_session.py", line 114, in session_to_nwb
    audio_interface = AudioInterface(file_paths=[usv_file_path])
  File "/Users/weian/anaconda3/envs/kind_lab_to_nwb_env/lib/python3.13/site-packages/pydantic/validate_call_decorator.py", line 60, in wrapper_function
    return validate_call_wrapper(*args, **kwargs)
  File "/Users/weian/anaconda3/envs/kind_lab_to_nwb_env/lib/python3.13/site-packages/pydantic/_internal/_validate_call.py", line 96, in __call__
    res = self.__pydantic_validator__.validate_python(pydantic_core.ArgsKwargs(args, kwargs))
  File "/Users/weian/anaconda3/envs/kind_lab_to_nwb_env/lib/python3.13/site-packages/neuroconv/datainterfaces/behavior/audio/audiointerface.py", line 59, in __init__
    raise ValueError(
    ...<2 lines>...
    )
ValueError: The currently supported file format for audio is WAV file. Some of the provided files does not match this format: ['.1'].
```

### Operating System

macOS

### Python Executable

Conda

### Python Version

3.12

### Package Versions

_No response_

### Code of Conduct

weiglszonja:
Fix #1326

### DIFF
--- a/tests/test_behavior/test_audio_interface.py
+++ b/tests/test_behavior/test_audio_interface.py
@@ -11,6 +11,7 @@ from dateutil.tz import gettz
 from numpy.testing import assert_array_equal
 from pydantic import FilePath
 from pynwb import NWBHDF5IO
+from pynwb.testing.mock.file import mock_NWBFile
 from scipy.io.wavfile import read, write
 
 from neuroconv import NWBConverter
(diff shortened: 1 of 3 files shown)


tests/test_behavior/test_audio_interface.py
```diff
@@ -258,3 +255,32 @@ class AudioTestNWBConverter(NWBConverter):
...
+
+            # Initialize a fresh AudioInterface with the file
+            interface = AudioInterface(file_paths=[file_path])
+            # Get metadata
+            metadata = interface.get_metadata()
+            metadata["NWBFile"].update(session_start_time=self.session_start_time)
+
+            # Run conversion
+            nwbfile_path = str(self.test_dir / "audio_test.1.nwb")
+            interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata, overwrite=True)
```
h-mayorquin, line 274: Can we just `add_to_nwbfile` to check that the error is not triggered so we don't write to file in this case? Just to make the tests faster saving somet time.

h-mayorquin (commented):
This looks good to me, it only needs a changelog. I would like to avoid writing to disk when is not necessary to keep our tests faster but this is a nitpick if you don't want to change it is OK.

h-mayorquin (approved):
LGTM.

PR #1342, opened 2025-05-14
Dallies modification so readme badge only displays as failure with critical tests

h-mayorquin:
Draft


h-mayorquin:
Ok, closing this in favor of #1344 as that is easier.

PR #1343, opened 2025-05-14
Testing: separate dailies and dev-dailies

h-mayorquin:
I think this is the simplest option. The other option in #1342 works but is convoluted (you need to define inputs to the called workflows so they continue on fail but also return the true state as the output. I think this is simpler.

### DIFF
--- a/.github/workflows/dailies.yml
+++ b/.github/workflows/dailies.yml
@@ -18,12 +18,6 @@ jobs:
       - id: load_os_versions
         run: echo "os_versions=$(cat ./.github/workflows/all_os_versions.txt)" >> "$GITHUB_OUTPUT"
 
-  build-and-upload-docker-image-dev:
-    uses: ./.github/workflows/build_and_upload_docker_image_dev.yml
-    secrets:
-      DOCKER_UPLOADER_USERNAME: ${{ secrets.DOCKER_UPLOADER_USERNAME }}
-      DOCKER_UPLOADER_PASSWORD: ${{ secrets.DOCKER_UPLOADER_PASSWORD }}
-
   run-daily-tests:
     needs: load_python_and_os_versions
     uses: ./.github/workflows/testing.yml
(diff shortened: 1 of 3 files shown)


h-mayorquin:
This will fix #1345

.github/workflows/dailies.yml
```diff
@@ -36,62 +30,6 @@ jobs:
...
-    with:
-      python-versions: ${{ needs.load_python_and_os_versions.outputs.ALL_PYTHON_VERSIONS }}
-      os-versions: ${{ needs.load_python_and_os_versions.outputs.ALL_OS_VERSIONS }}
-
-  run-daily-neuroconv-docker-testing:
-    uses: ./.github/workflows/neuroconv_docker_testing.yml
-    secrets:
-      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-      S3_GIN_BUCKET: ${{ secrets.S3_GIN_BUCKET }}
```
pauladkisson, line 67: This one should stay in the main dailies workflow since it just tests the latest release of neuroconv with docker.

pauladkisson (changes requested):
I agree that this is probably the best solution.

While I was looking at the badges, I noticed that they still say we support python 3.9, but we dropped it. Can you remove that as well?

.github/workflows/dailies.yml
h-mayorquin: I reverted but I does not look like it tests a release version.

h-mayorquin:
@pauladkisson the problem of python versions will be fixed when we release. The last release still supported python 3.9 and that is why is still there.

pauladkisson:
Caught a typo otherwise looks good.

.github/workflows/dailies.yml
```diff
@@ -154,36 +68,4 @@ jobs:
...
-        with:
-          server_address: smtp.gmail.com
-          server_port: 465
-          username: ${{ secrets.MAIL_USERNAME }}
-          password: ${{ secrets.MAIL_PASSWORD }}
-          subject: NeuroConv Daily Doc Link Check Failure
-          to: ${{ secrets.DAILY_FAILURE_EMAIL_LIST }}
-          from: NeuroConv
-          body: "The daily check for working links in the documentation failed, please check status at https://github.com/catalystneuro/neuroconv/actions/workflows/dailies.yml"
+          body: "The daily neuroconv docker test workflow failed, please check status at https://github.com/catalystneuro/neuroconv/actions/workflows/dev-dailies.yml"
```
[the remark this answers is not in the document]
h-mayorquin: Thanks.

pauladkisson (approved):

PR #1344, opened 2025-05-14
Add installation instructions for the read the docs

h-mayorquin:
It was noted that we don't have installation instruction on the read the docs and that many people are not familiar with github or know that there is usually a github repo attached to the read the docs.

Adding this info.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,7 @@ docs = [
     "pytest" # used to build the documentation for tools.testing
 ]
 
-dandi = ["dandi>=0.66.5", "dandischema>=0.11"]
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -19,6 +19,19 @@ Features:
 * Minimizes the size of the NWB files by automatically applying chunking and lossless compression.
 * Supports ensembles of multiple data streams, and supports common methods for temporal alignment of streams.
 
+Installation
+-----------
```
zm711, line 23: While I'm here you're missing a few :P

alessandratrapani (approved):
Look good!

Issue #1347, opened 2025-05-16
Ensure conversion gallery is properly tested


h-mayorquin:
We've encountered errors where the installation instructions for the conversion gallery fail to work properly. The root issue is that these instructions require more dependencies than are currently declared in the corresponding extras section of the `pyproject.toml`.

Here are some examples:

https://github.com/catalystneuro/neuroconv/issues/1304

https://github.com/catalystneuro/neuroconv/issues/1315#issuecomment-2879466456

https://github.com/catalystneuro/neuroconv/issues/1352

A failure of testing is that we currently test the conversion gallery only within the general workflow, which uses an environment that already includes many dependencies. This makes it a poor test case for verifying the minimal installation requirements.

Additionally, the specific test we do have for this is outdated:

https://github.com/catalystneuro/neuroconv/blob/7385018613e0c395e2f77c097f7a21ee23d9bba5/.github/workflows/formatwise-installation-testing.yml#L17-L18

and disabled.

That testing extra also includes `spikeinterface[qualitymetrics]' as part of its dependencies:

https://github.com/catalystneuro/neuroconv/blob/7385018613e0c395e2f77c097f7a21ee23d9bba5/.github/workflows/formatwise-installation-testing.yml#L33-L34

which smuggles some unwanted dependencies and gets us farther from the right context for testing.

https://github.com/catalystneuro/neuroconv/blob/7385018613e0c395e2f77c097f7a21ee23d9bba5/pyproject.toml#L68-L78

I propose the following steps:

1. Update the gallery testing workflow to be current and to run in a clean environment. [PR](https://github.com/catalystneuro/neuroconv/pull/1357)
2. Ensure that the testing extras do not include any unintended dependencies [PR](https://github.com/catalystneuro/neuroconv/pull/1380)
3. Run the conversion gallery tests before every release, possibly by including them in the release action workflow.

PR #1357, opened 2025-05-17
Clean test dependencies

h-mayorquin:
Moving forward to solve #1347. Also this was a request from Alessio as this extra will be removed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,6 @@ test = [
     "ndx-events==0.2.1",  # for special tests to ensure load_namespaces is set to allow NWBFile load at all times
     "parameterized>=0.8.1",
     "ndx-miniscope",
-    "spikeinterface[qualitymetrics]>=0.102.2",
     "zarr",
     "pytest-xdist"
 ]


pyproject.toml
```diff
@@ -70,10 +69,14 @@ test = [
     "pytest",
     "pytest-cov",
     "parameterized>=0.8.1",
-    "ndx-miniscope",
-    "spikeinterface[qualitymetrics]>=0.102.2",
-    "zarr",
-    "pytest-xdist"
+    "pytest-xdist"  # Runs tests on parallel
+]
+test_ecephys = [
```
pauladkisson, line 74: Why do we need these test_ecephys and test_ophys environments can't we just use ecephys and ophys?

h-mayorquin: Two reasons:

* Full ecephys and ophys might become too large (for example, if an extractor depends on a large library like PyTorch). I want to avoid installing unnecessary dependencies just for testing.
* Conversely, we might eventually need test dependencies that aren't used by any extractor. We shouldn't add these to the full ecephys package, as users might not need them.

h-mayorquin:
> Does this PR fix #1347? Or is that going to be a follow-up?

We will need a follow up.

pyproject.toml
pauladkisson: It just seems like unnecessary complexity to me (which will then add to maintenance burden). We don't currently need any ecephys testing dependencies that aren't covered by ecephys and test. Until we do, I suggest just using the normal environments. What do you think?

h-mayorquin: > We don't currently need any ecephys testing dependencies that aren't covered by ecephys and test (reason 2)

Scipy is needed for testing ecephys but none of the extractors includes it.

pauladkisson: I see. Because the sortingAnalyzer requires spikeinterface[qualitymetrics] which includes scipy is that right?

pauladkisson: But the only code that calls tools.spikeinterface.write_sorting_analyzer is the tests...

pauladkisson: The fact that scipy isn't a strictly test dependency but rather a dependency of something that gets tested seems like a bit of a code smell to me.

pauladkisson: For example, if a user tried to install neuroconv with the ecephys dependencies (`pip install "neuroconv[ecephys]"`) and then tried to use `neuroconv.tools.spikeinterface. write_sorting_analyzer`, it would fail bc scipy isn't installed.

In fact, if audio, caiman, and CNMFE all dropped scipy in a future release, then `neuroconv.tools.spikeinterface. write_sorting_analyzer` wouldn't work even in the full dependency environment (`pip install "neuroconv[full]"`).

I guess the crux of the problem is that tools.spikeinterface doesn't have a dedicated sub-dependency, but I would think it should be covered by ecephys.

h-mayorquin: > I see. Because the sortingAnalyzer requires spikeinterface[qualitymetrics] which includes scipy is that right?

Yes.
Right now, and at various points in the past, it has required Numba and scikit-learn (and it might require them again). The fact is, the sorting_analyzer as currently implemented is something of a black box. We don’t have much control over it, since it’s called here:

https://github.com/catalystneuro/neuroconv/blob/042a17dc3def4df33e8f227e01622b8a1d0cfec8/src/neuroconv/tools/spikeinterface/spikeinterface.py#L1919-L1930

and therefore brings along whatever dependencies it needs.

> The fact that scipy isn't a strictly test dependency but rather a dependency of something that gets tested

They’re the same to me, but maybe you can come up with a better name?

This block should include:

1. All dependencies needed to test the corresponding non-data modality tests.
2. It should not include all the extractor dependencies.

At the moment, that means scipy for ecephys and the fiber photometry extension for ophys.

As a side note, all of these should ideally be organized as dependency groups:
https://packaging.python.org/en/latest/specifications/dependency-groups/
and not as extras as we currently do now but we can update that later.

pauladkisson: > > The fact that scipy isn't a strictly test dependency but rather a dependency of something that gets tested
> 
> They’re the same to me, but maybe you can come up with a better name?
> 

They are not the same. Let me illustrate with some example code (courtesy of claude 4 sonnet 🚀):
```python
# functions.py

def f1(x, y):
    """Function that imports and uses scipy"""
    from scipy.stats import pearsonr
    correlation, _ = pearsonr(x, y)
    return correlation

def f2(x, y):
    """Function that does calculation without scipy"""
    n = len(x)
    mean_x = sum(x) / n
    mean_y = sum(y) / n
    
    num = sum((x[i] - mean_x) * (y[i] - mean_y) for i in range(n))
    den = (sum((x[i] - mean_x)**2 for i in range(n)) * 
           sum((y[i] - mean_y)**2 for i in range(n)))**0.5
    
    return num / den if den != 0 else 0


# test_functions.py

def test_f1():
    """Test f1 - scipy is imported inside f1"""
    x = [1, 2, 3, 4, 5]
    y = [2, 4, 6, 8, 10]
    result = f1(x, y)
    assert abs(result - 1.0) < 0.001

def test_f2():
    """Test f2 - scipy is imported in the test to verify f2's result"""
    from scipy.stats import pearsonr
    
    x = [1, 2, 3, 4, 5]
    y = [2, 4, 6, 8, 10]
    
    # Test our f2 implementation against scipy
    our_result = f2(x, y)
    scipy_result, _ = pearsonr(x, y)
    
    assert abs(our_result - scipy_result) < 0.001
```

Our situation is like f1, where the function that is exposed to the user is importing scipy, and then we test it. So, we really ought to have a user-exposed environment that works with tools.spikeinterface, not just a test environment. How about we add a sub-environment called spikeinterface or something that includes spikeinterface[qualitymetrics] so that SortingAnalyzer will remain covered?

> As a side note, all of these should ideally be organized as dependency groups: https://packaging.python.org/en/latest/specifications/dependency-groups/ and not as extras as we currently do now but we can update that later.

👍

h-mayorquin: > For example, if a user tried to install neuroconv with the ecephys dependencies (pip install "neuroconv[ecephys]") and then tried to use neuroconv.tools.spikeinterface. write_sorting_analyzer, it would fail bc scipy isn't installed.

>In fact, if audio, caiman, and CNMFE all dropped scipy in a future release, then neuroconv.tools.spikeinterface. write_sorting_analyzer wouldn't work even in the full dependency environment (pip install "neuroconv[full]").

Yes and that is OK to me. We are a conversion library and the dependencies we have and maintain are for doing conversions. 

> I guess the crux of the problem is that tools.spikeinterface doesn't have a dedicated sub-dependency, but I would think it should be covered by ecephys.

We could include a group called "sorting_analyzer" if you want to maintain it but I don't want to include that in ecephys because I think is out of scope:

* We don't have conversions projects that use it and is not used under the hood by any interface.
* We don't have correctness tests for it because of the black box behavior discussed in the previous comment.
* The maintenance is done by Alessio who adds features and fixes problems when he needs it.

The sorting analyzer in the tools as we currently have is a convenience that we have for Alessio. If that changes, then more though should be put into testing, dependency management ,etc.

pauladkisson: In that case, why does it live in NeuroConv instead of Spike Interface? @alejoe91

h-mayorquin: > They are not the same. Let me illustrate with some example code (

My bad, I understood what you meant without the example. I am saying that I don't find the distinction very useful. 

I am fine with a better name but `spikeinterface[qualitymetrics]` will be removed soon so we need to track that without that handle.

I would not like to expose that to the users for the reasons exposed in the previous response but there is no way of doing that without using dependency groups and I don't want to do the migration right now.

h-mayorquin: > In that case, why does it live in NeuroConv instead of Spike Interface? @alejoe91 

I don't want to derail this discussion.

Can you open an issue for that or bring it up in a meeting if you want to move it out (I disagree with doing it).

h-mayorquin: Great.

`sorting_analyzer` with only scipy works at the moment. We can add more stuff if they change it on spikeinterface. 

What about the ophys one? Do you also want a different name for the extra that includes only the fiber photometry extension?

pauladkisson (approved):
After discussion, this looks good! Thanks for taking the time to meet @h-mayorquin!

.github/workflows/testing.yml
```diff
@@ -71,29 +68,84 @@ jobs:
...
+          python -m venv env_ophys
+          # Activate ophys environment and install requirements
+          if [ "${{ runner.os }}" == "Windows" ]; then
+            . env_ophys/Scripts/activate
+          else
+            source env_ophys/bin/activate
+          fi
+          python -m pip install -U pip
+          python -m pip install ".[test,ophys_minimal]"
+          python -m pytest tests/test_ophys -vv -rsx -n auto --dist loadscope
```
pauladkisson, line 104:
```suggestion
          python -m pytest tests/test_modalities/test_ophys -vv -rsx -n auto --dist loadscope
```

h-mayorquin: Ah, thanks, just did.

PR #1321, opened 2025-05-07
Add Inscopix imaging and segmentation interfaces and tests 

daphnedequatrebarbes:
Fix #707, Replacing #714 and #856 : 

- [x] pyproject.toml
- [x] Inscopix Imaging
- [ ] Inscopix Segmentation 
- [ ] tests
- [x] update docs 
- [x] update ChangeLog

[ isx 2.0](https://pypi.org/project/isx/) is currently not supported on Python 3.13, which is causing all run-tests-latest-3.13 jobs to fail. Additionally, isx is not supported on macOS, leading to failures in the run-tests-macos jobs.
Not sure what the best way to handle this is. 

Fix segmentation : the conversion process tries to get ROI locations using integer indices (0, 1, 2...) but the Inscopix extractor returns string ROI IDs ('C0', 'C1', 'C2'...).

PR #856, opened 2024-05-13
add inscopix imaging extractor and test

bendichter:
to do:
* [x] conversion gallery
* [ ] better metadata in tests

PR #714, opened 2024-01-24
add inscopixdatainterface.py

bendichter:
fix #707

- [ ] tests
- [ ] gallery
- [ ] changelog

PR #1361, opened 2025-05-20
Inscopix Imaging Interface 

### LINKED ISSUE #707: [New Format]: Inscopix

Inscopix finally releases an open API

Not as far as I can tell, but could be inferred from some of the source code at https://github.com/inscopix/py_isx/blob/main/isx/__init__.py

### Existing APIs for format

https://github.com/inscopix/py_isx

Test files found with releases: https://github.com/inscopix/py_isx/releases

Yes.

### Code of Conduct

daphnedequatrebarbes:
Fix #707,  Replacing #714 and #856 : 
Clean version of #1321 

Inscopix Imaging Extractor : https://github.com/catalystneuro/roiextractors/pull/436

Inscopix Segmentation Interface at #1364


src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
```diff
@@ -0,0 +1,70 @@
...
+        -------
+        dict
+            Dictionary containing metadata including device information and imaging details
+            specific to the Scanbox system.
+        """
+        metadata = super().get_metadata()
+        metadata["Ophys"]["Device"][0]["description"] = "Inscopix imaging"
+        return metadata
+
+    def add_to_nwbfile(self, nwbfile, metadata=None, **kwargs):
```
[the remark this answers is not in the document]
daphnedequatrebarbes: Definitely will take it out, I was checking that modifying metadata through the Interface worked thx

daphnedequatrebarbes:
## Justification for Skipping Segmentation Tests with Empty Dataset

When testing `InscopixSegmentationInterface` with an empty dataset (no ROIs), certain tests are skipped because they fundamentally require operations that cannot succeed with empty ROI sets:

1. **`test_run_conversion_with_backend_configuration`**  
   During conversion, `get_roi_pixel_masks()`/`get_roi_image_masks()` attempt to stack empty arrays, triggering `ValueError: need at least one array to stack`.

2. **`test_configure_backend_for_equivalent_nwbfiles`**  
   Creating NWB files for comparison requires adding ROI data, which fails with the same empty mask errors.

3. **`test_all_conversion_checks`**  
   This comprehensive test calls all the problematic functions in sequence, including `add_to_nwbfile`.

I skipped these tests because they're designed to validate successful conversions, not to handle the expected errors from empty datasets that we test elsewhere.

h-mayorquin:
So the segmentation interface will need changes on roiextractors.

Can we divide this PR into one for imaging extraction and one for segmentation extraction? It will be easy to review this way and we don't need to wait for the fixes in roiextractors.

What do you think?

daphnedequatrebarbes:
Splitting it up makes sense—I’ll separate the PRs accordingly. Thanks for the suggestion!

tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -41,6 +43,15 @@
 except ImportError:
     from setup_paths import OPHYS_DATA_PATH, OUTPUT_PATH
 
+skip_on_darwin_arm64 = pytest.mark.skipif(
+    platform.system() == "Darwin" and platform.machine() == "arm64",
+    reason="Tests are skipped on macOS ARM64 due to platform limitations.",
+)
+skip_if_isx_not_installed = pytest.mark.skipif(
```
h-mayorquin, line 50: It is OK if the tests fail because we don't the import. We should check for that.

tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -41,6 +44,19 @@
...
 
+skip_on_darwin_arm64 = pytest.mark.skipif(
+    platform.system() == "Darwin" and platform.machine() == "arm64",
+    reason="Tests are skipped on macOS ARM64 because of incompatibility with the 'isx' module",
+)
+skip_if_isx_not_installed = pytest.mark.skipif(
+    not importlib.util.find_spec("isx"),
+    reason="Tests are skipped because the 'isx' module is not installed.",
+)
+skip_on_python_313 = pytest.mark.skipif(
```
h-mayorquin, line 55: You can add those test in the top of the test that is skipped just to keep content closer to use.

Also, can you reference the issue to keep track:

https://github.com/inscopix/pyisx/issues

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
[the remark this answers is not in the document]
h-mayorquin: You still need to remove this.

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
```diff
@@ -0,0 +1,57 @@
...
+    display_name = "Inscopix Imaging"
+    associated_suffixes = (".isxd",)
+    info = "Interface for handling Inscopix imaging data."
+
+    @validate_call
+    def __init__(
+        self,
+        file_path: FilePath,
+        verbose: bool = False,
+        photon_series_type: Literal["OnePhotonSeries", "TwoPhotonSeries"] = "OnePhotonSeries",
```
h-mayorquin, line 20: This is a conversion that will be inhertied from the add_to_nwbfile method. You don't need it at initialization:

https://github.com/catalystneuro/neuroconv/blob/5b6175415f35cbd47f4816d5f702649a25a0e0bb/src/neuroconv/datainterfaces/ophys/baseimagingextractorinterface.py#L140-L152

tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -903,3 +905,391 @@ def test_incorrect_folder_structure_raises(self):
...
+            imaging_plane=cls.imaging_plane_name,
+        )
+
+        cls.ophys_metadata = dict(
+            Device=[cls.device_metadata],
+            ImagingPlane=[cls.imaging_plane_metadata],
+            OnePhotonSeries=[cls.photon_series_metadata],
+        )
+
+    def check_extracted_metadata(self, metadata: dict):
```
h-mayorquin, line 969: Unless we define a method that in the interface that is called `get_metadata()` which we should it does not makes sense to override the parent methods here. This conversion is generic and the generic inherited methods should suffice.

h-mayorquin (commented):
I took the first look. 

Big question: is there any metadata that we can extract in this format from the source?

Examples:
* The session start time (the time at which recording took place)
* The specific name of the device.
* The size of the field of view 
* Any other ophys metadata?

If so, we should implement a `get_metadata` method that does this.

daphnedequatrebarbes:
> I took the first look.
> 
> Big question: is there any metadata that we can extract in this format from the source?
> 
> Examples:
> 
> * The session start time (the time at which recording took place)
> * The specific name of the device.
> * The size of the field of view
> * Any other ophys metadata?
> 
> If so, we should implement a `get_metadata` method that does this.
;

Good point—I was focused on fixing smaller issues and overlooked that part. There is Inscopix-specific metadata available, so I’ll work on adding a get_metadata method for that.

h-mayorquin:
Thanks, let me know when is read for me to do a second reading.

tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -903,3 +905,190 @@ def test_incorrect_folder_structure_raises(self):
...
+
+        # OnePhotonSeries checks
+        ops = metadata["Ophys"]["OnePhotonSeries"][0]
+        assert ops["name"] == "OnePhotonSeries"
+        assert ops["description"] == "Imaging data from one-photon excitation microscopy."
+        assert ops["unit"] == "n.a."
+        assert ops["imaging_plane"] == "ImagingPlane"
+        assert ops["dimension"] == [128, 128]
+
+        # Subject checks
```
h-mayorquin, line 969: In this case, when there is no real metadata about the subject maybe we should not add a Subject at all? What do you think?

tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -903,3 +905,190 @@ def test_incorrect_folder_structure_raises(self):
...
+    )
+    optical_series_name = "OnePhotonSeries"
+
+    def check_extracted_metadata(self, metadata: dict):
+        """Test metadata extraction for file with rich acquisition info."""
+
+        # NWBFile checks
+        nwbfile = metadata["NWBFile"]
+        assert isinstance(nwbfile["session_start_time"], str)
+        assert nwbfile["session_start_time"].startswith("2019-10-07T16:22:01")
```
h-mayorquin, line 997: You can create a datetime object and compare to them here.

tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -903,3 +905,190 @@ def test_incorrect_folder_structure_raises(self):
...
+        assert nwbfile["session_id"] == "4D_SAAV_PFC_IM7_20191007"
+
+        # Device checks
+        device = metadata["Ophys"]["Device"][0]
```
h-mayorquin, line 1001: can you check device_name here?

h-mayorquin (commented):
Looks very good, first pass.

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
```diff
@@ -0,0 +1,195 @@
...
+
+        # Update or create Subject metadata
+        if "Subject" in metadata and isinstance(metadata["Subject"], dict):
+            metadata["Subject"].update(subject_metadata)
+        else:
+            metadata["Subject"] = subject_metadata
+
+        return metadata
+
+    def add_to_nwbfile(
```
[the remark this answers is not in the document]
daphnedequatrebarbes: I realized just now that there is a nVista 2P as well (https://inscopix.com/nvista-2p-system/), should I find a way to handle both microscopes ?

tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -903,3 +905,190 @@ def test_incorrect_folder_structure_raises(self):
...
+        assert device["name"] == "Microscope"
+        assert "description" not in device or device.get("description", "") == ""
+
+        # ImagingPlane checks
+        imaging_plane = metadata["Ophys"]["ImagingPlane"][0]
+        assert imaging_plane["name"] == "ImagingPlane"
+        assert imaging_plane["device"] == "Microscope"
```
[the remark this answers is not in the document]
daphnedequatrebarbes: Yes most of the metadata is missing in most of the test files, movie longer than 3min is the most complete one.

tests/test_on_data/ophys/test_imaging_interfaces.py
daphnedequatrebarbes: Yes, I updated it to include the `Subject` only when subject information is detected.

docs/conversion_examples_gallery/imaging/inscopiximaging.rst
```diff
@@ -0,0 +1,29 @@
...
+    >>> from zoneinfo import ZoneInfo
+    >>> from pathlib import Path
+    >>> from neuroconv.datainterfaces import InscopixImagingInterface
+    >>>
+    >>> file_path = OPHYS_DATA_PATH / "imaging_datasets" / "inscopix" / "movie_128x128x100_part1.isxd"
+    >>> interface = InscopixImagingInterface(file_path=file_path, verbose=False)
+    >>>
+    >>> metadata = interface.get_metadata()
+    >>> # For data provenance we add the time zone information to the conversion
+    >>> session_start_time = datetime(1970, 1, 1, 0, 0, 0, tzinfo=ZoneInfo("US/Pacific"))
```
h-mayorquin, line 24: So, this interface gets the session start time in the metadata so no need to set it here. Check the spikeglx entry in the gallery for an example of an interface that does the same.

tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -903,3 +905,172 @@ def test_incorrect_folder_structure_raises(self):
...
+@skip_on_darwin_arm64
+class TestInscopixImagingInterfaceMovie128x128x100Part1(ImagingExtractorInterfaceTestMixin):
+    """Test InscopixImagingInterface with movie_128x128x100_part1.isxd (minimal metadata file)."""
+
+    data_interface_cls = InscopixImagingInterface
+    save_directory = OUTPUT_PATH
+    interface_kwargs = dict(
+        file_path=str(OPHYS_DATA_PATH / "imaging_datasets" / "inscopix" / "movie_128x128x100_part1.isxd")
+    )
+    optical_series_name = "OnePhotonSeries"
```
h-mayorquin, line 934: You have to write
```suggestion
     conversion_options = dict(optical_series_name="OnePhotonSeries") 
```

h-mayorquin: Probably has typos.

h-mayorquin: The only error that I see on the tests is related to the undefined datetime, where you are seeing this error? can you push so we can see the error on the CI?

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
```diff
@@ -0,0 +1,195 @@
...
+        ----------
+        file_path : str
+            Path to the .isxd Inscopix file.
+        verbose : bool, optional
+            If True, outputs additional information during processing.
+        """
+        super().__init__(
+            file_path=file_path,
+            verbose=verbose,
+            photon_series_type="OnePhotonSeries",
```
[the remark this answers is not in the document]
daphnedequatrebarbes: Also I believe that the test_metadata() fails because the interface was created without photon_series_type="OnePhotonSeries" in the initialization, so when interface.get_metadata() is called, it generates metadata structure for TwoPhotonSeries (the default) instead of OnePhotonSeries, and conversion_options cannot help here because they are only used during conversion while metadata generation happens before conversion and uses the interface's internal state from initialization. When the test tries to access metadata["Ophys"]["OnePhotonSeries"][0] expecting it to contain OnePhotonSeries metadata, it finds an empty dictionary instead, causing the assertion assert ops["name"] == "OnePhotonSeries" to fail with AssertionError: assert DeepDict({}) == 'OnePhotonSeries'.

To fix this, I am thinking of keeping photon_series_type parameter to the interface's __init__ method  and pass photon_series_type="OnePhotonSeries" in the test's interface_kwargs to configure the interface from creation.

tests/test_on_data/ophys/test_imaging_interfaces.py
h-mayorquin: You need to change the branc in pyproject to main now that your PR was merged.

docs/conversion_examples_gallery/imaging/inscopiximaging.rst
h-mayorquin: You marked this as resolved but did not implement or discussed the requested change. Why is that?

daphnedequatrebarbes: sorry, forgot to push the modif

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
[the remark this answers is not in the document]
daphnedequatrebarbes: I believe the issue comes down to timing in the test workflow. When tests run, the interface first generates metadata (deciding whether to create OnePhotonSeries or TwoPhotonSeries structures), and then later does the actual conversion. Since we're not specifying the photon series type during initialization, the interface defaults to TwoPhotonSeries metadata with the base class, but our tests are looking for OnePhotonSeries metadata - so they fail immediately when they can't find the expected structure. The key problem is that conversion_options can't help here because they only apply during the conversion step, but the test is already failing during the earlier metadata generation step. What would be the best approach to handle this - should I add the photon series type to initialization, modify how metadata generation works, or is there another way to structure this?

daphnedequatrebarbes:
### The Broken Pipeline Flow
• **Step 1 - Interface Creation:** `InscopixImagingInterface(file_path="...", photon_series_type="OnePhotonSeries")`
• **Step 2 - Metadata Generation:** Called `get_metadata()` → Created metadata structure expecting "OnePhotonSeries"
• **Step 3 - Conversion Process:** Called `run_conversion()` **WITHOUT** `photon_series_type` parameter → Defaulted to "TwoPhotonSeries"
• **Step 4 - Test Failure:** Test looked for "OnePhotonSeries" in NWB file → Found "TwoPhotonSeries" instead → `KeyError`

### Parameter Passing Failures
• **Missing Constructor Parameter:** Child class wasn't passing `photon_series_type` to parent constructor
• **Parent Initialization Error:** BaseImagingExtractorInterface initialized with default "TwoPhotonSeries" instead of intended "OnePhotonSeries"
• **Test Framework Gap:** `ImagingExtractorInterfaceTestMixin.check_nwbfile_temporal_alignment()` method ignored `conversion_options`
• **Conversion Default Override:** Interface-specific parameters like `photon_series_type` were lost during test conversion process

## Solution Implementation - Two-Part Fix

### Fix 1: Constructor Parameter Correction
```python
super().__init__(
    file_path=file_path,
    verbose=verbose,
    photon_series_type=photon_series_type,  # ← Now passed to parent
)
```

• Prevents Parent from defaulting to "TwoPhotonSeries" during metadata generation
• Parameter available during parent's initialization phase, not set after the fact

### Fix 2: Test Framework Gap Resolution
**File:** `src/neuroconv/tools/testing/data_interface_mixins.py`
**Method:** `ImagingExtractorInterfaceTestMixin.check_nwbfile_temporal_alignment()`
```python
# Use conversion_options if available
conversion_options = getattr(self, 'conversion_options', {})  # ← Defensive programming

interface.run_conversion(
    nwbfile_path=nwbfile_path, 
    overwrite=True, 
    metadata=metadata,
    **conversion_options  # ← KEY FIX: Ensures photon_series_type is passed
)
```
• **Framework Impact:** Fixes the gap for all interfaces using conversion options, not just Inscopix

### Fix 3: Test Class 
```python
interface_kwargs = dict(
    file_path=str(OPHYS_DATA_PATH / "imaging_datasets" / "inscopix" / "movie_128x128x100_part1.isxd"),
    photon_series_type="OnePhotonSeries",  # ← Interface initialization
)
optical_series_name = "OnePhotonSeries"     # ← Test expectation alignment
conversion_options = dict(photon_series_type="OnePhotonSeries")  # ← Conversion specification 
```
• Photon series type specified at interface creation, test expectations, AND conversion process

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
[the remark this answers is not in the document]
daphnedequatrebarbes: The only way I currently see to determine whether the data comes from a one- or two-photon series is if the microscope model is explicitly defined in the metadata.

From the stub NWB files available, only one include a microscope entry—one in the segmentation data and one in the imaging data. Both reference only "nVista" systems, and there is no example that explicitly mentions "nVista-2P", so it's unclear how it would be written (e.g., "NVista-2P", "NVista 2P", or "NVista2P").

Given that, if the microscope is not specified what should I default to ? I believe I cannot let the Photon Series empty in NWB.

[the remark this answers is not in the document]
daphnedequatrebarbes: I think it makes more sense to put the add_to_nwbfile workaround, what do you think ?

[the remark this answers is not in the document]
h-mayorquin: Yes, given the information you wrote above I agree that the `add_to_nwbfile` workaround is better.

tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -903,3 +905,201 @@ def test_incorrect_folder_structure_raises(self):
...
+        # NWBFile checks
+        nwbfile = metadata["NWBFile"]
+        assert nwbfile["session_start_time"] == datetime(2019, 10, 7, 16, 22, 1, 524186)
+        assert nwbfile["session_id"] == "4D_SAAV_PFC_IM7_20191007"
+
+        # Device checks
+        device = metadata["Ophys"]["Device"][0]
+        assert device["name"] == "NVista3"
+        device_desc = device.get("description", "")
+        if device_desc:
```
h-mayorquin, line 1006: Let's avoid the if here. If the description is available for this data, remove the if. If it is not, then just remove the assert.

I think this is important so we know exactly what we are testing for. Otherwise, we need to run the code to know that that information is there or know its contents already.

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
```diff
@@ -0,0 +1,194 @@
...
+                subject_metadata["date_of_birth"] = subject_info["date_of_birth"]
+                has_any_subject_data = True
+            if subject_info.get("weight") and subject_info["weight"] > 0:
+                subject_metadata["weight"] = str(subject_info["weight"])
+                has_any_subject_data = True
+
+        # Add Subject if we have ANY subject information, filling required fields with defaults
+        if has_any_subject_data:
+            if "subject_id" not in subject_metadata:
+                subject_metadata["subject_id"] = "Unknown"
```
h-mayorquin, line 163: Are those fields requiered in `Subject`? If they are not, we should avoid filling them.

We are a library centered in provenance. We should not make-up fields unless strictily necessary (the field is required)

Could you check?

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
```diff
@@ -0,0 +1,194 @@
...
+            else:
+                metadata["Subject"] = subject_metadata
+
+        return metadata
+
+    def add_to_nwbfile(
+        self,
+        nwbfile,
+        metadata: Optional[dict] = None,
+        photon_series_type: Literal["OnePhotonSeries", "TwoPhotonSeries"] = "OnePhotonSeries",
```
h-mayorquin, line 180: You should add kwargs here and proapgated them to the super().add_to_nwbfile(, **kwargs)
```suggestion
        **kwargs,
```

h-mayorquin: This way, the other kwargs from the add_to_nwbfile can be propagated.

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
```diff
@@ -0,0 +1,194 @@
...
+        # Add Subject if we have ANY subject information, filling required fields with defaults
+        if has_any_subject_data:
+            if "subject_id" not in subject_metadata:
+                subject_metadata["subject_id"] = "Unknown"
+            if "species" not in subject_metadata:
+                subject_metadata["species"] = "Unknown species"
+            if "sex" not in subject_metadata:
+                subject_metadata["sex"] = "U"
+
+            if "Subject" in metadata and isinstance(metadata["Subject"], dict):
```
h-mayorquin, line 169: the check for metadata["Subject] is instance dict here is unecessary I think. How are you thinking?

tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -903,3 +905,201 @@ def test_incorrect_folder_structure_raises(self):
...
+        assert subject["species"] == "Unknown species"
+        assert subject["sex"] == "M"
+        assert "description" not in subject
+        assert "weight" not in subject
+        assert "date_of_birth" not in subject
+
+
+@skip_on_python_313
+@skip_on_darwin_arm64
+class TestInscopixImagingInterfaceMovieU8(ImagingExtractorInterfaceTestMixin):
```
h-mayorquin, line 1053: I just realize, this one says in the docstring that is volumetric:

https://github.com/catalystneuro/roiextractors/blob/75afb5b4eedadf22a2a42cc22d8e5e2287a6e192/tests/test_inscopiximagingextractor.py#L111

But when you calculate `get_series` it seems that we are returning data as planar:

https://github.com/catalystneuro/roiextractors/blob/75afb5b4eedadf22a2a42cc22d8e5e2287a6e192/tests/test_inscopiximagingextractor.py#L136-L138

What is going on here?

pyproject.toml
```diff
@@ -304,6 +304,10 @@ hdf5 = [
 hdf5imaging = [
     "neuroconv[ophys_minimal]",
 ]
+inscopix = [
+    "isx>=2.0; (sys_platform != 'darwin' or platform_machine != 'arm64') and python_version < '3.13'",
+    "roiextractors>=0.6.0",
```
h-mayorquin, line 309:
```suggestion
    "neuroconv[ophys_minimal]",
```

src/neuroconv/tools/testing/data_interface_mixins.py
```diff
@@ -366,7 +366,11 @@ def check_nwbfile_temporal_alignment(self):
 
         metadata = interface.get_metadata()
         metadata["NWBFile"].update(session_start_time=datetime.now().astimezone())
-        interface.run_conversion(nwbfile_path=nwbfile_path, overwrite=True, metadata=metadata)
+
+        # Use conversion_options if available
+        conversion_options = getattr(self, "conversion_options", {})
```
h-mayorquin, line 371: I tested this locally and you don't really need it. I removed the conversion options here and in the interface.run_conversion below and things just work fine.

Why do we need this?

CHANGELOG.md
```diff
@@ -19,9 +19,11 @@
 * Fixed a bug with the Docker dev build [PR #1376](https://github.com/catalystneuro/neuroconv/pull/1376)
 
 ## Features
+* Added `InscopixImagingInterface` for converting Inscopix imaging data. [PR #707](https://github.com/catalystneuro/neuroconv/issues/707)
```
h-mayorquin, line 22: The PR number is wrong.

h-mayorquin (commented):
This is in good shape. Two remainign things. We should finish the discussion about the one_photon_series on the `__init__` and discuss the volumetric example in the metadata.

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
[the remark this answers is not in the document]
h-mayorquin: Ok, we still need to do this:
1) Remove the photon_series argument from the `__init__`.
2) Add logic so the correct metadata is calculated in `add_to_nwbfile`

Does this makes sense?

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
daphnedequatrebarbes: Yes, it is required when adding a Subject.

https://github.com/catalystneuro/neuroconv/blob/d4a7c104d2f2d2cb86ba6bc8e0d449a3718671ab/src/neuroconv/schemas/base_metadata_schema.json#L109

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
daphnedequatrebarbes: You're right, the check is unnecessary. It's just leftover from earlier debugging and can be removed.

src/neuroconv/tools/testing/data_interface_mixins.py
daphnedequatrebarbes: I need this because, on my machine, removing it causes tests to fail with KeyError: 'OnePhotonSeries'. This happens because the correct conversion_options (like photon_series_type="OnePhotonSeries") aren't passed to run_conversion, so it defaults to TwoPhotonSeries, which the tests don't expect.

These tests fail for me:

TestInscopixImagingInterfaceMovie128x128x100Part1::test_interface_alignment
TestInscopixImagingInterfaceMovieLongerThan3Min::test_interface_alignment
TestInscopixImagingInterfaceMovieU8::test_interface_alignment

Which make sense for me, because this code ensures the right options are passed so the interface produces the expected NWB output and the tests pass.

h-mayorquin: But those tests where you added that should not rely on the type of series. Can you remove it so I can check the failure in the CI?

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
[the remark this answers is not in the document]
h-mayorquin: Can you ping me here when this is done so I can take a look?

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
```diff
@@ -0,0 +1,197 @@
...
+        if subject_info and subject_info.get("animal_id"):
+            subject_metadata["subject_id"] = subject_info["animal_id"]
+            has_any_subject_data = True
+
+        species_value = None
+        strain_value = None
+
+        if subject_info and subject_info.get("species"):
+            species_raw = subject_info["species"]
+            # If it contains genotype info or matches NWB format, put it in species; otherwise strain
```
h-mayorquin, line 129: Can you write examples of these field looks in the metadata for the examples that we have so far. It will be easier for me to parse the logic of this if.

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
h-mayorquin: I see now. This makes sense. They are probably Dandi requirements. Thanks.

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
```diff
@@ -0,0 +1,197 @@
...
+            else:
+                metadata["Subject"] = subject_metadata
+
+        return metadata
+
+    def add_to_nwbfile(
+        self,
+        nwbfile,
+        metadata: Optional[dict] = None,
+        **kwargs,
```
h-mayorquin, line 181: Passing photon_series_type directly was OK. But this is also works. I was talking about the other conversion options.

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
```diff
@@ -0,0 +1,197 @@
...
+        ----------
+        nwbfile : NWBFile
+            NWB file to add the data to.
+        metadata : dict, optional
+            Metadata dictionary. If None, will be generated dynamically with OnePhotonSeries.
+        **kwargs
+            Additional keyword arguments passed to the parent add_to_nwbfile method.
+        """
+        kwargs["photon_series_type"] = "OnePhotonSeries"
+
```
h-mayorquin, line 196:
```suggestion
# TODO: add logic for determining whether the microscope is `nVista 2P` and change photon_series_type to TwoPhotonSeries accordingly.
```

Let's add a comment here along those lines.

src/neuroconv/tools/testing/data_interface_mixins.py
h-mayorquin: OK, great! when you push this let me know to mark this as resolved.

pyproject.toml
```diff
@@ -278,7 +278,7 @@ fiber_photometry = [
 
 ## Ophys
 ophys_minimal = [  # Keeps the minimal version of ophys dependencies in the repo
-    "roiextractors>=0.6",
+    "roiextractors>=0.5.13",
```
h-mayorquin, line 281: It should be 0.6

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
h-mayorquin: Ok.

tests/test_on_data/ophys/test_imaging_interfaces.py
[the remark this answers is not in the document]
h-mayorquin: Can you add an issue in roiextractor to add those more tests.

CHANGELOG.md
```diff
@@ -23,9 +23,11 @@
 * Fixed a bug with the Docker dev build [PR #1376](https://github.com/catalystneuro/neuroconv/pull/1376)
 
 ## Features
+* Added `InscopixImagingInterface` for converting Inscopix imaging data. [PR #1361](https://github.com/catalystneuro/neuroconv/issues/1361)
```
h-mayorquin, line 26: The link is wrong, it should be linking to a PR not an issue. 
```suggestion
* Added `InscopixImagingInterface` for converting Inscopix imaging data. [PR #1361](https://github.com/catalystneuro/neuroconv/pull/1361)
```

same in segmentation.

tests/test_on_data/ophys/test_imaging_interfaces.py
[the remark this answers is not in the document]
daphnedequatrebarbes: https://github.com/catalystneuro/roiextractors/issues/460, made it concise let me know if it is not detailed enough

tests/test_on_data/ophys/test_imaging_interfaces.py
h-mayorquin: Ok, thanks.

tests/test_on_data/ophys/test_imaging_interfaces.py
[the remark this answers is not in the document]
h-mayorquin: This is good. Thanks.

PR #1362, opened 2025-05-20
Remove Subject creation from `DeepLabCutInterface` unless explicitly specified

h-mayorquin:
From the discussion on https://github.com/catalystneuro/neuroconv/pull/1319#issuecomment-2860097871

### DIFF
--- a/tests/test_on_data/behavior/test_pose_estimation_interfaces.py
+++ b/tests/test_on_data/behavior/test_pose_estimation_interfaces.py
@@ -38,7 +38,6 @@ python_version = version.parse(python_version())
 ndx_pose_version = version.parse(importlib_version("ndx-pose"))
 
 
-@pytest.mark.skipif(ndx_pose_version < version.parse("0.2.0"), reason="Interface requires ndx-pose version >= 0.2.0")
 class TestLightningPoseDataInterface(DataInterfaceTestMixin, TemporalAlignmentMixin):
     data_interface_cls = LightningPoseDataInterface
     interface_kwargs = dict(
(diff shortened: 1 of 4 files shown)


h-mayorquin:
I will finish this after #1363

pauladkisson (changes requested):
Needs a test/check that skeleton.subject is None when the skeleton_subject doesn't match the nwbfile.subject.

Otherwise looks good!

h-mayorquin:
> Needs a test/check that skeleton.subject is None when the skeleton_subject doesn't match the nwbfile.subject.
> 
> Otherwise looks good!

@pauladkisson 
Added.

I am thinking about this. Reading the latest discussion on the ndx-pose repo it seems that multiple subjects in the same file is not desired so I am not sure how to think about this (see here https://github.com/rly/ndx-pose/issues/51). I will bring this up to the next group meeting to decide on how to handle this.

All that said, it is good that the library does not link by default.

tests/test_on_data/behavior/test_pose_estimation_interfaces.py
```diff
@@ -528,10 +527,25 @@ def check_read_nwb(self, nwbfile_path: str):
             skeleton = pose_estimation_container.skeleton
             assert skeleton.nodes[:].tolist() == ["snout", "leftear", "rightear", "tailbase"]
 
+    def test_subject_not_linked(self, setup_interface):
+        """
+        Test that the interface raises an error if the subject is not linked to the config file.
```
pauladkisson, line 532:
```suggestion
        Test that skeleton.subject is None if the subject_id in the metadata doesn't match the nwbfile.
```

pauladkisson (approved):
test docstring wasn't quite right, otherwise looks good.

pauladkisson:
> Reading the latest discussion on the ndx-pose repo it seems that multiple subjects in the same file is not desired so I am not sure how to think about this (see here https://github.com/rly/ndx-pose/issues/51). I will bring this up to the next group meeting to decide on how to handle this.

Yeah that's confusing.

tests/test_on_data/behavior/test_pose_estimation_interfaces.py
h-mayorquin: Thanks.

PR #1363, opened 2025-05-20
updated pose_estimation series names

pauladkisson:
See https://github.com/catalystneuro/neuroconv/pull/1319#discussion_r2078391267


docs/conversion_examples_gallery/behavior/deeplabcut.rst
```diff
@@ -60,29 +60,29 @@ use the following structure:
     ...     "original_videos": None,
     ...     "PoseEstimationSeries": {
     ...         "snout": {
-    ...             "name": "ind1_snout",
-    ...             "description": "Keypoint snout from individual ind1.",
+    ...             "name": "snout",
```
h-mayorquin, line 63: I would like to follow the best practices naming convention:

https://nwbinspector.readthedocs.io/en/dev/best_practices/general.html#naming-conventions

So this would be `PoseEstimationSeriesSnout` instead of just snout.

What do you think?

pauladkisson: Done!

h-mayorquin: Thanks, Paul. To be honest I don't like that convention very much but I rather stick to that standard.

h-mayorquin (approved):

PR #1364, opened 2025-05-21
Inscopix Segmentation Interface

daphnedequatrebarbes:


daphnedequatrebarbes:
## Justification for Skipping Segmentation Tests with Empty Dataset

When testing `InscopixSegmentationInterface` with an empty dataset (no ROIs), certain tests are skipped because they fundamentally require operations that cannot succeed with empty ROI sets:

1. **`test_run_conversion_with_backend_configuration`**  
   During conversion, `get_roi_pixel_masks()`/`get_roi_image_masks()` attempt to stack empty arrays, triggering `ValueError: need at least one array to stack`.

2. **`test_configure_backend_for_equivalent_nwbfiles`**  
   Creating NWB files for comparison requires adding ROI data, which fails with the same empty mask errors.

3. **`test_all_conversion_checks`**  
   This comprehensive test calls all the problematic functions in sequence, including `add_to_nwbfile`.

I skipped these tests because they're designed to validate successful conversions, not to handle the expected errors from empty datasets that we test elsewhere.

pyproject.toml
```diff
@@ -297,6 +298,10 @@ extract = [
 hdf5 = [
     "roiextractors>=0.5.13",
 ]
+inscopix = [
+    "isx>=2.0; python_version < '3.13'",
+    "roiextractors @ git+https://github.com/catalystneuro/roiextractors.git@inscopix_fix",
```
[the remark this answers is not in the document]
daphnedequatrebarbes: For sure! I realized I made a mistake in the previous roiextractors release, so I opened a new PR to fix it. I wasn’t sure how often releases are made.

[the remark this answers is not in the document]
bendichter: this is a good workflow, I just wanted to make a note so we don't accidentally pin to this on main

docs/conversion_examples_gallery/conftest.py
```diff
@@ -23,15 +23,33 @@ def add_data_space(doctest_namespace, tmp_path):
...
+                    "Installation instructions can be found at: "
+                    "https://github.com/inscopix/pyisx?tab=readme-ov-file#install"
+                )
+            # Skip on Python 3.13+
+            if version.parse(python_version) >= version.parse("3.13"):
+                pytest.skip(
+                    "Tests are skipped on Python 3.13 because of incompatibility with the 'isx' module "
+                    "Requires: Python <3.13, >=3.9) "
+                    "See: https://github.com/inscopix/pyisx/issues"
+                )
```
bendichter, line 55: This seems redundant with the

```python
@skip_on_darwin_arm64
@skip_on_python_313
``` statements in the tests

Comment (h-mayorquin) [reply docs/conversion_examples_gallery/conftest.py:55]: These are exclusive to the conversion gallery tests.

Comment (bendichter) [reply docs/conversion_examples_gallery/conftest.py:55]: ah ok thanks for the clarification

File: docs/conversion_examples_gallery/conftest.py [hunk 3]
```diff
@@ -23,15 +23,33 @@ def add_data_space(doctest_namespace, tmp_path):
...
+                    "The isx package is currently not natively supported on macOS with Apple Silicon. "
+                    "Installation instructions can be found at: "
+                    "https://github.com/inscopix/pyisx?tab=readme-ov-file#install"
+                )
+            # Skip on Python 3.13+
+            if version.parse(python_version) >= version.parse("3.13"):
+                pytest.skip(
+                    "Tests are skipped on Python 3.13 because of incompatibility with the 'isx' module "
+                    "Requires: Python <3.13, >=3.9) "
+                    "See: https://github.com/inscopix/pyisx/issues"
```
bendichter, line 54:
```suggestion
                    "See: https://github.com/inscopix/pyisx/issues/61"
```

let's cite the specific issue

h-mayorquin (commented):
First read looks good. I will come back to this once I read the roiextractors PRs more carefully.

tests/test_on_data/ophys/test_segmentation_interfaces.py
```diff
@@ -204,3 +210,301 @@ class TestSuite2pSegmentationInterfaceWithStubTest(SegmentationExtractorInterfac
...
+        assert self.plane_segmentation_name in metadata["Ophys"]["Fluorescence"]
+        assert "raw" in metadata["Ophys"]["Fluorescence"][self.plane_segmentation_name]
+        raw_traces_metadata = metadata["Ophys"]["Fluorescence"][self.plane_segmentation_name]["raw"]
+        assert raw_traces_metadata["name"] == self.roi_response_series_name
+
+
+@skip_on_darwin_arm64
+@skip_on_python_313
+class TestInscopixSegmentationInterfaceEmptyCellSet(SegmentationExtractorInterfaceTestMixin):
+    """Tests for InscopixSegmentationInterface with an empty cellset dataset."""
```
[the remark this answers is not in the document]
pauladkisson: I agree. An empty cell set with no data should raise an error.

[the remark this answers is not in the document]
daphnedequatrebarbes: Based on the discussion, I took a naive approach by adding a simple validation check in `add_to_nwbfile()` that detects when `get_num_rois() == 0` and raises a ValueError with an actionable message. This was my attempt to prevent the "silent error" scenario where users might think conversion succeeded when no meaningful segmentation data was actually written. I tried this quick fix approach, but please let me know if you expected something more comprehensive.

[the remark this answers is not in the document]
h-mayorquin: Thanks, @daphnedequatrebarbes. Base on the input I think this is the way and in parallel we should think on how to annotate a failure of the segmentation algorithm to find ROIs. I am not sure on how to do that but I think this is a bit out of scope here. Let's move this to another issue.

daphnedequatrebarbes:
wait on #1390 and [#449](https://github.com/catalystneuro/roiextractors/pull/449/files#diff-77d6a918d887055677717d11ed47f66ca02326ef2542d7e9af9bdd61193502aa) for failures.  
and on roiextractor release for pyproject isx dependency

pyproject.toml
```diff
@@ -304,6 +305,10 @@ hdf5 = [
 hdf5imaging = [
     "neuroconv[ophys_minimal]",
 ]
+inscopix = [
```
h-mayorquin, line 308: You can change this now that the correspondings PRs have been released.

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
```diff
@@ -0,0 +1,251 @@
...
+            # Sampling frequency
+            sampling_frequency = extractor.get_sampling_frequency()
+            if sampling_frequency:
+                imaging_plane_metadata["imaging_rate"] = sampling_frequency
+
+            # Optical channel
+            optical_channel = {
+                "name": "OpticalChannelDefault",
+                "description": "Inscopix optical channel",
+                "emission_lambda": float("nan"),
```
h-mayorquin, line 135: My impression reading the inscopix page is that most of them do in fact have fixed emission and excitation lambads:

https://inscopix.com/

Can you read it to see if you come away with the same impression?

If they do, when we have a way of determining the microscope we are using (we can't do that now because not enough sample sets) we should build a mapping and set it from there? What do you think?

See here:
https://info.inscopix.com/hubfs/nVista_Solutions_Spec_Sheet.pdf

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
```diff
@@ -0,0 +1,251 @@
...
+            if device_info.get("exposure_time_ms"):
+                desc_parts.append(f"Exposure: {device_info['exposure_time_ms']} ms")
+            if device_info.get("microscope_gain"):
+                desc_parts.append(f"Gain: {device_info['microscope_gain']}")
+
+            imaging_plane_metadata["description"] = "; ".join(desc_parts)
+
+            # Sampling frequency
+            sampling_frequency = extractor.get_sampling_frequency()
+            if sampling_frequency:
```
h-mayorquin, line 128: There should always be a sampling frequency.

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
```diff
@@ -0,0 +1,251 @@
...
+
+            desc_parts = [plane_desc]
+            if device_info.get("microscope_focus"):
+                desc_parts.append(f"Focus: {device_info['microscope_focus']} µm")
+            if device_info.get("exposure_time_ms"):
+                desc_parts.append(f"Exposure: {device_info['exposure_time_ms']} ms")
+            if device_info.get("microscope_gain"):
+                desc_parts.append(f"Gain: {device_info['microscope_gain']}")
+
+            imaging_plane_metadata["description"] = "; ".join(desc_parts)
```
h-mayorquin, line 124: For the imaging plane metadata, I would be good to get the grid spacing, see here:

https://pynwb.readthedocs.io/en/stable/pynwb.ophys.html#pynwb.ophys.ImagingPlane

But I am not sure if those are in fact recorded:
https://info.inscopix.com/hubfs/nVista_Solutions_Spec_Sheet.pdf

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
```diff
@@ -0,0 +1,251 @@
...
+            subject_metadata["subject_id"] = subject_info["animal_id"]
+            has_any_subject_data = True
+
+        species_value = None
+        strain_value = None
+
+        if subject_info and subject_info.get("species"):
+            species_raw = subject_info["species"]
+            # If it contains genotype info or doesn't match format, put it in strain instead
+            if " " in species_raw and species_raw[0].isupper() and species_raw.split()[1][0].islower():
```
h-mayorquin, line 177: Same comment about adding a comment (!) here with examples of species raw so future developers can understand why you are parsing like this.

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
```diff
@@ -0,0 +1,251 @@
...
+                subject_metadata["sex"] = "U"
+
+            if "Subject" in metadata and isinstance(metadata["Subject"], dict):
+                metadata["Subject"].update(subject_metadata)
+            else:
+                metadata["Subject"] = subject_metadata
+
+        return metadata
+
+    def add_to_nwbfile(self, nwbfile, metadata=None, **conversion_options):
```
h-mayorquin, line 224: Let's do this in general for segmentation interfaces so you don't need to override it here. You can remove this.

tests/test_on_data/ophys/test_segmentation_interfaces.py
```diff
@@ -204,3 +210,335 @@ class TestSuite2pSegmentationInterfaceWithStubTest(SegmentationExtractorInterfac
...
+
+        # Check session description includes key information
+        assert "session_description" in metadata["NWBFile"]
+        session_desc = metadata["NWBFile"]["session_description"]
+        assert self.expected_session_name in session_desc
+
+        # Check experimenter information
+        assert "experimenter" in metadata["NWBFile"]
+        experimenter = metadata["NWBFile"]["experimenter"]
+        if isinstance(experimenter, list):
```
h-mayorquin, line 279: I think this should always be a list?

tests/test_on_data/ophys/test_segmentation_interfaces.py
```diff
@@ -204,3 +210,335 @@ class TestSuite2pSegmentationInterfaceWithStubTest(SegmentationExtractorInterfac
...
+
+        assert device_metadata["name"] == self.expected_device_name
+        assert "description" in device_metadata
+        assert "Inscopix" in device_metadata["description"]
+        assert self.expected_device_serial in device_metadata["description"]
+
+        # Check subject information extraction
+        assert "Subject" in metadata
+        subject = metadata["Subject"]
+        assert subject["subject_id"] == self.expected_animal_id
```
h-mayorquin, line 299: Like here, it would be better in my view to just have the animal_id instead of self.expected_animal_id so I can check the value without having to navigate to other place in the code.

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
```diff
@@ -0,0 +1,251 @@
...
+        # Image segmentation (specific to segmentation interface)
+        segmentation_desc = "Inscopix cell segmentation"
+        if analysis_info and analysis_info.get("cell_identification_method"):
+            segmentation_desc += f" using {analysis_info['cell_identification_method']}"
+        if analysis_info and analysis_info.get("trace_units"):
+            segmentation_desc += f" with traces in {analysis_info['trace_units']}"
+
+        # Add number of ROIs to description
+        num_rois = extractor.get_num_rois()
+        segmentation_desc += f" ({num_rois} ROIs identified)"
```
h-mayorquin, line 158: No need to write this in the description. This is available to the user as the number of the rows in the plane segmentation.

tests/test_on_data/ophys/test_segmentation_interfaces.py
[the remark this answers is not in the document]
h-mayorquin: Let's remove this test. This should be a general behavior of the segmentation interface. We can test those with the MockSegmentationInterface. See here:

https://github.com/catalystneuro/neuroconv/pull/1364/files#r2135976644

Let's leave this out of this PR so we have less things to think about and this gets through faster.

tests/test_on_data/ophys/test_segmentation_interfaces.py
```diff
@@ -204,3 +210,335 @@ class TestSuite2pSegmentationInterfaceWithStubTest(SegmentationExtractorInterfac
...
+
+        # Check imaging plane metadata
+        assert "ImagingPlane" in metadata["Ophys"]
+        assert len(metadata["Ophys"]["ImagingPlane"]) == 1
+        imaging_plane = metadata["Ophys"]["ImagingPlane"][0]
+        assert imaging_plane["name"] == self.imaging_plane_name
+        assert imaging_plane["device"] == self.expected_device_name
+
+        # Check field of view is included in description
+        assert "field of view" in imaging_plane["description"]
```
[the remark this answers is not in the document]
h-mayorquin: This is in general. I think this gives us a better picture of what metadata we are actually storing. 

So:

```
value == expected_value
```
instead of
```
"part_of_value" in "expected_value"
```

Does that makes sense? What do you think?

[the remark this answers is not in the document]
daphnedequatrebarbes: Makes sense !

tests/test_on_data/ophys/test_segmentation_interfaces.py
```diff
@@ -204,3 +210,335 @@ class TestSuite2pSegmentationInterfaceWithStubTest(SegmentationExtractorInterfac
...
+class TestInscopixSegmentationInterfaceCellSet(SegmentationExtractorInterfaceTestMixin):
+    """Tests for InscopixSegmentationInterface."""
+
+    data_interface_cls = InscopixSegmentationInterface
+    interface_kwargs = dict(file_path=str(OPHYS_DATA_PATH / "segmentation_datasets" / "inscopix" / "cellset.isxd"))
+    save_directory = OUTPUT_PATH
+    conversion_options = dict(mask_type="pixel")
+
+    @pytest.fixture(scope="class", autouse=True)
+    def setup_metadata(self, request):
```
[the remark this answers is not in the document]
daphnedequatrebarbes: Yes, I copied the testing logic form the other tests in segmentation but I think it is much easier to have the values in check_extracted_metadata.

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
daphnedequatrebarbes: yes, I checked for the 2P and it is fixed as well (Wavelength : 920 nm, fixed)

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
daphnedequatrebarbes: not sure I understand this comment?

h-mayorquin: Yes. I was not very clear. I mean that this solution only works for this interface. But we should implement this for all the segmentation interfaces. So just remove this and let's do another PR were we implement this for all the segmentation interfaces:

https://github.com/catalystneuro/neuroconv/pull/1364#discussion_r2155432408

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
h-mayorquin: Yes, the main problem then is to find out which microscope are we dealing with, right?

From `get_acquisition_info` we get info like: 
```
Microscope Type': 'NVista3
```

But we don't know how that looks for the other microscopes.

Let's do two things so we can resolve this. 

1) Write a TODO comment here saying what are the excitation and emission wavelengths for each of the microscopes and that we don't know how to map the names that we get in `get_acquisition_info['MicroscopeType']` to the actual microscopes because we don't have examples.

2) Can you open an issue on the inscopix repository explaining some context and that for metadata provenance it would be great for us if the excitation and emission lambdas were recorded. We can get those from the factsheets but we don't know how `get_acquisition_info['MicroscopeType']` looks like for other microscopes (as all of our examples are `nVista`. 

Then we can add a link to the issue in the aforementioned comment just so we have a follow up.

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
h-mayorquin: Yes, I don't think those are recorded.

tests/test_on_data/ophys/test_segmentation_interfaces.py
```diff
@@ -204,3 +210,272 @@ class TestSuite2pSegmentationInterfaceWithStubTest(SegmentationExtractorInterfac
...
+        assert "imaging_rate" in imaging_plane
+        np.testing.assert_allclose(imaging_plane["imaging_rate"], 9.998700168978033, rtol=1e-3)
+
+        # Check field of view description exact match
+        expected_plane_desc = "Inscopix imaging plane with field of view 398x366 pixels"
+        assert imaging_plane["description"] == expected_plane_desc
+
+        # Check optical channel information
+        optical_channels = imaging_plane["optical_channel"]
+        if isinstance(optical_channels, list):
```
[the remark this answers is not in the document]
daphnedequatrebarbes: Yes: https://github.com/catalystneuro/neuroconv/blob/ae2fb24432bc4e15e660e287128f0f26dbba01f9/src/neuroconv/schemas/metadata_schema.json#L238

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
h-mayorquin: Thanks, this is great. I made a small suggestion for adding an explicit case in the TODO above but this solves this to me.

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
h-mayorquin: I think that the parsing of this is too brittle but let's leave it like this and fix it if and we get more examples.

tests/test_on_data/ophys/test_segmentation_interfaces.py
```diff
@@ -204,3 +210,169 @@ class TestSuite2pSegmentationInterfaceWithStubTest(SegmentationExtractorInterfac
...
+        expected_plane_desc = "Inscopix imaging plane with field of view 21x21 pixels"
+        assert imaging_plane["description"] == expected_plane_desc
+
+        # Check sampling rate extraction
+        assert "imaging_rate" in imaging_plane
+        np.testing.assert_allclose(imaging_plane["imaging_rate"], 10.0, rtol=1e-3)
+
+        # Check optical channel has default name
+        optical_channels = imaging_plane["optical_channel"]
+        if isinstance(optical_channels, list):
```
h-mayorquin, line 357: same about the list here.

docs/conversion_examples_gallery/conftest.py
h-mayorquin: Changed now.

tests/test_on_data/ophys/test_segmentation_interfaces.py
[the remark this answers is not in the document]
h-mayorquin: Ok, let's change the remaining case in the testing so we don't do the instance check to make the code as linear as possible.

src/neuroconv/datainterfaces/ophys/inscopix/inscopixsegmentationdatainterface.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        Retrieve the metadata for the Inscopix segmentation data.
+
+        Returns
+        -------
+        DeepDict
+            Dictionary containing metadata including device information, imaging plane details,
+            photon series configuration, and Inscopix-specific acquisition parameters.
+
+        TODO: Determine the excitation and emission wavelengths for each Inscopix microscope.
```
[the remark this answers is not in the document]
h-mayorquin: Thanks.

h-mayorquin (approved):
Thanks a lot for your hard work on this PR : )

PR #1373, opened 2025-05-27
Fix sync stream bug in `SpikeGLXConverterPipe`

h-mayorquin:
Now that the sync stream is separated on neo we need this safeguard:

https://github.com/NeuralEnsemble/python-neo/pull/1683


pauladkisson (approved):
Looks good to me!

PR #1399, opened 2025-06-16
Fix dallies

h-mayorquin:
The building of pyedflib in python 3.13.4 is failing. This is an error at the level of cpython and is already been fixed and release but the python action seems to be lagging behind:

https://github.com/python/cpython/issues/135151


.github/workflows/all_python_versions.txt
```diff
@@ -1 +1 @@
-["3.10", "3.13"]
+["3.10", "3.13.5"]
```
pauladkisson, line 1:
```suggestion
["3.10", "3.13.5"]  # TODO: switch back to 3.13 after setup-python uses 3.13.15 by default: https://github.com/actions/setup-python
```

pauladkisson (approved):
Thanks for tracking this down!

.github/workflows/all_python_versions.txt
pauladkisson: Good catch. I didn't think about that. We can just leave it w/o the todo then.

pauladkisson (approved):

PR #1422, opened 2025-07-02
Remove redundant arg in BlackrockSortingInterface docstring

rly:
Minor fix to a redundant doc that I noticed while debugging something else.

This also uses `optional` instead of `default: None`  in the docstring.

> Optional keyword parameters have default values, which are displayed as part of the function signature. They can also be detailed in the description [...] or as part of the type, instead of optional. If the default value would not be used as a value, optional is preferred.

https://numpydoc.readthedocs.io/en/latest/format.html


h-mayorquin (approved):
Thanks @rly

PR #1427, opened 2025-07-07
DeepLabCut add `get_ available subjects` 

h-mayorquin:
This will be useful for dealing with multiple subject files and necessary for when we built the multi-subject converter.


weiglszonja (approved):
LGTM, thanks!

Issue #1366, opened 2025-05-22
Use dependency groups for testing dependencies


h-mayorquin:
https://packaging.python.org/en/latest/specifications/dependency-groups/

Now that the latest version of pip supports them we should move our testing dependencies to dependency groups so we don't include them in the package in pypi.

We can add pre-commit and black as a dev dependency group as well.

PR #1434, opened 2025-07-11
Use PEP 735 dependency groups

h-mayorquin:
This should fix [issue #1366](https://github.com/catalystneuro/neuroconv/issues/1366).

I’ve been encountering dependency issues when trying to install neuroconv and solve the full installation. The problem is that the documentation dependencies are pinned to very specific versions (as they probably should be for reproducibility of the documentation style), but since they’re defined as extras, the package manager tries to resolve them and fails.

Fortunately, PEP 735 offers a clear solution by excluding these from the extras as dependency groups.

More concretely, this change reorganizes the testing, documentation, and development dependencies from [project.optional-dependencies] format into the new standardized [dependency-groups] section. Specifically, the test, docs, and dev (including pre-commit) dependencies are now defined as dependency groups, while domain-specific extras like ecephys, ophys, and behavior remain as extras dependencies.

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -19,7 +19,10 @@ jobs:
           python -m pip install -U pip  # Official recommended way
 
       - name: Install NeuroConv with full testing requirements
-        run: python -m pip install ".[full,test,docs]"
+        run: |
+          python -m pip install ".[full]"
+          python -m pip install --group test
+          python -m pip install --group docs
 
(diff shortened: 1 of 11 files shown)


h-mayorquin:
> Overall looks great. I have a few requests/suggestions:
> 
> * Can you update the copilot-setup-steps.yml workflow as well?
> * The "Testing Suite" page in the Developer Guide section of the docs needs to be updated
> * The "Building the Documentation" page in the Developer Guide section of the docs needs to be updated

Thanks for catching those. It should be updated now.

PR #1435, opened 2025-07-12
Add MockPoseEstimationInterface

h-mayorquin:
For testing.


pauladkisson (commented):
Looks good. We should probably add tests for this?

h-mayorquin:
@pauladkisson Ok, this is the basic version of a generic test for pose estimation (which also testes everything the base mixing does) Eventually, I would like to use this in DeepLabCut and Lighting Pose. I will be working on that later with the IBL project so I can work on unifying the testing of the interfaces.

Issue #789, opened 2024-03-26
[Feature] Handle all Intan modes and formats and add documentation


h-mayorquin:
The files from Intan come in a variety of formats:
* **Everything in one file**: This is the one that we currently support here.
* **One file per stream**: Here, every stream of data (i.e. digital, auxiliar, amplifier) comes in a file of its own
* **One file per channel**: here, every single channel comes in a file of its own.

For the last two we recently added suppport in Neo:
https://github.com/NeuralEnsemble/python-neo/pull/1402

And I will propagate this to Spikeinterface and then to neuroconv.

A further complication is that for the first format (the one that we already support) the files might be divided across time. That is, a file with all the streams and all their channels is produced every x minutes and to get the whole session you have to concatenate. See [this issue](https://github.com/catalystneuro/neuroconv/issues/519
) and specially [this comment](https://github.com/catalystneuro/neuroconv/issues/519#issuecomment-1983194576).

Regarding documentation, I think we could have a tutorial (maybe in the conversion gallery) where it is explained how to to handle the three cases and what to do if the files are scattered across time.

A lot of good information in this thread that I link here for provenance.
https://github.com/SpikeInterface/spikeinterface/issues/2620

zm711:
Related to this: 

All of Neo's code for rhs format's are stuck in version 1.0 when we should update to get to version 3.x (the current). So everything you've listed needs to be doubled for the .rhd AND for the .rhs.  :)

h-mayorquin:
Thanks for jumping in @zm711.

I think that rhs should be sort of straighfroward (fingers crossed!) but we would need some test files for that.

zm711:
I can make some ( still haven't set up my gin g-node account so if I share them would you put them on there :)? ) I think we've done most of the work with the rhd it would just be propagating. I might also try to share an intan merged file so we can get to the bottom of why merging with intan causes neo to fail. I'm pretty busy this week doing analysis, but I can share the files with you this weekend.

h-mayorquin:
OK, send them over when you have a chance. Consdering the neo release schedule and that now we have some resources on improving intan this is a good time to do it.

I still owe you the performance analysis that I will do against the data of the lab that I am working with. This will be important for them and maybe we can add further improvements before the next release.

Thanks again for jumping in!

CodyCBakerPhD:
>  still haven't set up my gin g-node account so if I share them would you put them on there

[...]

PR #1440, opened 2025-07-18
Add intan analog interface

h-mayorquin:
Expanding our support of Intan. 

Moving forward with  #789

### DIFF
--- a/tests/test_on_data/ecephys/test_intan_analog_interface.py
+++ b/tests/test_on_data/ecephys/test_intan_analog_interface.py
@@ -21,8 +21,8 @@ class TestIntanAnalogInterface:
         file_path = ECEPHY_DATA_PATH / "intan" / "rhs_stim_data_single_file_format" / "intanTestFile.rhs"
         interface = IntanAnalogInterface(file_path=file_path, stream_name="USB board ADC input channel")
 
-        assert interface.stream_name == "USB board ADC input channel"
-        assert interface.time_series_name == "TimeSeriesIntanADCInput"
+        assert interface._stream_name == "USB board ADC input channel"
+        assert interface._time_series_name == "TimeSeriesIntanADCInput"
         assert interface.recording_extractor is not None
 
         # Should have 8 ADC channels according to the README
@@ -37,8 +37,8 @@ class TestIntanAnalogInterface:
         file_path = ECEPHY_DATA_PATH / "intan" / "test_fcs_dc_250327_154333" / "info.rhs"
         interface = IntanAnalogInterface(file_path=file_path, stream_name="DC Amplifier channel")
 
-        assert interface.stream_name == "DC Amplifier channel"
-        assert interface.time_series_name == "TimeSeriesIntanDC"
+        assert interface._stream_name == "DC Amplifier channel"
+        assert interface._time_series_name == "TimeSeriesIntanDC"
         assert interface.recording_extractor is not None
 
     def test_interface_initialization_auxiliary_stream_rhd_data(self):
@@ -48,8 +48,8 @@ class TestIntanAnalogInterface:
         file_path = ECEPHY_DATA_PATH / "intan" / "intan_fps_test_231117_052500" / "info.rhd"
         interface = IntanAnalogInterface(file_path=file_path, stream_name="RHD2000 auxiliary input channel")
 
-        assert interface.stream_name == "RHD2000 auxiliary input channel"
-        assert interface.time_series_name == "TimeSeriesIntanAuxiliary"
+        assert interface._stream_name == "RHD2000 auxiliary input channel"
+        assert interface._time_series_name == "TimeSeriesIntanAuxiliary"
 
     def test_interface_initialization_adc_stream_rhd_data(self):
         """
@@ -58,8 +58,8 @@ class TestIntanAnalogInterface:
         file_path = ECEPHY_DATA_PATH / "intan" / "intan_fpc_test_231117_052630" / "info.rhd"
         interface = IntanAnalogInterface(file_path=file_path, stream_name="USB board ADC input channel")
 
-        assert interface.stream_name == "USB board ADC input channel"
-        assert interface.time_series_name == "TimeSeriesIntanADCInput"
+        assert interface._stream_name == "USB board ADC input channel"
+        assert interface._time_series_name == "TimeSeriesIntanADCInput"
 
     def test_invalid_stream_name(self):
         """Test that invalid stream names raise appropriate errors."""
@@ -144,11 +144,11 @@ class TestIntanAnalogInterface:
             nwbfile = io.read()
 
             # Check that the TimeSeries was added to acquisition
-            assert interface.time_series_name in nwbfile.acquisition
-            time_series = nwbfile.acquisition[interface.time_series_name]
+            assert interface._time_series_name in nwbfile.acquisition
+            time_series = nwbfile.acquisition[interface._time_series_name]
 
             # Check properties of the TimeSeries
-            assert time_series.name == interface.time_series_name
+            assert time_series.name == interface._time_series_name
             assert "USB board ADC input channels" in time_series.description
 
             # Check data dimensions - should have 8 ADC channels
@@ -174,8 +174,8 @@ class TestIntanAnalogInterface:
             nwbfile = io.read()
 
             # Check that the TimeSeries was added
-            assert interface.time_series_name in nwbfile.acquisition
-            time_series = nwbfile.acquisition[interface.time_series_name]
+            assert interface._time_series_name in nwbfile.acquisition
+            time_series = nwbfile.acquisition[interface._time_series_name]
 
             # Check that data is present but smaller (stub)
             assert len(time_series.data.shape) == 2


docs/conversion_examples_gallery/recording/intan_analog.rst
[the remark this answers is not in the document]
zm711: How will digital be handled. The dig-out and analog-out I believe are synced in some way (that is like an RHS main thing so I never use that). And what about the stim info for RHS? It is packed in a similar way to the DC amp no? The writing itself seems fine though. Did you want a line by line? Or just a vibe check?

docs/conversion_examples_gallery/recording/intan_analog.rst
```diff
@@ -0,0 +1,155 @@
+Intan Analog Data Conversion
+----------------------------
+
+This guide explains how to convert non-amplifier analog data from Intan systems.
+This includes signals from auxiliary inputs, ADC inputs, and DC amplifiers.
+If you need to convert the primary neural recording data from the amplifier channels, please refer to the :ref:`conversion_gallery_intan` guide.
```
zm711, line 6: This didn't show up as a link when I looked at the docs?

docs/conversion_examples_gallery/recording/intan_analog.rst
```diff
@@ -0,0 +1,155 @@
...
+    >>> # session_start_time is required for conversion. If it cannot be inferred
+    >>> # automatically from the source files you must supply one.
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=ZoneInfo("US/Pacific"))
+    >>> metadata["NWBFile"].update(session_start_time=session_start_time)
+    >>>
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> nwbfile_path = f"{path_to_save_nwbfile}"
+    >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
+
+RHD2000 auxiliary input channels
```
zm711, line 54: You can the stream channel below because I believe that is Intan term but you say channels up here.

docs/conversion_examples_gallery/recording/intan_analog.rst
```diff
@@ -0,0 +1,155 @@
...
+    >>> interface_aux = IntanAnalogInterface(
+    ...     file_path=file_path_aux,
+    ...     stream_name="RHD2000 auxiliary input channel",
+    ...     verbose=False
+    ... )
+    >>>
+    >>> # Extract what metadata we can from the source files
+    >>> metadata_aux = interface_aux.get_metadata()
+    >>> # session_start_time is required for conversion. If it cannot be inferred
+    >>> # automatically from the source files you must supply one.
```
zm711, line 79: I guess you put this note, but in the Intan specific case is this required or not?

docs/conversion_examples_gallery/recording/intan_analog.rst
```diff
@@ -0,0 +1,155 @@
...
+    >>> session_start_time = datetime(2020, 1, 1, 12, 30, 0, tzinfo=ZoneInfo("US/Pacific"))
+    >>> metadata_output["NWBFile"].update(session_start_time=session_start_time)
+    >>>
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> nwbfile_path_output = output_folder / "intan_adc_output_conversion.nwb"
+    >>> interface_output.run_conversion(nwbfile_path=nwbfile_path_output, metadata=metadata_output)
+
+.. note::
+    If your Intan data consists of the main amplifier channels (neural data),
```
zm711, line 154: maybe say the stream name or port channel example (A-001) etc.

docs/conversion_examples_gallery/recording/intan_analog.rst
```diff
@@ -0,0 +1,155 @@
...
+
+    pip install "neuroconv[intan]"
+
+Convert Intan analog channel data (non-amplifier streams) to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.intan.intananaloginterface.IntanAnalogInterface`.
+
+This interface supports analog streams including:
+
+* **USB board ADC input channel**: ADC input channels (analog signals -10V to +10V)
+* **RHD2000 auxiliary input channel**: Auxiliary input channels (e.g., accelerometer data)
+* **DC Amplifier channel**: DC amplifier channels (RHS system only)
```
zm711, line 20: You don't give a voltage here.... not that it is crucial.

zm711 (commented):
My only confusion is at what level do you handle digital stuff? Or will that be a separate extractor too? Because the analog digital divide is not completely clean here. Also where is the stim for rhs?

docs/conversion_examples_gallery/recording/intan_analog.rst
h-mayorquin: Thanks. Yes, I will remove it.

h-mayorquin:
> My only confusion is at what level do you handle digital stuff? Or will that be a separate extractor too? Because the analog digital divide is not completely clean here. Also where is the stim for rhs?

THat will be a separate interface as we require options to convert the data to events if the user requests it. Another PR it will be.

Yes, thanks for the reading, I just wanted to check that I was not saying something completely off or adding information that is wrong.

docs/conversion_examples_gallery/recording/intan_analog.rst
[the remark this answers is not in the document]
h-mayorquin: Yes, another interface, another PR.

We will require another for digital data with an option to convert the data to events and yet another interface for the stimuli.

docs/conversion_examples_gallery/recording/intan_analog.rst
h-mayorquin: Yes, thanks for the catch.

docs/conversion_examples_gallery/recording/intan.rst
```diff
@@ -1,13 +1,19 @@
...
+This guide covers the conversion of Intan data, including both amplifier data (primary neural recordings) and analog data (auxiliary inputs, ADC inputs, DC amplifiers) from RHD2000 and RHS2000 systems.
+
 Install NeuroConv with the additional dependencies necessary for reading Intan data.
 
 .. code-block:: bash
 
     pip install "neuroconv[intan]"
 
-Convert Intan data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.intan.intandatainterface.IntanRecordingInterface`.
+Intan Amplifier Data Conversion
```
weiglszonja, line 12: This is my preference, so ignore if you think the original is the better fit.
```suggestion
Intan Recording Data Conversion
```

docs/conversion_examples_gallery/recording/intan.rst
```diff
@@ -27,4 +33,144 @@ Convert Intan data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.int
     >>> metadata["NWBFile"].update(session_start_time=session_start_time)
     >>>
     >>> nwbfile_path = f"{path_to_save_nwbfile}"  # This should be something like: "./saved_file.nwb"
-    >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
+    >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata, overwrite=True)
+
+Intan Analog Data Conversion
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Convert non-amplifier analog data from Intan systems to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.intan.intananaloginterface.IntanAnalogInterface`.
```
weiglszonja, line 41: The `IntanAnalogInterface` link in the docs is not clickable, maybe it will be after merge?

image

src/neuroconv/datainterfaces/ecephys/intan/intananaloginterface.py
```diff
@@ -0,0 +1,208 @@
...
+        )
+
+        metadata["TimeSeries"] = {
+            self.time_series_metadata_key: {
+                self.time_series_name: dict(
+                    name=self.time_series_name,
+                    description=description,
+                )
+            }
+        }
```
weiglszonja, line 139: At first I thought `self.time_series_metadata_key` would be the key to look up in metadata directly—for example, always something like metadata["TimeSeries"]. But I now see that it’s actually nested like this:
```yaml
TimeSeries:
  TimeSeries:
    TimeSeriesIntanAuxiliary:
      name: TimeSeriesIntanAuxiliary
      description: x
    TimeSeriesIntanADCInput:
      name: TimeSeriesIntanADCInput
      description: x
    TimeSeriesIntanDC:
      name: TimeSeriesIntanDC
      description: x
```
Do we really need this extra layer of nesting (TimeSeries under TimeSeries)? Could we simplify the structure by flattening it one level?

pyproject.toml
```diff
@@ -226,6 +226,7 @@ ecephys = [  # Note that these are the requirements of the extractors as they ar
     "neuroconv[cellexplorer]",
     "neuroconv[edf]",
     "neuroconv[intan]",
+    "neuroconv[intan_analog]",
```
weiglszonja, line 229: Just curious — do both of these ultimately depend on [intanrawio.py from Neo](https://neo.readthedocs.io/en/latest/rawiolist.html#neo.rawio.IntanRawIO)? If so, is there a particular reason we’re keeping them separate, or could it make sense to consolidate them?

weiglszonja (approved):
Thanks for the updates — not requesting any changes, I had two comments just to better understand the design choices.

docs/conversion_examples_gallery/recording/intan.rst
h-mayorquin: The reason I choose this is that Recording is an spikeinterface object but all the data (both neural and non-neural) is loaded as a recording object so I wanted to disambiguate it. Within Intan parlance, the data that of this stream is called amplifier data and that's why I have chosen this term. What do you think?

docs/conversion_examples_gallery/recording/intan.rst
h-mayorquin: Thanks for pointing that out. I will fix it.

src/neuroconv/datainterfaces/ecephys/intan/intananaloginterface.py
h-mayorquin: Wait, I think it should be like this:

```
TimeSeries:
    TimeSeriesIntanAuxiliary:
      name: TimeSeriesIntanAuxiliary
      description: x
    TimeSeriesIntanADCInput:
      name: TimeSeriesIntanADCInput
      description: x
    TimeSeriesIntanDC:
      name: TimeSeriesIntanDC
      description: x
```

Is that what you are advocating for? If that's the case, I agree but that as my intention. Let me double check.

weiglszonja: Exactly, thanks!

docs/conversion_examples_gallery/recording/intan.rst
weiglszonja: Ah thanks for pointing this out, in this case I'd go with your suggestion and drop mine.

src/neuroconv/datainterfaces/ecephys/intan/intananaloginterface.py
h-mayorquin: Yes, there was an extra level of nested. Fixed.

docs/conversion_examples_gallery/recording/intan.rst
h-mayorquin: I forgot to add an entry on the API. It is clickable now.

src/neuroconv/datainterfaces/ecephys/intan/intananaloginterface.py
weiglszonja: Thanks!

docs/conversion_examples_gallery/recording/intan.rst
weiglszonja: Thanks!

PR #1448, opened 2025-07-25
Added optional metadata_file_path to get_metadata

### LINKED ISSUE #541: [Feature]: yaml file as input to converter.get_metadata()

What if `NWBConverter.get_metadata()` took the path of a yaml file as an optional input?

So this:

```python
# The metadata file path
metadata_path = "general_metadata.yml"
metadata_from_yaml = load_dict_from_file(metadata_path)

# The converter that combines the input streams into a single conversion
converter = IntoTheVoidNWBConverter(source_data=source_data)

# The converter can extract relevant metadata from the source files
metadata = converter.get_metadata()
# This metadata can be updated with other relevant metadata
metadata = dict_deep_update(metadata, metadata_from_yaml)
```

could become this:

```python
# The converter that combines the input streams into a single conversion
converter = IntoTheVoidNWBConverter(source_data=source_data)

# The converter can extract relevant metadata from the source files
metadata = converter.get_metadata("general_metadata.yml")
```

Internally, metadata would be loaded and would update the `metadata` object before returning.

_No response_

Yes.

### Code of Conduct

pauladkisson:
Fixes #541

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * Added `MockPoseEstimationInterface` for testing pose estimation workflows with deterministic Lissajous figure motion patterns [PR #1435](https://github.com/catalystneuro/neuroconv/pull/1435)
 * Added `images_container_metadata_key` parameter to `ImageInterface` to allow custom naming and organization of image containers in NWB files. This enables multiple image interfaces to coexist with distinct container names. [PR #1439](https://github.com/catalystneuro/neuroconv/pull/1439)
 * Added per-image metadata support to `ImageInterface` allowing users to specify individual `resolution` (pixels/cm), name and `description` for each image through metadata structure. [PR #1441](https://github.com/catalystneuro/neuroconv/pull/1441)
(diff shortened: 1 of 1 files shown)


h-mayorquin:
I am against this. I think this makes our codebase more complicated for not a lot of gain (saving some lines). I also think that that the behavior of implicitly doing a dict deep update when loading the data is a non-obvious side effect that will cause bad surprises to some of our users. 

Explicit is better than implicit.

pauladkisson:
I'm happy with splitting up metadata into source and default (although, I think that would take significantly more work to untangle the two different types from our current get_metadata methods). But I really don't see what's wrong with packaging that up into a high-level get_metadata function. For example, 

```python
def get_metadata(self, metadata_file_path: FilePath | None):
    default_metadata = self.get_default_metadata()
    source_metadata = self.get_source_metadata()
    metadata = dict_deep_update(default_metadata, source_metadata)
    if metadata_file_path is not None:
        file_metadata = load_dict_from_file(metadata_file_path)
        metadata = dict_deep_update(metadata, file_metadata)
    return metadata
```

PR #1450, opened 2025-07-30
feat: Add Context7 integration configuration

leej3:
## Add Context7 Configuration for NeuroConv

This pull request adds **first‑class Context7 support** for NeuroConv so that AI coding assistants (Cursor, Copilot Chat, Zed, etc.) can access version‑specific conversion guidance directly in prompts.

### Benefits for users
* **Accurate version-specific code examples** – agents retrieve *current* NeuroConv APIs instead of hallucinating.
* **Format‑aware recommendations** – rules instruct models to pick the right `*Interface` and to validate with `check_read()`.
* **Reproducibility** – version history lets downstream projects pin documentation to a specific tag.

### Details on the approach

I have tried to follow the best practices outlined in [their document](https://github.com/upstash/context7/blob/master/docs/adding-projects.md) for guiding this. Some of the highlights are:

| Item | Purpose |
|------|---------|
| **`context7.json`** | Using this config file in the project source helps to control and optimize how the MCP server provides help for Neuroconv |
| **Docs‑only crawl** | Source code, tests, CI files, and build artefacts are excluded, keeping responses focused and fast. |
| **Schema validation** | The `$schema` key enables auto‑completion in IDEs and guarantees the file adheres to the Context7 JSON‑Schema. |

### DIFF
--- a/context7.json
+++ b/context7.json
@@ -1,0 +1,54 @@
+{
+    "$schema": "https://context7.com/schema/context7.json",
+  
+    "projectTitle": "NeuroConv",
+    "description": "Convert 40+ neurophysiology data formats to NWB with automatic metadata extraction.",
+  
+    "folders": [
+      "docs/"
+    ],
+  
+    "excludeFolders": [
+      "src/",
+      "tests/",
+      "dockerfiles/",
+      ".github/",
+      ".venv/",
+      "build/",
+      "dist/",
+      "docs/_build/",
+      "docs/_static/"
+    ],
+  
+    "excludeFiles": [
+      "CHANGELOG.md",
+      "license.txt",
+      "MANIFEST.in",
+      "pyproject.toml",
+      "CITATION.cff",
+      "codecov.yml"
+    ],
+  
+    "rules": [
+      "Always use the data‑specific interface that matches your file format (e.g., SpikeGLXRecordingInterface for SpikeGLX).",
+      "Install optional extras with pip install \"neuroconv[format_name]\" when working with proprietary formats.",
+      "Combine multiple streams with NWBConverter to preserve temporal alignment.",
+      "Call get_metadata() and enrich the returned dict before running conversion.",
+      "Use run_conversion() with backend compression settings for large recordings.",
+      "Leverage chunking options to keep NWB files manageable.",
+      "Run check_read() on every interface before conversion to catch I/O issues early.",
+      "Prefer YAML conversion specifications for repeatable batch jobs.",
+      "Align timestamps from separate devices using set_aligned_timestamps().",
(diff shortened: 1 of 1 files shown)


h-mayorquin (commented):
Another similar solution is this:
https://github.com/ref-tools/ref-tools-mcp

On one hand, this is something extra to maintain but the document is really small and not very costly. Maybe we should just have it for the sake of experimentation. I am OK with it but we should improve the rules a bit.

@bendichter has been thinking hard about AI tools and I think he will be interested on this. Maybe @luiztauffer will be interested as well.

leej3:
> Another similar solution is this: https://github.com/ref-tools/ref-tools-mcp

Thanks, good to know. No doubt there will be others soon too.

> On one hand, this is something extra to maintain but the document is really small and not very costly. Maybe we should just have it for the sake of experimentation. I am OK with it but we should improve the rules a bit.

Projects can be added directly to context7.com (by "community members") but the configuration is detected automatically and errors etc. are more likely to occur. Given there are competing projects in this area it does seem like there is a risk of over-committing on support. The downside of not doing the experimentation you propose would be that end users think it's supported but the efficiency/functionality of a well specified set of rules won't be part of the MCP server.

I don't have enough experience with the project to specify the rules (clearly, given I didn't pick up on the hallucinations!). I hope we can get something along these lines merged though. If I can help just let me know.
  
> @bendichter has been thinking hard about AI tools and I think he will be interested on this. Maybe @luiztauffer will be interested as well.

Excellent.

context7.json
```diff
@@ -0,0 +1,54 @@
...
+      "docs/_static/"
+    ],
+
+    "excludeFiles": [
+      "CHANGELOG.md",
+      "license.txt",
+      "MANIFEST.in",
+      "pyproject.toml",
+      "CITATION.cff",
+      "codecov.yml"
```
luiztauffer, line 29: the Context7 project website says it does `version-specific documentation and code examples`. If that's the case, wouldn't it make sense to keep `CHANGELOG.md`?

Also, it feels like `pyproject.toml` is relevant, since agents might need to debug failures caused by missing dependencies and similar issues.

luiztauffer:
> The downside of not doing the experimentation you propose would be that end users think it's supported but the efficiency/functionality of a well specified set of rules won't be part o

I think the discussion on fully agentic conversions is one step beyond what is being proposed here by @leej3 

Context7 seems to be simply a way to efficiently parse and catalog documentations of open source projects which could be used by LLMs in general, e.g. for our daily work with cline. 

If this works (I’ve never tried it nor its alternatives) it would be a more sustainable approach than what I was trying to do by maintaining a vector database for NeuroConv.

PR #1468, opened 2025-08-19
Fix spikeinterface imports

### LINKED ISSUE #1346: SpikeInterface Extractor Import API change
I know @h-mayorquin  and @alejoe91  already know this but we are changing the import api for spikeinterface such that we only import "function" versions of the extractors rather than the class versions. Heberto said that shouldn't matter since you use the classes with more absolute import style, but we decided that we would change the storage places to the class to all be in a file called `extractor_classes`. So you may need to update your code base for this change:

so 
```python
from spikeinterface.extractors.extractor_classes import AnExtractor
```

or 
```python
from spikeinterface.extractors.an_extractor_file import AnExtractor
```

Just wanted you all to know for planning purposes moving forward. I haven't dug into neuroconv file organization enough to know if this matters. So if you're already doing absolute path down to the class itself then this shouldn't affect you, but our hope is moving forward for devs that want the class style rather than the function style (`read_xx`) they will be able to use the `extractor_classes` import mechanism.

PR: https://github.com/SpikeInterface/spikeinterface/pull/3898

h-mayorquin:
This will close #1346 

Plus, it fixes some other deprecations messages. 

This is an internal changes that does not need a changelog.

### DIFF
--- a/src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
+++ b/src/neuroconv/datainterfaces/ecephys/baserecordingextractorinterface.py
@@ -19,7 +19,7 @@ class BaseRecordingExtractorInterface(BaseExtractorInterface):
 
     keywords = ("extracellular electrophysiology", "voltage", "recording")
 
-    ExtractorModuleName = "spikeinterface.extractors"
+    ExtractorModuleName = "spikeinterface.extractors.extractor_classes"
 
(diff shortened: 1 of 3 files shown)


h-mayorquin:
> Looks good! I see that we're already on the latest version of spikeinterface (0.103.0). So that doesn't seem like it will be a problem here.

Yes, this will only eliminate the warnings that appear everywhere when we use neuroconv at the moment.

PR #1470, opened 2025-08-20
Non electrode channels in EDF. Add  conversion gallery example and `EDFAnalogInterface`

h-mayorquin:
This is a request from a user and I think we will be encountering more often with EDF. 

The current problem is that the ecephys pipeline has checks to avoid recordings with different offsets as that can not be represent in an NWB TimeSeries at the moment. For the EDF format, and for many others, this often indicates that recording mixes in its channels both neural and non-neural data so it is good that we error out. When the user tries to converts EDF they get an error saying that their data has multiple offsets (which is not clear) and they are unable to write their data. So, we give them an error and we have no current option for them to write their data.

This ameliorate this problem this PR adds:

1) The `EDFAnalogInterface` that converters non-electrode channels in the EDF format to a `TimeSeries` object in the NWB file.
2) Improve the conversion gallery with more documentation about the purpose of `EDFRecordingInterface`, its limitations and how to make it work when the source file contains non-electrode or analog channels.
3) Add an example to the conversion gallery for how to use the `EDFAnalogInterface` and how to combine both `EDFRecordingInterface` and `EDFAnalogInterface` to write all the source data.

The documentation /tutorial can be accessed here:
https://neuroconv--1470.org.readthedocs.build/en/1470/conversion_examples_gallery/recording/edf.html

Here is the gin file PR:
https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/163

The gin file is been merged and the tests are added. This is ready to go.


docs/conversion_examples_gallery/recording/edf.rst
```diff
@@ -16,16 +19,134 @@ Convert edf data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.edf.e
...
 
     # Choose a path for saving the nwb file and run the conversion
     nwbfile_path = f"{path_to_save_nwbfile}"
-    interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata)
+    interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata, overwrite=True)
+
+Combining Electrode and Non-Electrical Channels
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To convert both electrode and non-electrical channels into a single NWB file, use the ConverterPipe with multiple interfaces. Remember to group non-electrical channels by their unit types:
```
alessandratrapani, line 85: I find the "non-electrical" term a bit confusing in this context.

alessandratrapani (commented):
Looks good to me! I have just a comment on the choice "non-electrical" to define these signals, I would simply call the "auxiliary analog" signals.

h-mayorquin:
> Looks good to me! I have just a comment on the choice "non-electrical" to define these signals, I would simply call the "auxiliary analog" signals.

I made some changes in this direction, see above.

docs/conversion_examples_gallery/recording/edf.rst
```diff
@@ -16,16 +20,141 @@ Convert edf data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.edf.e
...
+    from zoneinfo import ZoneInfo
+    from pathlib import Path
+    from neuroconv import ConverterPipe
+    from neuroconv.datainterfaces import EDFRecordingInterface, EDFAnalogInterface
+    from neuroconv.utils import dict_deep_update
+
+    file_path = f"{ECEPHY_DATA_PATH}/edf/electrode_and_analog_data/electrode_and_analog_data.edf"
+
+    # Define the channels to process
+    all_non_electrical_channels = ["TRIG", "OSAT", "PR", "Pleth"]  # All non-electrical channels
```
alessandratrapani, line 105:
```suggestion
    all_non_electrical_channels = ["TRIG", "OSAT", "PR", "Pleth"]  # All auxiliary channels
```

docs/conversion_examples_gallery/recording/edf.rst
```diff
@@ -16,16 +20,141 @@ Convert edf data to NWB using :py:class:`~neuroconv.datainterfaces.ecephys.edf.e
...
+
+    # Create electrode interface (skip all non-electrical channels)
```
alessandratrapani, line 107:
```suggestion
    # Create electrode interface (skip all auxiliary channels)
```

docs/conversion_examples_gallery/recording/edf.rst
alessandratrapani: Agree, it makes sense to me

PR #1474, opened 2025-08-21
Add chatbot to docs (side panel) 

h-mayorquin:
This PR adds the chatbot currently used in the `pynwb` documentation to `neuroconv`. You can see it in action in the build here:  
<https://neuroconv--1474.org.readthedocs.build/en/1474/>  

Compared to the `pynwb` documentation, one improvement is that the chatbot now appears on **all pages**, rather than being limited to the index page.  

This is done by using a [Sphinx hook](https://docs.readthedocs.com/platform/stable/guides/adding-custom-css.html) that allows us to inject JavaScript code across the site. With this hook, the implementation remains relatively simple.  

One problem with the current approach is that it does not have content persistency when users navigate the documentation (that is, if you click in another section of the documentation the chatbot content rests). An approach to solve this is outlined in #1476.

There are still some UX/design details to refine. For example, on very wide screens the button is pushed too far away from the main content:  

image  

When the window is resized closer to common display widths, the layout issue resolves:  

image  

Rather than fixing these alignment details now, I suggest we first decide whether we want to continue with this anchored approach as a side panel or move toward the alternative discussed in #1476. That choice will likely affect how much effort we should put into polishing the layout itself.


h-mayorquin:
Closing as #1476 won.

PR #1476, opened 2025-08-21
Add chatbot to docs (external window)

h-mayorquin:
This is an alternative to https://github.com/catalystneuro/neuroconv/pull/1474 to add a chatbot.

This implementation displays the chatbot in an external window, which avoids the issues of content persistence when navigating through the Sphinx documentation.

You can check it out here:
https://neuroconv--1476.org.readthedocs.build/en/1476/

If we were to solve persistence within an anchored UI element and the current iframe setup, we would need to rely on JavaScript storage methods. Options here would include a URL-as-state mechanism or storage solutions such as localStorage or IndexedDB, both of which would introduce additional complexity and require changes in the nwb-assistant repository.

Using an external window instead has some potential advantages:
* Simplifies the implementation by avoiding custom storage logic.
* Provides a consistent, standalone UI that could be reused across repositories (pynwb, neuroconv, etc.).
* Follows a fairly common pattern for web-based chat interfaces.

Another alternative is to use the same side panel as in pynwb as in: https://github.com/catalystneuro/neuroconv/pull/1474


bendichter:
Yeah, I agree this approach works better in practice

bendichter (approved):

PR #1479, opened 2025-08-21
 Improve `linking_sorted_data` documentation

h-mayorquin:
For the benefit of the reviewer, to make their job easier, this PR extracts the core documentation improvements from the larger PR #1449 and focuses solely on documentation improvement:

* Adds `SortedRecordingConverter` rationale and use case
* Moves the electrode linking content from `user_guide` to `how_to` as a more practical reference.
* Adds a motivation section in the documentation that adds important context.
* Changes the example from SpikeGLX to Intan because SpikeGlx will have its own special solutio nin #1449 and the documentation was confusing users.


docs/how_to/linking_sorted_data.rst
```diff
@@ -0,0 +1,149 @@
...
+    sorting errors, identify artifacts, and maintain reproducibility by making the
+    sorting process transparent and auditable.
+
+
+Accessing Electrode Metadata from Units
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Once units are properly linked to electrodes and the electrodes table is well annotated,
+you can programmatically retrieve electrode-level metadata for any unit in your NWB file.
+
```
weiglszonja, line 53: I guess this is a personal preference, but I would change the order to first show how to link units to electrodes and then how to access them

h-mayorquin: Thanks for the review, I am gonna invert the order in a separate PR after https://github.com/catalystneuro/neuroconv/pull/1449 to avoid merge conflicts.

PR #1483, opened 2025-08-23
Fix version switcher in the documentation

h-mayorquin:
This PR solves two problems with the switcher.

First, at the moment, the two options we have are:
- stable (latest release)
- latest (development)

image

I think is confusing to have the word latest twice and overall there are too many words that don't really add much. This PR changes this to have only stable and development:

image
 

Second, in the current version, when you switch the label on the switcher always remains the same. That is, you click "latest (development)" and it correctly sends you to the documentation of main but the button does not change. This PR fix this as well.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,28 @@
+# Custom instructions for GitHub Copilot in the **NeuroConv** repository
+
+## 1 Environment awareness
+- **Read `.github/copilot-setup-steps.yml` first.**
+  It shows the Python version, pre-installed packages, and tools (pytest, pre-commit, ruff, mypy, etc.) already available on the runner.
+- **Prefer the existing environment.**
+  Install or upgrade a package **only** when the change truly cannot be completed otherwise.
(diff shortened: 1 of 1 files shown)


weiglszonja (commented):
Thanks, I think the phrasing is better now. Just one comment, the dropdown currently shows “Choose version” by default, even though the page already has an active version (I guess the default would be the stable version). It might be clearer to display the active version as the default selection, to immediately know which version we’re viewing without opening the dropdown.
Screenshot 2025-08-27 at 10 19 41 (2)

h-mayorquin:
@weiglszonja

Yes, this is a limitation of the switcher.json configuration. At runtime, the PR build doesn't have a corresponding entry in the switcher.json file, so no version match is found. This causes the switcher to display "Choose version" instead of highlighting the current PR version.

It will work fine on main and released versions but is a problem for PR builds and local builds.

I've implemented a fix that will display this correctly for PR builds:
https://github.com/catalystneuro/neuroconv/pull/1487

Please check it out.

For local builds I could not think on a good solution but I think is fine at the moment.

weiglszonja (approved):
Thanks for the clarification, makes sense now.

PR #1488, opened 2025-08-27
Fix more spikeinterface import warnings

h-mayorquin:
This makes progress on #1346 

Fixes (hopefully) the rest of the imports that I did not fix on #1468.

Also, adds a pytest directive to raise this warning to error.


pauladkisson (approved):
Looks good!

PR #1497, opened 2025-09-03
Use attestation in auto-publish

### LINKED ISSUE #1491: [Feature]: Switch to using Trusted Publisher on PyPI instead of a API token

The new standard for publishing a package to PyPI is using [trusted publishing](https://docs.pypi.org/trusted-publishers/).

All you have to do is 
1. Go to https://pypi.org/manage/project/neuroconv/settings/publishing/ and add the github workflow as a trusted publisher
2. Update the workflow to use 'attestation' instead of API token (see example [here](https://github.com/LernerLab/GuPPy/blob/packaging/.github/workflows/auto-publish.yml))

I'd do it myself, but I don't have the requisite permissions on PyPI. 

_No response_

Yes.

### Code of Conduct

h-mayorquin:
Should fix #1491

I also added two other things:
1. The release should only run for tags that start with "v" allowing the use of tags for other purposes if necessary.
2. A rule-set (changed in settings) that restricts who can create tags as tagging triggers releases.

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -1,5 +1,5 @@
-# This workflow will upload a Python Package using Twine when a release is created
-# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+# Simple: publish to PyPI when a GitHub Release is published AND the tag starts with 'v'.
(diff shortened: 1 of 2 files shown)


pauladkisson (approved):
Looks good to me. Probably worth having a changelog entry for this one, I would think.

h-mayorquin:
> Looks good to me. Probably worth having a changelog entry for this one, I would think.

Added

PR #1508, opened 2025-09-17
Add DOI badge to README

CodyCBakerPhD:
For the DANDI paper

Y'all get a DOI already, adding badge to README


h-mayorquin (approved):
Thanks, we can change or add the scipy one when it gets there.

@alejoe91 if you need one asap maybe this could be one?

I will let you know guys when we get the other one.

PR #1513, opened 2025-09-22
Remove support for Device object in metadata dict

h-mayorquin:
As in the title. The metadata is meant to be mostly serializable (see date time objects). Supporting this is an unnecessary complication.


weiglszonja (approved):
LGTM

PR #1519, opened 2025-09-26
Fork to merge #1486

h-mayorquin:
This is #1486 but needs to come from someone with secret access for this to actually be tested.


h-mayorquin:
@CodyCBakerPhD check out the error I am getting:

```
DEBUG    dandi:dandiapi.py:502 Using api key from DANDI_API_KEY environment variable
DEBUG    dandi:dandiapi.py:199 GET https://api-dandi.emberarchive.org/api/auth/token
DEBUG    dandi:dandiapi.py:270 Response: 401
ERROR    dandi:dandiapi.py:279 Error 401 while sending GET request to [https://api-](https://api-dandi.emberarchive.org/api/auth/token:)
```

So what happens if that this was working locally but then, in the same environemnt, I also set an API_DANDI_KEY and the service (incorrectly) uses it but then (correctly rejects it).

Looking at the code I can't find any reference to a variable called `EMBER_API_KEY` and it seems that the API_KEY always go by the same name `DANDI_API_KEY`.

https://github.com/dandi/dandi-cli/blob/2827711bdf2265673599cb4c6bbd91d1e4eeeb24/dandi/dandiapi.py#L499-L504

CodyCBakerPhD:
Let me try again locally then

h-mayorquin:
> @h-mayorquin I'm just curious if this works for you locally or if my machine is magical
> 
> Separate from GitHub workflow secret passing issues

I remember testing it locally and it work. I also opened another fork with less changes (as some of my trash got here):

https://github.com/catalystneuro/neuroconv/pull/1519

Let me try again.

PR #1520, opened 2025-09-26
Add Tuple -> tuple and Union[x, y] to x | y ruff rule

h-mayorquin:


h-mayorquin:
> Looks good. As with ROI extractors, I would suggest we add a rule for Optional as well: https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/

I am Ok with it but I remember that one caused problems at some point. Can we do it in another PR just so we don't these two rules stuck if the other proves itself more difficult?

pauladkisson (approved):
> Can we do it in another PR just so we don't these two rules stuck if the other proves itself more difficult?

Sure!

Issue #1511, opened 2025-09-22
Simplify metadata flow on roiextractors module


h-mayorquin:
When metadata flows in the roiextractors module from the top level down to the methods that write objects, the following pattern reoccurs:

https://github.com/catalystneuro/neuroconv/blob/e8928366459734366c7956d33e22c801ee0585bf/src/neuroconv/tools/roiextractors/roiextractors.py#L387-L394

1. We deep copy the metadata
2. We fetch the metadata defaults
3. We update the defaults with the metadata via dict.update

This same pattern is repeated in the other functions. For example, when `add_imaging_plane_to_nwbfile` is called by `add_photon_series_to_nwbfile`, the same steps are repeated even if they have already been done:

https://github.com/catalystneuro/neuroconv/blob/e8928366459734366c7956d33e22c801ee0585bf/src/neuroconv/tools/roiextractors/roiextractors.py#L271-L276

This approach has several disadvantages:
* Multiple inefficient deep copies of the entire metadata and defaults as the data flows through the stack.
* Silent modification of the metadata, so at deeper levels it’s unclear whether values come from the user or from defaults, making debugging harder.
* Multiple places where defaults are added, so no single source of truth for defaults.

The current motivation is clear: we don’t want to modify the user-provided metadata, and we want to ensure that data includes all defaults through deep updates.

However, I think this can be simplified to be less brittle by following these principles:
1. **Single source of truth**. Maintain one dictionary with all ophys metadata as the authoritative reference for defaults. This also is way more explicit and can be used as a reference.
2. **Immutable metadata**. Metadata should pass through the call stack without modifications or side effects.
3. **Targeted defaults**. Fetch defaults only at the point of object creation and only for the values that the object requires. Making default addition both more explicit and targeted. 

This would simplify the code by making defaults explicit, applying them only in the right scope, and improving efficiency by eliminating repeated deep copies and updates at every level.

Thoughts? @pauladkisson @weiglszonja @alessandratrapani

PR #1521, opened 2025-09-26
Simplify roiextractors metadata flow I 

h-mayorquin:
This is the first step to solve #1511. The idea here is to start from the top-down. First, define a immutable global reference for the ophys default metadata. Then, in other PRs, I will be clearing each of the writing branches of intermediate metadata mutations and only calling the defaults at the end.

### DIFF
--- a/src/neuroconv/tools/roiextractors/roiextractors.py
+++ b/src/neuroconv/tools/roiextractors/roiextractors.py
@@ -42,101 +42,104 @@ from ...utils import (
 from ...utils.str_utils import human_readable_size
 
 
-def _get_default_ophys_metadata() -> DeepDict:
-    """Fill default metadata for Device and ImagingPlane."""
+def _get_default_ophys_metadata():
+    """
+    Returns fresh ophys default metadata dictionary.
+
+    Single source of truth for all ophys default metadata.
+    Structure matches the output of get_nwb_imaging_metadata() and get_nwb_segmentation_metadata().
+    Each call returns a new instance to prevent accidental mutation of global state.
+    """
     metadata = get_default_nwbfile_metadata()
 
-    default_device = dict(name="Microscope")
-
-    default_optical_channel = dict(
-        name="OpticalChannel",
-        emission_lambda=np.nan,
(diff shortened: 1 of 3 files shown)


src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -42,101 +42,104 @@
...
-        description="The plane or volume being imaged by the microscope.",
-        excitation_lambda=np.nan,
-        indicator="unknown",
-        location="unknown",
-        device=default_device["name"],
-        optical_channel=[default_optical_channel],
-    )
+    Single source of truth for all ophys default metadata.
+    Structure matches the output of get_nwb_imaging_metadata() and get_nwb_segmentation_metadata().
+    Each call returns a new instance to prevent accidental mutation of global state.
```
pauladkisson, line 51: We should add tests for each of these two properties.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -42,101 +42,104 @@
...
-            ImagingPlane=[default_imaging_plane],
-        ),
-    )
+    metadata["Ophys"] = {
+        "Device": [{"name": "Microscope"}],
+        "ImagingPlane": [
+            {
+                "name": "ImagingPlane",
+                "description": "The plane or volume being imaged by the microscope.",
+                "excitation_lambda": np.nan,
```
pauladkisson, line 61: This pattern of filling essential metadata fields with NaN by default seems questionable to me, shouldn't we require either the extractor or the user to specify this value?

h-mayorquin: Maybe we should but I am keeping the current behavior.

src/neuroconv/tools/roiextractors/roiextractors.py
h-mayorquin: I added a test, I also call the `_get_default_ophys_metadata` in the other tests to add more complex checks that is not being modified under the hood.

PR #1524, opened 2025-09-26
Add documentation for miniscope interface

h-mayorquin:
Ok, this PR ports the work that @alessandratrapani has done in roiextractors to improve our miniscope conversion:

* To keep backwards compatibility it keeps the `MiniscopeConverter` working was it was before. It is still overfitting to the structure of the Tye Lab in two important ways that are discussed here: https://github.com/catalystneuro/roiextractors/issues/356. We can then slowly improve this to be better by removing the assumption of contigous sub-sessions and using the [configuration file to get the folder structure](https://github.com/catalystneuro/roiextractors/issues/356#issuecomment-3325184204) when available. To accomplish this, the PR introduces a private `_MiniscopeMultiRecordingInterface` which is the previous interface without any modifications so everything keeps working as it was.
* There should be a way of converting data and while we work on improving the converter we expose a now more flexible `MiniscopeImagingInterface` that can handle a single session recording when it has the expected structure with a `folder_path` and has a fallback `file_paths` parameter, together with an optional localization of the configuration file path and the timestamps for extra flexibility. The idea is that while we improve the converter this will allow users to conver their data and we can use the interface as the base for future improvmeents fo the converter.
* Finally, it is important to note that we don't have a flexible way of converting single session behavior data associated with Miniscope. This will have to wait.

### DIFF
--- a/src/neuroconv/datainterfaces/ophys/miniscope/miniscopeimagingdatainterface.py
+++ b/src/neuroconv/datainterfaces/ophys/miniscope/miniscopeimagingdatainterface.py
@@ -10,12 +10,12 @@ from ..baseimagingextractorinterface import BaseImagingExtractorInterface
 from ....utils import DeepDict, dict_deep_update
 
 
-class MiniscopeImagingInterface(BaseImagingExtractorInterface):
-    """Data Interface for MiniscopeImagingExtractor."""
+class _MiniscopeMultiRecordingInterface(BaseImagingExtractorInterface):
+    """Data Interface for MiniscopeMultiRecordingImagingExtractor."""
 
-    display_name = "Miniscope Imaging"
+    display_name = "Miniscope Multi-Recording Imaging"
     associated_suffixes = (".avi", ".csv", ".json")
-    info = "Interface for Miniscope imaging data."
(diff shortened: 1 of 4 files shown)


src/neuroconv/datainterfaces/ophys/miniscope/miniscopeimagingdatainterface.py
```diff
@@ -162,3 +162,180 @@ def add_to_nwbfile(
...
+                ├── metaData.json          # required configuration file
+                └── timeStamps.csv         # optional timestamps file
+        file_paths : list, optional
+            List of .avi file paths to be processed. These files should be from the same
+            recording session and must follow the naming convention (0.avi, 1.avi, 2.avi, ...).
+            Files will be concatenated in numerical order.
+            Use this for non-standard folder structures.
+        configuration_file_path : str, optional
+            Path to the metaData.json configuration file containing recording parameters.
+            Required when using file_paths parameter.
```
alessandratrapani, line 236: Should we add a deprecation message here in the doc string?

alessandratrapani (approved):
LGTM

PR #1526, opened 2025-09-30
Fix segmentation pipeline tests to only use public API of roiextractors

h-mayorquin:
While working on https://github.com/catalystneuro/roiextractors/pull/505 I noticed that there are some segmentation extractor tests that are failing because they are accessing private attributes of the segmentation extractor.

I am changing this here so only the public API is used.

In addition, I noticed a bug on roiextractors while fixing this:

https://github.com/catalystneuro/roiextractors/pull/508

So some tests will need to be temporarily commented out.


weiglszonja (approved):
Ok, this makes sense to me

Issue #1525, opened 2025-09-29
Avoid NaN in defaults


pauladkisson:
Currently, the imaging and segmentation interfaces use `np.nan` as default values for essential metadata fields such as `excitation_lambda`. To avoid creating NWB files that violate best practices that users might not notice until they run NWB Inspector, we should require that any _essential_ metadata be specified either at the extractor level or at the interface level. If essential metadata has not been specified, we should raise an informative error.

_Originally posted by @pauladkisson in https://github.com/catalystneuro/neuroconv/pull/1521#discussion_r2388550420_

h-mayorquin:
Yes, thanks for opening this. 

I think we have more cases like this where we prioritised convenience over correctness. I can't recall them from the top of the my head but we can collect them here.

This is the type of metadata that I had on mind on this comment:
https://github.com/catalystneuro/neuroconv/issues/1269#issuecomment-2758890826

h-mayorquin:
We ended up going in a different direction here by accepting conventional but identifiable placeholders in neuroconv and making the inspector responsibility for the the checks. Can we close this one?

PR #1527, opened 2025-09-30
Simplify roiextractors metadata flow I (imaging plane and add_devices)

h-mayorquin:
Moving forward to solve #1511

Next PR now that #1521 has been merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 * Support NIDQ analog streams in `OpenEphysBinaryAnalogInterface` [PR #1503](https://github.com/catalystneuro/neuroconv/pull/1503)
 
 ## Improvements
-* Refactored `add_devices_to_nwbfile` and `add_imaging_plane_to_nwbfile` to avoid `dict_deep_update` and metadata mutation, using defaults directly from single source of truth `_get_default_ophys_metadata()` [PR #1526](https://github.com/catalystneuro/neuroconv/pull/1526)
+* Refactored `add_devices_to_nwbfile` and `add_imaging_plane_to_nwbfile` to avoid `dict_deep_update` and metadata mutation, using defaults directly from single source of truth `_get_default_ophys_metadata()` [PR #1527](https://github.com/catalystneuro/neuroconv/pull/1527)
 * Refactored ecephys metadata functions to use a single source of truth pattern, eliminating hardcoded duplications and improving maintainability [PR #1522](https://github.com/catalystneuro/neuroconv/pull/1522)
 * Refactored ophys metadata functions to use a single source of truth pattern, preventing accidental mutation of global state and improving maintainability [PR #1521](https://github.com/catalystneuro/neuroconv/pull/1521)
 * Add ruff-rule to detect non-pep585 annotation [PR #1520](https://github.com/catalystneuro/roiextractors/pull/1520)


src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -227,39 +230,9 @@ def add_devices_to_nwbfile(nwbfile: NWBFile, metadata: dict | None = None) -> NW
     return nwbfile
 
 
-def _create_imaging_plane_from_metadata(nwbfile: NWBFile, imaging_plane_metadata: dict) -> ImagingPlane:
```
h-mayorquin, line 230: Now that are settings defaults having this extra private function obscures the logic so I am removing it.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -270,45 +243,60 @@ def add_imaging_plane_to_nwbfile(
...
+        (plane for plane in imaging_planes_list if plane["name"] == imaging_plane_name),
+        None,
+    )
+
+    # If user provided imaging planes but the requested name was not found, raise an error
+    has_user_requested_plane = user_plane_metadata is not None
+    has_imaging_planes = bool(imaging_planes_list)
+    if not has_user_requested_plane and has_imaging_planes:
+        raise ValueError(
+            f"Metadata for Imaging Plane '{imaging_plane_name}' not found in metadata['Ophys']['ImagingPlane']."
```
pauladkisson, line 280: I'm finding this block of code a little bit hard to read, maybe just because I'm tired, but I think the variable names might be a bit clearer.

```python
if imaging_plane_name is not None and imaging_plane_name_is_not_in_metadata:
...
```
I think, if I'm reading correctly, this is what the if statement is trying to accomplish here?

h-mayorquin: Thanks for the feedback, the process is the following:
1) the user provides an imaging_plane_name
2) We search the metadata to see if there is an entry whose name is `imaging_plane_name`.
3) if this is not find there, raise an error.

A more verbose version of the code is:

```python
    metadata_has_user_requested_plane = user_plane_metadata is not None
    metadata_has_imaging_planes_metadata = bool(imaging_planes_list)
    if not metadata_has_user_requested_plane and metadata_has_imaging_planes_metadata:
       raise error telling the user that they give us metadata but give us a name that is not there.
``` 

I had more verbose names like those at the beginning but it is my October wish to be less verbose x D

If you have name suggestions let me know.

h-mayorquin: Ah!

I think you are right.

We should throw an error when the user requests a name for which no metadata entry exists, right?

This is easier to read and does not have that but that you mention.

Am I getting you right?

h-mayorquin: I made some changes on that direction, check them out. I am probably tired too.

h-mayorquin: Wait, actually, I don't think that metadata should ever be None at this level. It is a function to build and add the imaging_plane from the metadata so metadata should always be passed.

It might happen that the metadata entry that the user request does not exist in the passed metadata but we already have a test for that:

https://github.com/catalystneuro/neuroconv/blob/5504e0a34829b131d9c1541231657bc6e8e3469d/tests/test_modalities/test_ophys/test_tools_roiextractors.py#L237-L246

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -227,39 +230,9 @@ def add_devices_to_nwbfile(nwbfile: NWBFile, metadata: dict | None = None) -> NW
...
-
-    imaging_plane = ImagingPlane(**imaging_plane_metadata)
-
-    return imaging_plane
-
-
 def add_imaging_plane_to_nwbfile(
     nwbfile: NWBFile,
-    metadata: dict,
+    metadata: dict | None = None,
```
pauladkisson, line 235: > Wait, actually, I don't think that metadata should ever be None at this level.

In that case, it should stay mandatory in the function signature and all the places where it's assumed that it could be None should be updated.

pauladkisson: But to be clear, metadata["Ophys"] could be an empty dictionary, right? And I don't think we are testing that case for this function.

src/neuroconv/tools/roiextractors/roiextractors.py
```diff
@@ -271,44 +244,65 @@ def add_imaging_plane_to_nwbfile(
...
 
-        imaging_plane = _create_imaging_plane_from_metadata(
-            nwbfile=nwbfile, imaging_plane_metadata=imaging_plane_metadata
-        )
-        nwbfile.add_imaging_plane(imaging_plane)
+        # Copy user metadata to avoid mutation
+        imaging_plane_kwargs = metadata_found.copy()
+
+        # Fill in any missing required fields with defaults
+        required_fields = ["name", "excitation_lambda", "indicator", "location", "device", "optical_channel"]
```
h-mayorquin, line 287: @pauladkisson 

I want to add this explicitly on the light of the discussion that you started on https://github.com/catalystneuro/neuroconv/issues/1525

I will be easier to modify if and when we reach an agreement there


I also improved the docstring and (imo) the general flow of the function based on our discussion. Let me know what you think.

pauladkisson (approved):
> I also improved the docstring and (imo) the general flow of the function based on our discussion. Let me know what you think.

Yes, this is so much easier for me to follow. Thank you!

h-mayorquin:

PR #1534, opened 2025-10-03
Simplify roiextractors metadata flow II

h-mayorquin:
Another step on the direction of #1511

Should come after #1527


pauladkisson (approved):
Looks good!

Issue #1549, opened 2025-10-14
Neuropixels are duplicating electrode table rows across ap and lf bands


h-mayorquin:
SpikeGLX recordings from Neuropixels probes can record multiple data streams from the same physical electrodes the AP band and the LF band. Currently, when converting these multi-stream recordings to NWB format, the electrode table creates duplicate rows for each physical electrode because the deduplication logic uses the combination of channel name and electrode group name as the unique identifier. Since AP and LF streams have different channel names (e.g., imec0.ap#AP0 vs imec0.lf#LF0), they are treated as distinct electrodes even though they originate from the same physical contact on the probe. This results in a 384-contact Neuropixels probe generating 768 electrode table rows instead of the correct 384 rows.

This duplication violates the semantic meaning of the electrode table, which should represent the physical recording hardware rather than the acquisition system channels. The inflated electrode table causes some problems like overestimation of the number of electrodes by downstream analysis tools and duplication of spatial information.

The key to fix this issue is 1) to change the `add_electrodes_to_nwbfile` function to support electrode-based deduplication and , 2) propagate the `contact_ids` to the electrodes table if there is a probe attached to the recording.

PR #1548 will implement this functionality while keeping backward compatibility with non-probe recordings. Then, in another PR, I will do localized changes to the spikeglx interface so the metadata is written correctly to the NWB file.

Once we solve https://github.com/catalystneuro/neuroconv/issues/1504 we can also do the same for neuropixels with open ephys.

PR #1553, opened 2025-10-15
Fix duplication of rows on Spikeglx Neuropixels 

h-mayorquin:
This will close https://github.com/catalystneuro/neuroconv/issues/1549 and has PR #1548 as a base. So #1548 should come first.

In spikeglx-neuropixels every electrode is associated with two channels, so we change the channel name to be reflect that so the mechanism on #1548 does its magic and avoids the duplication of electrode per band.


tests/test_on_data/ecephys/spikeglx_single_probe_metadata.json
h-mayorquin: This is a test, so I am just changing what the json would be.

src/neuroconv/datainterfaces/ecephys/spikeglx/spikeglx_utils.py
```diff
@@ -7,30 +7,6 @@
 from pydantic import FilePath
 
 
-def add_recording_extractor_properties(recording_extractor) -> None:
```
h-mayorquin: Moving away from the utils pattern here. I don't really want to have such a critical behavior (renaming the channel names) hidden away in a util.

PR #1554, opened 2025-10-17
Fix broken links

h-mayorquin:
I introduced this wrong link in https://github.com/catalystneuro/neuroconv/pull/1546

I am also removing the link to the catalyst neuro page which keeps failing intermittently.


weiglszonja (approved):
ok

PR #1577, opened 2025-11-07
Add multi-channel, volumetric and multi-file support to `TiffImagingInterface`

h-mayorquin:
This will replace #1457. the work [outlined there](https://github.com/catalystneuro/neuroconv/pull/1457#issuecomment-3208567275) is completed here

PR #1578, opened 2025-11-08
Add interactive visualization for dimension order in `TiffImagingInterface`

h-mayorquin:
Thinking on whether an interactive visualization widget embedded on the rest would help our users understand the concept of dimension order. This is in top of #1577 and has it as a base.

Check out the visualization here:
https://neuroconv--1578.org.readthedocs.build/en/1578/conversion_examples_gallery/imaging/tiff.html


alessandratrapani (approved):
Looks good, I think it's very clear

PR #1589, opened 2025-11-17
Remove video interface

h-mayorquin:
The time for removal has come. 

The `LightningPoseConverter` is still using this so let's remove this later to make this PR simpler.


h-mayorquin:
> This looks good. Could you write an issue for LightningPoseConverter to use ExternalVideoInterface and/or InternalVideoInterface? I don't think we should just leave a private legacy video interface hanging around forever.

Good idea, done here:
https://github.com/catalystneuro/neuroconv/issues/1594

PR #1600, opened 2025-11-22
Add `SpikeGLXSyncChannelInterface`

h-mayorquin:
This will be used in the IBL project.

### DIFF
--- a/tests/test_on_data/ecephys/test_spikeglx_converter.py
+++ b/tests/test_on_data/ecephys/test_spikeglx_converter.py
@@ -209,6 +209,44 @@ class TestMultiProbeSpikeGLXConverter:
         assert "NeuropixelsImec1" in nwbfile.electrode_groups
         assert len(nwbfile.electrode_groups) == 2
 
+    def test_streams_argument_filters_data(self):
+        """Test that the streams argument correctly filters which data is added to NWB."""
+        # Get all available streams first
+        all_streams = SpikeGLXConverterPipe.get_streams(folder_path=self.test_folder)
+
+        # Filter to only AP streams (exclude LF and sync)
+        ap_only_streams = [s for s in all_streams if ".ap" in s and "SYNC" not in s]
+
+        converter = SpikeGLXConverterPipe(
+            folder_path=self.test_folder,
+            streams=ap_only_streams,
+        )
+
+        nwbfile = converter.create_nwbfile()
+
+        # Verify AP streams ARE present (4 segments each for 2 probes = 8 total)
+        assert "ElectricalSeriesAPImec00" in nwbfile.acquisition
+        assert "ElectricalSeriesAPImec01" in nwbfile.acquisition
+        assert "ElectricalSeriesAPImec02" in nwbfile.acquisition
+        assert "ElectricalSeriesAPImec03" in nwbfile.acquisition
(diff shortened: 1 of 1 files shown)


tests/test_on_data/ecephys/test_spikeglx_converter.py
```diff
@@ -251,8 +283,10 @@ class TestSortedSpikeGLXConverter:
     def test_multi_probe_multi_stream_example(self, tmp_path):
         """dataset with two probes and both ap and lf streams"""
         # Initialize base SpikeGLX converter (notebook example 1)
+        # Disable sync channels since this test focuses on sorted units
         spikeglx_converter = SpikeGLXConverterPipe(
-            folder_path=SPIKEGLX_PATH / "multi_trigger_multi_gate" / "SpikeGLX" / "5-19-2022-CI0"
+            folder_path=SPIKEGLX_PATH / "multi_trigger_multi_gate" / "SpikeGLX" / "5-19-2022-CI0",
+            # include_sync_channels removed - use streams parameter to exclude
```
pauladkisson: I don't see any tests that actually use the streams parameter. Maybe I'm missing something, but that seems like a pretty critical parameter that should be tested.

pauladkisson: I agree, there should be one - and preferably only one - obvious way to do things. 

Even if we remove the streams parameter later, though, I think it should be tested in the interim. Since its behavior was changed by this PR, I think it's within the scope.

h-mayorquin: I am fine with adding test but where is that the behavior was changed? So I can focus the test on that.

tests/test_on_data/ecephys/test_spikeglx_converter.py
```diff
@@ -174,11 +142,110 @@ def test_multi_probe_spikeglx_converter(self):
...
+        assert "NeuropixelsImec0" in nwbfile.devices
+        assert "NeuropixelsImec1" in nwbfile.devices
+        assert len(nwbfile.devices) == 2
+
+        # Check electrode groups
+        assert "NeuropixelsImec0" in nwbfile.electrode_groups
+        assert "NeuropixelsImec1" in nwbfile.electrode_groups
+        assert len(nwbfile.electrode_groups) == 2
+
+    def test_streams_argument_filters_data(self):
```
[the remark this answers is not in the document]
pauladkisson: Beautiful!

tests/test_on_data/ecephys/test_spikeglx_converter.py
h-mayorquin: Ah, I see. Yes, before we did not have the feature and they were not included `SpikeGLXSyncChannelInterface`. Then, following the principle that the converter `SpikeGLXConverterPipe` should include EVERYTHING I aded it there.

Now, within the development of this branch, [at the start](https://github.com/catalystneuro/neuroconv/pull/1600/commits/913325744a371557e2d3d2a06c8d9c98cf471259), I did not add the sync channel to the convert by default. But then I thought that the converter should add everything available and that people could exclude that by using either of the two ways mentioned above so I eliminated that aforementioned bool.

Issue #1154, opened 2024-12-11
`SpikeGLXNIDQInterface` interface metadata 


h-mayorquin:
Currently, we are working on adding digital channels to SpikeGLXNIDQInterface and changing the data type from ElectricalSeries to TimeSeries (see #1152). The goal of the SpikeGLXConverterPipe is to automatically include as much information and metadata as possible, which includes handling NIDQ data.

NIDQ devices can acquire signals of very different natures. These can represent anything from temperature and motion to MEG or other non-neural physiological markers. Unfortunately, SpikeGLX metadata does not include channel-level descriptions ([reference](https://github.com/billkarsh/SpikeGLX/blob/master/Markdown/Metadata_30.md#nidq)), so we need to provide a way for users to specify this metadata and store it appropriately.

**Digital channels**
Each digital channel is written as its own ndx-events object, making it straightforward to annotate these channels individually. However, analog channels present a different issue.

**Analog channels**
Currently, all analog signals are written into a single TimeSeries. The problem is that TimeSeries does not support per-channel descriptions—only a single, general description. One possible workaround is to include a table in the TimeSeries description with details for each channel. Another option is to create one TimeSeries per channel, writing the channel’s description directly into that TimeSeries.

**Missing features**
* If we choose the one-TimeSeries-per-channel approach for analog signals, it would help to link all these TimeSeries objects to the same device.
* If we keep a single TimeSeries for all analog signals, it would be ideal to have per-channel descriptions rather than a single description for the entire dataset.
Neither linking multiple TimeSeries objects to a single device nor providing per-channel descriptions is currently supported by TimeSeries. I will open issues in pynwb to discuss these options.

PR #1601, opened 2025-11-23
Add support for metadata customization and channel separation in `SpikeGLXNIDQInterface`

h-mayorquin:
This feature will be used on the IBL project and should be useful for anyone that wants to properly annotate their NIDQ channels.

See https://github.com/catalystneuro/neuroconv/issues/1154


weiglszonja (approved):
Looks great, tried this in https://github.com/catalystneuro/IBL-widefield-to-nwb/pull/7

PR #1609, opened 2025-11-25
Stop propagating inter-sample shift on SpikeGLX interface

h-mayorquin:
On #1597 we added the true provenance information. On this PR we remove the derived quantity inter-sample-shift.

See:
https://github.com/SpikeInterface/probeinterface/issues/367#issuecomment-3419914104

For background context.


weiglszonja (approved):
Looks good to me

PR #1627, opened 2026-01-05
Fix Dev Tests

pauladkisson:
Added numba as dependency for sorting_analyzer.

See https://github.com/SpikeInterface/spikeinterface/issues/4290 for details about the bug.

See https://github.com/catalystneuro/neuroconv/actions/runs/20729573168 for dev tests


h-mayorquin (approved):
It is good that we separated this into its own dependency group. Thanks for sorting this one out.

PR #1636, opened 2026-01-22
Made dichroic_mirror optional

pauladkisson:
Depends on https://github.com/catalystneuro/ndx-fiber-photometry/pull/50 and subsequent release.


alessandratrapani (approved):
Agree, it can be optional.
Looks good to me. Maybe just check the failing tests before merging

PR #1651, opened 2026-01-29
Improve naming on test spikeinterface

h-mayorquin:
From a suggestion of @weiglszonja here:

https://github.com/catalystneuro/neuroconv/pull/1640#discussion_r2736078754


weiglszonja (approved):
Thanks for taking time to change this, I think the tests are much more readable this way

PR #1653, opened 2026-01-30
New ophys metadata structure: documentation

### LINKED ISSUE #1371: How to add planes for Ophys data (either generally or with caiman data)?
Hi,

I'm looking to see if there's a guide to using neuroconv with caiman data (cnmfe) with multiple planes (each plane in separate file and folder). 

I've been looking at the Suite2p data interface for some guide on how to implement multi-plane, but I'm still confused which metadata fields need to be updated as a general strategy. 

Could you give some pointers for either a general strategy to convert multi-plane ophys data, or a specific strategy with caiman data?  

Thanks!

h-mayorquin:
This PR is meant as a design document and to discuss. A chained PR with the implementation will be opened after this.

The two main documentation to explain the new schema is here:
https://neuroconv--1653.org.readthedocs.build/en/1653/how_to/annotate_ophys_metadata.html
And here:

https://neuroconv--1653.org.readthedocs.build/en/1653/developer_guide/ophys_metadata_structure.html

And can be consulted on the build of this PR documentation.

For specific illustration of the problems with the metadata here you can check #1371 and also the tricks that Suit2P has to implement to handle this:

https://github.com/catalystneuro/neuroconv/blob/57821fc145f8e548c161ff6d903e967b2ff21506/src/neuroconv/datainterfaces/ophys/suite2p/suite2pdatainterface.py#L10-L12

Once is merged it should close #1371

Other relevant issues are #269 #629 and #1280


pauladkisson (commented):
Added my thoughts.

docs/developer_guide/ophys_metadata_structure.rst
```diff
@@ -0,0 +1,326 @@
...
+
+       metadata["Ophys"]["ImagingPlanes"]["visual_cortex"]["indicator"] = "GCaMP6s"
+
+
+2. **Consistent metadata_key Across Interfaces**
+   Every ophys interface uses a single ``metadata_key`` parameter that propagates to all its
+   components (Device, ImagingPlane, PhotonSeries). This provides a consistent pattern across
+   all interfaces, making the API predictable and easier to learn.
+
+3. **Explicit References**
```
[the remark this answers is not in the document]
h-mayorquin: Good point. Here's a concrete example of the shared device case you're describing (two imaging planes in different locations sharing one microscope):

```python
metadata["Devices"]["shared_microscope"] = {
    "name": "Microscope",
    "description": "Two-photon microscope used for both planes",
    "manufacturer": "Thorlabs",
}

metadata["Ophys"]["ImagingPlanes"]["plane_area1"] = {
    "name": "ImagingPlaneArea1",
    "description": "Imaging plane in visual cortex area 1",
    "location": "V1",
    "device_metadata_key": "shared_microscope",
}

metadata["Ophys"]["ImagingPlanes"]["plane_area2"] = {
    "name": "ImagingPlaneArea2",
    "description": "Imaging plane in visual cortex area 2",
    "location": "V2",
    "device_metadata_key": "shared_microscope",
}
```

Since both imaging planes reference `"shared_microscope"` via `device_metadata_key`, both will be linked to that device. The device itself will be created in the NWB file by whichever interface writes its microscopy series first.

Importantly, devices are only created from entries in `metadata["Devices"]` that are actually referenced by other objects (e.g. via `device_metadata_key`). Entries that exist in the dict but are not referenced by anything will not be written to the file. This means you can have all the devices of a conversion defined in a shared YAML but only the ones that are actually linked will end up in the NWB file.

On the question of "who creates the device": devices are created at writing time. If an object (like an ImagingPlane) requires a device but no `device_metadata_key` is set, a default device will be created automatically. If a `device_metadata_key` is set, the referenced entry from `metadata["Devices"]` will be used instead.

This ties into how `get_metadata()` should work: it should be a provenance method. If the source data contains meaningful device information, `get_metadata()` will return an entry under the interface's `metadata_key` with everything linked automatically. If the source data has no device information, `get_metadata()` will not return a device link (it does not return things it doesn't know about). For shared resources across interfaces, the user (or the converter) needs to set those links explicitly. You can either create a new entry (like in the example above) or just re-map an existing entry from another interface using the metadata_key at the right level.

This design keeps `get_metadata()` as true provenance while also allowing the metadata dict to serve as a space for "metadata programming": you can define all your devices, planes, etc. in the YAML and then use the linking mechanism to select which elements are actually used depending on the situation.

docs/developer_guide/ophys_metadata_structure.rst
```diff
@@ -0,0 +1,326 @@
...
+   Components reference each other using explicit ``_metadata_key`` fields. This makes
+   relationships between components clear and enables validation.
+
+4. **Top-Level Devices**
+   Devices are stored at the top level (``metadata["Devices"]``) enabling device sharing
+   across ophys, ecephys, and other modalities.
+
+5. **Provenance-First get_metadata()**
```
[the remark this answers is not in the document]
h-mayorquin: Yes!

docs/developer_guide/ophys_metadata_structure.rst
```diff
@@ -0,0 +1,326 @@
...
+   Metadata is organized using dictionaries with meaningful keys. This structure makes metadata
+   easier to reference, organize, and extend. Dictionaries allow direct access to specific
+   components by name, which is clearer and less error-prone than positional access.
+
+   .. code-block:: python
+
+       metadata["Ophys"]["ImagingPlanes"]["visual_cortex"]["indicator"] = "GCaMP6s"
+
+
+2. **Consistent metadata_key Across Interfaces**
```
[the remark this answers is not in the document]
h-mayorquin: Thanks for mentioning this. I actually did not think very carefully about this before. I have improved the related section. Let me know what you think?

docs/developer_guide/ophys_metadata_structure.rst
```diff
@@ -0,0 +1,418 @@
...
+
+By adopting similar patterns (``MicroscopySeries``, ``RoiResponses``), NeuroConv's metadata
+structure will require minimal changes when ndx-microscopy is integrated into NWB core.
+This makes the eventual transition smoother for users.
+
+
+Linking and Object Creation
+---------------------------
+
+Each interface's goal is to create a main NWB object. For example, an imaging interface creates
```
pauladkisson:
```suggestion
Each interface's goal is to create its **primary object(s)** in NWB. For example, an imaging interface creates
```

pauladkisson (commented):
Beautiful! This is much clearer. I just made a couple of small nits.

docs/developer_guide/ophys_metadata_structure.rst
```diff
@@ -0,0 +1,418 @@
...
+            self,
+            *,  # Force keyword-only
+            verbose: bool = False,
+            metadata_key: Optional[str] = None,
+            **source_data,
+        ):
+            self.metadata_key = metadata_key
+            ...
+
+The argument name ``metadata_key`` is the same across all interfaces ensuring a common API.
```
pauladkisson, line 192: @h-mayorquin this is the section I believe you were referencing in #1708.

It seemed like a different case to me at first glance, since, for the interfaces in that PR, the metadata key is not dynamically generated. It might be worth adding a line to just mention that in the case that the interface has only one configurable output, then the metadata key may be hard coded or something like that.

h-mayorquin: Yes, for sure, thanks for checking.

docs/developer_guide/ophys_metadata_structure.rst
```diff
@@ -195,6 +195,14 @@ parameters that make the interface unique (e.g. stream name, channel name). When
 passes an explicit value, they take responsibility for uniqueness and can use it to
 intentionally share or customize metadata keys.
 
+The default is ``None`` rather than a hardcoded string (e.g. ``"caiman_segmentation"``) for
```
h-mayorquin, line 198: @pauladkisson

pauladkisson (commented):
Changelog?

h-mayorquin:
> Changelog?

Added

pauladkisson (approved):

PR #1658, opened 2026-02-11
Add how to for filtering ROIs in segmentation extractor

### LINKED ISSUE #1459: Removing rejected segmented ROIs (or ROIs with very bad quality) to save space
I'm using neuroconv with the caiman interface, and I'm looking for guidance on how to remove rejected ROIs when converting into NWB. 

**Rationale**: the reason for why such removal is preferred is because the folks who ran the Caiman chose the strategy to have high number of segmented ROIs (1000 - 1500 per plane), then QC them later to only accept 50 - 100 ROIs. Since this is 1-photon with 20Hz recording for ~30 min with 3 planes, the time series of these unwanted ROIs seem to take up quite a lot of space unnecessarily in the resulting NWB file. And since it's continuous values, I suspect compression doesn't help much?

**Potential strategy**: Hence, I was thinking of removing these ROIs in the `PlaneSegmentation` object / table. Would that be sufficient to automatically removing all "traces" of those ROIs from other objects as well, e.g. the time series in `DfOverF` and `Fluorescence`? I suspect it is possible, since they are linked, right? Is it recommended to do this by the way?

h-mayorquin:
Add roi_ids_to_add parameter to segmentation interfaces to select which ROIs to include when writing to NWB, along with a roi_ids property and a how-to guide.

Closes #1459

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Fixed `get_json_schema_from_method_signature` to skip `*args` (VAR_POSITIONAL) parameters, which was causing schema validation errors when methods used the `*args` pattern for deprecating positional arguments. [PR #1647](https://github.com/catalystneuro/neuroconv/pull/1647)
 
 ## Features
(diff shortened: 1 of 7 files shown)


pauladkisson (commented):
One minor comment. Otherwise, looks good.

pauladkisson (approved):

Issue #1659, opened 2026-02-12
Prevent encoding regressions by enabling EncodingWarning in test suite


h-mayorquin:
PR #1657 fixes all existing `open()` calls to include explicit `encoding="utf-8"`, but nothing currently prevents new code from reintroducing the same problem. Python 3.10+ supports `EncodingWarning` (PEP 597), which fires whenever `open()` is called without an explicit encoding. We can turn this into a test failure scoped to neuroconv code by adding `"error::EncodingWarning:neuroconv"` to the existing `filterwarnings` list in `pyproject.toml` and setting the `PYTHONWARNDEFAULTENCODING=1` environment variable in CI. This will catch any future `open()` call in neuroconv that omits encoding, without producing false positives from third-party dependencies.

PR #1660, opened 2026-02-12
Add regression testing for open without encoding

h-mayorquin:
Solves issue #1659 

Note that this failed (which means it works) until I fixed the fiber photometry for TDT


pauladkisson (commented):
The general idea makes sense.

The tests are failing for Python 3.13 -- looks like `subprocess.Popen` also needs an encoding specified: https://github.com/catalystneuro/neuroconv/blob/9f2c0e673d6b0c382502dd01767dd063feb0955f/src/neuroconv/tools/processes.py#L21

pauladkisson (approved):

PR #1662, opened 2026-02-12
Fix native timestamps logic on roiextractors.

h-mayorquin:
See https://github.com/catalystneuro/roiextractors/pull/552


pauladkisson (approved):
Looks good to me!

PR #1663, opened 2026-02-13
Deprecate positional conversion options in `add_to_nwbfile` and interface init

h-mayorquin:
Allows only nwbfile and metadata for `add_to_nwbfile`
Allows file_path, file_paths and folder_path for `__init__`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v0.9.2 (Upcoming)
 
 ## Removals, Deprecations and Changes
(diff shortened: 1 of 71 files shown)


pauladkisson (approved):
It was a lot of files, so I wanted to try out Copilot review to see if it would catch anything that I missed. As far as I can tell, the changes look good. Copilot's review seems to have caught a bunch of off-target stuff. Some of which may be legitimate and others that are really not. So I'll let you decide what to do with all that.

PR #1667, opened 2026-02-13
Fix conversion options validation failing on type and callable objects

bendichter:
## Summary
- `validate_conversion_options` was updated in #1139 to JSON-encode conversion_options before schema validation. This broke passing `progress_bar_class` (a `type` object) and callback functions in `progress_bar_options`, as `_NWBConversionOptionsEncoder` had no handler for these types.
- Add handling for `type` and `callable` objects in `_NWBConversionOptionsEncoder`, serializing them to their qualified module path string for validation purposes. The original objects are still passed through to the actual conversion code unchanged.

## How this manifests

NWB GUIDE sets `progress_bar_class=TQDMProgressSubscriber` (a class, not an instance) in `iterator_opts` within `conversion_options` at [`manage_neuroconv.py:1021`](https://github.com/NeurodataWithoutBorders/nwb-guide/blob/support-neuroconv-latest/src/pyflask/manageNeuroconv/manage_neuroconv.py#L1021):

```python
options_to_update["iterator_opts"] = dict(
    display_progress=True,
    progress_bar_class=TQDMProgressSubscriber,  # <-- type object
    progress_bar_options=progress_bar_options,   # <-- contains a callable
)
```

This dict ends up in `conversion_options`, which is passed to `NWBConverter.run_conversion()`. Inside `run_conversion` ([`nwbconverter.py:319`](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/nwbconverter.py#L319)):

```python
self.validate_conversion_options(conversion_options=conversion_options)
```

Which calls ([`nwbconverter.py:201`](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/nwbconverter.py#L201)):

```python
encoded_conversion_options = _NWBConversionOptionsEncoder().encode(conversion_options)
```

The encoder has no handler for `type` or `callable` objects, so it falls through to `json.JSONEncoder.default()` which raises:

```
TypeError: Object of type type is not JSON serializable
```

At runtime (after validation), the conversion_options dict is unpacked as kwargs into each interface's `add_to_nwbfile()` ([`nwbconverter.py:253`](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/nwbconverter.py#L253)):

```python
data_interface.add_to_nwbfile(nwbfile=nwbfile, metadata=metadata, **conversion_options.get(interface_name, dict()))
```

This passes `iterator_opts` as a kwarg, which flows through to `SpikeInterfaceRecordingDataChunkIterator` / `ImagingExtractorDataChunkIterator`, then to hdmf's `GenericDataChunkIterator`, where `progress_bar_class` is used as a constructor. This runtime path works fine — the issue is only in the validation step.

## Before vs after #1139

Before #1139, validation was:
```python
validate(instance=conversion_options or {}, schema=self.get_conversion_options_schema())
```
`jsonschema.validate()` traverses the dict but doesn't serialize it, so class objects were fine.

After #1139:
```python
encoded = _NWBConversionOptionsEncoder().encode(conversion_options)
decoded = json.loads(encoded)
validate(instance=decoded, schema=self.get_conversion_options_schema())
```
The JSON encode/decode round-trip fails on non-serializable types.

## Test plan
- [ ] Verify that passing `progress_bar_class` as a type object in `conversion_options` no longer raises `TypeError: Object of type type is not JSON serializable`
- [ ] Verify that passing callable objects in `progress_bar_options` works
- [ ] Verify that normal conversion options (strings, bools, dicts) still validate correctly


h-mayorquin (approved):
Needs a test and a changelog but LGTM.

h-mayorquin (approved):

PR #1669, opened 2026-02-20
Improve units table write performance with column-first strategy for new tables

h-mayorquin:
Previously, when writing a units table for the first time, spike times, waveform means, waveform standard deviations, electrode mappings, and all sorting properties were inserted row-by-row via HDMF's `add_unit()`. This is slow because each `add_unit()` call internally flattens and copies ragged array data (like spike times) one row at a time, creating unnecessary intermediate Python objects and memory copies that accumulate for large tables. This was particularly problematic for the IBL conversion pipeline, where a single session could consume more than 32 GB of RAM. This PR changes the first-write path to establish all rows in bulk and then add data as columns, avoiding the per-row overhead entirely.

I profiled peak RSS during units table writing on two sessions with large spike counts (1,366 and 2,185 units across 2 probes). The larger session has 3.3 GiB of per-spike ragged data on disk (spike times, amplitudes, and depths). Before these optimizations, it hit ~29.5 GB RSS and was OOM-killed on a 32 GiB machine. After this and https://github.com/hdmf-dev/hdmf/pull/1403 the peak RSS dropped to 13.8 GiB, a reduction of over 50%. On the smaller session (918 MiB on disk), I saw a similar 49% reduction (7,242 MiB to 4,597 MiB). With `np.concatenate`, each ragged column now costs exactly 1x its raw data size instead of 4.5x.

### DIFF
--- a/src/neuroconv/tools/spikeinterface/spikeinterface.py
+++ b/src/neuroconv/tools/spikeinterface/spikeinterface.py
@@ -2679,7 +2679,9 @@ def add_sorting_analyzer_to_nwbfile(
         recording, nwbfile=nwbfile, metadata=metadata, null_values_for_properties=null_values_for_properties
     )
     electrode_group_indices = _get_electrode_group_indices(recording, nwbfile=nwbfile)
-    unit_electrode_indices = [electrode_group_indices] * len(sorting.unit_ids)
+    unit_electrode_indices = (
+        [electrode_group_indices] * len(sorting.unit_ids) if electrode_group_indices is not None else None
+    )
 
     _add_units_table_to_nwbfile(
         sorting=sorting_copy,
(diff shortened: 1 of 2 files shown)


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -2092,33 +2167,89 @@ def _add_units_table_to_nwbfile(
...
+            [
+                sorting.get_unit_spike_train(
+                    unit_id=unit_ids[row_index],
+                    segment_index=segment_index,
+                    return_times=True,
+                )
+                for segment_index in range(num_segments)
+            ]
+        )
+        all_spike_times.append(spike_times)
```
alejoe91: I wonder if this could be sped up with the [`spike_vector_to_spike_train` numba accelerated function](https://github.com/SpikeInterface/spikeinterface/blob/main/src/spikeinterface/core/sorting_tools.py#L17)
This will give you in one shot all spike trains for all units and segments, that can then be concatenated and converted to times

alejoe91 (approved):
LGTM @h-mayorquin 

Definitely getting all data in bulk instead of retrieving it for each unit is a smart move :)

h-mayorquin:
18 % faster on the large dataset, better but not something to write home. At least we know there is no obvious regression.

PR #1670, opened 2026-02-20
Fix PEP 563 string annotations in JSON schema generation

bendichter:
## Summary
- When a module uses `from __future__ import annotations` (PEP 563), all type annotations are stored as strings at runtime
- `get_json_schema_from_method_signature()` passes these string annotations to `pydantic.create_model()`, which fails because pydantic cannot resolve types like `DirectoryPath` from their string form
- Uses `typing.get_type_hints()` to resolve string annotations back to their actual types before passing to pydantic
- When a class is passed (rather than a method), targets `__init__` for hint resolution since `inspect.signature` also uses `__init__` in that case

Currently affects `MiniscopeConverter`, which is the only interface using `from __future__ import annotations`.

## Test plan
- [ ] `MiniscopeConverter.get_source_schema()` succeeds instead of raising `PydanticUserError`
- [ ] All other interface schemas still generate correctly
- [ ] Existing tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Removals, Deprecations and Changes
 
 ## Bug Fixes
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
LGTM.

I added a test and the changelog.

PR #1682, opened 2026-03-03
Fix ndx-events version check to support v0.2.2

h-mayorquin:
This is coming from the IBL:

https://github.com/rly/ndx-events/issues/28

- Fixed `NotImplementedError` when using `ndx-events==0.2.2` with `configure_and_write_nwbfile`
- The version gate for ndx-events `Events` types was `<= 0.2.1`, excluding v0.2.2 which contains a critical bugfix ([rly/ndx-events#29](https://github.com/rly/ndx-events/pull/29))
- Changed to `< 0.3.0` to cover the entire v0.2.x series
- Bumped ndx-events pin from 0.2.1 to 0.2.2


alessandratrapani (approved):
Ok

PR #1677, opened 2026-03-03
Add dict-based metadata pipeline for imaging 

h-mayorquin:
Related to #269

Implements the dict-based metadata pipeline for imaging extractors as discussed in #1653.

`add_imaging_to_nwbfile` now accepts the new dict-based metadata format keyed by `metadata_key`. When old metadata is passed, the data goes through an alternative path, effectively segregating new and old formats. This keeps the functions backwards compatible and the diff simpler. New helper functions (`_add_imaging_plane_to_nwbfile`, `_add_photon_series_to_nwbfile`, `_add_device_to_nwbfile`) handle the dict-based format, while old list-based functions are renamed with an `_old_list_format` suffix to handle the old one.

We introduce `get_full_ophys_metadata()` which provides a user-facing example template (moving forward with #1557). When users provide metadata with missing required fields, a clear `ValueError` is raised listing the missing fields and their placeholders. See also https://github.com/NeurodataWithoutBorders/nwb-schema/issues/672 for some of the design rationale.

Closes #1653. Related: #1511, [nwb-schema#672](https://github.com/NeurodataWithoutBorders/nwb-schema/issues/672).

PR #1685, opened 2026-03-04
 Complete TestAddImaging tests for the dict-based metadata pipeline

h-mayorquin:
Completes the tests that were missing from #1677 to keep that diff focused on the implementation. The remaining tests are added here in the new pytest format so we can drop `TestAddPhotonSeries` entirely once we drop support for the list-based metadata format (September 2026).


pauladkisson (approved):
Looks good to me!

PR #1686, opened 2026-03-04
Add `OpenEphysBinaryConverter` for multi-stream OpenEphys binary data

### LINKED ISSUE #1504: Add `OpenEphysBinary` Converter
Currently we have two interfaces for the binary format of Open Ephys: the `OpenEphysBinaryRecordingInterface`, which handles neural signals, and the `OpenEphysBinaryAnalogInterface`, which handles analog signals.  

The Open Ephys binary format can describe a variety of experimental setups, such as multiple recording nodes, multiple streams (e.g., NIDQ and Neuropixels together), multiple probes, multiple experimental sessions in the same folder, and matching streams across recording nodes. See their docs:

https://open-ephys.github.io/gui-docs/User-Manual/Data-formats/Binary-format.html  

(I am also working on improving the docs [here](https://github.com/NeuralEnsemble/python-neo/pull/1774) as part of our ecosystem.)  

At the moment, users can aggregate data by pointing the neural and non-neural interfaces to the appropriate folders and then combining the interfaces themselves (see [here](https://github.com/catalystneuro/neuroconv/issues/1498#issuecomment-3287080912) for an example). However, this approach presents difficulties, such as needing to use the `es_key` parameter when more than one probe is present, and having to set the appropriate metadata so the electrode table and devices are written correctly.  

We should make this process easier for users by building a converter for the binary format, similar to what we did for the SpikeGLX acquisition system with `SpikeGLXConverterPipe`. The converter pipe would collect all available streams, parse the structure, add any necessary metadata for disambiguation of streams, and route them to the correct neurodata types in NWB.  

It appears that our GIN tests already contain data covering most of these cases, such as multiple blocks, multiple streams, and different types of acquisition plugins.

h-mayorquin:
Adds `OpenEphysBinaryConverter`, a converter subclass that auto-discovers all streams in an OpenEphys binary folder and routes neural streams to `OpenEphysBinaryRecordingInterface` and analog (ADC/NI-DAQ) streams to `OpenEphysBinaryAnalogInterface`. Follows the same pattern as `SpikeGLXConverterPipe`.

Closes #1504.

The conversion gallery entry is here:

https://neuroconv--1686.org.readthedocs.build/en/1686/conversion_examples_gallery/recording/openephys.html


alessandratrapani (approved):
Looks good to me! Nothing to add

PR #1689, opened 2026-03-09
Remove spike interface deprecations

h-mayorquin:


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -2510,7 +2434,7 @@ def write_sorting_to_nwbfile(
         if verbose:
             print(f"NWB file saved at {nwbfile_path}!")
 
-        return nwbfile  # Will return None in March 2026
+        return None
```
pauladkisson: Same comment as above.

src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -2920,7 +2786,7 @@ def write_sorting_analyzer_to_nwbfile(
...
```
pauladkisson: Same comment as above.

tests/test_modalities/test_ecephys/test_tools_spikeinterface.py
```diff
@@ -506,22 +505,8 @@ def setUp(self):
             session_description="session_description1", identifier="file_id1", session_start_time=testing_session_time
         )
 
-    def test_default_values_single_segment(self):
```
pauladkisson: I don't think that this test should be removed. You could just use a temp-path for the nwb-file path, and it seems like it would still be legitimate, or even swapping out write_recording to nwb-file with add_recording to nwb-file.

pauladkisson (commented):
Small comments; otherwise looks good.

tests/test_modalities/test_ecephys/test_tools_spikeinterface.py
h-mayorquin: I removed this test because it's already covered by [`TestAddElectricalSeriesWriting.test_default_values`](https://github.com/catalystneuro/neuroconv/blob/f00f81765b50d3ebf5a2fdf88610fbdfeec1efa5/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L59-L68). That test calls `add_recording_to_nwbfile` with a single-segment recording, asserts that `"ElectricalSeriesRaw"` is present in the acquisition module (verifying the default naming without numeric suffixes), and compares the extracted data against the original traces to check data integrity. These are the same three things that `test_default_values_single_segment` was testing.

pauladkisson (approved):
Looks good to me. Seems like a SpikeInterface update is breaking the tests.

Issue #1557, opened 2025-10-17
Provenance vs convenience: What should  `get_metadata()` return?


h-mayorquin:
Consider the common pattern for using `interface.get_metadata()`:

```python
metadata = interface.get_metadata()  # source metadata plus defaults
# user can modify metadata here
interface.add_to_nwbfile(nwbfile, metadata=metadata) # this would fail if we did not have defaults
```

Currently, this method returns **both** the metadata extracted from the source format and the defaults required to ensure that `add_to_nwbfile` does not fail. This is convenient because “the code just works.” However, this convenience has a downside: by adding defaults, we hide from users the fact that some required metadata fields are missing and have been filled with placeholders (see #1525).

After the refactor in #1511, this convenience is no longer necessary, since defaults are now added [explicitly and locally](https://github.com/catalystneuro/neuroconv/blob/6f10c989000e14e4016cbec1dceef1316776304e/src/neuroconv/tools/roiextractors/roiextractors.py#L286-L290) when constructing the neurodata type objects. We could therefore make `get_metadata()` a fully provenance-based method that returns only what is available in the source format. This approach is safer, as it avoids fabricating information and reduces the risk of missing metadata due to omission.

However, returning more complete metadata also has advantages. It gives users a kind of template they can easily modify. It is more intuitive to see a field like `emission_lambda` and update it than to see only the source metadata and later discover that other required fields are missing (often only after `add_to_nwbfile` fails).

This leads to a design decision about which role `get_metadata()` should play. Ordered from most provenance-faithful to most convenient, the options are:

- **Option 1 (Provenance):** Return only metadata extractable from the source format, without any defaults. This clearly reflects what came from the source files and nothing else.  
- **Option 2 (Current behavior):** Return source metadata merged with defaults for required values to prevent failures. This reflects what will actually be written to the NWB file.  
- **Option 3 (Discovery):** Return all possible NWB fields for the modality as a complete schema template. This maximizes discoverability by showing every field users could provide.  

I believe **Option 1 (Provenance)** is the right choice. The convenience of Options 2 and 3 is no longer necessary after #1511, and for a library focused on data conversion and provenance, we should avoid fabricating metadata whenever possible. A provenance-first approach ensures users know exactly what came from their data versus what they need to supply explicitly, following the principle that *explicit is better than implicit*.

[...]

PR #1694, opened 2026-03-12
Add `metadata_key` and `use_new_metadata_format` to imaging interfaces

h-mayorquin:
Continues with the project of #269.

 This is the first PR to implement the interface-layer conventions documented in #1653, building on the dict-based imaging pipeline added in #1677 and its functional tests in #1685.

This implements the changes for the following interfaces: `TiffImagingInterface`, `Hdf5ImagingInterface` and `SbxImagingInterface`, connecting them to the dict-based metadata pipeline. I have left out more complex interfaces (`ScanImageImagingInterface`, `ScanImageLegacyImagingInterface`, `BrukerTiffMultiPlaneImagingInterface`, `BrukerTiffSinglePlaneImagingInterface`, `ThorImagingInterface`, `FemtonicsImagingInterface`, `MicroManagerTiffImagingInterface`, `MiniscopeImagingInterface`, `InscopixImagingInterface`) to keep this diff easier to review. I will add them separately. `MockImagingInterface` was also updated to support both `metadata_key` and `use_new_metadata_format` so the unit test suite exercises the new paths.

Some design decisions:

`metadata_key` defaults to `None` in the base class and each subclass sets its own default. Interfaces with distinguishing parameters incorporate them into the key at construction time (e.g. `TiffImagingInterface` appends `_channel_{channel_name}` when a channel is specified). The parameter is always keyword-only across all interfaces.

For backwards compatibility, `get_metadata()` gains a `use_new_metadata_format` parameter that defaults to `False` so nothing breaks. Once all interfaces are migrated, a deprecation warning will be added before eventually removing the old format. When `True`, `get_metadata()` returns only provenance metadata that the source format natively provides. Formats without native metadata (like plain TIFF or HDF5) return an empty Ophys section, as discussed in #1557. Subclasses that do extract provenance from the source override this path: `SbxImagingInterface` populates both device metadata and a format-specific microscopy series description (`"Imaging data acquired with Scanbox."`), illustrating the pattern other vendor interfaces will follow. Ophys defaults are filled internally by `add_imaging_to_nwbfile()`. There is a corresponding TODO at `baseimagingextractorinterface.py` to switch `add_to_nwbfile` to use `self.get_metadata(use_new_metadata_format=True)` once all interfaces are migrated.

To keep the current metadata tests working, existing `check_extracted_metadata` overrides in on-data tests are renamed to `check_extracted_metadata_old_list_format` with TODO markers for removal. The `check_extracted_metadata` hook in the mixin now tests the new dict-based format. Interfaces not yet supporting the new format are skipped gracefully.


pauladkisson (approved):
Looks good to me!

Issue #562, opened 2023-09-12
[New Format]: multi-probe SpikeGLX


parviz-gh:
### What format would you like to see added to NeuroConv?

hello , it would be great to  add multi-shank/probe raw data selection (like lf or ap files that are generated by SpikGlx ) in nwb-guide.

### Does the format have any documentation?

No, as far as I know 

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

CodyCBakerPhD:
Hi!

If this request is specific to the NWB GUIDE, could you open it on that repo so it can be tracked?

https://github.com/NeurodataWithoutBorders/nwb-guide/issues/new/choose

As far as NeuroConv goes, the easiest thing for SpikeGLX is the `SpikeGLXConverter`

Instructions on how to use that class can be found here (can handle multiple probes, theoretically multiple gates as well - there are some planned improvements to metadata representation but the data itself should all make its way into the NWB file): https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/recording/spikeglx.html

CodyCBakerPhD:
Ah, but I see https://github.com/NeurodataWithoutBorders/nwb-guide/issues/356 has already been raised, strange I didn't receive notification of that one until I checked

CodyCBakerPhD:
@parviz-gh Have you had a chance to check out the `SpikeGLXConverter` yet? It should be capable of doing what you're requesting in this issue

parviz-gh:
@CodyCBakerPhD . Yes , I tried it . It works for single stream (one probe ). As far as I undrestood , all the functions aceept address of one stream NOT list of streams .

CodyCBakerPhD:
This was finally solved recently by #957

PR #1696, opened 2026-03-24
Fix compatibility with roiextractors PR #562

h-mayorquin:
The `get_channel_names()` method on imaging extractors internally calls `get_num_channels()`, which is removed in [roiextractors PR #562](https://github.com/catalystneuro/roiextractors/pull/562). This PR replaces those calls with a hardcoded `["OpticalChannel"]` default (imaging extractors are single-channel), switches the segmentation metadata path to use `_channel_names` directly, and removes the `get_num_channels` monkey-patch hack in `FemtonicsDataInterface`. The volumetric chunking test is also updated to compute the expected chunk shape from the extractor's dtype rather than hardcoding a value that assumed `uint16`.


pauladkisson (approved):
Looks good to me.

Issue #1536, opened 2025-10-03
Improvement and support roadmap for Miniscope


h-mayorquin:
On 2025-10-02 I had a meeting with @GarrettBlair, who has worked [very closely with Miniscope](https://www.science.org/doi/10.1126/sciadv.adg3918).

Garrett was very generous with his time and I am writing this issue to keep this information in a public place and to provide context for improving NeuroConv's Miniscope support in the future.

### 1. Sampling rate
The sampling rate we are currently extracting from the configuration file is unreliable. The sampling value in the configuration file is set by the user, and the DAQ only follows it on a best-effort basis. Crucially, the DAQ always provides timestamps, so we should use those as the source of truth.  

I have a PR in ROIExtractors that implements these changes:  
https://github.com/catalystneuro/roiextractors/pull/509

### 2. Configuration file and folder structure
From version 1.0 onward, the DAQ provides a configuration file (according to @GarrettBlair, earlier versions might have had a similar file, but we need to double-check). This [configuration file describes the folder structure](https://github.com/catalystneuro/roiextractors/issues/356#issuecomment-3325184204).  

Thanks to the work of @alessandratrapani, we now have a more [flexible Miniscope Extractor](https://github.com/catalystneuro/roiextractors/pull/441).  I am using these changes in #1528 to  automatically parse the folder structure and read all the data. See the PR for more details.

When the configuration file is not available, I fall back to the structure used in the Tye Lab for backward compatibility. This assumes that all the time folders represent back-to-back recordings, which is not generally true. We might want to revisit this assumption, but for now I am keeping it for backward compatibility.

In addition, I am adding documentation for users to build their own converter using the new flexible interface and adjust this themselves in #1524. This will be useful for cases when our assumptions does not hold and should work as a general fallback for this format.

We are in good shape here. The work of @alessandratrapani provides both (1) a flexible core that allows us to use the configuration file in the `MiniscopeConverter` #1528  and (2) a flexible base that users can adapt to build mostly automated solutions for edge cases #1524.

### 3. Head orientation
According to @GarrettBlair, the `headOrientation.csv` file is present most of the time.  
In [this issue](https://github.com/catalystneuro/neuroconv/issues/1535) I outlined the structure of this file, and I think we should add support to the converter so it is automatically written for the user’s benefit.

### 4. Metadata considerations
We discussed the particularities of Miniscope metadata. Most of it is already covered in the work by @weiglszonja with **ndx-miniscope**:  
[...]

Issue #509, opened 2023-07-10
[Bug]: Testing Suite failing after latest SI release


CodyCBakerPhD:
### What happened?

Our tests have been failing over the weekend since the latest SpikeInterface release

@h-mayorquin said he knew what was going on - do you think you can fix it quickly? It's starting to hold up other PRs now

### Steps to Reproduce

```python
Logs https://github.com/catalystneuro/neuroconv/actions/runs/5500781517/jobs/10023933116
```

### Operating System

Windows

### Python Executable

Python

### Python Version

3.9

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [X] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

CodyCBakerPhD:
Oky doky, going with option 2 for now

Thanks for clarifying

PR #1706, opened 2026-04-07
Deprecate legacy folder discovery mode in `MiniscopeConverter`

h-mayorquin:
When we recently improved Miniscope support across neuroconv ([#1528](https://github.com/catalystneuro/neuroconv/pull/1528), [#1604](https://github.com/catalystneuro/neuroconv/pull/1604)) and roiextractors ([#441](https://github.com/catalystneuro/roiextractors/pull/441), [#509](https://github.com/catalystneuro/roiextractors/pull/509)), we kept the changes small and preserved the legacy folder discovery path for backward compatibility. However, there is no reason to support it indefinitely. The configuration file is produced automatically by the Miniscope DAQ software (see [roiextractors#356](https://github.com/catalystneuro/roiextractors/issues/356) for background and [#1536 Section 2](https://github.com/catalystneuro/neuroconv/issues/1536) for a summary), so requiring it is not an extra burden on users.

In this PR I am adding a generous deprecation date for the legacy mode of `MiniscopeConverter` (when no `user_configuration_file_path` is passed). After December 2026 the parameter will become required. The legacy path assumes that all recordings within a session are back-to-back, which does not hold in general. Even in cases where a configuration file is not available, the current fallback behavior would be a silent error for users who did not have the Tye Lab folder structure in mind. Making the parameter required surfaces this explicitly rather than guessing wrong.  I have also updated the Miniscope conversion gallery entry to reflect this deprecation and to clarify that the "Combining Multiple Acquisitions" section is the recommended approach when a configuration file is not available. This covers the current Tye Lab workflow and any other workflow where the `configuration_file` is not available.


alessandratrapani (approved):
Yeah make sense to me

PR #1695, opened 2026-03-12
Add summary images to dict-based segmentation pipeline

h-mayorquin:
Closes the summary images gap left open by #1692. The dict-based segmentation pipeline now writes mean and correlation images alongside `PlaneSegmentation` and `RoiResponses`, matching the capability of the old list-based `_add_summary_images_to_nwbfile` in `roiextractors_pending_deprecation.py`.

The implementation follows the design already documented in #1653: a single shared `Images` container (fixed name `"SegmentationImages"`) holds all summary images from all segmentation pipelines. Per-image names and descriptions are configurable through `metadata["Ophys"]["SegmentationImages"][metadata_key]`, using the same `metadata_key` that indexes `PlaneSegmentations` and `RoiResponses`. When no metadata is provided, placeholder names (`correlation_image`, `mean_image`) are used from `_get_ophys_metadata_placeholders()`.

I have left the `Images` container name and description non-configurable to keep this consistent with `ImageSegmentation` and `Fluorescence`, which also use fixed names.

Related: #269, #1511, #1692, #1653, #1693

PR #1692, opened 2026-03-10
Add dict-based segmentation pipeline

h-mayorquin:
Continues with the plan of [#1653](https://github.com/catalystneuro/neuroconv/pull/1653). Adds new dict-based metadata functions for writing segmentation objects (`_add_plane_segmentation_to_nwbfile`, `_add_roi_response_traces_to_nwbfile`) with dual routing in `add_segmentation_to_nwbfile`. Note that a lot of related but non-core changes were done in [#1690](https://github.com/catalystneuro/neuroconv/pull/1690) (moving old functionality to new functions without changing behavior) so we can focus on the new functionality here.

This PR implements the most basic functionality, leaving background segmentation and summary images for a future PR so we can focus on the core of the changes. Like the imaging PRs ([#1677](https://github.com/catalystneuro/neuroconv/pull/1677) for implementation, [#1685](https://github.com/catalystneuro/neuroconv/pull/1685) for full tests), the comprehensive tests that are not strictly related to the metadata handling will be added in a follow-up PR.

Some design decisions I would like to highlight:
- **Omitted mask choice functionality**: the current pipeline supports this choice as a conversion option. Here, the minimal functionality of writing the image mask in the native format of the segmentation algorithm is implemented. We can discuss how to add support for format conversion, if desirable, later. Again, so we can focus on the core of the change.
- **Omitted centroids and acceptance**: we have moved to a system of properties that writes the properties of the native format as PlaneSegmentation columns (see [roiextractors#467](https://github.com/catalystneuro/roiextractors/pull/467) and [roiextractors#533](https://github.com/catalystneuro/roiextractors/pull/533)). The accepted list has been deprecated (see [roiextractors#524](https://github.com/catalystneuro/roiextractors/pull/524)) and [I think](https://github.com/catalystneuro/roiextractors/issues/556) we should support centroids as a property (not as a conversion option) but if you don't agree can add that later and focus on the main change here.
- **Single Fluorescence container**: all traces (including dff) go into one `Fluorescence` container instead of splitting between `Fluorescence` and `DfOverF`. This follows the direction of [nwb-schema#616](https://github.com/NeurodataWithoutBorders/nwb-schema/issues/616) and matches ndx-microscopy's single-container pattern (`MicroscopyResponseSeriesContainer`). We can discuss if we want to add support for the split but again, this simplifies the diff.

PR #1708, opened 2026-04-08
Add metadata_key support to base and simple segmentation interfaces

h-mayorquin:
Continues with the project of #269.

This is the segmentation counterpart of PR #1694, applying the same approach to the simplest segmentation extractors. It builds on the dict-based segmentation pipeline added in #1692 and the summary images support from #1695, which together provide the infrastructure this PR connects to at the interface layer.

`BaseSegmentationExtractorInterface` gains a `metadata_key` parameter stored as `self.metadata_key`, and `get_metadata` gains a keyword-only `use_new_metadata_format: bool = False` flag. When `True`, `get_metadata` returns only what the source file natively provides (provenance-first), leaving structure to the pipeline. `add_to_nwbfile` forwards `self.metadata_key` to `add_segmentation_to_nwbfile`, which already dispatches to the dict-based path when the key is present.

This PR migrates the four structurally simple interfaces (`CaimanSegmentationInterface`, `CnmfeSegmentationInterface`, `ExtractSegmentationInterface`, `SimaSegmentationInterface`) and `MockSegmentationInterface`. I have left out more complex interfaces (`Suite2pSegmentationInterface`, `MinianSegmentationInterface`, `InscopixSegmentationInterface`) to keep this diff easier to review. The `SegmentationExtractorInterfaceTestMixin` gains the same `test_get_metadata` / `check_extracted_metadata` / `check_extracted_metadata_old_list_format` pattern as `ImagingExtractorInterfaceTestMixin`.


src/neuroconv/datainterfaces/ophys/caiman/caimandatainterface.py
```diff
@@ -34,7 +34,7 @@ def get_extractor_class(cls):
         return CaimanSegmentationExtractor
 
     def __init__(
-        self, file_path: FilePath, *args, verbose: bool = False
+        self, file_path: FilePath, *args, verbose: bool = False, metadata_key: str | None = None
```
pauladkisson, line 37: Why are you using `None` here when you could just specify the default value directly in the `__init__`?

pauladkisson (commented):
Not sure why metadata key defaults to None instead of the actual string value of its default. For example, "caiman_segmentation".

src/neuroconv/datainterfaces/ophys/caiman/caimandatainterface.py
h-mayorquin: Consistency with multi-channel and multi-stream interfaces.

For multi-channel and multi-stream interfaces, this information should be encoded in metadata_key so that each instance has a distinct value. In these cases, stream_name or channel_name are provided at initialization, so metadata_key cannot have a fixed default in the signature, as its value is computed from that information. Setting `metadata_key = None` indicates that the default should be used, but this default can only be determined at construction time.

I make this consistent across all function signatures so they share the same type and semantics.

pauladkisson: I see. That seems very reasonable to me.

I was briefly considering if this information might be relevant to include in a comment or docstring somewhere, but I don't think there's any great place to put it, so I think it's probably just fine as is.

pauladkisson (approved):

src/neuroconv/datainterfaces/ophys/caiman/caimandatainterface.py
h-mayorquin: I think I added on the design docs:

https://github.com/catalystneuro/neuroconv/pull/1653

I will double check. At this point, after we merge this PR. I think we should merged the docs and do the final improvements as we add support for the last interfaces. 

Thanks a lot for helping me review all of this btw. This has been a lot of work on your side.

pauladkisson: No problem! :)

PR #1712, opened 2026-04-09
Fix recording segment start time retrieval for `spikeinterface>=0.104.1`

MGAMZ:
https://github.com/SpikeInterface/spikeinterface/pull/4462 introduces breaking change, `BaseRecording._recording_segments` no longer exists, and a new property `BaseRecording.segments` is available.

This issue causes NeuroConv `_add_recording_segment_to_nwbfile` function to fail.

This PR introduces a compatibility function `_get_recording_segment_start_time` which can handle both `spikeinterface<=0.104.0` and `spikeinterface>=0.104.1`.

The test unit is updated too.


h-mayorquin:
Has the change on SpikeInterface been released?

MGAMZ:
> Has the change on SpikeInterface been released?

Oh, not yet released, maybe we should wait until then? I discovered this issue on the its dev branch.

h-mayorquin:
I think we can simplify the fix. SpikeInterface already has a public `recording.get_start_time(segment_index=segment_index)` method available since `0.104.0` ([permalink](https://github.com/SpikeInterface/spikeinterface/blob/0.104.1/src/spikeinterface/core/baserecording.py#L432)), so we can replace the internal `recording._recording_segments[segment_index].t_start or 0` with that directly. This avoids the need for the `_get_recording_segment_start_time` compatibility shim entirely.

h-mayorquin (approved):
Thanks for catching this, and sorry that it sat for so long.

On the question you raised back in April: it turns out we don't need to wait for the release after all. The removal of `_recording_segments` is still unreleased (0.104.8 is the latest, and its `get_start_time` still reads that attribute internally), but the public `get_start_time` itself has been there since 0.104.0, and we now pin `spikeinterface>=0.104.7`. So we can move to the public API today and be correct on every version we support, instead of carrying a compatibility helper through the transition.

I hope you don't mind that I pushed to your branch directly: I merged main in and swapped the `_get_recording_segment_start_time` helper for a direct `recording.get_start_time(segment_index=...)` at both call sites, which let the helper and its fallback test go away. Your two `test_shifted_recording_uses_starting_time` tests and the removal of the direct writes to `_recording_segments` in the test suite are exactly what was needed here, so those stayed as you wrote them. I also added a CHANGELOG entry.

I ran into the same breakage independently in #1828 while fixing our dev dailies, but you called it out first, so I would rather this land as your PR and remove the duplicate changes from mine. Merging as soon as the tests pass. This should credit you properly.

MGAMZ:
@h-mayorquin
thanks for the detailed reply.
Glad I could help.

PR #1703, opened 2026-04-06
Add metadata_key support to Inscopix imaging interface

h-mayorquin:
Has #1702 as its base.

Adds metadata key support for the `InscopixImagingInterface`.

PR #1702, opened 2026-04-06
Add metadata_key support to MicroManager imaging extractor

h-mayorquin:
Has #1694 as its base.

Adds metadata key support for the `MicroManagerTiffImagingInterface`

PR #1716, opened 2026-04-20
Add metadata_key support to Suite2pSegmentationInterface and deprecate plane_segmentation_name

h-mayorquin:
Part of the ongoing migration of ophys interfaces to the dict-based metadata format. See PR #1694, PR #1702, PR #1703, PR #1704, PR #1705, and PR #1708 for prior work in this series.

Suite2p joins the migration. The new format surfaces source-provided `imaging_rate` on `ImagingPlanes`, links `MicroscopySeries` to the plane via `imaging_plane_metadata_key`, and populates `RoiResponses` and `SegmentationImages` entries for whatever trace types and summary images the extractor happens to have (raw / neuropil / deconvolved / dff for traces; correlation / mean for images). I have left the old list-based path untouched so backward compatibility is preserved until the format is removed.

I have also folded in a fix for `MicroManagerTiffImagingInterface`: its new format path was dropping `imaging_rate`, which the old path exposes. Added an `ImagingPlanes` entry and linked it from `MicroscopySeries`. Small PR on its own, made more sense to bundle.

`plane_segmentation_name` on `Suite2pSegmentationInterface.__init__` and `add_to_nwbfile` now emits a `DeprecationWarning` when passed explicitly. The parameter was introduced in PR #623 (issue #603) as an intermediate step: it served as both the NWB object name and the lookup key into nested metadata (Fluorescence, SegmentationImages) for multi-channel/multi-plane Suite2p output. In the dict-based format, `metadata_key` generalizes the pattern-discovery role across all ophys object types, and the NWB name becomes just a field inside the entry, so `plane_segmentation_name` is redundant. The warning points users to edit `metadata["Ophys"]["PlaneSegmentations"][metadata_key]["name"]` directly, and ties removal to the old-format removal. I have left the corresponding kwargs on the public `add_segmentation_to_nwbfile` tool function with a docstring deprecation rather than a runtime warning: interfaces still pass them through internally, and a runtime warning there would fire on every normal conversion.


pauladkisson (approved):
Looks good to me!

PR #1718, opened 2026-04-21
Fix: wrong default device added by `add_electrodes_to_nwbfile`

alejoe91:
This PR fixes a weird bug that creates a "ghost" default `Device` when adding electrodes.

If the user provides `metadata` with `Device` and `ElectrodeGroup` to the `add_electrodes_to_nwbfile`, these are completely ignored [here](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/spikeinterface/spikeinterface.py#L949-L954), which results in the creation of a `Device` and `ElectrodeGroups` addociated to that device.

The fix takes in consideration existing metadata and adds only groups which are in the recording and not in the provided metadata

### TODO:

- [x] Add test

### DIFF
--- a/src/neuroconv/tools/spikeinterface/spikeinterface.py
+++ b/src/neuroconv/tools/spikeinterface/spikeinterface.py
@@ -946,11 +946,25 @@ def add_electrodes_to_nwbfile(
         data = np.full(recording.get_num_channels(), fill_value=default_location)
         data_to_add["location"] = dict(description="location", data=data, index=False)
 
-    # Add missing groups to the nwb file
-    groupless_names = [group_name for group_name in group_names if group_name not in nwbfile.electrode_groups]
-    if len(groupless_names) > 0:
-        electrode_group_list = [dict(name=group_name) for group_name in groupless_names]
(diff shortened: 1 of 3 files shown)


src/neuroconv/tools/spikeinterface/spikeinterface.py
```diff
@@ -946,11 +946,25 @@ def add_electrodes_to_nwbfile(
...
+            missing_group_metadata["Ecephys"]["ElectrodeGroup"] = electrodes_group_metadata
+        else:
+            missing_group_metadata = dict()
+            electrodes_group_metadata = list()
+            missing_group_metadata["Ecephys"] = dict()
+            missing_group_metadata["Ecephys"]["ElectrodeGroup"] = electrodes_group_metadata
+        # Fill in missing groups
+        for group in electrode_group_list:
+            if group["name"] not in [g["name"] for g in electrodes_group_metadata]:
+                electrodes_group_metadata.append(group)
```
alejoe91, line 967:
```suggestion
                missing_group_metadata["Ecephys"]["ElectrodeGroup"].append(group)
```

PR #1719, opened 2026-04-22
Make `add_electrodes_to_nwbfile` private

h-mayorquin:
I thought about it and I think you are right. Thanks.

@alejoe91


alejoe91:
Replaces #1718

alejoe91 (approved):

PR #1721, opened 2026-04-22
Add metadata_key support to ScanImageImagingInterface

h-mayorquin:
Has [#1705](https://github.com/catalystneuro/neuroconv/pull/1705) as its base, continuing the imaging-interface migration chain that began with [#1694](https://github.com/catalystneuro/neuroconv/pull/1694) and passed through [#1702](https://github.com/catalystneuro/neuroconv/pull/1702), [#1703](https://github.com/catalystneuro/neuroconv/pull/1703), and [#1704](https://github.com/catalystneuro/neuroconv/pull/1704). This PR adds `metadata_key` and `use_new_metadata_format` support to `ScanImageImagingInterface` (v3/4/5).

The default `metadata_key` is derived from the runtime parameters so a multi-channel / multi-plane ScanImage dataset produces unique, non-colliding metadata entries out of the box, for example `scan_image_imaging_channel_1_plane_0`. When only `channel_name` is set the plane suffix is dropped, and when neither is set the key falls back to `scan_image_imaging`. `get_metadata(use_new_metadata_format=True)` emits a `Devices` entry whose description includes the ScanImage version when available, an `ImagingPlanes` entry carrying `imaging_rate` and, when the ScanImage header supplies them, `grid_spacing` and `origin_coords` in meters to match the other migrated interfaces, and a `MicroscopySeries` entry linked back to the imaging plane via `imaging_plane_metadata_key` and carrying `scan_line_rate` and a per-channel description.

The three modern ScanImage test classes now inherit from `ImagingExtractorInterfaceTestMixin` to match Thor, Femtonics, and the other migrated interfaces, with new-format coverage driven by the standard `check_extracted_metadata` hook. Getting there required one small fix to `ImagingExtractorInterfaceTestMixin.check_nwbfile_temporal_alignment`: it used to assert `series.starting_time == aligned_starting_time`, which presumes the written series stores timing as a `(starting_time, rate)` pair and implicitly assumes the original starting time was zero. The interleaved multi-plane ScanImage fixtures write timing as a full `timestamps` array instead, so that assertion fails with `None == 1.23`. The check now captures `interface.get_original_timestamps()[0]` before alignment and verifies whichever representation the series ended up with: `series.timestamps[0] == original_first_timestamp + aligned_starting_time` for the timestamps path, or `series.starting_time == original_first_timestamp + aligned_starting_time` for the rate path. This is backward compatible with every other migrated interface (their original first timestamp is 0, so `0 + 1.23 == 1.23` still holds) and removes the need for any per-class override on the ScanImage classes.


pauladkisson (approved):
Looks good to me!

PR #1723, opened 2026-04-23
Add metadata_key support to ScanImageLegacyImagingInterface

h-mayorquin:
This PR migrates `ScanImageLegacyImagingInterface` (v3.8 TIFFs) to the dict-based ophys metadata format, closing the ScanImage family started in [PR #1721](https://github.com/catalystneuro/neuroconv/pull/1721). The constructor gains a keyword-only `metadata_key` argument that defaults to `"scan_image_legacy_imaging"` and is forwarded to the base class so the same key threads through `add_to_nwbfile`.

`get_metadata` now accepts `use_new_metadata_format`. The default path preserves the existing behavior verbatim, dumping the raw ScanImage EXIF metadata as a JSON string into the `TwoPhotonSeries` description. When `use_new_metadata_format=True`, the method emits `Devices` keyed by `metadata_key` (with a static "Microscope controlled by ScanImage (legacy v3.8)" description), `Ophys.ImagingPlanes` (with `imaging_rate` linked to the Device via `device_metadata_key`), and `Ophys.MicroscopySeries` (with the JSON-dumped image metadata as description, linked to the ImagingPlane via `imaging_plane_metadata_key`). Session start time extraction from `state.internal.triggerTimeString` is shared between both branches.


pauladkisson (approved):
Looks good to me.

PR #1724, opened 2026-04-24
Support Intan recordings saved across multiple files

### LINKED ISSUE #789: [Feature] Handle all Intan modes and formats and add documentation

h-mayorquin:
Intan RHX can be configured to "create a new save file every N minutes" when recording in Traditional format, producing several rotated `.rhd` or `.rhs` files in one session folder. Pointing a single-file interface at one of them silently drops the rest of the session. This was first surfaced in [#519](https://github.com/catalystneuro/neuroconv/issues/519) and tracked under [#789](https://github.com/catalystneuro/neuroconv/issues/789); @weiglszonja's [comment](https://github.com/catalystneuro/neuroconv/issues/519#issuecomment-1983194576) describes the community workaround of building a custom interface on top of SpikeInterface's `ConcatenateSegmentRecording`. This PR adds a `saved_files_are_split: bool = False` parameter to `IntanRecordingInterface`, `IntanAnalogInterface`, and `IntanStimInterface` that delegates to `IntanSplitFilesRecordingExtractor` from SpikeInterface to concatenate the full session in filename order. When the flag is left at its default and sibling rotation files are detected, the interface emits a non-fatal warning pointing to it.

I chose a boolean flag on the existing interfaces over a separate `IntanSplitFilesRecordingInterface` or a `folder_path` parameter because the user's mental model is "I have an Intan recording that happens to span several files," not a different kind of input; the passive-voice name makes the `False` default read as a state description rather than a disabled action. I also added a "File formats and save modes" orienting section at the top of the Intan gallery page and a new "Converting a session saved as multiple files" section with a working doctest, since the three on-disk save modes were already supported transparently but the gallery did not surface what to pass as `file_path` for each.

Closes #789.


weiglszonja (approved):
Looks good to me, thanks!

Issue #572, opened 2023-09-18
Choosing an interface to store in NWB file


MKVarun:
I was wondering if there's a way to choose an interface class to store in NWB if I have several interface classes (OpenEphys, PhySorting, KiloSortSorting).

CodyCBakerPhD:
When you have multiple data streams to put into a single NWB file, you can easily choose which interfaces to combine in NeuroConv as shown in our galleries for combining interfaces, such as this one for SpikeGLX (NeuroPixels) + Phy

https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/combinations/spikeglx_and_phy.html

MKVarun:
So I actually created the interface this way (as in the documentation):

```
class ExampleNWBConverter(NWBConverter):
    data_interface_classes = dict(
        SpikeGLXRecording=SpikeGLXRecordingInterface,
        PhySorting=PhySortingInterface
    )
source_data = dict(
    SpikeGLXRecording=dict(
        file_path="raw_dataset_path"
    ),
    PhySorting=dict(
        folder_path="sorted_dataset_path"
    )
)

example_nwb_converter = ExampleNWBConverter(source_data)
```

But I wanted to select only one of the interfaces while converting to NWB, say PhySorting only.

CodyCBakerPhD:
Turns out, that's even easier!

Just don't include the interface in the `source_data` dictionary

All classes defined in the `data_interface_classes` are optional and only instantiated if defined by the `source_data` entries

CodyCBakerPhD:
@MKVarun Any luck getting this to work on your end?

I guess my big TODO from this issue would be to enhance the documentation to make this feature more apparent

MKVarun:
Yes, one of the solutions you provided works for me. 

My idea was to read metadata across the two folders (Spike and Phy) and then provide an option for users to include one or both of the folders, instead of having to create an interface again and repeat reading folders. 

For example, in my dataset, the metadata is stored in the Spike folders and I wanted to create NWB files with Phy data. What I am currently doing is create a converter for Spike and Phy folders get the metadata across the folders, create a converter for Phy/Spike and use the metadata extracted in the first case to create NWB files. However, this requires two Interface calls. It would be faster if I could call the Interface function on both folders at once, get the metadata across folders and then allow my user option to choose which of the two folders' data to include.

h-mayorquin:
[...]

PR #1725, opened 2026-04-24
Stop importing Miniscope format parsers from ndx-miniscope

h-mayorquin:
I am in the middle of the ophys metadata refactor (#1677, #1694, #1702, #1703, #1704, #1705, #1708, #1721) and the Miniscope interfaces are next on the list. One blocker I hit while scoping that work is that `MiniscopeBehaviorInterface` reaches into `ndx_miniscope.utils` to parse raw Miniscope files, which puts part of the interface's behavior inside an NWB extension package. This PR pulls that responsibility back into neuroconv, which is where it belongs: the extension package should be about defining the `Miniscope` neurodata type and its NWB constructors, not about reading vendor file formats.

V3 Miniscope references remain scattered across the imaging interface and converter, but we have no V3 test fixtures and none of the active paths exercise them. I am adding an explicit error inviting users to contribute data so we can add proper support. This is consistent with the cleanup work already in flight (roadmap #1536, legacy converter deprecation #1706, roiextractors #572): I would rather wait for real V3 data than ship speculative support against code we cannot test.


alessandratrapani (approved):
That's great! LGTM

PR #1728, opened 2026-05-07
Add warning when DeepLabCutInterface uses fallback timestamps

weiglszonja:
When no config file is provided DeepLabCutInterface uses DataFrame integer index ([0, 1, 2, ...]) as timestamps. This causes NWB files to be written with rate = 1.0 Hz. While this makes sense for a fallback, this should not be done silently but the very least let the user know that they should check the timestamps. 

This PR adds a `UserWarning` in that fallback path explaining that 1 Hz is being assumed and directing the user to `set_aligned_timestamps` to fix it. A test is added to verify the warning fires as expected.


h-mayorquin:
Do you think we should throw an error instead and force the user to pass a sampling rate in that case? I am not sure that writing a wrong sampling rate is the right behavior, what do you think?

h-mayorquin:
Closing this as #1937 has done the work.

This is a bug and not a missing warning. A DeepLabCut file's rows are video frames, and we write that row index into the pose series as if it were seconds. So any conversion without timestamps came out timed at 1 Hz, and that is the path the gallery documents.

I wanted the fix in the release so I finished it. Your commit is on it as co-author (thanks for catching this!).

PR #1731, opened 2026-05-07
Add metadata_key support to `MiniscopeImagingInterface`

h-mayorquin:
Continues the imaging-interface migration chain that began with [#1694](https://github.com/catalystneuro/neuroconv/pull/1694) and most recently passed through [#1721](https://github.com/catalystneuro/neuroconv/pull/1721). This PR adds `metadata_key` (default `"miniscope_imaging"`) and `use_new_metadata_format` to `MiniscopeImagingInterface`, sourcing `Devices`, `ImagingPlanes`, and `MicroscopySeries` entries from the extractor and `metaData.json`. 

The deprecated `_MiniscopeMultiRecordingInterface` is left on the existing list-based path since it disappears with the legacy `MiniscopeConverter` mode in December 2026 per [#1706](https://github.com/catalystneuro/neuroconv/pull/1706).


tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -980,6 +980,13 @@ def test_incorrect_folder_structure_raises(self):
         ):
             self.data_interface_cls(folder_path=folder_path)
 
+    def check_nwbfile_temporal_alignment(self):
```
pauladkisson, line 983: This change seems off target. Where is it coming from?

h-mayorquin: Originally `MiniscopeImagingInterfaceMixin` was maintained independently of the general `ImagingExtractorInterfaceTestMixin` because the miniscope interface was very different from the other imaging interfaces. I have been working to unify it across #1523, #1524, #1593, and #1725. The previous metadata-migration PRs already used the base mixin's routing mechanism to test both the new dict-based format and the old list-based one, and I wanted to reuse that here. Instead of duplicating it inside `MiniscopeImagingInterfaceMixin`, I took the logical step and finally made miniscope testing join the fold. Because of this, I also had to inherit the temporal-alignment testing as it is, which was doing nothing in the [old testing class](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/testing/data_interface_mixins.py#L322-L324). On `_MiniscopeMultiRecordingInterface` I am reproducing that previous behavior. 

See the comment on the test for more details.

pauladkisson: I see, thanks for the clarification.

pauladkisson (approved):

tests/test_on_data/ophys/test_imaging_interfaces.py
h-mayorquin: Thanks to you for the review and noticing this kind of thing that might be an oversight. Reviewing time is the scarce resource of the LLM era.

PR #1734, opened 2026-05-09
Migrate LightningPoseConverter to `ExternalVideoInterface`

h-mayorquin:
The May 2026 deprecation window on `LightningPoseConverter` has passed, so this PR removes the `external_mode` parameter, the warning about the old `Videos` metadata layout, and the `_convert_new_metadata_to_old` shim. The converter is also switched from the private `_VideoInterface` to the public `ExternalVideoInterface`, which is what the converter has effectively been doing all along (videos were always written external; `external_mode=False` was deprecated and then a no-op).

`get_metadata` now returns the `ExternalVideos` dict structure natively rather than the old `Videos` list, so the user-facing metadata shape matches everywhere else in neuroconv. `LightningPoseDataInterface` still reads video names from the old `Behavior.Videos` list, so the converter synthesizes a small stub `Videos` list before delegating to it. Migrating the data interface itself is a separate change and was not part of the May 2026 deprecation set. 

I have also kept the `add_to_nwbfile` positional-argument fallback in place: although its warning text said "May 2026", the codebase-wide deadline established by PR #1663 is August 2026, so it will be removed on that schedule alongside every other interface.


pauladkisson (approved):
Looks good.

PR #1735, opened 2026-05-09
Fix Inscopix multiplane detection

### LINKED ISSUE #1732: Inscopix Interface flags single-plane videos as multiplane
We have a series of recordings that are single-plane videos; however, the current InscopixImagingDataInterface throws the multiplane error. Since the pyisx API still does not have robust multiplane support, I was looking for alternatives. (I'm curious if the underlying isxd file format now includes a "multiplane" key? I didn't want to take the time to byte read one of our giant videos to find out.)

While the pyisx API does return `... 'Microscope Focus': 610, ...`, it is not clear whether this will be a list if there are multiple focal planes. Additionally, the `isx.utils.verify_deinterleave` function does not seem to have a c implementation yet, so that avenue still wont work.

As a very brute-force alternative, I added a `skip_multiplane_check` parameter to circumvent this issue until the `isx` API properly can detect these issues.

h-mayorquin:
This closes #1732. The interface previously decided whether a recording was multiplane by reading the file as bytes and searching for the literal string "multiplane". This is wrong. Most Inscopix files embed a `multiplane` block in the JSON metadata at the end of the file to record user-interface configuration state, regardless of how many planes were actually acquired, so the check rejected single-plane recordings from current acquisition software. A false positive that made the interface fail for most fails.

This PR fixes this by parsing the JSON metadata block directly from binary and reads the `microscope.multiplane.enabled` boolean instead of grepping for the substring. The dual-color test fixture, which is single-plane but carries the configuration block, had been asserted as expected-to-raise under the old behavior and now loads, with its device and subject metadata checked.  In addition, the reporter of #1732 shared sample recordings and confirmed that the interface loads their previously-rejected single-plane files.


tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -1240,90 +1240,81 @@ class TestInscopixImagingInterfaceMovieLongerThan3Min:
...
+        file_path = str(OPHYS_DATA_PATH / "imaging_datasets" / "inscopix" / "multiplane_movie.isxd")
+        with pytest.raises(NotImplementedError, match="Multiplane ISXD file detected"):
+            InscopixImagingInterface(file_path=file_path)
 
 
 @skip_on_python_313
 @skip_on_darwin_arm64
-class TestInscopixImagingInterfaceDualColorMovieWithDroppedFrames:
-    """Test InscopixImagingInterface with movie_longer_than_3_min.isxd (multiplane file that should raise NotImplementedError)."""
+class TestInscopixImagingInterfaceDualColorMovieWithDroppedFrames(ImagingExtractorInterfaceTestMixin):
```
h-mayorquin, line 1287: The changes here. This test was incorrectly flagged as multi-plane and therefore not read. Now, with the changes we can read it and we implement the full test with metadata.

tests/test_on_data/ophys/test_imaging_interfaces.py
```diff
@@ -1240,90 +1240,81 @@ class TestInscopixImagingInterfaceMovieLongerThan3Min:
...
-            "Multiplane ISXD file detected (found 'multiplane' in file).\n"
-            "This is a hacky check (not an official ISX API method) and may not be robust.\n"
-            "Proper separation logic is not yet implemented in roiextractors.\n"
-            "Loading as 2D would result in incorrect data interpretation.\n\n"
-            "Please open an issue at:\n"
-            "https://github.com/catalystneuro/roiextractors/issues\n\n"
-            "Reference: https://github.com/inscopix/pyisx/issues/36"
-        )
-        assert str(exc_info.value) == expected_message
+        file_path = str(OPHYS_DATA_PATH / "imaging_datasets" / "inscopix" / "multiplane_movie.isxd")
```
h-mayorquin, line 1270: This is the true multiplane data.

src/neuroconv/datainterfaces/ophys/inscopix/inscopiximagingdatainterface.py
```diff
@@ -6,25 +6,77 @@
...
+    # shown to be a real input.
+    metadata_tail = _read_isxd_json_metadata_tail(file_path)
+    microscope = (metadata_tail.get("extraProperties") or {}).get("microscope") or {}
+    multiplane_block = microscope.get("multiplane")
+    if multiplane_block is not None:
+        return bool(multiplane_block.get("enabled"))
+
+    # No `microscope.multiplane` block: an older file, or one rewritten by
+    # isx.Movie.write (which drops the whole extraProperties tree). We cannot
+    # tell, so assume single-plane and warn.
```
alessandratrapani, line 70: Agree!

alessandratrapani (approved):
Looks good to me!

h-mayorquin:
Thanks for the review @alessandratrapani

Issue #1269, opened 2025-03-26
New Interfaces should include metadata in their conversion gallery entries


pauladkisson:
Currently, most interfaces in the conversion gallery only provide the code to run the interface, but leave the metadata unspecified (which then get filled with defaults during the conversion).

In real conversions, I almost always need to edit some metadata (add a description here, a device there, etc.). The metadata for each interface is not stored in a standard way, but rather is idiosyncratic for each interface. As a result, in order to get the information I need to actually use an interface, I have to go into the code and figure out how it stores its metadata, which defeats the whole purpose of the conversion gallery (and as a result I basically never use it).

As an alternative, I recently added the [TDTFiberPhotometryInterface](https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/fiberphotometry/tdt_fp.html), which includes an example metadata.yaml file. This allows users to quickly and easily edit whichever metadata field they want without needing to dig into the code. I should also point out that it is much easier to take a complete document like this and omit certain entries than it is to start with an incomplete document (or no metadata example at all) and add metadata in the right format.

With these considerations in mind, I propose that new interfaces should follow the convention established by TDTFiberPhotometryInterface and include the metadata structure in their conversion gallery entry.

Clearly the conversion gallery entries are how-tos, so diataxis [docs](https://diataxis.fr/how-to-guides/) is an important reference here.

h-mayorquin:
Separation of Concerns.
I prefer to have separate how-tos with two different functions:

1. **Conversion Gallery (Current)**  
The purpose of this is to showcase that conversion is easy. It should look friendly and execute with the minimal amount of metadata required to simply write the file. Since we are a conversion automation library, this gallery demonstrates and conveys that conversion is easy and gets people code to hit the ground running.

2. **Per-Format Metadata Editing And Customization**  
This would be a new how-to for each format, focused on how to edit specific metadata or do any other common customization. We should link to this from the conversion gallery (or include it directly underneath). These guides would provide instructions for handling more detailed or format-specific requirements.

I agree that 2 is missing but I never intended the conversion gallery to be for users like @pauladkisson. Maybe it should be but I don't find it odd and I would rather have more documentation.

[...]

PR #1737, opened 2026-05-18
Add VAME interface

weiglszonja:
# Summary

Adds VameInterface (datainterfaces/behavior/vame/vamedatainterface.py) wrapping the ndx-vame NWB extension to write VAME (Variational Animal Motion Encoding) behavioral segmentation outputs to NWB.

Note: temporarily pinned to `ndx-vame @ git+https://github.com/catalystneuro/ndx-vame.git@main` until a new  release is available.

  - Writes per-frame motif labels (`MotifSeries`), with multiple algorithm runs supported per session (e.g. kmeans + hmm) — each becomes its own MotifSeries in the same VAMEProject
  - Writes latent-space embeddings (`LatentSpaceSeries`), shared across all motif series and linked via `latent_space_series`
  - Writes community labels (`CommunitySeries`), one per algorithm run, each optionally linked to its corresponding MotifSeries via `motif_series_key`
  - Serializes the full VAME config.yaml as JSON into VAMEProject.vame_config, and stores `time_window_samples`/`vame_version` on the container
  - Auto-discovers file paths from the standard VAME results layout (results/<session>/VAME/<algo>-<n_clusters>/...) when given just session_name — reads segmentation_algorithms/n_clusters/vame_version from config.yaml, with explicit path arguments taking precedence. Warns when the config is missing required
  fields, uses an unverified VAME version (< 0.13.0), or no motif files are found for the session
  - Supports `sampling_frequency_hz` for regular-rate timestamp generation, correctly offsetting `starting_time` by VAME's time_window / 2 to account for the sliding-window labeling
  - Supports `set_aligned_timestamps` for temporal alignment with other data streams
  - Supports multiple independent VAME projects in the same NWB file via `metadata_key`
  - Supports soft-linking to an upstream PoseEstimation container via `pose_estimation_metadata_key`, with clear errors listing available containers if the name doesn't match
  - Adds stub_test support (truncates all series to min(100, n_frames) for fast test runs)
  - Uses the dict-based metadata pipeline (metadata["Behavior"]["VAMEProjects"][metadata_key]), consistent with the rest of the new metadata format
  - Adds ndx-vame as an optional extra (neuroconv[vame])
  - Adds a conversion gallery example (docs/conversion_examples_gallery/behavior/vame.rst)


weiglszonja:
@luiztauffer 
This is ready for a first review.
I also have an open PR for the example data in GIN: https://gin.g-node.org/CatalystNeuro/behavior_testing_data/pulls/18
 Could you take a look when you have time?
I’ll keep this PR in draft until the example data PR gets merged.

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
```diff
@@ -0,0 +1,329 @@
...
+            "When provided, project metadata are serialized into the NWB file. Optional."
+        )
+        return source_schema
+
+    @validate_call
+    def __init__(
+        self,
+        motif_labels_file_path: FilePath,
+        latent_vectors_file_path: FilePath | None = None,
+        community_labels_file_path: FilePath | None = None,
```
luiztauffer: These three paths could be automatically inferred from the VAME project config. I think it's fine to leave them as optional kwargs to be overwrittten, as this might be usefu in some cases, e.g. older VAME versions that had a different folder structure

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
```diff
@@ -0,0 +1,329 @@
...
+        vame_config_file_path: FilePath | None = None,
```
luiztauffer: I think this should be the only required argument, the other file paths can be inferred from it

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
weiglszonja: Thanks for the suggestion! 
  
 For latent vectors and community labels, I think automatic inference is feasible: given the project root (derived from config_file_path.parent) and an optional `session_name`, I can find exactly one file in both the test data (nested VAME ≥0.13 layout) and the Meletis lab data (older VAME version). 

For motif labels, automatic inference can be ambiguous. `ndx-vame` spec allows at most one MotifSeries per VAMEProject (quantity: '?'), so when multiple segmentation algorithms are run — as in our test data which has both `hmm-15/15_hmm_label_{session_name}.npy` and
  `kmeans-15/15_kmeans_label_{session_name}.npy` — The `motif_labels_file_path` is precisely the argument that tells each interface which algorithm run it represents — two algorithms means two `VameInterface` instances, each pointing to a different motif label file.

If you agree with this, I would keep `motif_labels_file_path` required too. But I might be missing something here.

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
```diff
@@ -0,0 +1,329 @@
...
+        sampling_frequency_hz: float | None = None,
+        vame_project_metadata_key: str = "VAMEProject",
```
h-mayorquin: can use metadata_key instead of `vame_project_metadata_key` I am trying to unify that here:
https://neuroconv.readthedocs.io/en/main/developer_guide/ophys_metadata_structure.html
https://neuroconv--1729.org.readthedocs.build/en/1729/developer_guide/ecephys_metadata_structure.html



```suggestion
        metadata_key: str = "VAMEProject",
```

Check this links so you can pass that info as context to an agent.

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
```diff
@@ -0,0 +1,329 @@
...
+                data=community_data,
+                **community_metadata,
+                **timing_kwargs,
+                motif_series=motif_series,
+            )
+            community_series = CommunitySeries(**community_kwargs)
+
+        # Optional link to an upstream PoseEstimation container
+        pose_estimation = None
+        behavior_module = get_module(nwbfile, name="behavior", description="Processed behavioral data.")
```
weiglszonja: Make this more flexible to allow retrieving pose estimation from other modules as well and raise when it cannot find by that name but there are other PoseEstimation types there

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
```diff
@@ -0,0 +1,329 @@
...
+        if self.pose_estimation_name is not None:
+            pose_estimation = behavior_module.data_interfaces.get(self.pose_estimation_name)
+
+        vame_project = VAMEProject(
+            name=project_name,
+            pose_estimation=pose_estimation,
+            latent_space_series=latent_series,
+            motif_series=motif_series,
+            community_series=community_series,
```
weiglszonja: Change this in `ndx-vame` to allow multiple motif_series and community_series per `VAMEProject`: https://github.com/catalystneuro/ndx-vame/pull/4

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
weiglszonja: Thanks, this makes sense.

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
```diff
@@ -0,0 +1,349 @@
...
+            "Path to the VAME 'config.yaml' project file. "
+            "The full configuration is serialized as JSON and stored in the VAMEProject.vame_config field."
+        )
+        return source_schema
+
+    @validate_call
+    def __init__(
+        self,
+        vame_config_file_path: FilePath,
+        motif_labels_file_path: FilePath | None = None,
```
h-mayorquin: We should use file_path to be consistent with the other interfaces, most of them use file_path or folder_path. Also, everything after that should be ** kwargs only.

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
```diff
@@ -0,0 +1,349 @@
...
+            record the upstream pose data used by VAME. Raises ``ValueError`` if the name is
+            not found, listing any available ``PoseEstimation`` containers to help.
+        verbose : bool, default False
+            Controls verbosity of the conversion process.
+        """
+        import ndx_vame  # noqa: F401 – ensure ndx-vame namespace is registered
+
+        self.vame_config_dict = load_dict_from_file(vame_config_file_path)
+
+        self.motif_labels_file_path = Path(motif_labels_file_path) if motif_labels_file_path else None
```
h-mayorquin: Could we keep the attributes of the interface private unless we need to make them public for a good reasons. Makes the API documentation easier and makes it easier to modify the internals in the future as we don't need to be ourselves concerned with changes.

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
```diff
@@ -0,0 +1,349 @@
...
+            )
+        reference_file = self.motif_labels_file_path or self.latent_vectors_file_path or self.community_labels_file_path
+        if reference_file is None:
+            raise ValueError(
+                "VameInterface cannot generate original timestamps without any data file. "
+                "Provide at least one of motif_labels_file_path, latent_vectors_file_path, or "
+                "community_labels_file_path, or call set_aligned_timestamps()."
+            )
+        n_frames = np.load(reference_file).shape[0]
+        return np.arange(n_frames) / self.sampling_frequency_hz
```
h-mayorquin: VAME crops `time_window/2` frames from each end of its output, so the labels actually start half a window into the recording. I think `get_original_timestamps` should offset the start by half a window, matching VAME's own exporter (`vame/io/nwb.py`). The crop is already correct, so this is a metadata-only change to `starting_time`, and a test should assert it.

```python
    ...
    n_frames = len(np.load(self.motif_labels_file_path))
    time_window = self.vame_config_dict.get("time_window")
    offset_frames = int(time_window // 2) if time_window else 0
    start = offset_frames / self.sampling_frequency_hz
    return start + np.arange(n_frames) / self.sampling_frequency_hz
```

I think the published 001828 files have this same offset, since they were made by the local `open_field_test/vameinterface.py` in `meletis-lab-to-nwb`, not this PR. Maybe worth double-checking, and since it is still embargoed it would be cheap to re-export with the corrected `starting_time` before release.

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
```diff
@@ -0,0 +1,349 @@
...
+class VameInterface(BaseTemporalAlignmentInterface):
+    """DataInterface for VAME behavioral segmentation data using the ndx-vame NWB extension.
+
+    EthoML/VAME (Variational Animal Motion Encoding) segments animal behavior into discrete motifs
+    by training a variational autoencoder on pose estimation time series. This interface
+    writes the per-frame outputs—motif labels, optional latent-space embeddings, and optional
+    community labels to NWB using the ``ndx-vame`` extension.
+
+    The NWB container name and all series descriptions are controlled via the metadata dict
+    under ``metadata["VAME"][metadata_key]``.
```
h-mayorquin: I don't like `metadata["VAME"]` at the top level. NeuroConv is converging on a dict-based layout where everything lives under a modality key (`Behavior`, `Devices`), as registries keyed by `metadata_key` and cross-referenced via `*_metadata_key`. Since VAME output is behavioral, it should move under `metadata["Behavior"]` like this:


```python
metadata["Behavior"]["VAMEProjects"][metadata_key] = {     # metadata_key = the project
    "name": "VAMEProject",
    "LatentSpaceSeries": {"name": ...},                    # one, shared, optional
    "MotifSeries": {                                       # keyed by algorithm/run
        "hmm":    {"name": "MotifSeriesHmm",    "algorithm": "hmm"},
        "kmeans": {"name": "MotifSeriesKmeans", "algorithm": "kmeans"},
    },
    "CommunitySeries": {
        "hmm": {"name": "CommunitySeriesHmm", "motif_series_key": "hmm"},   # cross-ref within the project
    },
    "pose_estimation_metadata_key": ...,                   # cross-ref to Behavior/Pose, later
}
```

Note that this already takes into account the multiplicity that your PR https://github.com/catalystneuro/ndx-vame/pull/4 enables. The top-level key is a pluralized container name, as discussed in https://github.com/catalystneuro/neuroconv/issues/1280. The `pose_estimation_metadata_key` will be used to link the pose estimation, see the [mock pose estimation PR](https://github.com/catalystneuro/neuroconv/pull/1745). You can define it like this and use it to mean the name for now (until we migrate the interfaces, then I will take care of the logic), but the schema name should be correct now.

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
```diff
@@ -0,0 +1,349 @@
...
+            per frame). Typically located inside a ``community/`` sub-directory of the algorithm directory.
+        sampling_frequency_hz : float, optional
+            Video acquisition rate in Hz (frames per second). Required when not providing aligned
+            timestamps via :meth:`set_aligned_timestamps`.
+        metadata_key : str, default "VAMEProject"
+            Key used to look up this interface's metadata inside
+            ``metadata["VAME"][metadata_key]``. Change this when storing
+            results from multiple VAME runs in the same NWB file so each run has a unique
+            metadata entry.
+        pose_estimation_name : str, optional
```
h-mayorquin: I am not sure this should be at init argument, this might depend on running time, in any case it should be a conversion option or as I proposed in another comment a metadata concern for a converter.

h-mayorquin (commented):
Some commments.

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
weiglszonja: Thank you, this is very important. It would be useful to have the reference file/code line from VAME documentation on this PR as well, can you point me to it? Also wondering whether this is worth mentioning in `ndx-vame` too.

h-mayorquin: https://github.com/catalystneuro/ndx-vame/pull/5

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
weiglszonja: Thanks, added in [0059eb7](https://github.com/catalystneuro/neuroconv/pull/1737/commits/0059eb7c25bcf9f28b071af1049cb301e6d19cae)

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
weiglszonja: Thanks, added in [10e17b2](https://github.com/catalystneuro/neuroconv/pull/1737/commits/10e17b25999eab925bff9a817868720a70134168)

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
weiglszonja: Thanks, added in [fd76544](https://github.com/catalystneuro/neuroconv/pull/1737/commits/fd76544d743e3c717d9c0333f990455b2e75b488) and tests in [993f157](https://github.com/catalystneuro/neuroconv/pull/1737/commits/993f15713883d8e1bd8dbec1fe3ed25c20e51ca4)

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
weiglszonja: Added auto-discovery of file paths in [6845717](https://github.com/catalystneuro/neuroconv/pull/1737/commits/68457175784dd11450ad309737783f57250e8ecf) and in [f239511](https://github.com/catalystneuro/neuroconv/pull/1737/commits/f239511c0fa61d5d4cd129eaa413413be0fc2f70)

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
weiglszonja: Thanks, moved this to metadata

src/neuroconv/datainterfaces/behavior/vame/vamedatainterface.py
weiglszonja: Thanks, I changed this too.

pyproject.toml
```diff
@@ -116,6 +116,9 @@ lightningpose = [
 medpc = [
     "ndx-events==0.2.2",
 ]
+vame = [
+    "ndx-vame @ git+https://github.com/catalystneuro/ndx-vame.git@main",
+]
```
weiglszonja, line 121: This temporary pin needs to be changed to `ndx-vame==0.3.1`, I'll make this release soon.

h-mayorquin (changes requested):
More comments. This is looking really good now, almost everything from the earlier rounds is in. 

The `algorithm` attribute on `MotifSeries`/`CommunitySeries` never gets set, so everything we write ends up with the `n/a` default. We already know the algorithm at `get_metadata` time, the run key is exactly it (in the auto-discovery path it comes straight from `segmentation_algorithms` in the config), so I think we can just populate it there:

```python
if self._motif_labels_file_paths is not None:
    vame_project_metadata["MotifSeries"] = {
        run_key: dict(
            name=f"MotifSeries{run_key.capitalize()}",
            description="VAME behavioral motif labels.",
            algorithm=run_key,
        )
        for run_key in self._motif_labels_file_paths
    }
```

and the same for `CommunitySeries`. The test currently asserts `algorithm == "n/a"`, so that one would flip too. Right now the algorithm only lives in the series name (`MotifSeriesKmeans`/`MotifSeriesHmm`), and I'd rather have it in the field a reader actually queries, otherwise the attribute is there but always empty.

The pose-estimation link I'd like to sort out here rather than defer, now that the DLC PR (#1750) is open and gives us a real producer to point at. Right now `_get_pose_estimation` finds the container by matching its object name, but the field is `pose_estimation_metadata_key`, so it should resolve by the registry key the same way `device_metadata_key`/`skeleton_metadata_key` already do: read `metadata["Behavior"]["Pose"]["PoseEstimations"][pose_estimation_metadata_key]["name"]` to get the object name, then fetch that from the file. It passes today only because the mock happens to name its container after its key; with DLC the key is `deep_lab_cut_metadata_key` while the name is `PoseEstimationDeepLabCut`, so handing VAME the key finds nothing. `add_to_nwbfile` already receives the full merged metadata in a converter, so the Pose registry is right there to look up (a name-match fallback is fine if the key isn't in the registry). Could you add a test that links VAME to a DLC-style PoseEstimation where the key and name differ, so this can't silently regress?

A get_available_sessions(file_path)` helper would be great. `session_name` is required for the auto-discovery path and a single `config.yaml` usually lists several sessions (our test config has five), so right now the only way a user finds the valid values is to open the YAML by hand. DeepLabCut already has the exact parallel with `get_available_subjects`, so I'd mirror it:

```python
@staticmethod
def get_available_sessions(file_path: FilePath) -> list[str]:
    """Return the session names declared in a VAME config.yaml."""
    config = load_dict_from_file(file_path)
    return config.get("session_names", [])
```

(`load_dict_from_file` is already imported.) Then the flow is: call it to see the options, pass one as `session_name`. Worth a one-liner in the gallery too.

On the gallery doc itself, I'd drop the YAML metadata blocks and the `load_dict_from_file` snippet and keep the in-Python dict form. This is the convention we settled on in #1269: a minimal quick-start, then a "Specifying Metadata" section that builds a plain dict and merges it with `dict_deep_update`, no separate file to save and load by path. `recording/whitematter.rst` is the pattern to mirror. Right now the file-based YAML approach only lingers in `tdt_fp.rst` (the original, pre-discussion) and this PR; everything added since uses the dict form. While you're in there, adding `metadata["Subject"] = dict(subject_id=..., species=..., sex=..., age=...)` to the quick-start would make the example DANDI-ready, which is what the other recent entries do.

h-mayorquin (approved):
LGTM. The last point of requests have been addressed.

Note, I put cap on python-neo which has been released but requires a SpikeInterface release to work properly.

PR #1739, opened 2026-05-18
Add IntanConverter for one-call multi-stream Intan conversion

### LINKED ISSUE #519: [Feature]: IntanRecordingInterface - digital channels and aux channels

h-mayorquin:
Intan recordings routinely contain several streams in a single file: the main amplifier plus some combination of ADC inputs/outputs, auxiliary, DC amplifier, and stimulation channels (the last two on RHS only). Today users must construct three or four interfaces by hand, knowing in advance which streams are present in their file. `IntanConverter` parses the file header once (via neo, without loading data) and routes each present stream to the appropriate sub-interface, so a single call writes the whole session to NWB. This brings the Intan format in line with the multi-stream converter pattern already established for `SpikeGLXConverterPipe` and `OpenEphysBinaryConverter`.

The Digital channels are TTL events whose natural NWB representation depends on the new events model proposed in [NWBEP001](https://github.com/nwb-extensions/nwbep-review/issues/4) is waiting for the next pynwb release.; once that lands and the digital interface is implemented, adding it to the converter is a one-line entry in the routing dictionary. That digital interface is also the only piece still needed to close [#519](https://github.com/catalystneuro/neuroconv/issues/519), which is the only outstanding Intan-related issue.


weiglszonja (approved):
Looks got to me, the failing CI tests look unrelated to this PR.

PR #1743, opened 2026-05-22
Add dict-based metadata pipeline for ecephys

h-mayorquin:
This adds the ecephys counterpart of the dict-based metadata pipeline that landed for ophys ([PR #1677](https://github.com/catalystneuro/neuroconv/pull/1677) for imaging, [PR #1692](https://github.com/catalystneuro/neuroconv/pull/1692) for segmentation). It moves ecephys devices to a top-level `metadata["Devices"]` and reshapes electrode groups and electrical series into dict-based sections keyed by a `metadata_key`. The goal is a single top-level `Devices` shared across modalities, so a recording and a pose estimation series ([PR #1745](https://github.com/catalystneuro/neuroconv/pull/1745)) from the same session can reference a common device instead of each modality carrying its own. That shared `Devices` is the prerequisite for combining ecephys and ophys metadata in a single conversion.

The old list-based format keeps working unchanged: the metadata shape is detected automatically and routed by the dispatcher in `add_recording_metadata_to_nwbfile`, with the previous list-based helpers preserved, so existing callers see no behavior change. This PR is the pipeline only, kept at parity with what the old format already supported; the recording interfaces still emit list-based metadata, and migrating them is a follow-up done the same way as for ophys (the base interface first, then per-interface).


h-mayorquin:
Here is a table to navigate the tests for the dict-based metadata pipeline (`add_recording_to_nwbfile`), to make reviewing easier. Each test owns a distinct path rather than re-checking what another already covers. The non-mutation tests came from a closer look at the metadata contract: the default-generation path was mutating the caller's metadata (violating the immutability principle from [issue #1511](https://github.com/catalystneuro/neuroconv/issues/1511)), now fixed to resolve defaults locally as the roiextractors pipeline does.

| Test | What it verifies | Why it earns its place |
| --- | --- | --- |
| [`test_basic`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2440-L2468) | With no metadata, defaults are generated: a default device, an electrode group per channel group, and the electrical series. | Only test of the fully-default path (no metadata supplied). |
| [`test_full_metadata_specification`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2470-L2540) | Every user-supplied field lands on its object and the whole chain resolves: device to group to series to each electrode row. | Owns the all-specified path and end-to-end cross-link resolution. |
| [`test_missing_device_metadata_key_falls_back_to_default`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2542-L2574) | An electrode group with no `device_metadata_key` gets the default device. | Owns the absent-device-reference branch. |
| [`test_partially_specified_electrode_groups`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2791-L2851) | Specified and default-generated groups coexist in one call: specified groups keep their fields and device, unspecified groups get defaults, every channel maps to the right group. | Owns the partition branch (user entries and default entries both present). |
| [`test_shared_device_two_recordings`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2576-L2635) | Two recordings pointing at the same `Devices` entry share one device; group and electrode rows are not duplicated across calls. | Owns cross-call deduplication. |
| [`test_multiple_devices`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2853-L2919) | Channels split across two devices and two groups; each row maps to the correct group and device. | Owns multi-group channel-to-row mapping with distinct devices. |
| [`test_multiple_groups_share_one_device`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2921-L2986) | Several electrode groups referencing one `Devices` entry resolve to a single shared device object. | Owns within-call device reuse across groups. |
| [`test_missing_required_electrode_group_field_raises`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2637-L2673) | A group entry missing required fields raises a clear error listing them. | Owns the required-field validation branch. |
| [`test_missing_metadata_key_raises`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2675-L2694) | An unknown `metadata_key` raises an error listing the available keys. | Owns the series-key lookup error branch. |
| [`test_metadata_passed_without_metadata_key_raises`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2696-L2714) | Passing metadata without a `metadata_key` raises (the key selects which `ElectricalSeries` to write). | Owns the contract that metadata requires a `metadata_key`; the only metadata-free path is passing no metadata. |
| [`test_metadata_not_mutated`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2716-L2757) | Fully-specified metadata is left unchanged by the call. | Owns non-mutation on the strict-lookup path. |
| [`test_metadata_not_mutated_when_generating_defaults`](https://github.com/catalystneuro/neuroconv/blob/2e8a5013455ed7ec15885006b2bb4825f1ad25da/tests/test_modalities/test_ecephys/test_tools_spikeinterface.py#L2759-L2789) | Generating defaults for unspecified metadata leaves the caller's dict unchanged. | Owns non-mutation on the default-generation path (the regression behind the #1511 fix); the full-spec test above never walks this path. |

pauladkisson (approved):
Code looks good.

I wanted to flag here that I still think it's kinda awkward to have to specify the electrode groups by digging into the recording extractor. It would be nice to have a metadata-driven path or at least a convenience function on the interface to be able to do that without digging into the internals. 

However, that kind of improvement is definitely out of scope for this PR, so, it'll have to wait.

h-mayorquin:
> Code looks good.
> 
> I wanted to flag here that I still think it's kinda awkward to have to specify the electrode groups by digging into the recording extractor. It would be nice to have a metadata-driven path or at least a convenience function on the interface to be able to do that without digging into the internals.
> 
> However, that kind of improvement is definitely out of scope for this PR, so, it'll have to wait.

I agree with you. It really bothers me and I am thinking on it just right now because we have the same problem on icephys (thought less severe) with the intracellular recordings table. 

I want to get neuroconv to the point where we can do a release as it has been some time so I want to ship a minimal version first. Thanks again for reviewing all of this. It has been a lot of work.

PR #1744, opened 2026-05-26
Fix Image interface dtype

h-mayorquin:
A bug where the original dtype was overwritten.


src/neuroconv/datainterfaces/image/imageinterface.py
```diff
@@ -97,7 +108,7 @@ def recommended_data_shape(self):
     @property
     def dtype(self):
         """Define the data type of the array"""
-        return np.dtype(float)
+        return self._dtype
```
h-mayorquin, line 111: This is the core of the change.

PR #1745, opened 2026-05-26
Migrate MockPoseEstimationInterface to the dict-based metadata shape

h-mayorquin:
This PR is in the same line of work as [PR #1677](https://github.com/catalystneuro/neuroconv/pull/1677) and [PR #1692](https://github.com/catalystneuro/neuroconv/pull/1692) (the roiextractors imaging and segmentation pipelines) and [PR #1743](https://github.com/catalystneuro/neuroconv/pull/1743) (the ecephys pipeline).

This ports `MockPoseEstimationInterface` onto the unified dict-based metadata layout the rest of NeuroConv is converging on: a top-level `Devices` registry plus a `Behavior/Pose` sub-modality holding parallel `Skeletons` and `PoseEstimations` registries keyed by `metadata_key`, with cross-component links expressed as `device_metadata_key` and `skeleton_metadata_key` references.

The idea of this PR is to start with the mock as a prototyping playground so we can see what a place where we control everything would look like. Once we converge on this, we can then discuss how DeepLabCut, Lightning Pose, and SLEAP will migrate. This is analogous to those imaging and segmentation PRs and the current ecephys PR, but this case is different because there is no common pipeline.

The metadata shape is the following:

```python
metadata = {
    "Devices": {
        metadata_key: {
            "name": "CameraMockPoseEstimation",
            "description": "Mock camera device for pose estimation testing.",
        },
    },
    "Behavior": {
        "Pose": {
            "Skeletons": {
                metadata_key: {
                    "name": "SkeletonMockPoseEstimation",
                    "nodes": ["head", "neck", "left_shoulder"],
                    "edges": [[0, 1], [1, 2]],
                },
            },
            "PoseEstimations": {
                metadata_key: {
                    "name": "MockPoseEstimation",
                    "description": "Mock pose estimation data from MockSourceSoftware.",
                    "source_software": "MockSourceSoftware",
                    "scorer": "MockScorer",
                    "dimensions": [[640, 480]],          # per-camera, paired with the device
                    "original_videos": ["mock_video.mp4"],  # per-camera, paired with the device
                    "device_metadata_key": metadata_key,    # -> metadata["Devices"][...]
                    "skeleton_metadata_key": metadata_key,  # -> metadata["Behavior"]["Pose"]["Skeletons"][...]
                    "PoseEstimationSeries": {
                        "head": {
                            "name": "PoseEstimationSeriesHead",
                            "description": "Mock pose estimation series for head.",
                            "unit": "pixels",
[...]

Issue #1654, opened 2026-02-05
[Feature]: ABF Converstion Structure


mcoe-tolo:
### What would you like to see added to NeuroConv?

Is there a way to have ABF files with sweeps import as one continuous Tsd for each channel (i.e. one for current and one for voltage) with an IntervalSet for the corresponding start and end for each sweep? 

### Is your feature request related to a problem?

Files are currently being saved with Tsds for each sweep as a separate Tsd, which can make it challenging to identify, without prior knowledge, which Tsd corresponds to which recording protocol. (i.e. stimulus-01-ch-1, stimulus-02-ch-1, voltage_clamp-response-01-ch-1, voltage_clamp-response-02-ch-1)

### Do you have any interest in helping implement the feature?

No.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [x] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

h-mayorquin:
Hi, @mcoe-tolo, sorry for the delay. The machinery for doing this was already merged on #1746. Now, what I am unsure is whether pynapple will use the intracellular table as an interval set. I will test that.

h-mayorquin:
Hi @mcoe-tolo, following up with a concrete answer after testing pynapple end to end.

The NWB structure you asked for now ships in `AxonIntracellularInterface` (#1746, merged) and `AxonIntracellularConverter` (#1761, merged): each channel is written as a single continuous `PatchClampSeries` (one for current, one for voltage), and each sweep is one row in the `IntracellularRecordings` table that addresses its slice of that continuous series by `(idx_start, count)`. No more `stimulus-01-ch-1` / `response-02-ch-1` pile.

On pynapple specifically, I loaded a converted file and checked what it yields:

- The continuous series loads directly as a `Tsd`, one per channel, so the "one continuous Tsd per channel" part works out of the box.
[...]

Issue #1564, opened 2025-10-21
`add_icephys_recordings`  wrongly assumes that every neo channel is a response


h-mayorquin:
The `add_icephys_recordings()` function assumes that every Neo signal channel corresponds to a response signal and automatically creates a corresponding stimulus from the raw protocol data. In [`add_icephys_recordings()`](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/neo/neo.py#L333-L387), the code loops over all neo channels and treats each as a response, then pairs it with stimulus data indexed by the same channel number from `protocol[0][segment][ei]`.

For the Khaliq lab data that I have, the first channel represents the Voltage response and the second signal represents the current (a monitor signal). When using `add_icephys_recordings` with this configuration two errors occur: (1) The monitor current is stored using the wrong NWB data type (`CurrentClampSeries`, a response type instead of `CurrentClampStimulusSeries`, the stimulus type), and (2) The code creates a paired stimulus from an inactive/empty DAC protocol, resulting in a `CurrentClampStimulusSeries` filled with zeros that has no experimental meaning.

The interface should provide a way for users to explicitly specify what each ADC channels represents. This would allow correct data type assignment (response series vs. stimulus series) and proper stimulus-response pairing. We could also potentially improve the automatic detection heuristics by analyzing channel units and matching them to active DAC configurations, though this may not cover all cases and user-specified mapping would be more reliable.

h-mayorquin:
Fixed by `AxonIntracellularInterface` (#1746, merged). The new interface never blind-pairs every channel with a stimulus. The stimulus source is declared explicitly: either a recorded monitor channel (`stimulus_channel_name`) or a reconstructed command (`stimulus_command`), or none at all. So the current monitor is no longer written as a `CurrentClampSeries` response, and there is no zero-filled bogus stimulus from an empty command. Closing as resolved.

Issue #1563, opened 2025-10-21
Bug: ABF interface incorrectly counts number of electrodes


h-mayorquin:
The ABF interface incorrectly assumes that each Neo channel, which corresponds to an ADC signal channel, represents a separate physical electrode. This assumption is made in [`get_number_of_electrodes()`](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/tools/neo/neo.py#L55-L65). For the Khaliq lab conversion this does not hold, and I don't think it holds in general. A monitor signal of the protocol is also propagated to the acquisition system:

<img width="995" height="749" alt="Image" src="https://github.com/user-attachments/assets/07e9568a-0ae9-41f2-9b14-68254ba2e0c1" />

Where the IN0 channel with mV represents the voltage response and the IN1 channel with pA represents the current monitor signal of the same electrode. The current implementation of the ABF interface counts these as two separate electrodes.

Looking at the format, I don't think there is a bulletproof automatic method to determine electrode count from ABF files alone. I think the right thing to do is to provide a way for the user to pass this specification, either as metadata or as conversion options (I need to think more about it and come with a proposal) and error out if this is not provided, as making up electrodes is a serious provenance violation.

We could though improve the heuristic by counting active DAC channels (those with defined epoch protocols in `dictEpochInfoPerDAC`, see [neo/rawio/axonrawio.py](https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/axonrawio.py) where this structure is populated from the ABF file's EpochPerDACSection) as each electrode requires its own stimulus generator. For files where this information is unavailable (ABF version 1) or ambiguous (e.g., multiple DACs active but unclear ADC-to-electrode mapping), the electrode count could be inferred by grouping ADC channels with complementary units (mV/pA pairs in current clamp, or pA/mV pairs in voltage clamp) under the assumption that each pair monitors a single electrode. Not sure if this is worth the added complexity and the possibility of a provenance error though.

h-mayorquin:
[...]

PR #1746, opened 2026-06-05
Add `AxonIntracellularInterface` for Axon Binary Format (ABF) intracellular conversion

h-mayorquin:
The current `AbfInterface` has a lot of problems:

- It assumes one recorded channel equals one electrode, but in the Khaliq lab conversion we have seen that Clampex supports encoding one channel as a monitor. This wrong assumption leads to doubled electrode counts and recording-table rows, and a monitor channel is stored as a bogus response ([#1563](https://github.com/catalystneuro/neuroconv/issues/1563), [#1564](https://github.com/catalystneuro/neuroconv/issues/1564)).
- It pairs every channel with a stimulus indexed by the same number, so an empty command channel is written as a zero-filled stimulus ([#1564](https://github.com/catalystneuro/neuroconv/issues/1564)).
- The main way to configure the clamp type (voltage versus current) does not work: the `icephys_experiment_type` argument is silently ignored, and the value is instead read from metadata, where it defaults to `None` and has problems of its own ([#1561](https://github.com/catalystneuro/neuroconv/issues/1561)).
- It ignores the recorded channels' real names and labels everything by position (`icephys_electrode_0`, `…-ch-0`), so the names the acquisition software stored never reach the NWB file ([#1482](https://github.com/catalystneuro/neuroconv/issues/1482)). 
- It writes one `PatchClampSeries` per sweep, producing a flat pile of series. The data supports a cleaner representation: a single continuous series per channel, with each sweep registered as a row in the intracellular-recordings table that points to its `(start_index, count)` slice ([#1654](https://github.com/catalystneuro/neuroconv/issues/1654)).

I started fixing some of these in the current interface. Two are patchable in place: the clamp-config crash ([#1561](https://github.com/catalystneuro/neuroconv/issues/1561)) and the per-sweep layout ([#1654](https://github.com/catalystneuro/neuroconv/issues/1654)). The other three share one structural root, the model has no notion of channel roles, which is what produces the electrode doubling, the bogus stimulus, and the lost channel names ([#1563](https://github.com/catalystneuro/neuroconv/issues/1563), [#1564](https://github.com/catalystneuro/neuroconv/issues/1564), [#1482](https://github.com/catalystneuro/neuroconv/issues/1482)): nothing tells the interface which recorded channel is the response, which is a monitor, and which command actually drove the cell. And underneath all of them is a second, more strategic limitation: the interface is built on the old list-based metadata, whereas neuroconv is moving to dict-based metadata keyed by name (see [#269](https://github.com/catalystneuro/neuroconv/issues/269), [#1677](https://github.com/catalystneuro/neuroconv/pull/1677), [#1743](https://github.com/catalystneuro/neuroconv/pull/1743), [#1745](https://github.com/catalystneuro/neuroconv/pull/1745)). That list-based `icephys_metadata` / `recording_sessions` shape is incompatible with the dict model (positional list items cross-referenced by name-string lookup, with no `metadata_key` to address an entry and no `_metadata_key` reference to share a device or electrode), so migrating it in place would be highly disruptive.

So two fundamental pieces of the current interface would have to be changed. At this point, I think it is better to start from scratch, as the new interface will have a different scope and that will avoid breaking current working workflows. So this PR creates a new interface that does:

- Takes the response channel by name and its mode (voltage versus current) as arguments, so channel roles and the clamp type are declared, not guessed, and the real channel names are kept ([#1563](https://github.com/catalystneuro/neuroconv/issues/1563), [#1564](https://github.com/catalystneuro/neuroconv/issues/1564), [#1482](https://github.com/catalystneuro/neuroconv/issues/1482), [#1561](https://github.com/catalystneuro/neuroconv/issues/1561)).
- Writes one continuous `PatchClampSeries` per channel with one `IntracellularRecordings` row per sweep `(start_index, count)`, keeping the sweep grouping in the table ([#1654](https://github.com/catalystneuro/neuroconv/issues/1654)).
- Uses the unified dict-based metadata keyed by `metadata_key`, consistent with the other modalities.


h-mayorquin:
https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/177

luiztauffer (approved):

PR #1749, opened 2026-06-09
Zarr v3 backend configuration

alejoe91:
Depends on v3 port of:

- hdmf-zarr: https://github.com/hdmf-dev/hdmf-zarr/pull/325
- probeinterface: https://github.com/SpikeInterface/probeinterface/pull/387
- spikeinterface: https://github.com/SpikeInterface/spikeinterface/pull/4260


h-mayorquin:
I took a first look, very high level:


1. The validator should check `shard_shape` against `buffer_shape`, not just `chunk_shape`. As far as I understand a shard is assembled whole in memory on every flush, so `shard_shape` sets a per-flush memory footprint a user can push until they run out of memory, and a shard bigger than the buffer also forces a read-modify-write since we write buffer by buffer. So I'd enforce `shard_shape <= buffer_shape`, ideally that it evenly tiles the buffer. 
2. There's no user-facing documentation, we should do this and make this easy to review.
3. What do you think about defaults shard values? Should we write defaults equal bo the buffer size? non-shared by defaults? 

@alejoe91 let me know if this is the type of feedback that you wanted or if I should look into other aspects of it.

alejoe91:
> I took a first look, very high level:
> 
> 1. The validator should check `shard_shape` against `buffer_shape`, not just `chunk_shape`. As far as I understand a shard is assembled whole in memory on every flush, so `shard_shape` sets a per-flush memory footprint a user can push until they run out of memory, and a shard bigger than the buffer also forces a read-modify-write since we write buffer by buffer. So I'd enforce `shard_shape <= buffer_shape`, ideally that it evenly tiles the buffer.
> 2. There's no user-facing documentation, we should do this and make this easy to review.
> 3. What do you think about defaults shard values? Should we write defaults equal bo the buffer size? non-shared by defaults?
> 
> @alejoe91 let me know if this is the type of feedback that you wanted or if I should look into other aspects of it.

@h-mayorquin I'm picking this up again and will address all the great comments above shortly :)

alejoe91:
Ok first issue is that a standalone `Shuffle` is not implemented in Zarr V3, but only embedded in Blosc. However, Blosc doesn't natively support Gzip.

Two options:
1. Drop default shuffling
2. Use another codec through blosc (LZ4 or Zstd)

alejoe91:
> I took a first look, very high level:
> 
> 1. The validator should check `shard_shape` against `buffer_shape`, not just `chunk_shape`. As far as I understand a shard is assembled whole in memory on every flush, so `shard_shape` sets a per-flush memory footprint a user can push until they run out of memory, and a shard bigger than the buffer also forces a read-modify-write since we write buffer by buffer. So I'd enforce `shard_shape <= buffer_shape`, ideally that it evenly tiles the buffer.

This should be solved by https://github.com/hdmf-dev/hdmf-zarr/pull/374, which supports buffer sizes smaller than shard sizes.

> There's no user-facing documentation, we should do this and make this easy to review.
>

Maybe worth adding a "How to configure NWB backend" page? I suggest in a separate PR

> What do you think about defaults shard values? Should we write defaults equal bo the buffer size? non-shared by defaults?

As a start, I think that setting shard sizes as buffer sizes is reasonable. Not sure if it should be up to `neuroconv` or `hdmf-zarr` though (see [comment](https://github.com/hdmf-dev/hdmf-zarr/pull/374#issuecomment-5634548859))

h-mayorquin:
1) Yes, here, I would like to actually constrain ourselves and only handle buffer shapes that are aligned with shard size. 

2) Yes, I agree, but this was me when I understood this less now I think I need it less : P. Once zarr 3.0 lands I will think on better documentation. Another PR as you say.

3) Related to 1. I think here is the right place. HDMF-zarr as a flexible pass through, here we have good defaults and automated suggestions as we can iterate quicker here.

PR #1750, opened 2026-06-14
Migrate DeepLabCutInterface to the unified dict-based metadata format

h-mayorquin:
Neuroconv is converging on a single dict-based way of describing metadata across modalities (a shared top-level device registry plus per-modality registries cross-referenced by key) instead of each interface carrying its own bespoke shape. This PR brings `DeepLabCutInterface` into that scheme so DeepLabCut pose data is described the same way the imaging and electrophysiology pipelines already are, and the way the mock pose interface established for behavior. It is opt-in and backward compatible: existing conversions are unchanged and produce identical files, users adopt the new shape explicitly, and the interface's metadata-key argument is renamed for consistency with the rest of the pipeline (the old name kept as a deprecated alias).

Related PRs:
- Pose anchor (mock interface): https://github.com/catalystneuro/neuroconv/pull/1745
- Imaging interfaces (cross-cutting setup): https://github.com/catalystneuro/neuroconv/pull/1694
- Imaging interface example (Suite2p): https://github.com/catalystneuro/neuroconv/pull/1716
- Ecephys dict-based pipeline: https://github.com/catalystneuro/neuroconv/pull/1743


weiglszonja (approved):
LGTM

Issue #1748, opened 2026-06-09
Which ndx-events version should new interfaces use?


pauladkisson:
Currently, the latest version of ndx-events is 0.4.0: https://github.com/rly/ndx-events.

But several interfaces still use ndx-events 0.2.2: MedPC and SpikeGLX.

I am planning on developing several event-based interfaces for the GuPPy NWB export project. Should I be using ndx-events 0.2.2 to avoid conflicts with existing interfaces, or ndx-events 0.4.0 so that brand new interfaces use the latest version?

h-mayorquin:
Last time I checked the latest version of ndx-events was overriding the NWBFIle object. I thought that was a brittle move and that would require some code churn so I thought on delaying it till the events were added to core. Now that this is closer (is it?) I still hold that position but I guess if you have a very strong case we can think on something.

pauladkisson:
Nope, I do not have any strong reason to use ndx-events 0.4.0. Personally, I'd probably rather just use ndx-events 0.2.2. If our current stance is we're just gonna wait till it gets merged to the core and then switch everything over at that time, that sounds great to me.

h-mayorquin:
Closing this. Now we support pynwb native events and a base event interface that should be used to write them:

https://github.com/catalystneuro/neuroconv/pull/1774

PR #1751, opened 2026-06-15
Added TDT events

pauladkisson:
This PR adds an interface for TDT events stored as simple ndx-events.Events, split off from #965 as requested.

I also updated https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/46 to include strobed event data from TDT


h-mayorquin (changes requested):
Hi, thanks for starting this and for splitting it off from #965. Some points.

As we don't have great coverage of events interfaces and NWBEP0001 is about to release, I would like to start with the unified metadata for this at the moment. Can you check PR #1759 that I just opened? There is the proposed metadata schema for the events interfaces. I took care of making the metadata backwards compatible, so only the pieces that are valid under ndx-events 0.2.x are used, but once events is merged into pynwb we won't need to break anything. This implies some things for this PR, like:
* Add metadata_key
* remove the epoc from the metadata
* The metadata will be nested under "Events" and not "behavior" as we are doing.

One thing that bothers me is that TDT events are not really behavior: epocs are also stimulus markers and sync/TTL pulses, not just behavioral events. This has two separate sides.

First, the NWB container. Writing them into a `behavior` processing module over-claims, since it asserts the events are behavioral. We could go the parent container route (see #1760, where I would like your review) to let the caller choose, though that becomes obsolete with NWBEP0001's native events location (https://github.com/NeurodataWithoutBorders/nwb-schema/pull/690) anyway. Unsure in how to move forward here, adding a parmater that will be removed soon seems annoying. Maybe we should change the view on #1748 but to avoid the annoynace of overiding the NWW objec itself we can just develop these interfaces against pynwb dev? What do you think?

Second, and separately, where the interface code should live in the repo. It is under `behavior/` now, which has the same problem. The deeper issue is that our per-modality file organization breaks down once acquisition systems record across modalities; it would have been better to divide by acquisition format, but now that we are committed to modality-based folders maybe we should just add an `events` folder as its own modality. The problem goes deep, as our testing infra is coupled to the file organization in the same way. Any better idea?


On the conversion itself is some information loss here. The TDT spec shows an epoc carries onset (STRON), offset (STROFF, a duration), and a strobe value. `get_events()` reads all three, but `add_to_nwbfile` writes only onset as a plain `Events`, dropping offset and value. Again, if we had NWBEP0001 there is the natural duration that takes this into account, but given that we don't, maybe we should write those onset and offset as a separate LabeledEvents for the moment. What do you think?

pauladkisson:
Thanks for the awesome comprehensive feedback! Lots to work through here...

pauladkisson:
> One thing that bothers me is that TDT events are not really behavior: epocs are also stimulus markers and sync/TTL pulses, not just behavioral events. This has two separate sides.

Yeah, I've run into this problem in some form before with TDT fiber photometry (since TDT can be both a FP and ephys format). It's definitely a little awkward, but the approach that we went with there is to just split out the two different interfaces depending on the interpretation of the data. So my first thought would be to give one interpretation of the TDT event epocs (ndx-events, behavioral), and then if we need to add another one, we could have another interface that reads the data in the same way but writes it to NWB differently.  I guess the thing that makes this a little more complicated is that ndx-events itself doesn't specify a super concrete interpretation for its "timestamped events". 

> First, the NWB container. Writing them into a behavior processing module over-claims, since it asserts the events are behavioral. We could go the parent container route (see https://github.com/catalystneuro/neuroconv/pull/1760, where I would like your review) to let the caller choose, though that becomes obsolete with NWBEP0001's native events location (https://github.com/NeurodataWithoutBorders/nwb-schema/pull/690) anyway. Unsure in how to move forward here, adding a parmater that will be removed soon seems annoying. Maybe we should change the view on https://github.com/catalystneuro/neuroconv/issues/1748 but to avoid the annoynace of overiding the NWW objec itself we can just develop these interfaces against pynwb dev? What do you think?

In light of https://github.com/NeurodataWithoutBorders/nwb-schema/pull/690, I think we should just pick a container (acquisition, processing/behavior, processing/events), accept the potential mild inaccuracy, and change the container to events when we switch over to 0.4.x. I'd rather not develop against pynwb dev. 

> On the conversion itself is some information loss here. The TDT spec shows an epoc carries onset (STRON), offset (STROFF, a duration), and a strobe value. get_events() reads all three, but add_to_nwbfile writes only onset as a plain Events, dropping offset and value. Again, if we had NWBEP0001 there is the natural duration that takes this into account, but given that we don't, maybe we should write those onset and offset as a separate LabeledEvents for the moment. What do you think?

Yes, good catch. Guppy actually only reads the onsets (which is probably a limitation/bug on the guppy side), and sometimes onset = offset, in which case the offset is meaningless. But in this interface we should write both with appropriate labels. This might have implications for the metadata formatting situation, I will have to implement to see...

h-mayorquin:
> In light of https://github.com/NeurodataWithoutBorders/nwb-schema/pull/690, I think we should just pick a container (acquisition, processing/behavior, processing/events), accept the potential mild inaccuracy, and change the container to events when we switch over to 0.4.x. I'd rather not develop against pynwb dev. 

Seems like a fine compromise to me, I think the one that would suffer the code churn is you for the guppy NWB reader so I wanted to warn you. 

Let me know when this is ready for a second look.

pauladkisson:
> Seems like a fine compromise to me, I think the one that would suffer the code churn is you for the guppy NWB reader so I wanted to warn you.

The GuPPy NWB reader actually doesn't depend on location in the NWB file. It searches for all relevant `ndx-events` types regardless of where they live.

pauladkisson:
> Let me know when this is ready for a second look.

Still working through a few things, but a couple of decisions that I can already highlight for discussion:
1. I opted to have a generic interface-based metadata key ("TDTEvents") instead of the recommended event-based metadata key bc a single interface usu has multiple events. Maybe we should consider updating the recommendation in #1759
2. None of the testing data I have access to has any legitimate offset times. I prefer not to develop features without real data to test it against, so I left that as not implemented. I do have some test data with strobe values, so I will implement that.

pauladkisson:
@h-mayorquin this should be ready for another round of review.

h-mayorquin:
One thing I want to check before going further: the metadata shape diverges from the contract in #1759, and I want to know whether that was deliberate.

What `get_metadata()` produces now:

```python
metadata["Events"] = {
    "TDTEvents": {                       # metadata_key, hardcoded default
        "PrtR": {
            "name": "PrtR",
            "description": "Onset times of the TDT epoc 'PrtR'.",
        },
        "PAB_": {                        # a strobe store
            "name": "PAB_",
            "description": "Onset times of the TDT epoc 'PAB_', labeled by strobe value.",
            "labels": {0: "0", 16: "16", 2064: "2064"},
        },
    },
}
```

What #1759 specifies (same data):

```python
metadata["Events"] = {
    "<metadata_key>": {                  # source-derived snake_case (the tank/block name), not a hardcoded constant
        "event_columns": {
            "PrtR": {
                "column_name": "PrtR",
                "description": "Onset times of the TDT epoc 'PrtR'.",
            },
            "PAB_": {                    # a strobe store
                "column_name": "PAB_",
                "description": "Onset times of the TDT epoc 'PAB_', labeled by strobe value.",
                "column_categories": {
                    "labels": {0: "0", 16: "16", 2064: "2064"},
                    # "meanings": {...}  # optional
                },
            },
        },
    },
}
```

So the differences are:

- the `event_columns` wrapper around the per-event entries
- `name` to `column_name`
- the flat `labels` to a nested `column_categories.labels`
- the hardcoded `metadata_key` default (`"TDTEvents"`) to a source-derived one

Was the divergence intentional?

pauladkisson:
> One thing I want to check before going further: the metadata shape diverges from the contract in #1759, and I want to know whether that was deliberate.

Fixed! Sorry for the sloppy code. All the divergences were unintentional except for the hard-coded metadata_key

src/neuroconv/datainterfaces/events/tdt_events/tdteventsdatainterface.py
```diff
@@ -0,0 +1,251 @@
...
+    display_name = "TDTEvents"
+    info = "Data Interface for converting discrete events (epocs) from TDT files."
+    associated_suffixes = ("Tbk", "Tdx", "tev", "tin", "tsq")
+
+    @validate_call
+    def __init__(
+        self,
+        folder_path: DirectoryPath,
+        *,
+        event_names: list[str] | None = None,
```
h-mayorquin: I would like to flip this to `exclude_events` rather than `event_names`. I think is more common to drop specific events rather than to specify only the desired ones which  makes you spell out the long way.

pauladkisson: Done!

h-mayorquin:
Thanks for the improvements. One question, one small request and a comment. I think this is ready to go after the request.

1) Question: The hardcoded `metadata_key`.

    > I opted to have a generic interface-based metadata key ("TDTEvents") instead of the recommended event-based metadata key bc a single interface usu has multiple events.

    I still don't understand this. #1759 already keys per interface (`metadata_key`) and per event below it (`event_columns`), so that part isn't the recommendation. The main reason for deriving `metadata_key` in events and the other modalities is to avoid a collision if the interface is instantiated twice and added to a converter for whatever reason. I think that's unlikely here, but it does happen in roiextractors because they take channel names, and I just like keeping the typing consistent across interfaces. Can you restate your reasoning?

2) Request: The gallery entry. Could you add a link to the TDT events example from the two existing TDT entries (`recording/tdt` and `fiberphotometry/tdt_fp`)? That way a TDT user finds it whether they came in through ephys or fiber photometry. I am genuinely torn about what to do here, other event interfaces live on the acquisition but TDT acquisition living in both places complicates things. Previously @bendichter has voiced his preference for keeping everything that belongs to a format in a single place but with TDT that ship has sailed. As this is not blocking anything, let's defer and unify if the need arises later.

3) Comment: Offsets are unblocked now: I added TDT blocks with real onset/offset (`STROFF`) durations to the ephys test data (https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/178, please take a look when you have a chance). So the offset path that currently raises `NotImplementedError` can be implemented as an enhancement. I think we should do this in a follow up as this is only a stop-gap for the real thing that you want (the GuPPy export) and we can improve this later.

pauladkisson:
> 1) Question: The hardcoded `metadata_key`.
> 
>     > I opted to have a generic interface-based metadata key ("TDTEvents") instead of the recommended event-based metadata key bc a single interface usu has multiple events.
> 
>     I still don't understand this. #1759 already keys per interface (`metadata_key`) and per event below it (`event_columns`), so that part isn't the recommendation. The main reason for deriving `metadata_key` in events and the other modalities is to avoid a collision if the interface is instantiated twice and added to a converter for whatever reason. I think that's unlikely here, but it does happen in roiextractors because they take channel names, and I just like keeping the typing consistent across interfaces. Can you restate your reasoning?

No, _I_ don't fully understand this. I thought that the developer documentation was recommending that the default key should be derived from one of the events converted by the interface. Is that not right? If not "TDTEvents" nor "Dv1A" (or something like that) what _should_ the `metadata_key` be here?

pauladkisson:
> 3) Comment: Offsets are unblocked now: I added TDT blocks with real onset/offset (`STROFF`) durations to the ephys test data (https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/178, please take a look when you have a chance). So the offset path that currently raises `NotImplementedError` can be implemented as an enhancement. I think we should do this in a follow up as this is only a stop-gap for the real thing that you want (the GuPPy export) and we can improve this later. 

Awesome. Sounds good.

pauladkisson:
> 2) Request: The gallery entry. Could you add a link to the TDT events example from the two existing TDT entries (`recording/tdt` and `fiberphotometry/tdt_fp`)? That way a TDT user finds it whether they came in through ephys or fiber photometry. I am genuinely torn about what to do here, other event interfaces live on the acquisition but TDT acquisition living in both places complicates things. Previously @bendichter has voiced his preference for keeping everything that belongs to a format in a single place but with TDT that ship has sailed. As this is not blocking anything, let's defer and unify if the need arises later.

Done!

h-mayorquin (approved):
Per the docs ([events_metadata_structure.rst](https://github.com/catalystneuro/neuroconv/blob/7c1713068c5147d0af8a98405a46e37124e40f46/docs/developer_guide/events_metadata_structure.rst?plain=1#L135)), `metadata_key` should be `None` in the signature and built at construct time as a snake_case key that avoids metadata collisions. Something like:

```python
def __init__(self, folder_path, *, metadata_key: str | None = None, ...):
    ...
    self.metadata_key = metadata_key or "tdt_events"
```

The reason it's `None`-in-the-signature-then-built-in-`__init__` (rather than a constant default) is that some interfaces get instantiated twice with different parameters in one conversion, and their metadata shouldn't collide. **But that's not valid here**: as far as I understand, no one would instantiate two event interfaces in the same conversion, would they? So that concern doesn't apply.

So the [current pattern you have](https://github.com/catalystneuro/neuroconv/blob/285fe8c1ffc8e9465a1418e7f68cd10069117960/src/neuroconv/datainterfaces/events/tdt_events/tdteventsdatainterface.py#L78) works fine. It seems that we were talking pass each other and I was just asking why were you going against the docs here. I'd *like* it `None` in the signature and built as a snake_case in `__init__` for consistency's sake, but that's a nitpick. I'll approve this. 

Separately: there's a bug with the latest Python release that's currently breaking things. I added the fix in the VAME interface PR (#1737), which should merge soon, so we can merge this one after that lands.

pauladkisson:
> Separately: there's a bug with the latest Python release that's currently breaking things. I added the fix in the VAME interface PR (https://github.com/catalystneuro/neuroconv/pull/1737), which should merge soon, so we can merge this one after that lands.

Yes, did you see #1763?

pauladkisson:
> as far as I understand, no one would instantiate two event interfaces in the same conversion, would they?

That's correct. I don't think there's any reason to instantiate two TDTEvents interfaces. 

> It seems that we were talking pass each other 

Yes, I didn't realize it was just a pure formatting fix. That should be no problem at all.

PR #1752, opened 2026-06-18
Split output files by heterogeneous offset

AlexN-SFV:
Related to #1106

Added BaseRecordingExtractorInterface.split_by_offset and BaseRecordingExtractorInterface.run_conversion_split_by_offset to support recordings (e.g. some EDF files) whose channels have heterogeneous offsets, which cannot be represented in a single NWB ElectricalSeries. The channels are partitioned by offset value and each partition is written to its own NWB file (paths derived by inserting an _offset{index} suffix). Recordings with a single offset are written to one file unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v0.9.4 (Upcoming)
 
 ## Features
+* Added `BaseRecordingExtractorInterface.split_by_offset`, which partitions a recording's channels by their offset value (using only the offsets present in the recording, independent of any external metadata) and returns one sub-interface per distinct offset, each with a distinct `es_key`. A single NWB `ElectricalSeries` can only carry one scalar `offset`, so recordings with heterogeneous offsets cannot be written to one series; the returned sub-interfaces can be combined in a `ConverterPipe` to write one `ElectricalSeries` per offset into a single NWB file (or converted individually). Recordings with a single offset return the interface unchanged. [PR #1752](https://github.com/catalystneuro/neuroconv/pull/1752)
 * Added `DoricFiberPhotometryInterface` for converting fiber photometry data from Doric Neuroscience Studio `.doric` HDF5 files. [PR #1727](https://github.com/catalystneuro/neuroconv/pull/1727)
 
 ## Removals, Deprecations and Changes
--- a/docs/conversion_examples_gallery/recording/edf.rst
+++ b/docs/conversion_examples_gallery/recording/edf.rst
@@ -48,8 +48,55 @@ Other auxiliary signal channels should be excluded using the ``channels_to_skip`
     nwbfile_path = f"{path_to_save_nwbfile}"
     interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata, overwrite=True)
 
+Handling Channels with Heterogeneous Offsets
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The NWB ``ElectricalSeries`` stores a single scalar ``offset`` that is shared by all of its channels.
+Some EDF files contain electrode channels whose physical-to-digital offsets differ from one another,
+which cannot be represented in a single ``ElectricalSeries``. In that case a normal ``run_conversion`` call
+raises ``ValueError: Recording extractors with heterogeneous offsets are not supported.``
+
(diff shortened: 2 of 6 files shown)


h-mayorquin (changes requested):
Thanks for picking this up. Before we settle on a mechanism I'd like to understand what angle you're coming from, because there may already be a path for part of this. In our earlier European Data Format (EDF) conversions we've used `channels_to_skip` together with a `ConverterPipe` to carve a single file into channel groups and write more than one series from it, so a single file fans out into multiple `ElectricalSeries` (with auxiliary channels grouped by unit). That pattern is documented here: [Combining Electrode and Auxiliary Channels](https://neuroconv.readthedocs.io/en/main/conversion_examples_gallery/recording/edf.html#combining-electrode-and-auxiliary-channels). So there's a path today.

That said, I do think it's worth making this more automatic, which is what I take you to be going for here. Ideally, we want to automatically create different ElectricalSeries for each signal kind (e.g. [fill here]).  My worry is that the channel offset isn't a good key to partition the `ElectricalSeries` on automatically. The offset is a derived value (it falls out of each channel's physical min/max), so it doesn't track signal kind: different kinds can land on the same offset, and channels of the same kind can land on different offsets. Partitioning on it therefore both over-splits coherent montages and can lump unrelated channels together. Concretely, in the file that originally motivated #1106 the differing offsets all come from non-neural channels (the DC channels and a plethysmography channel) that happen to be labeled in microvolts, while the 256 actual electrode channels share a single offset, so there `channels_to_skip` is really all that's needed.

The way I'd like to improve this is upstream, in python-neo (the library we use to read these files), by making the separation of signals by kind smarter, so the grouping is meaningful instead of keyed on a symptom. Separately, I did find that some files genuinely have the same signal kind with different per-channel offsets (some per-channel-autoscaling exporters do this). I think the NWB schema should support that and I'll follow up there, but in the meantime I agree we should offer an escape hatch to write that edge case without rescaling the data, even if the result is slightly sub-optimal.

So the thing that would help me most: what kind of data are you coming from? Do you genuinely have the same signal kind with different offsets in one file, or are the differing offsets coming from different kinds of channels bundled together? This would be useful to set a course.

AlexN-SFV:
Thanks for the feedback. I am particularly interested in files with the same kind of data with different offsets, such as both extra-cranial and intra-cranial EEG signals. Considering the "Combining Electrode and Auxiliary Channels" path that already exists, I will scale back this PR to just the split_by_offset() function to identify heterogeneous offsets so they can be handled using that path.

h-mayorquin:
Thanks Alex. I'd rather not add a new method to `BaseRecordingExtractorInterface` as I don't want to add a contract to a very general function with a very specific use case. I put up #1814 instead. Between that and what already exists, there are now two workflows for channels with heterogeneous offsets:

1. **Split the channels and write the groups separately** : already supported through channel selection (`channels_to_skip`, and the "Combining Electrode and Auxiliary Channels" example).
2. **Write all the channels into a single `ElectricalSeries` in physical units** rather than digital counts: pass `data_representation="physical_units"` to `add_to_nwbfile()` / `run_conversion()`. This folds each channel's gain and offset into float microvolts, which is what lets differently-offset channels share one series — an `ElectricalSeries` fixes its `unit` to volts and allows only a single scalar `offset`, so the per-channel offset has to go into the data.

The one tradeoff with (2) is that folding into float gives up the lossless integer round-trip. The fully faithful fix would be to keep the raw ints and store a *per-channel* offset, which needs a schema change; I've proposed that at NeurodataWithoutBorders/nwb-schema#694, so `physical_units` is the workaround until (or unless) that lands.

Would (2) be useful for your case?

h-mayorquin:
@AlexN-SFV, an update before I close this.

Three things landed since we last talked here. #1890 makes both offset errors print which channels carry which offset, so the identification half of what you scaled this PR down to comes out of the error itself, and there is a how-to that separates the two cases:
https://neuroconv.readthedocs.io/en/main/how_to/handle_heterogeneous_offsets.html

#1894 fixes a bug I found while working on EDF. `data_representation="physical_units"` was writing the folded values at the recording's own integer dtype, so they came back rounded to whole microvolts. It writes float32 now, which matters for the extra-cranial and intra-cranial case you described.

#1892 adds `stream_name` and `get_stream_names` to `EDFRecordingInterface` and `EDFAnalogInterface`, so an EDF that sampled some of its signals at a different rate than the rest can be read at all.

Between those and #1814 both workflows are in main, so I am closing this one. Say something here if your own files are not covered and I will reopen.

PR #1753, opened 2026-06-18
Support conversion from EEGLAB file format

AlexN-SFV:
* Added `EEGLABRecordingInterface` for converting EEGLAB (`.set` / `.fdt`) data to NWB. The reader is dependency-light (`pymatreader` for the `.set` struct and NumPy for the `.fdt` binary; no MNE) and supports both EEGLAB layouts from the `.set` path: a self-contained `.set` with inline data, and a `.set` paired with an external `.fdt`. The continuous signal is written as an `ElectricalSeries` with channel names and locations, and `EEG.event` markers are written to a `TimeIntervals` table (disable with `write_events=False`). Epoched datasets (`EEG.trials > 1`) are converted with `run_conversion_split_by_epoch` (or `split_by_epoch`), which writes one NWB file per epoch — each with an epoch-relative time axis (the time-locking event at `t = 0`) and only its own events; calling `run_conversion` directly on an epoched dataset raises an informative error.


h-mayorquin (changes requested):
Thanks. It will be useful to support this andit seems you went through the trobule of using a pure implementation instead of mne. Is there a reason why you kept it pure? just out curiosity.

The most critical thing is: Do you have a real `.set` file (or a few) we could use as test data? We keep test data in the gin repo ([ephy_testing_data](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data)) and run the interfaces against it in CI (continuous integration), so we catch regressions on genuine files rather than only synthetic ones. Right now the tests build their fixtures with `savemat`, which validates the logic but is a bit circular (written and read by the same tooling) and won't surface the quirks of an actual EEGLAB export.

One thing I want to understand about the subject metadata: the code reads `EEG.BIDS.pInfo` from inside the `.set`. My understanding is that the usual EEGLAB BIDS export (`bids-matlab-tools` / `bids_export`) writes the participant demographics to a `participants.tsv` at the dataset root, not into the `.set` struct itself, so I'm curious how the `EEG.BIDS` field ended up in your files. How are you producing these, and what does your export actually contain: do you end up with a `participants.tsv` alongside the data, an `EEG.BIDS` field inside the `.set`, or both? I'd like to understand the export before we settle on where to pull subject info from.

AlexN-SFV:
For sample data I used the data in the EEGLAB repo
https://github.com/sccn/eeglab/tree/develop/sample_data

The two reasons I didn't use MNE is that pymatreader is already a dependency and MNE would be a large new one that's currently unused in neuroconv, and MNE's Epochs doesn't expose the original urevent latencies.

An example of EEGLAB .set files that carry an EEG.BIDS.pInfo can be found in this dataset below:
https://openneuro.org/datasets/ds004588/versions/1.2.0

The EEG.BIDS.pInfo is populated by pop_importbids (https://github.com/sccn/EEG-BIDS/blob/master/pop_importbids.m)

And you're right, there is a participants.tsv file in the root that has only participant_id and no demographics, so pInfo adds nothing in this case. I'll make reading the EEG.BIDS.pInfo an opt-in option for those cases when there are no exported sidecar files.

AlexN-SFV:
I'm also removing the run_conversion_split_by_epoch function that created a separate NWB file per epoch, and instead having each epoch as its own ElectricalSeries within one NWB file.

h-mayorquin:
@AlexN-SFV, an update on the MNE side of this. I put up the base I mentioned:
https://github.com/catalystneuro/neuroconv/pull/1898

It covers the continuous half. A subclass builds the `mne.io.Raw` and the base writes it, one interface per channel type, streaming so a `Raw` opened with `preload=False` is never materialized. An EEGLAB subclass on top of it should be small. The epoched half stays yours, since MNE does not expose the `urevent` latencies your shared-timeline reconstruction needs.

Do you want to split this PR along that line, or wait for #1898 to land first?

PR #1754, opened 2026-06-19
Add CSVFiberPhotometryInterface for raw CSV fiber photometry

pauladkisson:
## Motivation

Part of expanding the GuPPy fiber-photometry conversion in NeuroConv beyond TDT to the other acquisition sources GuPPy supports. This adds the **CSV** raw-acquisition interface. The CSV events interface and the CSV ↔ GuPPy converter land in separate PRs, so this one branches cleanly off `main`.

## What this adds

`CSVFiberPhotometryInterface` — a single-series interface (inherits `BaseFiberPhotometryInterface`, writing one `FiberPhotometryResponseSeries`). Point it at one or more CSV files and name the `timestamps_column` and `data_columns` — by header name, or a 0-based index for header-less files. The named data columns read across the files are column-stacked into the series (the multi-file case, e.g. GuPPy's per-channel CSVs, shares the first file's timebase). Like the other fiber-photometry interfaces, `get_metadata()` returns the placeholder scaffold and does not set `session_start_time` (CSV carries none).

Also included: the `csv_fp` extra (`pip install "neuroconv[csv_fp]"`) and interface registration, a conversion-gallery page + API docs, and tests over dynamically generated CSVs built on `FiberPhotometryInterfaceTestMixin`.

## Testing

- `pytest tests/test_modalities/test_fiber_photometry/test_csv_fiber_photometry_interface.py` → 20 passed
- Gallery doctest (`csv_fp.rst`) → passed


pauladkisson (commented):
This looks good to me and is ready for review.

pauladkisson:
@h-mayorquin this is now ready for review.

docs/conversion_examples_gallery/fiberphotometry/csv_fp.rst
```diff
@@ -0,0 +1,71 @@
...
+    >>> # should be replaced before archiving. add_to_nwbfile warns about any that remain unset.
+
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> nwbfile_path =  f"{path_to_save_nwbfile}"
+    >>> # stub_test writes only the first stub_samples samples, which is useful for quick tests
+    >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata, stub_test=True, overwrite=True)
+
+The full metadata format (device models, devices, indicators, the ``FiberPhotometryTable``, and the
+per-interface response series) is shared across the fiber photometry interfaces and documented at
+:ref:`fiber_photometry_metadata_structure`.
```
h-mayorquin, line 67: I would like to remove the file generation logic here. This can be done in the conftest like we do for mda. It is a good idea to keep the header and a couple of rows so the user can see, at glance, what is the input we expect, but I think the code generation detracts from the spirit of the gallery.

pauladkisson: Makes sense to me. Done!

h-mayorquin (changes requested):
Discussion:

I'd like to make this a single `file_path` interface. Supporting multiple file paths requires checks we aren't currently doing: we'd need to verify that the `data_columns` match across the files and that their timestamps line up, because only channels on a common timebase can be aggregated into one `FiberPhotometryResponseSeries`. And if they're not being aggregated into one series, then two single-`file_path` interfaces in a `ConverterPipe` already do the same job, more modularly. Right now there's no such check, so if someone points `file_paths` at two files that don't share a common timebasis, they get a silently mis-timed series (every channel pinned to the first file's timestamps) instead of an error.

That said, maybe there is a case to keep multi-file support. I know GuPPy happens to write one file per region, and those do meet the criteria above (same structure, shared timebase), so it will be a consumer of this. Do you think there will be others? Either way my preferred shape is the same: keep `file_path` as the primary argument, and if we support the multi-file case add it as an optional argument named for its purpose, `file_paths_for_additional_streams`. Each extra file becomes one more stream/channel of the same series (it is stacked), and the argument would assert the extra files share the primary's `data_columns` and timestamps, otherwise it raises and points the user to the converter pipe workflow (channels on different timebases belong in separate series). 

The reason I'm protective about `file_paths` as a single argument here is consistency and defending simplicity for the common case. By consistency I mean that almost everywhere else in the library `file_paths` is used to concatenate files across time (video segments, tiff-files), whereas here it would concatenate across channels, a different operation under the same name. And since I expect the common usage is a single file, I'd rather keep `file_path` as the primary argument so it doesn't surprise or pull users away, and add the multi-file case as an extra, properly documented `file_paths_for_additional_streams`.

pauladkisson:
> I'd like to make this a single file_path interface.

The problem is it is already a single-series interface (one FP series), and there are cases like in Guppy where you want multiple CSV files to be aggregated into one series. 

> Supporting multiple file paths requires checks we aren't currently doing: we'd need to verify that the data_columns match across the files and that their timestamps line up

Actually, we are checking that the data columns match across the files -- see `_assert_columns_present`. The timestamps are assumed to match, which is a documented constraint in the docstrings of the interface. And I don't think it would be appropriate to assert equality because the secondary CSV files could omit the timestamps column entirely, and it would still work just fine as long as they're on the same time basis. 

> That said, maybe there is a case to keep multi-file support. I know GuPPy happens to write one file per region, and those do meet the criteria above (same structure, shared timebase), so it will be a consumer of this. Do you think there will be others? Either way my preferred shape is the same: keep file_path as the primary argument, and if we support the multi-file case add it as an optional argument named for its purpose, file_paths_for_additional_streams. Each extra file becomes one more stream/channel of the same series (it is stacked), and the argument would assert the extra files share the primary's data_columns and timestamps, otherwise it raises and points the user to the converter pipe workflow (channels on different timebases belong in separate series).
> 
> The reason I'm protective about file_paths as a single argument here is consistency and defending simplicity for the common case. By consistency I mean that almost everywhere else in the library file_paths is used to concatenate files across time (video segments, tiff-files), whereas here it would concatenate across channels, a different operation under the same name. And since I expect the common usage is a single file, I'd rather keep file_path as the primary argument so it doesn't surprise or pull users away, and add the multi-file case as an extra, properly documented file_paths_for_additional_streams.

I understand the desire for simplicity here, but to me, adding an extra constructor argument for additional CSV files seems like it would be a more complicated solution. 

If you'd really like to insist on a simpler interface, perhaps we could refactor out the multi-file case into a dedicated multi-file CSV fiber photometry interface?

h-mayorquin:
Ok, yes, I misspoke above: what I meant is that the *timestamps* check is not present (the `data_columns` one is, via `_assert_columns_present`, you're right). And I did not know about the case where a secondary CSV file can omit the timestamps column entirely. That does change it, you can't assert timestamp equality when there is nothing to compare. I guess we can do that conditionally on the existence of the timestamps but I think that's up to you if you want to separate. 

Honestly, my first reaction was exactly what you propose: keep this interface simple and single-file, and use a dedicated aggregator for the GuPPy multi-file case. I thought that handling everything in one class (the optional `file_paths_for_additional_streams` argument) would be less trouble for you. But if you think the dedicated multi-file interface is the better design, then we are aligned, let's go that way.

pauladkisson:
> Honestly, my first reaction was exactly what you propose: keep this interface simple and single-file, and use a dedicated aggregator for the GuPPy multi-file case. I thought that handling everything in one class (the optional file_paths_for_additional_streams argument) would be less trouble for you. But if you think the dedicated multi-file interface is the better design, then we are aligned, let's go that way.

Done!

docs/conversion_examples_gallery/fiberphotometry/csv_fp.rst
```diff
@@ -0,0 +1,71 @@
...
+    >>> metadata["NWBFile"]["session_start_time"] = datetime.now(tz=ZoneInfo("US/Pacific"))
+    >>> # Add subject information (required for DANDI upload)
+    >>> metadata["Subject"] = dict(subject_id="subject1", species="Mus musculus", sex="M", age="P30D")
+    >>> # get_metadata() returns an editable scaffold; the required fiber photometry fields (excitation/
+    >>> # emission wavelengths, indicator, location, ...) are pre-filled with placeholder values that
+    >>> # should be replaced before archiving. add_to_nwbfile warns about any that remain unset.
+
+    >>> # Choose a path for saving the nwb file and run the conversion
+    >>> nwbfile_path =  f"{path_to_save_nwbfile}"
+    >>> # stub_test writes only the first stub_samples samples, which is useful for quick tests
```
h-mayorquin: we usually don't have this on the other galleries, the stub_test thingy.

docs/conversion_examples_gallery/fiberphotometry/csv_fp.rst
```diff
@@ -0,0 +1,71 @@
...
+    >>> available_columns = CSVFiberPhotometryInterface.get_available_columns(file_path=csv_signal_channel_path)
+
+    >>> interface = CSVFiberPhotometryInterface(file_path=csv_signal_channel_path, data_columns="data", timestamps_column="timestamps", metadata_key="calcium_signal", verbose=False)
+    >>> metadata = interface.get_metadata()
+    >>> # CSV recordings have no embedded start time, so it must be set explicitly.
+    >>> metadata["NWBFile"]["session_start_time"] = datetime.now(tz=ZoneInfo("US/Pacific"))
+    >>> # Add subject information (required for DANDI upload)
+    >>> metadata["Subject"] = dict(subject_id="subject1", species="Mus musculus", sex="M", age="P30D")
+    >>> # get_metadata() returns an editable scaffold; the required fiber photometry fields (excitation/
+    >>> # emission wavelengths, indicator, location, ...) are pre-filled with placeholder values that
```
h-mayorquin: I think having a how-to would be a good idea. Another PR of course.

src/neuroconv/datainterfaces/fiber_photometry/csv/multifilecsvfiberphotometrydatainterface.py
```diff
@@ -0,0 +1,130 @@
...
+        BaseFiberPhotometryInterface.__init__(
+            self,
+            stream_names=self._file_paths,
+            timestamps_column=timestamps_column,
+            metadata_key=metadata_key,
+            verbose=verbose,
+        )
+        self._assert_timestamps_aligned(timestamps_column)
+
+    def _assert_timestamps_aligned(self, timestamps_column: str | int) -> None:
```
h-mayorquin: I think that you could start by checking if they have same number of rows and only if they do and they have a timestmaps then check that right? the number of rows check is also useful in the case where there are no timestamps in other files.

src/neuroconv/datainterfaces/fiber_photometry/csv/multifilecsvfiberphotometrydatainterface.py
```diff
@@ -0,0 +1,130 @@
...
+        Only channels on a common time axis can be column-stacked into one series. Secondary files may
+        omit their (redundant) timestamps column; those that keep it must match the first file, so
+        misaligned files fail loudly instead of being silently pinned to the first file's timestamps.
+        """
+        reference_timestamps = self._get_stream_timestamps(stream_name=self._file_paths[0])
+        for file_path in self._file_paths[1:]:
+            if not self._file_has_column(file_path, timestamps_column):
+                continue
+            timestamps = self._get_stream_timestamps(stream_name=file_path)
+            assert timestamps.shape == reference_timestamps.shape and np.allclose(timestamps, reference_timestamps), (
```
h-mayorquin: `np.allclose`'s default `rtol=1e-5` is too loose here. Fiber photometry timestamps are large-magnitude seconds (GuPPy ~`10251`, NPM ~`57932`), so the effective tolerance is `rtol*|b| ≈ 0.1 s`, larger than the sub-sample interleave offset this guard needs to catch.

Concretely, from the Guppy data `npm_phat_donaldson` data: `chev1` vs `chpr1` are the same ROI under two different interleaved LEDs, so they differ by ~`dt/3` (`10251.836512` vs `10251.828192`, i.e. `0.00832 s`). Since `0.00832 < 0.1`, `np.allclose` returns `True`, the two get stacked despite being on different timebases, and `chpr1` is written on `chev1`'s time axis, shifted ~8 ms.

src/neuroconv/datainterfaces/fiber_photometry/csv/multifilecsvfiberphotometrydatainterface.py
```diff
@@ -0,0 +1,130 @@
...
+
+import numpy as np
+import pandas as pd
+from pydantic import FilePath, validate_call
+
+from .csvfiberphotometrydatainterface import CSVFiberPhotometryInterface
+from ..basefiberphotometryinterface import BaseFiberPhotometryInterface
+
+
+class MultiFileCSVFiberPhotometryInterface(CSVFiberPhotometryInterface):
```
h-mayorquin, line 12: Is this meant to be used by the GuppyConverter when the input format is CSV? If so, maybe this should be private, have a guppy relevant name and be called within the converter, or what kind of workflows are you envisioning?

If we want to go with this public I would not call it `MultiFileCSVFiberPhotometryInterface`, for the reasons discussed previously: in the library multi-file is used to stack across time, whereas here it is stacked across channels. Something like `ChannelPerFileCSVFiberPhotometryInterface` names the actual layout (one channel per file) and pairs cleanly with `CSVFiberPhotometryInterface` (channels as columns of one file).

h-mayorquin (changes requested):
I did a second reading, some other points.

Playing with this I realize that by default this is creating a fiber photometry table (plus the devices and an indicator). I don't think it should, the source carries no provenance to write, and the response series is valid without a table (`fiber_photometry_table_region` is optional). This is base-level behavior, so nothing to fix in this PR; I opened #1789 to track it. What do you think?

src/neuroconv/datainterfaces/fiber_photometry/csv/multifilecsvfiberphotometrydatainterface.py
pauladkisson: I prefer not to make it private nor Guppy-specific, as I envision this interface would be generally useful for the case that you want to aggregate columns from multiple CSV files into a single fiber photometry response series.

I played around with some different names, and none of them seemed particularly good. `ChannelPerFile` doesn't quite work because you could actually have multiple channels per file, all named according to the same data column names.  

Is `MultiFile` really so bad? I'm not even sure which interface you're referring to that is named multi-file stacking across time.

pauladkisson:
Other than the one outstanding point of feedback above, all feedback has been addressed.

src/neuroconv/datainterfaces/fiber_photometry/csv/multifilecsvfiberphotometrydatainterface.py
h-mayorquin: Not terrible, fair. To your question about which interface I mean: none are literally named `MultiFile`. The convention lives in the `file_paths` argument itself, and on every time-series interface that takes it, multiple files means concatenate across time, in order:

- `VideoInterface` / `ExternalVideoInterface`: *"segment a sequence of videos over the course of the experiment... in sorted, consecutive order."*
- `AudioInterface`: *"the file paths to the audio recordings in sorted, consecutive order."*
- `TiffImagingInterface` / `ScanImageImagingInterface`: more files means more frames along the time axis.
- `ABFInterface`: multiple files means sequential recordings.

The only exception is `ImageInterface`, where `file_paths` is an unordered collection of stills and "across time" does not apply, so still not our case. 

This one would be the first where multiple files stack across *channels* on a shared timebase, the opposite axis, which is why `MultiFile` reads slightly against the grain to me. `ColumnStackedCSVFiberPhotometryInterface` names the actual operation and sidesteps the overload.

That said, this is a preference, not a blocker. The timebase-alignment check means a user who misreads it and passes time-consecutive files gets a loud error, not a mis-shaped series, so nothing silently breaks either way. I lean `ColumnStacked`, but if you still like `MultiFile` after this, let's agree to disagree and keep what you have built. Your call.

src/neuroconv/datainterfaces/fiber_photometry/csv/multifilecsvfiberphotometrydatainterface.py
h-mayorquin: @pauladkisson 
Half of this is addressed: for files that carry the timestamps column you now check `timestamps.shape == reference_timestamps.shape` before comparing values, which is the rows-first ordering I asked for.

What's still missing is the second half: the row-count check for files that have **no** timestamps column. Those hit the `continue` at the top of the loop and get no length check at all, so a timestamp-less file of the wrong length passes construction and only fails later, deep in `_read_response_data`, as an opaque numpy error. Reproduced on this branch (file A has `timestamps`+`data` with 20 rows, file B has only `data` with 10 rows):

src/neuroconv/datainterfaces/fiber_photometry/csv/csvfiberphotometrydatainterface.py
```diff
@@ -0,0 +1,145 @@
...
+        # the only source config the seams read back from source_data.
+        super().__init__(
+            stream_names=[file_path],
+            timestamps_column=timestamps_column,
+            metadata_key=metadata_key,
+            verbose=verbose,
+        )
+
+    @staticmethod
+    def _assert_columns_present(file_path: str, columns: list[str | int]) -> None:
```
h-mayorquin: I think you are not propagating `read_kwargs` to these checks, right? `_read_dataframe` applies them (line 133-134), but `_assert_columns_present`, `get_available_columns`, and `_file_has_column` all call `pd.read_csv` bare with the default comma parser. 

Since the CI is passing, this means we have no test exercising `read_kwargs`, as this path is not failing here.

h-mayorquin:
> Other than the one outstanding point of feedback above, all feedback has been addressed.

Two additional comments, everything else looks Ok.

src/neuroconv/datainterfaces/fiber_photometry/csv/multifilecsvfiberphotometrydatainterface.py
pauladkisson: I see. Sorry, I missed that second request.

pauladkisson: Fixed!

src/neuroconv/datainterfaces/fiber_photometry/csv/multifilecsvfiberphotometrydatainterface.py
pauladkisson: Let's agree to disagree.

src/neuroconv/datainterfaces/fiber_photometry/csv/csvfiberphotometrydatainterface.py
pauladkisson: Thanks for the catch! `read_kwargs` was not getting propagated to all of the `pd.read_csv` calls.

I fixed that and also added a test to cover it.

h-mayorquin (approved):
LGTM

PR #1755, opened 2026-06-19
Add CSVEventsInterface for discrete events from CSV

pauladkisson:
## Motivation

We're expanding the GuPPy fiber-photometry conversion support in NeuroConv horizontally to the CSV acquisition format. The first piece, `CSVFiberPhotometryInterface` (#1754), reads the continuous signal/control CSVs and deliberately **excludes** the single-column event/TTL CSVs that GuPPy sessions also contain (its stream discovery skips any CSV without a `data` column). This PR adds the missing piece for those event streams.

It's scoped to just the events interface — the CSV↔GuPPy converter that would compose these interfaces is a separate PR. This one has no dependency on `CSVFiberPhotometryInterface`, so it branches off `main`.

## What this adds

`CSVEventsInterface` — the CSV parallel to `TDTEventsInterface`. It reads discrete event timestamps from single-column `timestamps` CSV files (one CSV per event stream, named `<event_name>.csv`) and writes each as an `ndx_events.Events` object into a `behavior` ProcessingModule.

- **Folder-based** with an optional `event_names` list; when omitted, it auto-discovers the event CSVs in the folder (single `timestamps` column — the complement of what `CSVFiberPhotometryInterface` reads).
- **No invented session start time:** like the fiber photometry interface, `get_metadata()` does not set `session_start_time` (CSV recordings carry none); the user supplies it.
- Targets the `ndx-events==0.2.2` `Events(name, description, timestamps)` API, matching `TDTEventsInterface`/`MedPCInterface` and the existing pin.

Also included: the `csv_events` optional-dependency extra (`pip install "neuroconv[csv_events]"`), interface registration, an on-data interface test, and conversion-gallery + API docs.

## Test data

The test and gallery doctest read the `Sample_TTL.csv` event stream from the stubbed `CSV/sample_data_csv_1` fixture in the `ophys_testing_data` dataset (already uploaded to AWS, so CI passes). For reference, the data was added via GIN PR [CatalystNeuro/ophys_testing_data#47](https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/47).

## Testing

- `pytest tests/test_on_data/behavior/test_csv_events_interface.py` → 7 passed (construction/discovery, no-`session_start_time`, schema validation, `add_to_nwbfile`, and round-trip; timestamp values hard-coded from `Sample_TTL.csv`)
- Gallery doctest (`csv_events.rst`) executes against the gin stub data → passed
- `test_keyword_only_arguments.py[CSVEventsInterface]` → passed


pauladkisson (commented):
Code looks good to me and is ready for review.

h-mayorquin (changes requested):
Hi, thanks for this one. What I would like is a general CSV events interface, flexible enough to serve as the basis for other CSV-based event interfaces like NPM (#1757), which would then become a thin wrapper over it. That fits the modular split you already lay out in the PR: you keep the GuPPy-to-NWB converter as a separate piece, and that converter is the right place to walk a session folder and aggregate the per-file interfaces (this events one plus the fiber-photometry one) at the end. So the idea is to have a flexible interface that will be composed by another layer and the following comments go in that direction:


**I think this should be a `file_path` interface.** A general interface should point to a `file_path` rather than the `folder_path` we use now: the most reliable way to identify an event CSV is to have the user point at one explicitly. Auto-detecting a single timestamps column is fragile too, it misses events that carry a payload, and the time column is not always named predictably. Better to make these choices explicit, and in the same vein...

**The `__init__` should specify the timestamps column and the event types.** There are three possible layouts here:

* csv with only timestamps
* csv with extra value columns that represent a single event.
* csv with extra value columns where one of the columns specifies the event type (and therefore requires a group-by separation).

How to provide knobs for the user to specify this? My proposal:

- `timestamps_column` (`str | int`): which column holds the event onset times. A column name for a CSV with a header, or a positional index for a header-less one.

  Why we ask for it explicitly: a CSV has no convention for which column is time, and the time column is not always named predictably, so naming it is more robust than guessing (and guessing breaks the moment a file has more than one numeric column).

- `event_type_column` (`str | int | None`): which column, if any, names the type of each event. Required, keyword-only. Pass a column name or index when the file holds several event types told apart by that column, so each distinct value becomes its own event type (the one-file-many-types case, which is what NPM is). Pass `None` when the file is a single event type with no such column, in which case the event type id comes from the filename (renamable later in the metadata, which is exactly today's behavior).

    Why it is required rather than optional: the two cases produce structurally different NWB output (one events table per value vs a single table), and the bytes do not say which is meant. If the argument were optional, a multi-type file would be silently written as one undifferentiated event type whenever the caller simply forgot to pass it. Making it required, with `None` as the explicit "single type" answer, forces the caller to state which file they have rather than fall into the wrong default.


**Metadata.** Same ask as on the TDT PR: please start from the unified events metadata schema in #1759 (add a `metadata_key`, and nest the metadata under `Events` rather than `behavior`). I wrote the full argument on #1751, so I'm just linking #1759 here as well.

pauladkisson:
This should be ready for another round of review @h-mayorquin

.github/workflows/testing.yml
```diff
@@ -160,6 +160,27 @@ jobs:
             deactivate
           fi
 
+      - name: Test events machinery
```
h-mayorquin, line 163: This looks good.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -0,0 +1,222 @@
...
+from neuroconv.tools import get_package
+from neuroconv.utils import DeepDict
+
+
+class CSVEventsInterface(BaseDataInterface):
+    """Data Interface for converting discrete events from a single CSV file.
+
+    This is a general-purpose CSV events reader: the caller points at one CSV file and names the
+    column holding the event onset times (``timestamps_column``) and, optionally, the column that
+    tells the event types apart (``event_type_column``). The file is written into
```
h-mayorquin: Let's remove the "the file is written into acquisition bit. We don't need to justify internal design choices in a user facing docstring and moreover this will be moot now that the pynwb 4.0 is supported and this will go to events.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -0,0 +1,222 @@
...
+from neuroconv.basedatainterface import BaseDataInterface
+from neuroconv.tools import get_package
+from neuroconv.utils import DeepDict
+
+
+class CSVEventsInterface(BaseDataInterface):
+    """Data Interface for converting discrete events from a single CSV file.
+
+    This is a general-purpose CSV events reader: the caller points at one CSV file and names the
+    column holding the event onset times (``timestamps_column``) and, optionally, the column that
```
h-mayorquin:
```suggestion
    column holding the event timestamps (``timestamps_column``) and, optionally, the column that
```
not necessarily events with durations.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -0,0 +1,222 @@
...
+    tells the event types apart (``event_type_column``). The file is written into
+    ``nwbfile.acquisition``. Acquisition is used because these CSV streams are raw acquired markers
+    (TTLs, sync pulses) whose interpretation is not necessarily behavioral.
+
+    Two layouts are supported:
+
+    - **A single event type** (``event_type_column=None``): every row is one occurrence of the same
+      event, and the file becomes one ``ndx_events.Events`` object (onset timestamps only) named
+      after the file stem.
+    - **Several event types in one file** (``event_type_column`` set): each row carries a label in
```
h-mayorquin, line 26: This probably should be changed. I will make a longer comment as part of the general review as it is an overarching issue.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -0,0 +1,222 @@
...
+        verbose: bool = False,
+    ):
+        """Initialize the CSVEventsInterface.
+
+        Parameters
+        ----------
+        file_path : FilePath
+            The path to the CSV file holding the events.
+        timestamps_column : str or int
+            The column holding the event onset times (seconds). A column name for a CSV with a header
```
h-mayorquin:
```suggestion
            The column holding the event timestamps (seconds). A column name for a CSV with a header
```

What about if the user's timestamps are not in seconds? should we provide a way of them specifying that or do you think we should force the input format to be on seconds?

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -0,0 +1,222 @@
...
+
+    @validate_call
+    def __init__(
+        self,
+        file_path: FilePath,
+        *,
+        timestamps_column: str | int,
+        event_type_column: str | int | None,
+        metadata_key: str | None = None,
+        verbose: bool = False,
```
h-mayorquin, line 54: we should have a `exclude_columns` so they can omit some columns from a csv in case they are corrupted, represent something else, etc. What do you think?

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -0,0 +1,222 @@
...
+        the rows are not grouped). Returns ``(timestamps, None)`` when there is no event-type column.
+        """
+        timestamps_column = self.source_data["timestamps_column"]
+        event_type_column = self.source_data["event_type_column"]
+        # An int column specifier means a header-less file (positional columns); a str means a header row.
+        header = None if isinstance(timestamps_column, int) else 0
+        # float_precision="round_trip" uses an exact, platform-independent float parser; pandas's
+        # default C parser rounds the final ULP differently across platforms (Linux/Windows vs macOS).
+        dataframe = pd.read_csv(self.source_data["file_path"], header=header, float_precision="round_trip")
+        timestamps = dataframe[timestamps_column].to_numpy()
```
h-mayorquin: Do you think we should do something about missing timestamps (np.nan) or wait for the inspector to flag this? Do we have flags on the inspector for this?

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -0,0 +1,222 @@
...
+            as a single ``LabeledEvents`` with one label per distinct value. Pass None when the file
+            is a single event type with no such column, in which case the file is written as a plain
+            ``Events`` named after the file stem.
+        metadata_key : str, optional
+            The key under ``metadata["Events"]`` that namespaces this interface's events metadata.
+            If None (default), ``"csv_events"`` is used.
+        verbose : bool, optional
+            Whether to print status messages, default = False.
+        """
+        super().__init__(
```
h-mayorquin, line 83: First, a quick request: let's expose a `read_kwargs` here and forward it to `pd.read_csv`, the way `CsvTimeIntervalsInterface` does ([its `_read_file` forwards `read_kwargs` straight to `pd.read_csv`](https://github.com/catalystneuro/neuroconv/blob/965eb660b22dc440c6db4eb15602d4214ec7003f/src/neuroconv/datainterfaces/text/csv/csvtimeintervalsinterface.py#L14-L15)). It keeps us consistent with our other CSV interface and lets people handle the usual format quirks (`sep`, `encoding`, `decimal`, `skiprows`, ...) without us trying to anticipate them.

The thing I actually want to discuss is how we handle missing values by default. pandas' `read_csv` treats a whole set of tokens as missing by default (`''`, `None`, `null`, `NA`, `N/A`, `NaN`, ...), and we feed the label column straight into `LabeledEvents`. So a real category value that happens to be one of those (a `reward` column with a `None` no-reward level, an outcome with `hit`/`miss`/`none`) is silently read as missing. A file like:

```
onset,reward
1.0,small
2.0,None
3.0,NA
4.0,null
5.0,
6.0,large
```

currently comes out as `LabeledEvents(labels=['small', 'nan', 'large'], data=[0, 1, 1, 1, 1, 2])`. `None`, `NA`, `null` and a genuinely blank cell all collapse into a single `nan` label, and renaming it in the metadata cannot split them back.

Using `keep_default_na=False` (a `read_csv` option, `True` by default, so it is one of the things `read_kwargs` can carry) reads them literally and avoids this. Since we are exposing `read_kwargs` anyway, a caller could always set it themselves, but the default (someone passing nothing) still collapses silently, and whoever hits it is the least likely to know to change it. So the question is really just the default: do we leave pandas' default (and treat `read_kwargs` as the opt-out), or default to `keep_default_na=False` ourselves so it is safe out of the box (still overridable through `read_kwargs`)? And either way, should we add a couple of tests for these cases (a token-valued label, a blank cell) so the behavior is pinned down?

Curious what you think.

h-mayorquin (changes requested):
This looks good, I added comments but there is one big thing we should discuss.

My understanding is that the semantics of LabeledEvents ([definition in ndx-events](https://github.com/rly/ndx-events/blob/0.2.2/spec/ndx-events.extensions.yaml#L28-L60)) is that you have labels of a single event type. For example, the type of reward (e.g. "water", "sweet", etc.) or size of a stimulus ("large", "small", or a number) or which port a nose-poke happened at ("left", "center", "right"). As far as I understand it was not meant to be an event-type differentiator (e.g. making one label "frame_start", another "lick_time", another "TTL pulse") the way you are doing it here. What I initially had in mind is that we should separate into distinct Events (if we only have the timestamps and the `event_type_column`) or LabeledEvents if there are more columns in the csv table that are genuine event values. Note that now that pynwb 4.0 is merged this is not a big concern for writing, but it might be a concern for your NWB reader on the GuPPy side where you will find data with ndx-events on dandi, right?

Now that pynwb 4.0 is supported, we should modify this (not necessarily here) in the following way:
* In the case when `event_type_column=None`, everything is written to a single EventsTable.
* In the case when `event_type_column` names a column of event types, each unique value in that column gets assigned its own table.

So it is always the same object type (an EventsTable): any extra value columns just become other columns in that table.

After this round of comments is fixed, two options: I can take care of the migration to the new events for this and TDT, or if you want to do the pynwb 4.0 events migration here that is fine by me as well. What's your call?

pauladkisson:
> This looks good, I added comments but there is one big thing we should discuss.
> 
> My understanding is that the semantics of LabeledEvents ([definition in ndx-events](https://github.com/rly/ndx-events/blob/0.2.2/spec/ndx-events.extensions.yaml#L28-L60)) is that you have labels of a single event type. For example, the type of reward (e.g. "water", "sweet", etc.) or size of a stimulus ("large", "small", or a number) or which port a nose-poke happened at ("left", "center", "right"). As far as I understand it was not meant to be an event-type differentiator (e.g. making one label "frame_start", another "lick_time", another "TTL pulse") the way you are doing it here. What I initially had in mind is that we should separate into distinct Events (if we only have the timestamps and the `event_type_column`) or LabeledEvents if there are more columns in the csv table that are genuine event values. Note that now that pynwb 4.0 is merged this is not a big concern for writing, but it might be a concern for your NWB reader on the GuPPy side where you will find data with ndx-events on dandi, right?

Your understanding is right, but the difference is fuzzy ("left" nose_poke or left_nose_poke?). In particular, some formats, like NPM, use coded values to refer to different event types. So the difference can't be straightforwardly inferred from the source data. 

ex. 
```
timestamps, values
1, 0
2, 0
3, 1
4, 1
5, 2
6, 2
```
left_nose_poke = 0, right_nose_poke = 1, unrelated experimental stimulus = 2

Guppy flattens everything into a single array per event type (no values allowed). So the real challenge is going to be adding support for the new pynwb 4.0 events format.

> Now that pynwb 4.0 is supported, we should modify this (not necessarily here) in the following way:

> In the case when event_type_column=None, everything is written to a single EventsTable.
> In the case when event_type_column names a column of event types, each unique value in that column gets assigned its own table.
> So it is always the same object type (an EventsTable): any extra value columns just become other columns in that table.

This makes sense to me from the metadata/nwb perspective, but it's still unclear how that would translate to reading from ambiguous source data.

> After this round of comments is fixed, two options: I can take care of the migration to the new events for this and TDT, or if you want to do the pynwb 4.0 events migration here that is fine by me as well. What's your call?

Ugh, a bit annoying to have to redo the events work. I think a separate PR would be good, but I can take care of it, bc I probably need to develop guppy NWB read support alongside the events PR.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
pauladkisson: I think that's a "solve it when we see it" kind of problem.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
pauladkisson: Agree.

pauladkisson: On second look, the interface only looks for the timestamps column and the event type column, so there's no need to exclude columns since we only include those two.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
pauladkisson: I don't think that we do, since ndx-events would have to implement them. 

Should be pretty straightforward to exclude the Nans though.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
pauladkisson: Yeah, good questions. I think we should probably make it safe out of the box.

pauladkisson:
Responded to all feedback, ready for another round of review @h-mayorquin

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
h-mayorquin: We should add it it in the documentation that we expect the times to be in seconds then. This will take generic csv input and people stores their timestamps in all sorts of ways.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
h-mayorquin: It should not include only those two columns. Events might and do have multi-value payloads and you can write each of those as specific columns in the new EventsTable, but we can do that in the migration PR.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
h-mayorquin: Yes, I meant to clean or error, you went for cleaning which is fine. I think you should add a warning.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
h-mayorquin: Thanks.

h-mayorquin:
> Guppy flattens everything into a single array per event type (no values allowed). So the real challenge is going to be adding support for the new pynwb 4.0 events format.

You mean adding support for Guppy to read events? In both cases it just needs to read the timestamps right? In the new format that is just a column. So is either reading the timestamps from an object (Events and/or LabeledEvents in one case) and 
from a column of an EventsTables in the other. Is this what you meant?


> Your understanding is right, but the difference is fuzzy ("left" nose_poke or left_nose_poke?). In particular, some formats, like NPM, use coded values to refer to different event types. So the difference can't be straightforwardly inferred from the source data.

I'm confused here. I was pointing out that this is a declarative parameter: `event_type_column` is the knob for the user to tell us which values are distinct event types. We do a group-by on it and write each type to its own EventsTable. The fact that it's ambiguous from the bytes/source data is exactly why it's a user knob rather than something we infer. Did I get you wrong?

docs/conversion_examples_gallery/events/csv_events.rst
```diff
@@ -0,0 +1,47 @@
+CSV Events data conversion
+--------------------------
+
+:py:class:`~neuroconv.datainterfaces.events.csv_events.csveventsdatainterface.CSVEventsInterface`
+is a general-purpose reader for discrete events (e.g. TTL pulses) stored in a CSV file. You point it
+at one CSV and name the column that holds the event onset times (``timestamps_column``). Pass
```
h-mayorquin: this still say event onset times, we should express that is on seconds.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -0,0 +1,247 @@
...
+        read_kwargs: dict | None = None,
+        verbose: bool = False,
+    ):
+        """Initialize the CSVEventsInterface.
+
+        Parameters
+        ----------
+        file_path : FilePath
+            The path to the CSV file holding the events.
+        timestamps_column : str or int
```
h-mayorquin, line 62: Mention that this is in seconds.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -0,0 +1,247 @@
...
+            "keep_default_na": False,
+            **self._read_kwargs,
+        }
+        dataframe = pd.read_csv(self.source_data["file_path"], **read_kwargs)
+        # Coerce the timestamps column to numbers directly: keep_default_na=False leaves a blank
+        # timestamp cell as the literal '', so recover the missing values here (blank or non-numeric
+        # -> NaN) independent of the na settings the label column relies on.
+        timestamps = pd.to_numeric(dataframe[timestamps_column], errors="coerce").to_numpy(dtype="float64")
+        labels = None if event_type_column is None else dataframe[event_type_column].to_numpy()
+        # Drop rows with a missing timestamp, keeping the labels aligned.
```
h-mayorquin, line 123: A warning somehow I think here would be important.

h-mayorquin (changes requested):
Ok, some small request and this should be ready to go.

pauladkisson:
> > Guppy flattens everything into a single array per event type (no values allowed). So the real challenge is going to be adding support for the new pynwb 4.0 events format.
> 
> You mean adding support for Guppy to read events? In both cases it just needs to read the timestamps right? In the new format that is just a column. So is either reading the timestamps from an object (Events and/or LabeledEvents in one case) and from a column of an EventsTables in the other. Is this what you meant?

Yes, and it's not really relevant for this PR, so let's not worry about it here.

> > Your understanding is right, but the difference is fuzzy ("left" nose_poke or left_nose_poke?). In particular, some formats, like NPM, use coded values to refer to different event types. So the difference can't be straightforwardly inferred from the source data.
> 
> I'm confused here. I was pointing out that this is a declarative parameter: `event_type_column` is the knob for the user to tell us which values are distinct event types. We do a group-by on it and write each type to its own EventsTable. The fact that it's ambiguous from the bytes/source data is exactly why it's a user knob rather than something we infer. Did I get you wrong?

I feel like maybe we are talking past each other a little bit here. I was trying to point out that there is inherent ambiguity in the label versus annotation distinction (which I can elaborate on more if you want). But I agree that the approach you suggested is the best we can do.

pauladkisson:
This should be ready to go @h-mayorquin

h-mayorquin (approved):
Thanks for bearing with me. 

This should make the migration easier:
https://github.com/catalystneuro/neuroconv/pull/1774

Here is a migration example
https://github.com/catalystneuro/neuroconv/pull/1775

PR #1756, opened 2026-06-22
Add NPMFiberPhotometryInterface for raw Neurophotometrics fiber photometry

pauladkisson:
This PR adds an FP interface for NPM.

One important design decision worth noting is that each NPM interface only writes one LED state at a time to the NWB file. Since different LED states have different timestamps, they cannot easily be combined into a single fiber photometry response series, so such an operation is not possible with this interface, in contrast to some of the others that have more flexibility.


src/neuroconv/datainterfaces/ophys/npm_fp/npmfiberphotometrydatainterface.py
```diff
@@ -0,0 +1,651 @@
...
+        DeepDict
+            The metadata dictionary for this interface.
+        """
+        metadata = super().get_metadata()
+        return metadata
+
+    # ------------------------------------------------------------------
+    # Demultiplexing (ported from GuPPy's NpmRecordingExtractor, self-contained)
+    # ------------------------------------------------------------------
+    def _candidate_data_paths(self) -> list[Path]:
```
pauladkisson: This approach is going to be brittle because it depends on excluding what NPM is not. It would be better to have a characteristic structure of the CSV that makes it clear that it's NPM.

src/neuroconv/datainterfaces/ophys/npm_fp/npmfiberphotometrydatainterface.py
```diff
@@ -0,0 +1,651 @@
...
+
+        Only rows 2-11 are examined to skip potential startup artefacts (e.g. an all-LEDs-on first
+        frame).
+        """
+        state = state.astype(int)
+        unique_state = np.unique(state[2:12])
+        if unique_state.shape[0] > 3:
+            raise ValueError(
+                f"NPM file contains {unique_state.shape[0]} unique channel states ({unique_state.tolist()}), "
+                "but only 1-3 channels are supported."
```
pauladkisson: Why do we only support one to three channels?

h-mayorquin:
We can use this:
https://github.com/catalystneuro/neuroconv/pull/1808

Here to make it a simple coordinate wrapper I think.

pauladkisson:
> We can use this: #1808
> 
> Here to make it a simple coordinate wrapper I think.

Sick, thanks.

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
```diff
@@ -0,0 +1,212 @@
...
+    display_name = "NPMFiberPhotometry"
+    info = "Interface for raw fiber photometry data from modern Neurophotometrics files."
+    associated_suffixes = ("csv",)
+
+    @validate_call
+    def __init__(
+        self,
+        file_path: FilePath,
+        *,
+        led_state: int,
```
h-mayorquin:
```suggestion
         excitation_wavelength: Literal[415, 470, 560],
```

Could we use fiber photometry vocabulary here instead?

LED-state, from the FP3002 user guide (p96), can have the following values:

Image

Those three are the lamps the instrument actually has, so a wavelength names the same thing without the user having to know the encoding. This also implies `get_available_excitation_wavelengths` rather than `get_available_led_states`. 

The reason I would rather not expose the LED state directly is that it admits values that are not fiber photometry channels at all. `get_available_led_states` on the test file returns `[16, 17, 18]` and advertises 16 as a channel, but 16 is the initialization frame: pass it and you get no error, just a one-sample series with a value of 0.0. With a wavelength there is nothing invalid to pass.

In addition, this workflows would allow us to build a converter as we can iterate over all the relevant signals available. 

Finally, LED state numbers do not map cleanly onto a single fiber photometry stream as I have elaborated on the main review thread so this is also wrong in that account.

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
```diff
@@ -0,0 +1,212 @@
...
+        )
+
+    @staticmethod
+    def _default_metadata_key(file_path: FilePath, led_state: int, data_columns: list[str]) -> str:
+        stem = Path(file_path).stem.replace(" ", "_").strip("_").lower()
+        regions = "_".join(str(column).replace(" ", "_").lower() for column in data_columns)
+        return f"fiber_photometry_{stem}_ledstate{led_state}_{regions}"
+
+
+class NPMLegacyFiberPhotometryInterface(CSVFiberPhotometryInterface):
```
h-mayorquin: How do you know this one is a legacy version?

Looking at the examples we have, it reads less like an older format and more like a lab that wired the photometry stream into Bonsai's stock `CsvWriter` node instead of the NPM's `PhotometryWriter`. `CsvWriter.IncludeHeader` defaults to `false`, and its `Selector` property chooses which fields get written, which is exactly what we see here: no header, no state column, and whichever columns that workflow happened to pick. `PhotometryWriter` writes both the header and the state column unconditionally
([source](https://github.com/bruno-f-cruz/neurophotometrics/blob/master/Neurophotometrics/PhotometryWriter.cs),
Apache-2.0), so I don't think it could have emitted this file in any version.

If that is right, there isn't much structure here for us to maintain, since the  structure is whatever the workflow author chose that day. And the class reflects that: it is mostly pass-through to `CSVFiberPhotometryInterface`, with one extra argument, `skip_rows`, which `read_kwargs` can carry anyway.

So I would drop it, and instead note on the NPM gallery page that Bonsai output which did not come through `PhotometryWriter` should be read with the generic CSV interface. If you have seen this is common we can add a how-to I think.

h-mayorquin (changes requested):
I went to the manual and could not find the 17 we use in the test data anywhere; it
only ever enumerates 0 to 7. So I went looking at the implementation instead, and the `LedState` column is not a
state at all. It is a packed word whose three lowest bits carry the excitation. Here is the most
complete mapping I could put together: from the manual, which only goes up to 7; from what we have
seen in our own test data; and from
[`FrameFlags.cs`](https://github.com/bruno-f-cruz/neurophotometrics/blob/master/Neurophotometrics/FrameFlags.cs)
in NPM's own Bonsai package (Apache-2.0), cross-checked against files that write the digital lines as
their own columns.

| LedState | in binary | wavelength bits | meaning |
|---------:|-----------|----------------:|---------|
| 0 | 0000000**000** | **000** | no LED on |
| 1 | 0000000**001** | **001** | 415 nm |
| 2 | 0000000**010** | **010** | 470 nm |
| 4 | 0000000**100** | **100** | 560 nm |
| 6 | 0000000**110** | **110** | 470 nm and 560 nm together |
| 7 | 0000000**111** | **111** | all three LEDs on |
| 16 | 0000010**000** | **000** | no LED on, Output 0 high |
| 17 | 0000010**001** | **001** | 415 nm, Output 0 high |
| 18 | 0000010**010** | **010** | 470 nm, Output 0 high |
| 20 | 0000010**100** | **100** | 560 nm, Output 0 high |
| 273 | 0100010**001** | **001** | 415 nm, Output 0 and Input 1 high |
| 274 | 0100010**010** | **010** | 470 nm, Output 0 and Input 1 high |
| 276 | 0100010**100** | **100** | 560 nm, Output 0 and Input 1 high |
| 529 | 1000010**001** | **001** | 415 nm, Output 0 and Input 0 high |
| 530 | 1000010**010** | **010** | 470 nm, Output 0 and Input 0 high |
| 785 | 1100010**001** | **001** | 415 nm, Output 0 and both inputs high |
| 786 | 1100010**010** | **010** | 470 nm, Output 0 and both inputs high |

Seventeen distinct values, but still only three excitation wavelengths, because everything above the
lowest three bits is a digital line rather than an excitation.

So 17 and 273 are the same LED, and what differs between them is that a TTL is in a different state.
Which means your current test with `led_state=17` reads only part of that channel: on the full
session it picks up 12,379 rows where the 415 nm channel has 15,183. Our fixture is only the first 41
rows of that session, which is before any of this happens, so nothing goes red in CI.

What we should do is mask the three lowest bits and use that to determine the wavelength. I have also
cut a second 60-row excerpt from the same recording, positioned so that it spans the moment the input
goes high and carries all four codes, which I will add to gin so a test can catch this.

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
h-mayorquin: Thinking more about this I would make it fully specific and add nm: `excitation_wavelength_in_nm`. That is already the `FiberPhotometryTable` column name, so the user types the same string in the constructor and in the metadata.

This also dovetails with your option 3 in https://github.com/catalystneuro/ndx-fiber-photometry/issues/55: if a response series is one
wavelength, the parameter selecting it just is the wavelength, and `data_columns` becomes which fibers went into it. NPM is a clean case for that. Every ROI is read off the same camera frame, so all fibers of one excitation share timestamps exactly, while 415 and 470 frames are interleaved and do not, which rules out one series per session physically rather than by taste.

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
pauladkisson: In the Guppy NPM read code, this headerless version of the NPM data is referred to as "data_np", whereas the version with flags or LED state columns is referred to as "data_np_v2". The headered version with flags/LED_state columns is also the more common one in the example data that I have available. 

From these facts, I basically just guessed that the data_np format was some kind of legacy. To get a definitive account of where these files came from, we would have to ask Venus.

> Looking at the examples we have, it reads less like an older format and more like a lab that wired the photometry stream into Bonsai's stock CsvWriter node instead of the NPM's PhotometryWriter. CsvWriter.IncludeHeader defaults to false, and its Selector property chooses which fields get written, which is exactly what we see here: no header, no state column, and whichever columns that workflow happened to pick. PhotometryWriter writes both the header and the state column unconditionally
(source,
Apache-2.0), so I don't think it could have emitted this file in any version.

This seems like a perfectly plausible hypothesis, especially with access to the source code. 

> So I would drop it, and instead note on the NPM gallery page that Bonsai output which did not come through PhotometryWriter should be read with the generic CSV interface.

Very reasonable. It's going to make the GuppyConverter more complicated, but that's okay.

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
pauladkisson: Happy to implement these requests. 

It's worth noting that Guppy already treats the LED states of 17 and 273 as identical by simply using a stride-based demultiplexing. I was already aware of this fact, but I wasn't sure if it was a feature or a bug. The npm manual seems like a very valuable resource on these points. Could you share a link?

pauladkisson:
> I have also cut a second 60-row excerpt from the same recording, positioned so that it spans the moment the input
goes high and carries all four codes, which I will add to gin so a test can catch this.

Please let me know when that testing data is up, and I will add a test for this.

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
h-mayorquin: Here you go:
https://www.mbfbioscience.com/app/uploads/2026/04/FP3002_User_Guide_067.pdf

Marking this as resolved.

h-mayorquin:
The tests are here:
https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/55

Note that this will need some re-org after I merged them and upload to s3.

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
```diff
@@ -0,0 +1,150 @@
...
+from pydantic import FilePath, validate_call
+
+from ..csv.csvfiberphotometrydatainterface import CSVFiberPhotometryInterface
+
+# The three lowest bits of a Flags/LedState word encode the excitation LED; the higher bits are
+# digital TTL lines. Mask to these bits to recover the wavelength regardless of the TTL state.
+_EXCITATION_BITS = 0b111
+_WAVELENGTH_TO_CODE = {415: 1, 470: 2, 560: 4}
+_CODE_TO_WAVELENGTH = {code: wavelength for wavelength, code in _WAVELENGTH_TO_CODE.items()}
+
```
h-mayorquin, line 33: This misses frames with two LEDs on. The FP3001 strobes 470 and 560 together in
`TriggerMode.Trigger1` (`Neurophotometrics/FP3001.cs#L99`), and there is a CC0 dataset whose only
470 nm code is `6`, no `2` anywhere, so the assertion below fires and the file cannot be read. It is
on gin as `multi_wavelength_per_led_state`.

Test the bit rather than the whole field, excluding the other LED in the same emission band:

```python
_WAVELENGTH_TO_CODE = {415: 1, 470: 2, 560: 4}
# 415 and 470 both excite the green band, so a frame with both lit belongs to neither. 560 excites
# red and can share a frame with either.
_SAME_BAND_CODE = {415: 2, 470: 1, 560: 0}
_CODE_TO_WAVELENGTH = {code: wavelength for wavelength, code in _WAVELENGTH_TO_CODE.items()}
```

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
```diff
@@ -0,0 +1,150 @@
...
+        verbose : bool, default: False
+            Whether to print status messages.
+        """
+        data_columns_list = [data_columns] if isinstance(data_columns, str) else list(data_columns)
+        state_column = self._detect_state_column(file_path, read_kwargs)
+
+        code = _WAVELENGTH_TO_CODE[excitation_wavelength_in_nm]
+        state_values = self._read_state_values(file_path, state_column, read_kwargs)
+        matching_states = [value for value in state_values if value & _EXCITATION_BITS == code]
+        assert matching_states, (
```
h-mayorquin, line 166: Related to the comment above:

```python
        code = _WAVELENGTH_TO_CODE[excitation_wavelength_in_nm]
        same_band_code = _SAME_BAND_CODE[excitation_wavelength_in_nm]
        state_values = self._read_state_values(file_path, state_column, read_kwargs)
        matching_states = [
            value
            for value in state_values
            if value & code
            and not value & same_band_code
            and value & _EXCITATION_BITS != _EXCITATION_BITS
        ]
```



`get_available_excitation_wavelengths` needs the same treatment, otherwise the two disagree. It maps
distinct masked values back through `_CODE_TO_WAVELENGTH`, so on that file it reports `[415]` while
the selector would return 470 nm rows.

h-mayorquin:
The data is available on s3 now.

tests/test_on_data/fiber_photometry/test_npm_fiber_photometry_interfaces.py
```diff
@@ -0,0 +1,1172 @@
...
+
+    data_interface_cls = NPMFiberPhotometryInterface
+    interface_kwargs = dict(
+        file_path=MULTI_WAVELENGTH_PER_LED_STATE_FILE,
+        excitation_wavelength_in_nm=470,
+        data_columns=["G0", "G2"],
+        timestamps_column="SystemTimestamp",
+        metadata_key="green_signal",
+    )
+    save_directory = OUTPUT_PATH
```
pauladkisson, line 965: One design decision worth noting is that, strictly speaking, nothing is stopping a user from specifying the wrong excitation wavelength in cases like these. 

In this file, G0 and G2 refer to the 470 nm channel, and R1 and R3 refer to the 560 nm channel, but they share rows (and LED states).

So with the current design, a user could specify `excitation_wavelength_in_nm=470`, but then pick out `data_columns=["R1", "R3"]`. This would load the data without warning or error but be misleading because the actual excitation wavelength was 560nm.

I considered adding an assertion on the values of the data columns to ensure that they contain "G" if the excitation wavelength is 415 or 470, or "R" if the excitation wavelength is 560. But my feeling is that this kind of check would be very brittle.

After taking a little bit more of a detailed look at the user guide, it doesn't look like there is a configuration option to manually specify region labels, which is good news for any assertion on the data columns. There are at least two different formats, though: `RegionG1`, `RegionR2`, etc. and `G1`, `R2`, etc.
 
I still think that the potential brittleness is not worth the extra safety in this case, but I am open to being convinced otherwise.

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
pauladkisson: This should be addressed now.

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
pauladkisson: This should be addressed now.

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
pauladkisson: The NPM legacy interface has been dropped, so I am resolving this thread.

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
h-mayorquin: The `LedState` 6 case is fixed, thanks, `simultaneous_470_and_560.csv` reads now. What didn't come across is `_SAME_BAND_CODE`, which was doing separate work from the bit test, and I don't think I made that clear.

Testing the bit alone is right when the lit LEDs excite different indicators and wrong when they excite the same one. 415 nm and 470 nm both excite the same GCaMP molecule, which is the entire point of the isosbestic, so a frame with both lit gives the green ROI a single number that is the sum of the calcium-dependent and calcium-independent responses, and nothing in the file lets you recover either part. This is a problem.

Concretely, for a GCaMP recording. Say the green ROI reads 100 under 415 nm alone and 300 under 470 nm alone. A frame with both lit reads about 400, and under the current predicate that one number is written into both series: four times the isosbestic trace's own baseline, and 1.33 times the signal trace's, which messes up the fit when doing the control.

So we should:

```python
_SAME_BAND_CODE = {415: 2, 470: 1, 560: 0}
matching_states = [
    value for value in state_values
    if value & code and not value & _SAME_BAND_CODE[excitation_wavelength_in_nm]
]
```

plus the same clause in `get_available_excitation_wavelengths`, or the two disagree. Existing tests should be unaffected.

src/neuroconv/datainterfaces/__init__.py
```diff
@@ -252,6 +256,8 @@
         if "Segmentation" in interface.__name__
     },
     fiber_photometry={
+        "TDTFiberPhotometry": TDTFiberPhotometryInterface,
```
h-mayorquin: I think this is an accident.

tests/test_on_data/fiber_photometry/test_npm_fiber_photometry_interfaces.py
h-mayorquin: Yes, it seems premature to add an assertion in the code with what we have, but I do think we should help users get to the regions and make the interface easier to use.

I did another round over the old codebase and the manuals, and I think we can get the regions by subtraction rather than by matching. The metadata columns are more or less closed: `FrameCounter`, `Timestamp`, `SystemTimestamp`, `ComputerTimestamp`, `TriggerEvents`, `Flags`, `LedState`, plus `Stimulation`, `Output0`, `Output1`, `Input0` and `Input1`, which only existed between 2021-05-20 and 2021-11-07. They're hardcoded in `PhotometryWriter` with `MetadataOffset = 3`, and the guide corroborates it positionally with "Column 5+, Region Data". To be safe against a variant we haven't seen I'd also drop anything ending in `Timestamp` or `Counter`. Everything left over is a region.

With that we can add a `get_available_region_columns` that returns all the columns minus the metadata ones. The output is then directly pastable into the constructor, rather than `['FrameCounter', 'Timestamp', 'Flags', 'Region0G', ...]` with the user working out which of them are the data. Along those lines I also think we should rename `data_columns` to `region_columns` to make this more transparent: `data_columns` is the base's generic name because a generic CSV interface can't know what its columns are, but here we do know they're regions, which is the same reasoning that took us from `led_state` to `excitation_wavelength_in_nm`.

Now, we could go even further and get the regions associated with a specific band. Two spellings are in the wild, one per software generation: the open Bonsai package writes `Region0G`, `Region1R`, building the name as `RegionLabel + Region.Index + modeLabel`, while the current closed MBF package writes `G0`, `R1`, which is the only form the 0.6.7 guide uses. Either way the band is in the name as a `G` or an `R`, appended in the first spelling and prepended in the second, and it isn't user text, the software assigns it from which half of the sensor the ROI sits on ([`Photometry.cs:91-92`](https://github.com/h-mayorquin/neurophotometrics/blob/9544d8accb34b51d4ad2afb6ab5f2094dd5910b8/Neurophotometrics/Photometry.cs#L91-L92), and p. 27 of the guide says the same in words). 415 nm and 470 nm are the green band, 560 nm is the red one. The caveats are that the prefix *is* user text when a `GroupRegions` operator is upstream, which gives you `Mouse10G`, that the letter is missing entirely when the user drew no ROIs at all, in which case the whole frame becomes a single unlabelled `Region0` spanning both halves, and that older packages spelled the prefix `Channel` rather than `Region`.

None of this is a blocker for me here, but if we want a converter we would need something of the sort, since it has to work out on its own which columns go with each wavelength it iterates over. What do you think?

src/neuroconv/datainterfaces/fiber_photometry/npm/npmfiberphotometrydatainterface.py
pauladkisson: Just to be clear, this really only comes up with LED state 7, right?

pauladkisson: I'm not so sure that excluding is the right thing to do here. It really sounds like a kind of post-processing, analysis-time decision.

src/neuroconv/datainterfaces/__init__.py
pauladkisson: Yes, crept in during merge resolution. Should be fixed now.

tests/test_on_data/fiber_photometry/test_npm_fiber_photometry_interfaces.py
pauladkisson: I think the first suggestion to add `get_available_regions` instead of data columns and subtract off the known metadata column names is a good one. 

> Now, we could go even further and get the regions associated with a specific band.

This is the part where I get a little bit worried about reliability. And I think I'd like to wait on it until we really need it.

pauladkisson: > With that we can add a get_available_region_columns that returns all the columns minus the metadata ones. The output is then directly pastable into the constructor, rather than ['FrameCounter', 'Timestamp', 'Flags', 'Region0G', ...] with the user working out which of them are the data. Along those lines I also think we should rename data_columns to region_columns to make this more transparent: data_columns is the base's generic name because a generic CSV interface can't know what its columns are, but here we do know they're regions, which is the same reasoning that took us from led_state to excitation_wavelength_in_nm.

Implemented!

doric_fiber_photometry.nwb
h-mayorquin, line 1: This should be removed.

h-mayorquin (changes requested):
LGTM. 

I am not approving because you added to git a nwb file that should be removed. This is something that we need to fix on the fiber photometry for Doric and I will push a PR soon.

tests/test_on_data/fiber_photometry/test_npm_fiber_photometry_interfaces.py
h-mayorquin: Thanks.

h-mayorquin:
Following in the discussion above:
https://github.com/catalystneuro/neuroconv/pull/1756#discussion_r3669936855

This is not a blocker

> Just to be clear, this really only comes up with LED state 7, right?

It is more about 3, where the signal and the isosbestic control are on at the same time and we might want to separate them. 7 has two meanings: the initialization frame, which is what every 7 in our data is and which your `skip_rows` already removes, and all three lamps genuinely lit, which is a variety of 3 with a legitimate 560 nm alongside. That second meaning, and 3 itself, are hardware capabilities I checked in the firmware and the Bonsai sources, and we have no data for either.

> I'm not so sure that excluding is the right thing to do here. It really sounds like a kind of post-processing, analysis-time decision.

I am not sure post-processing is the right frame, because I don't think there is a neutral choice here. Let me explain by just enumerating the options:

**A) Separate every combination:** as IBL does, each distinct set of lit LEDs becomes its own series ([`neurophotometrics.py:252-272`](https://github.com/int-brain-lab/ibl-photometry/blob/c5f0bae9a54b726a6e36f507e81bf3c4d3e1cf41/src/iblphotometry/neurophotometrics.py#L252-L272), state table at [`:21`](https://github.com/int-brain-lab/ibl-photometry/blob/c5f0bae9a54b726a6e36f507e81bf3c4d3e1cf41/src/iblphotometry/neurophotometrics.py#L21)).

**B) Separate per LED:** as you are doing, a sample joins a series when that LED's bit is set.

**C) Separate per LED, within an emission band:** what I proposed, a sample joins a series when that LED was the only one of its band lit. On a GCaMP recording that is what keeps the 470 nm signal and the 415 nm isosbestic control apart, since both land in the same green columns.

Why is the choice not neutral? 

From the provenance perspective A keeps everything, which is great, but it fuses series that are genuinely separable and leaves them with no wavelength to write: a `LedState` 6 frame is 470 nm in the green columns and 560 nm in the red ones, two independent measurements on one timestamp, and A collapses them into a single series with `excitation_wavelength_in_nm = NaN` (this is what our gin example `multi_wavelength_per_led_state/simultaneous_470_and_560.csv` is for testing). The approach A of the IBL reduces the archival value of the file and makes analysis harder. 

B keeps the series usable, but if we annotate the `FiberPhotometryTable` saying the excitation wavelength was 415 nm, that is wrong for any sample where 470 nm was also on, so the metadata is not completely faithful. C, my proposal, has the downside of not storing the original data as faithfully as possible, dropping samples where the camera integrated two LEDs from the same band.

Now, I guess a post-processing step could clean B up. Under B a compound sample lands in both series with the same timestamp, so in principle the user can recover it by intersecting the two timebases and masking whatever coincides. But that is a lot to ask: it only works if both channels were converted, and it is not obvious to me how one would annotate either the procedure or the need for it, since it is not standard practice. I did check the downstream cost by injecting such samples into real GuPPy recordings and refitting, and the corrected trace ends up dominated by the artefact rather than mildly degraded, happy to go into it if useful.

At this point I am unsure what is best myself, and I think we should leave this to the user to decide. Concretely, a `Literal` argument on the interface, `same_band_samples: Literal["include", "exclude"] | None = None`, and if a file turns out to contain same-band samples while it is still unset we raise and ask for a policy. That keeps it out of everyone's way on the happy path, since nothing we have would ever trigger it, and forces a choice about the data representation only if the case actually arrives. It also costs no extra reading, since the state column is already loaded in `__init__`. What do you think?

This last point I am happy to implement myself after this PR, but I would like us to agree first, which is why I laid all of this out.

tests/test_on_data/fiber_photometry/test_npm_fiber_photometry_interfaces.py
h-mayorquin: Closing this as resolved as I think this unblocks this PR to me. Follow up here: https://github.com/catalystneuro/neuroconv/pull/1756#issuecomment-5122095159

h-mayorquin (approved):
LGTM.

doric_fiber_photometry.nwb
pauladkisson: That's so weird. I thought I had removed that NWB file. I even raised an issue for it here #1837

pauladkisson:
> At this point I am unsure what is best myself, and I think we should leave this to the user to decide. Concretely, a Literal argument on the interface, same_band_samples: Literal["include", "exclude"] | None = None, and if a file turns out to contain same-band samples while it is still unset we raise and ask for a policy. That keeps it out of everyone's way on the happy path, since nothing we have would ever trigger it, and forces a choice about the data representation only if the case actually arrives. It also costs no extra reading, since the state column is already loaded in __init__. What do you think?

Leaving it up to the user to determine the policy sounds like the right way to go about this. I am totally on board.

> This last point I am happy to implement myself after this PR, but I would like us to agree first, which is why I laid all of this out.

Sounds like you have a concrete vision for this, so I will leave it to you to implement.

PR #1757, opened 2026-06-22
Add NPMEventsInterface for discrete events from Neurophotometrics

pauladkisson:
This interface is a very thin wrapper around the base CSV events interface. It is worth considering, perhaps, whether or not this interface is worth including in NeuroConv at all, since it provides very little extra functionality. I think the discovery value of a dedicated, format-specific interface is worth the trouble, but could be convinced otherwise.


pauladkisson (commented):
The code looks good to me and is ready for review.

h-mayorquin (changes requested):
Let's do #1755 first since I think that can be used as a base for this. See the review there.

pyproject.toml
```diff
@@ -137,11 +137,17 @@ csv_events = [
     # CSV events now write native pynwb 4.0 EventsTable objects and need only core pandas + pynwb>=4.0;
     # kept as a stable install alias so `neuroconv[csv_events]` and `neuroconv[events]` remain valid.
 ]
+npm_events = [
```
h-mayorquin, line 140: Yes, this makes sense.

docs/api/interfaces.events.rst
```diff
@@ -5,6 +5,14 @@ CSV Events
 ----------
 .. automodule:: neuroconv.datainterfaces.events.csv_events.csveventsdatainterface
 
+Doric Events
```
h-mayorquin, line 8: Thanks.

src/neuroconv/datainterfaces/events/npm_events/npmeventsdatainterface.py
```diff
@@ -0,0 +1,63 @@
...
+        metadata_key : str, optional
+            The key under ``metadata["Events"]`` that namespaces this interface's events metadata.
+            If None (default), the file stem is used (inherited from :class:`.CSVEventsInterface`).
+        verbose : bool, optional
+            Whether to print status messages, default = False.
+        """
+        super().__init__(
+            file_path=file_path,
+            timestamps_column=0,
+            event_type_column=1,
```
h-mayorquin, line 84: Remind me here, how are we sure that this is always an `event_type` and not a payload and therefore a `value_column`.

docs/conversion_examples_gallery/events/npm_events.rst
```diff
@@ -0,0 +1,53 @@
...
+
+:py:class:`~neuroconv.datainterfaces.events.npm_events.npmeventsdatainterface.NPMEventsInterface`
+converts discrete events from Neurophotometrics (NPM) recordings. NPM stores its events in a raw,
+headerless two-column stimuli CSV: the first column holds the event onset time (in the recording's
+raw time base) and the second column holds the event type label (e.g. ``whitenoise``, ``pinknoise``,
+a boolean ``True``/``False`` annotation, or a numeric code). Each distinct label is split out and
+written as its own ``pynwb.event.EventsTable`` (onset timestamps) into ``nwbfile.events``. The raw
+onset times are scaled to seconds by ``time_unit``.
+
+How the event types map onto tables -- one table per type by default, or several types merged into a
```
h-mayorquin: Could we remove the double dashes, I think it has become for some people a signature of people publish this without reading and I would like to avoid giving away that impression.

docs/conversion_examples_gallery/events/npm_events.rst
```diff
@@ -0,0 +1,53 @@
+Neurophotometrics (NPM) Events data conversion
+----------------------------------------------
+
+:py:class:`~neuroconv.datainterfaces.events.npm_events.npmeventsdatainterface.NPMEventsInterface`
+converts discrete events from Neurophotometrics (NPM) recordings. NPM stores its events in a raw,
```
h-mayorquin, line 5: This supports the Bonsai CSV but NPM can also stores events on the photometry/signal CSV.  I think we should be clear about that?

h-mayorquin (changes requested):
Mostly looks ready. One question (inline comments) and one request:

I think we should use the gin datasets here that we prepare for this, why did you go for generated here?

src/neuroconv/datainterfaces/events/npm_events/npmeventsdatainterface.py
pauladkisson: It is the case for all of the example files that I have seen.

Beyond that, I don't know.

docs/conversion_examples_gallery/events/npm_events.rst
pauladkisson: Fixed!

docs/conversion_examples_gallery/events/npm_events.rst
pauladkisson: I was not aware of this. Do you have an example?

pauladkisson:
> Mostly looks ready. One question (inline comments) and one request:
> 
> I think we should use the gin datasets here that we prepare for this, why did you go for generated here?

I figured since it is a CSV format that is relatively easy to generate on the fly, that it would follow the example of the other CSV interfaces.

h-mayorquin:
> > Mostly looks ready. One question (inline comments) and one request:
> > I think we should use the gin datasets here that we prepare for this, why did you go for generated here?
> 
> I figured since it is a CSV format that is relatively easy to generate on the fly, that it would follow the example of the other CSV interfaces.


Let's use the real fixtures here and reserve generated data for scenarios we don't have a fixture for. We already prepared these for exactly this interface, organized for each of the cases (label dtype plus cardinality). 

Check out: https://gin.g-node.org/CatalystNeuro/ophys_testing_data/src/main/events_datasets/NPM/README.md

They should be already on s3.

docs/conversion_examples_gallery/events/npm_events.rst
h-mayorquin: The `multifile_split` on [gin](https://gin.g-node.org/CatalystNeuro/ophys_testing_data/src/main/fiber_photometry_datasets/NPM/README.md#multi-file-split-multifile_split
) shows it directly in the header:

```
FrameCounter,Timestamp,LedState,Stimulation,Output0,Output1,Input0,Input1,Region0G,Region1G
```

`Region0G`/`Region1G` are the fluorescence signal, but `Stimulation`, `Output0`/`Output1` and `Input0`/`Input1` are digital I/O (TTL) lines: one `0`/`1` per frame recording whether that line was high. So a Neurophotometrics photometry CSV can carry discrete events (a stimulation marker, TTL outputs/inputs) right alongside the signal, in their own columns. 

That is the older firmware's layout; newer firmware bit-packs the same lines into the `Flags`/`LedState` column instead, but either way events can live in the photometry/signal CSV, not only in the standalone stimuli file. I confirmed this in the bonsai catalogue when I was stubingg those fixtures.

pauladkisson: Yep, you're right. I will clarify that.

src/neuroconv/datainterfaces/events/npm_events/npmeventsdatainterface.py
h-mayorquin: Ok, yeah, basically what I wanted to confirm was whether you had a manual or a schema claim for this. I went through the vendor manual with an agent, and I could not find any clean documentation that pins the two-column format. The manual does document a separate 5-column `Digital IOs` event log, and it mentions the two-column `(time, label)` case (e.g. the `KeyDown` example), but it never specifies that the second column is the event type rather than a value. So the format genuinely does not fix it.

Given that every file you have seen behaves this way, I have three requests:

1. In the docstring, state that we assume the second column is the event type.
2. Add an internal comment, not in the docstring, noting we keep this default because it matches every example we have but we could not find ironclad proof in the manual, so our future selves do not take it as gospel.
3. If a file comes in with more than two columns, error out cleanly. Something like: "`NPMEventsInterface` expects a headerless two-column CSV (onset time, event type), but this file has N columns. Neurophotometrics also produces richer event files (e.g. the 5-column `Digital IOs` log), which this interface does not support. Please open an issue at https://github.com/catalystneuro/neuroconv/issues if you need it." That keeps us from silently misreading the richer 5-column `Digital IOs` layout.

tests/test_on_data/events/test_npm_events_interface.py
```diff
@@ -0,0 +1,188 @@
...
+            rtol=0,
+            atol=1e-6,
+        )
+
+    def test_numeric_labels(self, number_interface):
+        """Numeric code labels: codes 1 and 3 split into two tables kept verbatim as "1" and "3"."""
+        assert list(number_interface.get_metadata()["Events"]["ttls"]["event_types"]) == ["1", "3"]
+
+        nwbfile = mock_NWBFile()
+        number_interface.add_to_nwbfile(nwbfile=nwbfile, metadata=number_interface.get_metadata())
```
h-mayorquin, line 74: nit: I don't think th get_meatdata() here is necessary.

src/neuroconv/datainterfaces/events/npm_events/npmeventsdatainterface.py
h-mayorquin: Let's document this and add the assertion (no test necessary) and we should be ready to go.

pauladkisson: Very reasonable. Thank you for taking the time to track down the vendor manual. -- implemented

h-mayorquin: Ok. Thanks.

h-mayorquin (approved):
LGTM

PR #1765, opened 2026-06-25
Skip additionalProperties-only nodes in fill_defaults

### LINKED ISSUE #1764: fill_defaults crashes with KeyError: 'properties' on schema nodes that only use additionalProperties
## Context

While working on the GuPPy converter (#965), I peeled the `TDTEventsInterface` off into its own PR (#1751), which has since merged. In the reworked interface, the events metadata is namespaced by a dynamic `metadata_key` and then keyed by the (dynamic) TDT epoc-store names, so `TDTEventsInterface.get_metadata_schema()` describes its `Events` block with `additionalProperties` rather than a fixed set of `properties`:

```python
metadata_schema["properties"]["Events"] = {
    "type": "object",
    "additionalProperties": {        # keyed by metadata_key
        "type": "object",
        "properties": {
            "event_columns": {
                "type": "object",
                "additionalProperties": { ... },   # keyed by epoc store name
            },
        },
    },
}
```

This worked fine for the interface in isolation. But as soon as I wired the GuPPy converter (a `ConverterPipe`) to it and called `converter.get_metadata_schema()`, it blew up with `KeyError: 'properties'`.

## Cause

`NWBConverter.get_metadata_schema()` calls `fill_defaults(metadata_schema, default_values)` ([utils/json_schema.py](https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/utils/json_schema.py)). `fill_defaults` walks the schema tree by `properties`, with one special-case for `patternProperties`:

```python
properties_reference = "properties"
if properties_reference not in schema and "patternProperties" in schema:
    properties_reference = "patternProperties"

for key, val in schema[properties_reference].items():   # KeyError here
```

When it recurses into a node that the defaults dict has an entry for, but whose schema is validated **only** by `additionalProperties` (no `properties`, no `patternProperties`), `schema[properties_reference]` raises `KeyError: 'properties'`. A node like that has no named properties to attach defaults to, so it should simply be skipped — the same way the `patternProperties` case is handled.

This is latent for the interface alone (`interface.get_metadata_schema()` never goes through `fill_defaults`); it only surfaces once an `NWBConverter`/`ConverterPipe` wraps an interface whose schema uses `additionalProperties`.

## Repro

Minimal, no data required:

```python
from neuroconv.utils.json_schema import fill_defaults

schema = {
    "properties": {
        "Events": {
            "type": "object",
            "additionalProperties": {"type": "object"},
        }
    }
}
defaults = {"Events": {"tdt_events": {"event_columns": {}}}}

fill_defaults(schema, defaults)   # KeyError: 'properties'
```

## Fix

Guard `fill_defaults` so a node with neither `properties` nor `patternProperties` is skipped rather than indexed by `"properties"`. I'll open a small standalone PR for this.

pauladkisson:
Fixes #1764.

<Details>
## Problem

While wiring the GuPPy converter (#965) to the `TDTEventsInterface` that was peeled into #1751, I hit `KeyError: 'properties'` the moment a `ConverterPipe` wrapped that interface and called `get_metadata_schema()`.

`fill_defaults` (invoked by `NWBConverter.get_metadata_schema()`) walks the schema by `properties`, with a special case for `patternProperties`:

```python
properties_reference = "properties"
if properties_reference not in schema and "patternProperties" in schema:
    properties_reference = "patternProperties"

for key, val in schema[properties_reference].items():   # KeyError here
```

`TDTEventsInterface` namespaces its events metadata under a dynamic `metadata_key`, so its `Events` schema node is validated by `additionalProperties` rather than a fixed `properties` set. When `fill_defaults` recurses into that node (because the defaults dict has an `Events` entry) it indexes `schema["properties"]` and crashes.

This is latent for an interface in isolation — `interface.get_metadata_schema()` never goes through `fill_defaults` — and only surfaces once an `NWBConverter`/`ConverterPipe` wraps such an interface.

## Fix

A node with neither `properties` nor `patternProperties` has no named properties to attach defaults to, so `fill_defaults` now skips it (consistent with the existing `patternProperties` handling).

## Repro

```python
from neuroconv.utils.json_schema import fill_defaults

schema = {"properties": {"Events": {"type": "object", "additionalProperties": {"type": "object"}}}}
defaults = {"Events": {"tdt_events": {"event_columns": {}}}}
fill_defaults(schema, defaults)   # was: KeyError: 'properties'
```

## Tests

Added `test_fill_defaults_skips_additional_properties_node` in `tests/test_minimal/test_utils/test_json_schema_utils.py`; the full `test_json_schema_utils.py` suite passes.

<\Details>


pauladkisson:
Looks like we missed some nuances with the schema @h-mayorquin

h-mayorquin (approved):
Make sense, skip levels with not defined properties.

PR #1770, opened 2026-06-30
Add a curated ethogram to `VameInterface` and align VAME metadata with dict metadata

h-mayorquin:
`VameInterface` now writes, for each motif run, a curated [`ndx-ethogram`](https://github.com/catalystneuro/ndx-ethogram) product alongside the faithful `ndx-vame` series: each per-frame `MotifSeries` is run-length-encoded into an `EthogramBouts` timeline with an `Ethogram` catalogue (one row per motif id, the modal community recorded as each motif's `category`), and the bouts link back to the producing `MotifSeries` and the upstream `PoseEstimation`. The faithful series stay; the ethogram is the downstream-consumable curated layer, so both are kept side by side. The run-length-encoding and the ethogram builder live in tool-agnostic `tools/` helpers so future segmenters (keypoint-MoSeq, B-SOiD) can reuse them, and the curated layer lives in a shared `metadata["Behavior"]["Ethograms"]` registry rather than nested under VAME.

This PR also restructures the VAME metadata into flat, addressable registries under `metadata["Behavior"]["Vame"]` (a `VameProjects` container registry plus per-type `MotifSeries` / `CommunitySeries` / `LatentSpaceSeries` registries, each series linking up to its project via `vame_project_metadata_key`), matching the unified pattern ophys ([PR ophys](https://github.com/catalystneuro/neuroconv/pull/1677)) and ecephys ([PR ecephys](https://github.com/catalystneuro/neuroconv/pull/1743)) already use instead of nesting the series inside the container. I folded this into the same PR because VAME has not shipped yet (the interface is unreleased in v0.9.4), so reshaping the metadata now is a free, non-breaking change, whereas doing it after release would require a deprecation cycle.


pyproject.toml
```diff
@@ -118,6 +118,7 @@ medpc = [
 ]
 vame = [
     "ndx-vame==0.3.1",
```
weiglszonja: Since we merged https://github.com/catalystneuro/ndx-vame/pull/6 and released [0.3.2](https://pypi.org/project/ndx-vame/)
```suggestion
    "ndx-vame==0.3.2",
```

docs/conversion_examples_gallery/behavior/vame.rst
weiglszonja, line 1: Maybe it would be worth mentioning the `EthogramBouts` in the docs?

h-mayorquin: I think that's a good idea.

h-mayorquin: Done.

pyproject.toml
h-mayorquin: Thanks, I put this as a floor.

docs/conversion_examples_gallery/behavior/vame.rst
weiglszonja: I don't see it?

h-mayorquin: My bad, I forgot to push.

h-mayorquin: It should be up now.

weiglszonja (approved):
looks good, thanks!

PR #1769, opened 2026-06-30
Update pynwb minimum to 4.0

h-mayorquin:
Bumps the minimum `pynwb` to `>=4.0.0` now that 4.0 has been released. The only version pin lives in `pyproject.toml`; the dev-testing workflow installs pynwb from its `dev` branch and needs no change.

This required a release on ndx-sound that I just did.

PR #1772, opened 2026-07-02
HERD integration: species ontology terms

### LINKED ISSUE #1771: [Feature]: HERD integration

HERD now allows you to map term in NWB files to ontological entries see https://nwb-overview.readthedocs.io/en/latest/external_resources_entity_guide.html.

I would like to add a feature within NeuroConv, so we can recommend terms, e.g. for species and species-aware anatomical labels (https://github.com/dandi/dandi-archive/pull/2768)

This could perhaps use termsets: https://hdmf.readthedocs.io/en/stable/tutorials/plot_term_set.html#sphx-glr-tutorials-plot-term-set-py

_No response_

Yes.

### Code of Conduct

bendichter:
Closes #1771.

## What

Adds `neuroconv.tools.ontology` and wires it into `make_nwbfile_from_metadata` so that a subject's species is both **recommended** in a standardized form and **annotated** with a machine-readable NCBITaxon reference in the written file.

Two behaviors:

1. **Suggestion (non-blocking).** `validate_species` warns with a concrete recommendation when `Subject.species` is a recognized common name (e.g. `"mouse"`) or a likely typo (e.g. `"Homo sapien"`), pointing at the canonical Latin binomial and its NCBITaxon id (`Mus musculus` / `NCBITaxon:10090`). Valid-but-uncommon binomials pass silently.
2. **HERD annotation.** `add_species_external_resource` attaches an external-resource reference mapping a recognized `Subject.species` binomial to its NCBITaxon entity, stored **in-file** under `/general/external_resources` via HDMF's HERD. This is the HERD integration the issue asked for — downstream tools (e.g. DANDI) can resolve the term without guessing.

The species table is a small curated, offline, dependency-free lookup of common neuroscience species (no network, no ontology client).

## Notes

- In-file HERD storage requires **pynwb ≥ 4.0.0**, already the minimum pin (#1769).
- The HERD is constructed with pynwb's type map, which is required for the `Subject.species` attribute spec to resolve; without it `add_ref` raises. The write is idempotent and extends a pre-existing `external_resources` HERD in place rather than replacing it.

## Testing

- `tests/test_minimal/test_tools/test_ontology_species.py` — table integrity, common-name/typo suggestions, `get_species_term` resolution, warning behavior.
- `tests/test_minimal/test_tools/test_ontology_external_resources.py` — no-op cases, annotation, idempotency, extend-in-place, and a full write→read round trip confirming the reference persists in the NWB file.


src/neuroconv/tools/ontology/_species.py
```diff
@@ -0,0 +1,228 @@
...
+
+    @property
+    def entity_uri(self) -> str:
+        """Resolvable URI for the NCBITaxon entity (usable as a HERD ``entity_uri``)."""
+        taxon_number = self.ncbitaxon_id.split(":", 1)[1]
+        return f"http://purl.obolibrary.org/obo/NCBITaxon_{taxon_number}"
+
+
+# Canonical Latin binomial -> NCBITaxon identifier (CURIE) for common neuroscience species.
+SPECIES_TERMS: dict[str, SpeciesTerm] = {
```
oruebel, line 42: First of all, it's great to see this effort to add species ontologies to NeuroConv 👍 

In HDMF and PyNWB we can configure the terms via configurations. Here is an example of how this could look there https://github.com/oruebel/nwb_termsets/tree/main

HDMF uses LinkML YAML files to define this kind of sets. E.g., see here for a corresponding NCBI term set. https://github.com/oruebel/nwb_termsets/blob/main/src/nwb_termsets/term_sets/subject_species_ncbitaxon_termset.yaml  . With this defined in YAML, you could then also use it directly with Species by configuring the `TermSetWrapper` to validate terms automatically when populating the file, and automatically collect all references for the NWB in HERD via: 

```
herd = nwbfile.get_external_resources()
herd.add_ref_container(nwbfile)
```

Regarding `_COMMON_NAME_TO_CANONICAL: dict[str, str] = {`, LinkML also allows definition of `aliases:` for each term, which can be used to defined those kind of mappings. I don't think HDMF supports aliases yet, but that would be useful to add and shouldn't be too complicated since that really only affects the `TermSet` (and maybe `TermSetWrapper`) class. 

In general, I think it could be useful here to use the HDMF `TermSet` class here and if necessary expand it, instead of creating custom functionality for this here. E.g. `TermSet` and `TermSetWrapper` already validate terms. 

Management of aliases and lookup of matches as defined below look like mostly general functionality that would be useful to add to `TermSet` since they seem to mainly perform look-up of terms and aliases and close matches. 

@bendichter @rly  thoughts

bendichter: would this approach work for annotating an existing NWB file?

alessandratrapani: Yes. Opening `NWBHDF5IO(path, "r+")`, running inference/annotation against the loaded nwbfile, and `io.write(nwbfile)` works the same as on a freshly created file. Added a working example to the docs on #1776

alessandratrapani:
Closing this in favor of #1776, which is built on top of this branch and carries it forward (species + brain-region ontology together, using LinkML TermSet YAML files as you suggested). Didn't want to merge #1772 separately since #1776 already supersedes it.
@oruebel your TermSet/LinkML point did shape the design, #1776 backs its vocabularies with LinkML TermSet YAML files rather than hardcoded dicts. If you want to continue that discussion, #1776 or #1771 (the roadmap issue) are the live threads.

oruebel:
> @oruebel your TermSet/LinkML point did shape the design, #1776 backs its vocabularies with LinkML TermSet YAML files rather than hardcoded dicts. If you want to continue that discussion, #1776 or #1771 (the roadmap issue) are the live threads.

Thank you @alessandratrapani  I'll take a look at those.

PR #1759, opened 2026-06-23
Add documentation for metadata dict in the event interfaces

h-mayorquin:
This is like https://github.com/catalystneuro/neuroconv/pull/1729 and https://github.com/catalystneuro/neuroconv/pull/1653 but for events

The how-to:
https://neuroconv--1759.org.readthedocs.build/en/1759/how_to/annotate_events_metadata.html

The developer documentation:
https://neuroconv--1759.org.readthedocs.build/en/1759/developer_guide/events_metadata_structure.html

PR #1774, opened 2026-07-03
Add `BaseEventsInterface` and `MockEventsInterface` for discrete events

h-mayorquin:
pynwb 4.0 brings `EventsTable` into the core schema (NWBEP001), so I want every discrete-event interface to share one write path instead of each reimplementing it against the old `ndx-events` extension. This adds `BaseEventsInterface`, which owns the `Events` metadata schema and provides a writer that maps arbitrary readers into an internal class that `add_to_nwbfile` uses to route the data to the right place.

For testing, this PR adds a `MockEventsInterface`, a data-free mock that spans the different axes of the acquisition formats for events:

- Event extent: point events versus events with a duration.
- Payload cardinality: timestamps only, a single value per event, or a multi-field struct.
- Number of event types: mimicking acquistion syttems with multiple event types.

This is the base work. I will follow it with a PR migrating the Tucker-Davis Technologies (TDT) interface on top of it to illustrate the pattern in practice.

The `Events` metadata dict shape is documented in [PR #1759](https://github.com/catalystneuro/neuroconv/pull/1759), which can probably be merged after this.


src/neuroconv/datainterfaces/events/baseeventsinterface.py
```diff
@@ -0,0 +1,298 @@
...
+    an ``event_type`` discriminator column naming each row's ``event_type_source_id``. This structure
+    is documented in the events taxonomy and PR #1759
+    (https://github.com/catalystneuro/neuroconv/pull/1759).
+
+    Subclasses must set ``self.metadata_key`` and implement :meth:`_load_event_data_dict`.
+    """
+
+    keywords = ("events",)
+
+    def _load_event_data_dict(self) -> dict[str, _EventInternalClass]:
```
pauladkisson: This is an abstract method, right? So, `@abstractmethod`?

src/neuroconv/datainterfaces/events/baseeventsinterface.py
```diff
@@ -0,0 +1,298 @@
...
+            # Pooled types concatenate in blocks; sort by timestamp for a true chronological timeline
+            # (a no-op for a single type, whose onsets are already ordered).
+            rows.sort(key=lambda row: row[0])
+
+            table_metadata = event_tables_metadata[table_metadata_key]
+            table = EventsTable(name=table_metadata["table_name"], description=table_metadata["description"])
+            # Only the native EventsTable columns go through add_event: timestamp, plus duration when any
+            # pooled type has one. Every derived column (the event_type discriminator and the value
+            # columns) is attached full-length afterwards.
+            for timestamp, _event_type_source_id, duration, _cells in rows:
```
pauladkisson: Shouldn't `_event_type_source_id` and `_cells` just be plain `_` if they are unused?

src/neuroconv/datainterfaces/events/baseeventsinterface.py
```diff
@@ -0,0 +1,298 @@
...
+        collisions = {name: keys for name, keys in table_metadata_keys_by_name.items() if len(keys) > 1}
+        assert not collisions, (
+            f"Duplicate EventTables 'table_name' values found in metadata: {collisions}. "
+            "Each EventsTable must have a unique name; give these table_metadata_key entries distinct table_name values."
+        )
+
+        for table_metadata_key, event_type_source_ids in event_type_source_ids_by_table.items():
+            # More than one event type pooled into one table is a merge, told apart by an event_type
+            # column. A single event type (however many columns it fans into) needs no discriminator.
+            is_merge = len(event_type_source_ids) > 1
```
pauladkisson: This covers table merges within the same interface, but not table merges across two different interface instances with different metadata keys, which should in principle be possible, or, if we want to disallow it, should error out.

src/neuroconv/datainterfaces/events/baseeventsinterface.py
```diff
@@ -0,0 +1,298 @@
...
+                    event_kwargs["duration"] = duration
+                table.add_event(**event_kwargs)
+
+            # A merge tells its pooled types apart with an event_type discriminator: one value per row,
+            # filled on every row, so it is just another full-length column like the value columns below.
+            if is_merge:
+                table.add_column(
+                    name="event_type",
+                    description="The event type of each event.",
+                    data=[event_type_source_id for _, event_type_source_id, _, _ in rows],
```
pauladkisson: Shouldn't this event type column contain the labels of each event rather than the source ID? Source IDs are generally uninformative in the NWB file (ex. PrtN) whereas the labels are more informative (ex. port_entries).

src/neuroconv/datainterfaces/events/baseeventsinterface.py
```diff
@@ -0,0 +1,298 @@
...
+                )
+
+            # Attach each value column full-length, filling rows from other types, plus a MeaningsTable
+            # for a categorical column that supplies meanings.
+            for column_name, column_spec in column_specs.items():
+                categories = column_spec.get("column_categories")
+                fill = "" if categories is not None else np.nan
```
pauladkisson: I'd like to throw something out for discussion. Maybe it's not worth the trouble to actually implement, but I thought I would flag it here. 

Creating an EventsTable with mutually incompatible columns that require these fill values seems like a best practice suggestion problem to me. In these cases, it seems like we ought to suggest that mutually incompatible columns each live in their own EventsTable object. Otherwise, we will get tables that look like this. (filled with NaNs and empty strings)

| timestamp | event_type | port | volume_ul | freq |
|-----------|------------|------|-----------|------|
| 1.0       | lick       | left | nan       | ""   |
| 2.0       | reward     | ""   | 5.0       | ""   |
| 3.0       | tone       | ""   | nan       | high | 

Curious what you think.

tests/test_minimal/test_base_events_interface.py
```diff
@@ -0,0 +1,274 @@
...
+        # A columns entry whose field_source_id is absent from the data payload is a mismatch and must
+        # fail with a clear message.
+        interface = MockEventsInterface(event_payload="timestamps only")
+        metadata = interface.get_metadata()
+        metadata["Events"]["mock_events"]["event_types"]["events"]["columns"]["ghost"] = {"column_name": "ghost"}
+        expected_error = "Event type 'events' declares a column for payload field 'ghost', but its payload has no such field (has [])."
+        with pytest.raises(AssertionError, match=re.escape(expected_error)):
+            interface.add_to_nwbfile(nwbfile=mock_NWBFile(), metadata=metadata)
+
+    def test_merge_event_types_in_single_table(self):
```
pauladkisson, line 155: Per my comment above, I think it would be helpful to have a testing merge from two separate interface instances onto the same table.

pauladkisson (commented):
Looks good, mostly minor comments.

src/neuroconv/datainterfaces/events/baseeventsinterface.py
h-mayorquin: Yes, I think this make sense.

src/neuroconv/datainterfaces/events/baseeventsinterface.py
h-mayorquin: Yes, you are right. This should be abstract. I changed the name to `_get_events_data_dict()` I added caching in case it needs to be called more than once and made it abstract.

src/neuroconv/datainterfaces/events/baseeventsinterface.py
h-mayorquin: Yes, that's a feature that is not implemented at the momtnt. The ture is that I did consider it, but I couldn't come up with a natural, meaningful example to design the merge against, so I deferred it to keep this PR scoped. T

he closest case I could think of is auxiliary event hardware, a NIDQ board or an Arduino, running alongside a main acquisition system like Neuralynx or TDT, where you could argue the auxiliary events and the system's own events belong in one table. 

That said, the metadata already anticipates this (the global `EventTables` block is meant to be shared across interfaces), so we could use the mock interface to impelment this if you think is worth it but I did not want to make the review survace larger.

pauladkisson: Yeah, I think it is worth the trouble, and I don't mind the slightly larger review surface in this case.

h-mayorquin: Ok, great. Happy to add, let me ping you back when I do.

src/neuroconv/datainterfaces/events/baseeventsinterface.py
h-mayorquin: Done.

src/neuroconv/datainterfaces/events/baseeventsinterface.py
h-mayorquin: I agree. The problem was that the event-type label had no proper home in the metadata before, a type's identity only surfaced through the table name. I just changed this. Now, every event type gets `event_name` and an `event_description` with a dual role by layout:

- Default (one `event_type_source_id` per table, single- or multi-valued payload): `event_name` becomes the table's object name (CamelCased) and `event_description` its table description.
- Merge (several types pooled into one table): `event_name`/`event_description` become the label and its meaning in the `event_type` column's `MeaningsTable`, and the table's own name/description come from the `EventTables` entry, which has the last word.

What do you think?

src/neuroconv/datainterfaces/events/baseeventsinterface.py
h-mayorquin: Are you saying that we should not allow this to happen? 
Currently, the default is that every event type would go to its own table so this does not happen by default. This is the original intent of NWBEP0001 (each event should go to its own table) However, they [design doc](https://docs.google.com/document/d/1qcsjyFVX9oI_746RdMoDdmQPu940s0YtDjb1en1Xtdw/edit) does allow for non-compatible columns (see example 5) so I leave this as an opt-in option by using the table_metadata_key mechanism and create a saprse table like the one that you are showing.

Is that what you meant, or am I reading you wrong?

src/neuroconv/datainterfaces/events/baseeventsinterface.py
h-mayorquin: Done.

tests/test_minimal/test_base_events_interface.py
h-mayorquin: This should be in.

src/neuroconv/datainterfaces/events/baseeventsinterface.py
pauladkisson: That makes sense to me.

src/neuroconv/datainterfaces/events/baseeventsinterface.py
pauladkisson: Thanks for the reference. It's very helpful. I think we _should_ allow this using the interface, especially since the design doc has an example specifically demonstrating it. 

I was primarily considering whether it would be worth adding an inspector check for this.

pauladkisson (approved):
Looks great.

src/neuroconv/datainterfaces/events/baseeventsinterface.py
h-mayorquin: I agree with the inspector the idea.

PR #1775, opened 2026-07-03
Migrate TDTEventsInterface to the native pynwb 4.0 EventsTable model

h-mayorquin:
Now that the NWBEP001 events types ship in `pynwb` 4.0 core, `TDTEventsInterface` (for Tucker-Davis Technologies tanks) no longer needs the `ndx-events` extension, so this migrates it to write native `pynwb.event.EventsTable` objects into `nwbfile.events` and drops `ndx-events` from the `tdt_events` extra. It builds on the shared `BaseEventsInterface` added in [PR #1774](https://github.com/catalystneuro/neuroconv/pull/1774): `TDTEventsInterface` is now just the thin reader that maps each epoc into the shared internal representation and inherits the metadata schema and writer from the base. The `Events` metadata dict shape is documented in [PR #1759](https://github.com/catalystneuro/neuroconv/pull/1759).


pauladkisson (approved):
Looks great.

PR #1776, opened 2026-07-04
HERD integration: brain-region ontology terms (Allen Mouse Brain Atlas)

bendichter:
Builds on the species HERD work (#1772). **Stacked on `add-species-herd-ontology`** so the diff shows only the brain-region changes; retarget/rebase onto `main` once #1772 merges.

## What

Extends `neuroconv.tools.ontology` so that, for a **mouse** subject, anatomical `location` strings are annotated with machine-readable [Allen Mouse Brain Atlas](https://bioregistry.io/registry/mba) (MBA) references stored in-file via HDMF's HERD. Downstream tools (e.g. DANDI) can then resolve the exact brain structure instead of guessing from an acronym.

Two-step resolution per distinct `location` string:

1. **Metadata mapping (takes precedence).** `metadata["BrainRegions"]` maps a location string to an MBA identifier — a CURIE (`"MBA:382"`), bare id (`"382"`), full URI, or a dict with `mba_id` and optional `acronym`/`name`. This is how a user annotates a region the offline table doesn't recognize, or overrides one it does.
2. **Offline lookup.** Otherwise a curated, dependency-free table of ~98 common mouse structures resolves an exact Allen acronym (`"CA1"`, `"VISp"`), a canonical structure name (`"caudoputamen"`), or a common informal name/abbreviation (`"hippocampus"`, `"V1"`).

Locations resolving to neither (including the `"unknown"` placeholder) are left untouched. The whole pass is a no-op for non-mouse subjects (the MBA vocabulary is mouse-specific), gated via the existing species resolver.

## Where it's annotated

- electrodes table `location` column (ecephys)
- `ElectrodeGroup.location`
- `ImagingPlane.location` (ophys)

## Notes

- Runs at write time (end of `create_nwbfile` for both `BaseDataInterface` and `NWBConverter`, and the in-memory `run_conversion` path) — **after** interfaces populate the electrodes table and imaging planes, unlike species annotation which can run at file creation. Idempotent and extends a pre-existing `external_resources` HERD in place, so it composes with the species reference.
- MBA ids/URIs (`https://purl.brain-bican.org/ontology/mbao/MBA_<id>`) were verified against the Allen CCFv3 structure graph, not hardcoded from memory.
- Requires **pynwb ≥ 4.0.0** for in-file HERD (already the minimum pin).

## Testing

- `tests/test_minimal/test_tools/test_ontology_brain_regions.py` — table integrity, acronym/name/alias resolution, case-sensitivity, identifier parsing.
- `tests/test_minimal/test_tools/test_ontology_brain_region_external_resources.py` — no-op/gating cases, electrodes/group/imaging-plane annotation, metadata mapping + precedence, idempotency, extend-in-place, and a write→read round trip.

## Docs

- User guide page `docs/user_guide/brain_region_ontology.rst`; the `neuroconv.tools.ontology` API page picks up the new functions automatically.


alessandratrapani:
## New changes:

- Annotate FiberPhotometryTable location column with brain-region HERD
- Add term_sets/uberon_common_regions.yaml — 17 common regions (cortex subdivisions, hippocampus, striatum, substantia nigra, PAG, locus coeruleus, cerebellum, trapezius muscle), every UBERON ID verified live against the EBI Ontology Lookup Service.
- `_atlas_for_species` now falls back to this generic UBERON atlas for any recognized species without a dedicated Allen atlas (mouse and human keep their MBA/HBA atlases unchanged). 
- New export UBERON_TERMS, alongside MBA_TERMS/HBA_TERMS.
- `_subject_atlas_species` now gates only on "is the species recognized at all" rather than "does it have a dedicated atlas," since every recognized species resolves through some atlas now.

docs/user_guide/ontology.rst
```diff
@@ -0,0 +1,290 @@
...
+        "brain_regions": {
+            "CA1": {"id": "MBA:382", "uri": "https://purl.brain-bican.org/ontology/mbao/MBA_382"},
+        },
+    }
+
+Brain-region terms are keyed by the free-text ``location`` string and live under the modality block
+whose objects carry that location: ``metadata["Ecephys"]["ontology"]["brain_regions"]`` (electrodes
+table and electrode groups), ``metadata["Ophys"]["ontology"]["brain_regions"]`` (imaging planes),
+and ``metadata["FiberPhotometry"]["ontology"]["brain_regions"]`` (the ``FiberPhotometryTable``). To
+annotate one value with **several** ontologies, map it to a list of terms:
```
bendichter, line 73: Is there an advantage to splitting brain regions over different modalities? I don't think we'd see a situation where a single file holds a single brain region name across multiple modalities and means something *different* for each modality.

alessandratrapani: This came from @h-mayorquin's point on [#1771](https://github.com/catalystneuro/neuroconv/issues/1771#:~:text=The%20terms%20should%20sit%20next%20to%20the%20thing%20they%20annotate.%20Take%20for%20example%20the%20subject.%20Right%20now%20the%20terms%20live%20in%20their%20own%20top%2Dlevel%20block%2C%20away%20from%20the%20values%20they%20describe%3A): terms next to what they annotate, reusing existing blocks instead of new top-level keys. For Subject.species/strain that fits, one subject, one term.
For brain regions it got the same treatment mechanically, ontology.brain_regions inside each modality block, but I can see the problem with this approach. 
So should I roll back to to a flattened metadata["ontology"]["brain_regions"] map?

alessandratrapani: @bendichter I went ahead and made this change: 10bf393d8 flattens `ontology.brain_regions` to one file-wide `metadata["ontology"]["brain_regions"]` map and removes the per-modality split and its conflict-warning code. it's what you had in mind?

h-mayorquin: Yes, I did not think on the case of multiple objects annotated with one thing.

For that case, I think the `metadata_key` concept is better. You keep the annotation local and explicit but link to a global group.Let me see what you implemented on that commit to see if I have a more concrete suggestion.

h-mayorquin: Ok, I took a look.

My view is that no `metadata_key` is necessary here. The link only goes from the HERD object to the objects that mark the anatomical region, and those are completely specified by the name, so there is nothing to disambiguate. If the objects were linked to HERD, then a `metadata_key` would make sense.

Btw, I think you can also move the anatomy to a global space, as the same argument applies. I think this is best.

If you decide to keep it localy, note that the anatomy is not sitting in the right place for pose estimation currently: it is sitting in `metadata["PoseEstimation"]` but it should sit in `metadata["Pose"]`.

h-mayorquin: Yes, thinking more about it, I think my initial suggestion is wrong.

HERD does not point to rows in dynamic tables and only links implicitly by text so specifying the metadata very locally and explicitly at the local level is not something that the data model does anyway. This I did not know. 

As you said, this still makes sense for single objects like `Subject` but if you look at [ndx-multisubjects](https://github.com/nehatk17/ndx-multisubjects) they decided to go for a dynamic table approach there so it will not make sense in those cases either. So I think subject is better handled at the global ontology level as well.

At this point, we already have 4 entries on ontology and might as well go with that as a convention that you have for brain regions already:

```python
metadata["ontology"] = {
    "species": {"Mus musculus": {"id": "NCBITaxon:10090", "uri": "http://purl.obolibrary.org/obo/NCBITaxon_10090"}},
    "strain": {"C57BL/6J": {"id": "RRID:IMSR_JAX:000664", "uri": "https://scicrunch.org/resolver/RRID:IMSR_JAX:000664"}},
    "brain_regions": {"CA1": {"id": "MBA:382", "uri": "https://purl.brain-bican.org/ontology/mbao/MBA_382"}},
    "anatomy": {"Snout": {"id": "UBERON:0002536", "uri": "http://purl.obolibrary.org/obo/UBERON_0002536"}},
}
```

My bad.

alessandratrapani: Done, all four now live under `metadata["ontology"]`:

0259746d5 species (#1776)
93335a29c strain (#1884)
9292db0ed anatomy (#1900), moved out of `metadata["PoseEstimation"]`, so the Pose/PoseEstimation issue goes away too

PR #1778, opened 2026-07-08
Added BaseFiberPhotometryInterface

pauladkisson:
## Summary

Adds a base `FiberPhotometryInterface`, migrates `TDTFiberPhotometryInterface` onto it, and standardizes the fiber photometry interfaces around a single coherent design.

## Design decisions

**Single-series interfaces.** This PR settles on "one interface writes one response series," taking a cue from the ecephys interfaces. 

I was a little bit dissatisfied with the breadth of the original TDTFiberPhotometryInterface, and I felt like in the standardization process it would be helpful to focus it a little bit. This change maintains maximum flexibility in how the user wants to structure the individual fiber photometry response series objects with the stream names and the stream indices, allowing full flexibility in terms of what data actually gets written to a fiber photometry response series, but clarifies the scope a little bit of the interface to keep it from being quite so sprawling.

This change also allows the interface to be constructor-driven instead of metadata-driven, which matches the behavior of other interfaces in NeuroConv and helps avoid having to cart around a big metadata dictionary every time you want to use a fiber photometry interface. 

**Fiber photometry is its own category.** It moved out from under `ophys` into a dedicated `fiber_photometry` section, mirroring the way ecephys and icephys are split even though both are technically "ephys."

**Metadata follows the Ophys `metadata_key` system.** Rather than keep fiber photometry's older name-based metadata, this PR adopts the dict-keyed `metadata_key` structure the rest of NeuroConv is migrating toward, and lifts it to a top-level `metadata["FiberPhotometry"]` block (like `Icephys`).

**Runnable defaults, loud about placeholders.** `get_metadata()` returns a scaffold that runs on zero user metadata, with unfilled required fields marked by sentinels that warn at write time.

This decision is a second crack at resolving the tension between easy writing and high-quality metadata. Before, I simply required the `FiberPhotometryInterface` to have all of the metadata that it needs specified, which is just kind of a pain and deviates significantly from the way all of the other interfaces work in NeuroConv. I thought it was worth a shot to be more strict, but after using it several times, I think it's more trouble than it's worth. With that being said, I still thoroughly dislike the prospect of users naively writing crappy placeholders into NWB files and not noticing that the defaults that they're inheriting from NeuroConv are insufficient for a high-quality NWB file. So in this case, I am using dedicated placeholders that are caught in the add to NWB file method and warned so that a naive user is alerted to the fact that they need to edit the metadata. You can also even run the interface with a strict mode so that it errors out in the case of such placeholders. But because we have the defaults available, it should make running the interface in cases like the conversion gallery and tests a lot easier. 

## Scope

Doric is intentionally left for a follow-up PR, and the deprecated multi-series path keeps the old metadata format until it's removed.


alessandratrapani (approved):
Looks great! Very much needed! I refactor the DoricInterface to inherit from this base class and it works (see https://github.com/catalystneuro/neuroconv/pull/1779))

h-mayorquin:
I have some comments regarding the metadata. I will cover them soon.

h-mayorquin:
The one thing I want us to align on is the device metadata. I'd like to finally add general, cross-modality support for device models (this has been a long time coming), and I think now is a good place to discuss it, as the fiber photometry modality makes heavy use of it. The proposal is in #1780.

The main difference between #1780 and this is that both the device and the device model live at the top level instead of under the modality. There are two reasons for this:
* They get a unified way of being addressed through the `metadata_key`. Otherwise, when looking for a specific device, we also have to tell the function where to look. Keeping them top level lets us use a single function for adding devices of all sorts.
* A device or device model that belongs to two modalities would otherwise have two rightful places, compounding the problem above. For example, an `ExcitationSource` shared by a fiber photometry rig and a two-photon microscope on the same setup could sit under either `metadata["FiberPhotometry"]` or `metadata["Microscopy"]`, and neither is more correct. These are ndx-ophys-devices classes that ndx-microscopy already reuses.

Could you take a look at #1780 so we're aligned on the shape before fiber photometry commits to device containers under `metadata["FiberPhotometry"]`?

How to move forward? If you need this merged to unblock Doric, merge it as-is and we refactor onto the shared registry before a release. Otherwise we can align on the shape first and then merge. Whatever's easier for both of you.

pauladkisson:
> The one thing I want us to align on is the device metadata. I'd like to finally add general, cross-modality support for device models (this has been a long time coming), and I think now is a good place to discuss it, as the fiber photometry modality makes heavy use of it. The proposal is in #1780.
> 
> The main difference between #1780 and this is that both the device and the device model live at the top level instead of under the modality. There are two reasons for this:
> 
> * They get a unified way of being addressed through the `metadata_key`. Otherwise, when looking for a specific device, we also have to tell the function where to look. Keeping them top level lets us use a single function for adding devices of all sorts.
> * A device or device model that belongs to two modalities would otherwise have two rightful places, compounding the problem above. For example, an `ExcitationSource` shared by a fiber photometry rig and a two-photon microscope on the same setup could sit under either `metadata["FiberPhotometry"]` or `metadata["Microscopy"]`, and neither is more correct. These are ndx-ophys-devices classes that ndx-microscopy already reuses.
> 
> Could you take a look at #1780 so we're aligned on the shape before fiber photometry commits to device containers under `metadata["FiberPhotometry"]`?
> 
> How to move forward? If you need this merged to unblock Doric, merge it as-is and we refactor onto the shared registry before a release. Otherwise we can align on the shape first and then merge. Whatever's easier for both of you.

Sounds good to me. This PR is absolutely the place to get consensus on all these sorts of issues. I will take a look.

pauladkisson:
> > The one thing I want us to align on is the device metadata. I'd like to finally add general, cross-modality support for device models (this has been a long time coming), and I think now is a good place to discuss it, as the fiber photometry modality makes heavy use of it. The proposal is in #1780.
> > The main difference between #1780 and this is that both the device and the device model live at the top level instead of under the modality. There are two reasons for this:
> > 
> > * They get a unified way of being addressed through the `metadata_key`. Otherwise, when looking for a specific device, we also have to tell the function where to look. Keeping them top level lets us use a single function for adding devices of all sorts.
> > * A device or device model that belongs to two modalities would otherwise have two rightful places, compounding the problem above. For example, an `ExcitationSource` shared by a fiber photometry rig and a two-photon microscope on the same setup could sit under either `metadata["FiberPhotometry"]` or `metadata["Microscopy"]`, and neither is more correct. These are ndx-ophys-devices classes that ndx-microscopy already reuses.
> > 
> > Could you take a look at #1780 so we're aligned on the shape before fiber photometry commits to device containers under `metadata["FiberPhotometry"]`?
> > How to move forward? If you need this merged to unblock Doric, merge it as-is and we refactor onto the shared registry before a release. Otherwise we can align on the shape first and then merge. Whatever's easier for both of you.
> 
> Sounds good to me. This PR is absolutely the place to get consensus on all these sorts of issues. I will take a look.

OK, I went ahead and incorporated the new top-level devices and device models in a data structure, so this PR should be good with respect to that comment. But see #1782

PR #1780, opened 2026-07-09
Add top-level DeviceModel support to the device metadata registry

h-mayorquin:
This adds `metadata["DeviceModels"]` alongside the existing top-level `metadata["Devices"]`, so device models (core `pynwb.DeviceModel` and extension subclasses such as `OpticalFiberModel`) can be written and referenced from device instances via a `device_model_metadata_key`. It is the shared foundation the fiber photometry migration ([#1778](https://github.com/catalystneuro/neuroconv/pull/1778)) builds on: device models are a cross-modality primitive, not a fiber-photometry detail (the ndx-ophys-devices classes are already reused by ndx-microscopy's rig, and probes will follow), so rather than let fiber photometry grow its own parallel device-writing path this lands the registry once and every modality writes devices the same way.

Both models and instances live top-level, always, with no per-modality nesting. That choice is deliberate: any "this device belongs to one modality" rule is an unstable prediction (already falsified, since `ExcitationSource`/`Photodetector` were fiber-photometry's devices until ndx-microscopy reused them), and a flat top-level shape mirrors the NWB file's own flat `nwbfile.devices` / `nwbfile.device_models`. An entry's optional `type` resolves to its concrete class by lazily importing the providing extension through `get_package`, so an optional extension is imported only when an entry of that type is actually written and a pure-ecephys conversion never touches ndx-ophys-devices. The add helpers follow NeuroConv's canonical `(nwbfile, metadata, metadata_key)` shape, with a device instance pulling in its referenced model on demand (idempotently), so the key does its own resolution against the full metadata. To keep this PR's diff to the registry itself, `_add_device_to_nwbfile` also retains a transitional `device_metadata` form: its existing ecephys, ophys, and behavior callers are left untouched here and migrate to the canonical form one call site at a time, after which the transitional parameter is removed.


src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
```diff
@@ -189,38 +195,123 @@ def add_device_from_metadata(nwbfile: NWBFile, modality: str = "Ecephys", metada
...
-    nwbfile : NWBFile
-        The NWB file to add the device to.
-    device_metadata : dict
-        Dictionary describing the device. Must contain at least a ``"name"`` key.
+    * **Canonical** ``(nwbfile, metadata, metadata_key)`` — the shape all callers should converge on,
+      matching the rest of NeuroConv. Resolves ``metadata["Devices"][metadata_key]`` and, if the entry
+      carries a ``device_model_metadata_key``, adds that model (idempotently, on demand) and links it.
+    * **Transitional** ``(nwbfile, device_metadata)`` — a pre-resolved entry dict, for callers that
+      build the entry themselves (e.g. placeholder or deprecated-nested-device fallbacks). This form
+      will be removed once those callers can pass ``metadata`` + ``metadata_key``.
```
pauladkisson, line 255: What is wrong with the device_metadata way of doing it that we need to change it here?

pauladkisson (commented):
I get the advantage of wanting to put all devices and device models in top-level metadata sections. That makes perfect sense to me. 

I'm not sure I understand the rationale behind needing to change the add device to NWB file function. Seems like it would work fine as is.

src/neuroconv/tools/nwb_helpers/_metadata_and_file_helpers.py
h-mayorquin: Mostly consistency: the other `add_*_to_nwbfile` methods pass the full metadata and resolve by `metadata_key`. This also allows the device model metadata to be resolved inside `add_device_to_nwbfile`. Otherwise we would need to either pass the device model metadata as another parameter, nested inside the metadata or pass also the device_metadata_key which all are fine options, but I'd rather just keep a consistent workflow across the codebase.

pauladkisson: I see, sounds good, thank you.

h-mayorquin:
Ready to review and finished @pauladkisson. I fixed the changelog that got scrambled in some previous PR I think.

pauladkisson (approved):
Looks good to me.

PR #1781, opened 2026-07-10
Support events with durations in TDTEventsInterface and migrate its tests onto gin datasets

h-mayorquin:
Two changes.

**1. `TDTEventsInterface` now writes event durations.** It previously raised `NotImplementedError` on any event type whose source carried a real end time (an interval), because there was nowhere to record a duration. The `EventsTable` model has a native `duration` column and the base events interface already writes it from `_EventsData.durations`, so an event type made of intervals is now written with a per-event duration. Timestamp-only event types are unchanged.

**2. The tests move onto gin-tracked datasets.** Off the fiber-photometry datasets under `ophys_testing_data` (imaging fixtures that only ever lived on the S3 mirror) and onto dedicated, anonymized event datasets now tracked in `ephy_testing_data` on gin. Two datasets cover the three event-type shapes the interface handles:

- **Timestamp-only** (no value column, no duration).
- **With a categorical value column** (a label per event).
- **With durations** (each event is an interval), now written with a per-event duration. This is what lets the duration support above be exercised on real data, replacing the fabricated stand-ins the old test used.


pauladkisson (approved):
Looks good to me.

PR #1783, opened 2026-07-13
GuppyConverter

pauladkisson:
This PR adds a converter for the Guppy format, which bundles together the Guppy interface, an events interface, and a fiber photometry interface.

Gin PR for testing data here: https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/56

One important design decision to highlight here: Guppy has some settings that filter on the events that it processes, specifically all events must occur after `baselineCorrectionStart`("Baseline Correction Start time(int)
") And must be at least `timeInterval` ("Time Interval (s)") apart. So I updated the DTRs of the events registry table to only reference the events that Guppy actually processes.

The following edge cases are left as follow-ups:
- NWB as an input data format (#1877)
- mixed format support (#1876)


pauladkisson:
Approved informally by @h-mayorquin

PR #1785, opened 2026-07-14
Refactor CSVEventsInterface onto BaseEventsInterface

pauladkisson:
Migrates `CSVEventsInterface` onto `BaseEventsInterface` (#1774), so it writes native pyNWB 4.0 `EventsTable` objects into `nwbfile.events` instead of `ndx_events` `Events`/`LabeledEvents` into `nwbfile.acquisition`. Follows the events-interface migration begun with `TDTEventsInterface` (#1775, #1781). Since `CSVEventsInterface` (#1755) is unreleased, this is a clean break.

## Scope

- Columns are assigned roles at construction: `timestamps_column`, `event_type_column`, and the two new knobs `value_columns` (per-event payload) and `durations_column` (per-event durations, blank cells → `NaN`).
- Each distinct `event_type_column` value becomes its own event type and, by default, its own `EventsTable`; merging into one table with an `event_type` discriminator is opt-in via `table_metadata_key` in metadata.
- Presentation (categorical vs numeric, via `column_categories`) and table grouping (merge vs split) stay metadata-driven; a non-numeric `value_columns` column is seeded categorical, a numeric one plain.
- Drops the `ndx-events` dependency for CSV events (now core `pandas` + `pynwb>=4.0`); the `csv_events` extra is kept as an empty install alias.
- Long/tidy format only; wide format and onset/offset-pair durations are documented as explicit non-goals.

Tests rewritten to assert against `EventsTable`/`nwbfile.events`; docs gallery and CHANGELOG updated.


pauladkisson:
Awaiting #1787

pauladkisson:
This PR also trims the GetMetadata method down to only what the source CSVs provide (i.e., the names) to conform with the new metadata principles.

docs/conversion_examples_gallery/events/csv_events.rst
```diff
@@ -3,14 +3,18 @@ CSV Events data conversion
...
-apart by a label column, name that column with ``event_type_column`` and the file is written as an
-``ndx_events.LabeledEvents`` object carrying the per-event labels. Either way the object lands in
-``nwbfile.acquisition``.
-
-Install NeuroConv with the additional dependencies necessary for reading CSV event data.
+at one CSV and assign each column a role. Every row is one event occurrence at ``timestamps_column``
+(seconds). Pass ``event_type_column=None`` when the file is a single event type, in which case it is
+written as one ``pynwb.event.EventsTable`` named after the file stem. When the file holds several event
+types told apart by a label column, name that column with ``event_type_column``: each distinct value
+becomes its own event type and, by default, its own ``EventsTable``. Merging several types into one
```
h-mayorquin: Maybe we can point them to the how-to here.

docs/conversion_examples_gallery/events/csv_events.rst
```diff
@@ -3,14 +3,18 @@ CSV Events data conversion
...
+table with an ``event_type`` discriminator column is opt-in by pointing their ``table_metadata_key`` at
+a shared key in the editable metadata. Optionally, ``value_columns`` carries extra columns along as
```
h-mayorquin: We should mention that by default the columns that are not specified are omitted (which I think is a good choice by the way, only add an event payload if the user opts-in.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -1,40 +1,43 @@
...
-    Columns other than ``timestamps_column`` and ``event_type_column`` are ignored; only timestamps
-    (and, for the labeled case, the event labels) are written.
+    Columns without an assigned role are ignored.
 
     Notes
     -----
     CSV recordings carry no embedded recording-start timestamp, so :meth:`get_metadata` does NOT
     populate ``NWBFile/session_start_time``. The user must supply it via editable metadata.
+
+    Two source layouts are anticipated but not yet supported: a *wide* format that spreads one event
```
h-mayorquin, line 37: Ok, make sense.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -49,6 +52,8 @@ def __init__(
         *,
         timestamps_column: str | int,
         event_type_column: str | int | None,
+        value_columns: list[str | int] | None = None,
```
h-mayorquin, line 55: payload_columns instead of `value_columns`?

I think value is less specific but you clarify this anyway on the documentation. This is a slight preference of mine. Keep what you think is best.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -149,108 +198,69 @@ def get_metadata(self) -> DeepDict:
...
-            event_type_column = self.source_data["event_type_column"]
-            column["description"] = (
-                f"Timestamps of the '{file_stem}' events from CSV, labeled by the " f"'{event_type_column}' column."
-            )
-            # A label per distinct value (first-appearance order), seeding LabeledEvents. The map is
-            # raw value -> display label; the user can rename the display labels in editable metadata.
-            column["column_categories"] = {"labels": {str(value): str(value) for value in pd.unique(labels)}}
-        metadata["Events"][self.metadata_key]["event_columns"][file_stem] = column
-        return metadata
+        timestamps, labels, _, _ = self._read_source()
```
h-mayorquin, line 234: does this re parses the csv file eagain? can that be avoided?

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -84,23 +96,41 @@ def __init__(
             file_path=file_path,
             timestamps_column=timestamps_column,
             event_type_column=event_type_column,
+            value_columns=value_columns,
+            durations_column=durations_column,
             verbose=verbose,
         )
         self.metadata_key = metadata_key or "csv_events"
```
h-mayorquin: Paul, do you think more than one CSV events interface will commonly be used in the same conversion? 

If so, I would derive the metadata_key in __init__ from the file stem so instances avoid collisions by default. If you think the multi-CSV case is rare, I would keep the current fixed literal for readability, since a single interface then has a stable, documentable key (metadata["Events"]["csv_events"]) that a user can name and type.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -115,12 +145,19 @@ def _read_timestamps_and_labels(self) -> tuple[np.ndarray, np.ndarray | None]:
...
+        # settings the label / value columns rely on.
         timestamps = pd.to_numeric(dataframe[timestamps_column], errors="coerce").to_numpy(dtype="float64")
         labels = None if event_type_column is None else dataframe[event_type_column].to_numpy()
-        # Drop rows with a missing timestamp, keeping the labels aligned.
+        durations = (
+            pd.to_numeric(dataframe[durations_column], errors="coerce").to_numpy(dtype="float64")
+            if durations_column is not None
+            else None
+        )
+        values = {column: dataframe[column].to_numpy() for column in value_columns}
```
h-mayorquin:
```python
values = {column: dataframe[column].to_numpy() for column in value_columns}
```

Payload columns are read raw, unlike `durations`/`timestamps` (which use `pd.to_numeric(..., errors="coerce")`). With `keep_default_na=False`, one blank cell in a numeric payload column reads as `''` and pandas keeps the whole column as object strings, the valid numbers included. Verified:

```
amplitude 0.5, <blank>, 0.9  ->  array(['0.5', '', '0.9'], dtype=object)   # strings, not floats
```

I don't think we can do to numeric here because the pyaload might be cateogircal, instead we need to check the type per column:

```python
coerced = pd.to_numeric(raw, errors="coerce")
unparseable = coerced.isna() & (raw.astype(str).str.strip() != "")  # a NaN on a non-blank cell = a real string
values[column] = raw.to_numpy() if unparseable.any() else coerced.to_numpy()
```

Either way, worth a test feeding a numeric payload column with a blank.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
```diff
@@ -84,23 +96,41 @@ def __init__(
...
+        (``None`` when ``durations_column`` is not set). Rows whose timestamp is missing (``NaN``) are
+        dropped, and the label / value / duration arrays are filtered in lockstep so every returned
+        array stays row-aligned.
         """
         timestamps_column = self.source_data["timestamps_column"]
         event_type_column = self.source_data["event_type_column"]
+        durations_column = self.source_data["durations_column"]
+        value_columns = self.source_data["value_columns"] or []
         # An int column specifier means a header-less file (positional columns); a str means a header row.
         header = None if isinstance(timestamps_column, int) else 0
```
h-mayorquin: Every column specifier is either a header name (`str`) or a positional index (`int`), and right now nothing checks they are all the same kind. The read mode comes from `timestamps_column` alone (`header = None if isinstance(timestamps_column, int) else 0`), so if you mix a name and an index it just blows up later as an opaque pandas `KeyError`. The duplicate-role guard does not catch it either: it is an `assert` (stripped under `python -O`) that compares the raw specifiers, so a name and an index for the same column do not collide.

Fix: up front, require all specifiers to be one kind and `raise` if mixed, derive `header` from that (store it), then run the duplicate check now that we are comparing like with like. Both `raise ValueError`, not `assert`:

```python
specifiers = [s for s in [timestamps_column, event_type_column, durations_column, *(value_columns or [])] if s is not None]
by_name = all(isinstance(s, str) for s in specifiers)
by_index = all(isinstance(s, int) for s in specifiers)
if not (by_name or by_index):
    raise ValueError(f"Column specifiers mix names and positional indices ({specifiers}); use one or the other.")
self._header = 0 if by_name else None
if len(specifiers) != len(set(specifiers)):
    raise ValueError(f"Each column may fill only one role, but the same column was assigned more than once: {specifiers}.")
```

h-mayorquin (changes requested):
The design looks good, one big question about the metadata_key convention and two suggestions for integrity checks.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
pauladkisson: I prefer to just leave it as is with value_columns.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
pauladkisson: I added a dynamic defaulting metadata_key, so that should cover this case.

src/neuroconv/datainterfaces/events/csv_events/csveventsdatainterface.py
h-mayorquin: Seems good. Marking this as resolved.

pauladkisson:
@h-mayorquin, all feedback has been addressed.

h-mayorquin (approved):
LGTM pending tests passed. Thanks for handling the feedback.

PR #1787, opened 2026-07-15
Allow a value column shared across merged event types

### LINKED ISSUE #1786: BaseEventsInterface cannot write a value column shared across event types merged into one EventsTable
## Background

We're partway through migrating the existing events interfaces onto `BaseEventsInterface` (#1774), the new base that writes native pyNWB 4.0 `EventsTable` objects into `nwbfile.events`. `TDTEventsInterface` went first (#1775, #1781), and while migrating `CSVEventsInterface` onto it (#1785) a gap in the base surfaced that is orthogonal to CSV — it lives entirely in `BaseEventsInterface`, so it's worth its own dedicated fix rather than something the CSV interface should work around.

The shape that exposes it is the most ordinary CSV events file there is: one timestamp column, one event-type column, and one or more payload columns shared by every row, e.g.

```
onset, kind,  outcome
1.0,   left,  go
2.0,   right, no_go
3.0,   left,  no_go
```

The natural NWB target is a single `EventsTable` with `timestamp`, an `event_type` discriminator (`left`/`right`), and one shared `outcome` column. Both the `EventsTable` model and the source data support this perfectly — you cannot, however, express it through the base.

## What actually blocks it

The metadata schema hangs `columns` off each event type (`event_types[<source_id>]["columns"]`), so a column that is logically shared across types has to be declared redundantly, once under each type. When those types are merged into one table, `_append_events_to_table` walks both types' `columns` blocks and trips its uniqueness guard:

https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/events/baseeventsinterface.py#L312

```python
assert column_name not in column_specs, (
    f"Two event columns write the same column_name '{column_name}'. "
    "Give each event column a unique column_name."
)
```

So the only way to satisfy the writer today is to give each type a *distinct* column name (`outcome_left`, `outcome_right`), which is precisely *not* a shared column — you get two half-empty columns backfilled with fill values on each other's rows, instead of one `outcome` column.

Worth being precise about the scope:

- **It is merge-only.** `column_specs` is scoped to a single `_append_events_to_table` call, i.e. one table. The default layout (each event type in its own table) is completely fine — separate tables can each carry their own `outcome` column independently. The problem is exclusively the opt-in merge path (types sharing a `table_metadata_key`).
- **Neither NWB nor the data is the obstacle.** `column_specs` and the column-creation loop are both keyed by `column_name`, so a shared column would be created exactly once and filled by both types — no duplicate `add_column`, no pynwb-level collision. The assertion is a purely *semantic* guard, not a technical necessity.
- **The guard is catching a real (but different) hazard.** It exists so that two types accidentally reusing a `column_name` with *different* specs don't silently collapse. If the assertion were simply removed, cell values would still be correct (they're computed per type from each type's own label map), but the shared column's single `description` and its `MeaningsTable` would be last-write-wins — one type's annotation silently dropped. So a blanket removal trades a loud failure for a silent one.

## Repro

```python
from neuroconv.tools.testing.mock_interfaces import MockEventsInterface

interface = MockEventsInterface(num_event_types=2, num_events=2, event_payload="single value")
metadata = {
    "Events": {
        "EventTables": {"pooled": {"table_name": "Trials", "description": "Pooled trials."}},
        "mock_events": {
            "event_types": {
                "events_0": {
                    "event_name": "left", "event_description": "Left-port events.",
                    "table_metadata_key": "pooled",
                    "columns": {"outcome": {"column_name": "outcome",
                                            "column_categories": {"labels": {0: "go", 1: "no_go"}}}},
                },
                "events_1": {
                    "event_name": "right", "event_description": "Right-port events.",
                    "table_metadata_key": "pooled",
                    "columns": {"outcome": {"column_name": "outcome",
                                            "column_categories": {"labels": {0: "go", 1: "no_go"}}}},
                },
            }
        },
    }
}
interface.add_to_nwbfile(nwbfile=mock_NWBFile(), metadata=metadata)
# AssertionError: Two event columns write the same column_name 'outcome'. Give each event column a unique column_name.
```

h-mayorquin:
This closes #1786.

When several event types are pooled into one `EventsTable` (they share a `table_metadata_key`), the writer could not give them a value column in common: two types declaring the same `column_name` tripped a uniqueness assert, so the only way through was to invent distinct names (`outcome_left`, `outcome_right`) and end up with two half-empty columns instead of the single `outcome` column the data actually wants. 

The guard now errors only on a genuine conflict:
* Conflicting meanings for the same would-be column
* Conflicting description for the same would-be-column.

This PR fixes the within-interface case (the repro in the issue); a shared column across interfaces already worked.

Edit: I've folded the cross-interface work into this PR. I originally meant to ship the single-interface merge here and the multi-interface case as a follow-up, but once I sat with it the two turned out to be the same code path: validating a shared column is one pass over the whole `Events` metadata, and it doesn't care whether the contributing event types come from one interface or several (a converter hands every interface's block to each write anyway


pauladkisson (commented):
This is a workable solution, but it seems like it would be better to declare shared columns once in a central location, the same way that shared EventsTables are declared. That way mismatches are structurally impossible.

Are there any reasons not to rework the metadata in this way?

h-mayorquin:
@pauladkisson 

I think this is a good idea. The best thing about it is that it makes mismatches structurally impossible, which is good ([making illegal states unrepresentable](https://functional-architecture.org/make_illegal_states_unrepresentable/) is a good principle). 

Let's discuss it. In the current PR we need at least two checks for merging across event tables:

* Meanings agree
* Descriptions agree

So I am thinking on something like:

```
EventTables:
  trials:
    table_name: Trials
    description: ...
    columns:                              # shared semantics, declared once (labels are NOT here)
      outcome:
        description: ...
        meanings: {go: ..., no_go: ...}   # keyed by label

event_types:
  acquisition:                            # key = event_type_source_id
    table_metadata_key: trials            # routes into the shared table
    columns:
      <field_source_id>:                  # the payload field this type reads
        column_name: outcome              # matches the central column by name (no separate pointer)
        labels: {0: go, 1: no_go}         # this source's own code -> label
  auxiliary:
    table_metadata_key: trials
    columns:
      <field_source_id>:
        column_name: outcome
        labels: {5: go, 7: no_go}         # a different source: same labels, different codes
```
Note the following:
* The checks are not necessary because that is explicit on the top
* The matching is done by column name here, to avoid introducing yet another metadata_key.
* The label remapping remains at the event level, as that is an acquisition-system to human-readable mapping and needs to be there.

What do you think of this design? I honestly can't find any flaw in it. The EventTables part is not filled by default, so it is only filled by the user to add meanings or a description. If they want to merge two columns within an interface they add the same column name, and by default it is written with no description or meaning. If the meanings table does not cover all the values of the merged table then we error. And if they want to merge across interfaces, the process is the same: it will point to the same table_metadata_key, and column_name should do the job.

Let me think a bit (and play with) on how merging across interfaces and in converter pipeline looks like but I like this so far.

pauladkisson:
Yes, this seems great to me.

h-mayorquin:
I thought more about this, and I do see now a real downside of the proposal:

As discussed previously, the intent of NWBEP0001's `EventsTable` was that each event type should get its own table. Merging multiple event types is supported and documented here on neuroconv, but I think it should take the back seat in terms of support. The current structure is designed for this:

```yaml
Events:
  my_interface:                       # metadata_key
    event_types:
      trial:                          # event_type_source_id
        columns:
          outcome_field:              # field_source_id
            column_name: outcome
            description: "Trial outcome."
            column_categories:        # a categorical column: possible to have labels and meanings annotated here
              labels:   {0: go, 1: no_go}   # has to be here, it labels the acquisition data
              meanings: {0: "...", 1: "..."}  # remains close to other fields
```

Note that there is no need for an `EventTables` entry, or a `table_metadata_key` at all, as an input from the user if they want to annotate a column. Everything is in the same place. The `EventTables` part of the metadata and the `table_metadata_key` only come into play if the user wants to merge event types into one table. That is, it keeps the common path local.

The proposal here would split the fields of a single event across two places, forcing every user who wants to annotate meanings to deal with `EventTables` and `table_metadata_key`, which is more complex:

```yaml
Events:
  my_interface:
    event_types:
      trial:
        table_metadata_key: trials    # now required
        columns:
          outcome_field:
            column_name: outcome
            column_categories:
              labels: {0: go, 1: no_go}          # stays here (source-specific)
  EventTables:
    trials:
      table_name: Trials
      columns:
        outcome:
          description: "Trial outcome."          # moved here
          meanings: {go: "...", no_go: "..."}    # moved here
```

In short, we would be making the merge case more concise and avoiding the checks here at the expense of making the common case more complicated, which I don't think is a good trade-off. So my inclination is to keep the single-place (inline) contract and treat merging as either a documented edge case which can be handled as an internal responsibility of an `NWBConverter` subclass. In any case this PR produces clear, loud errors on genuine conflicts.

pauladkisson:
I get the desire to prioritize support for the more common case, but I felt that your previous proposal was much more elegant for the more general case where a user might want to share events in the same table. 

We already essentially have a two-tiered system for the metadata structure, with the extra complexity of EventTables and table_metadata_keys being required if a user wants to merge events into the same table. So maybe those two different tiers are just structured a little bit differently to reflect their different requirements. i.e.

The metadata follows this structure if the user wants to merge two or more events into the same table.
```yaml
EventTables:
  trials:
    table_name: Trials
    description: ...
    columns:                              # shared semantics, declared once (labels are NOT here)
      outcome:
        description: ...
        meanings: {go: ..., no_go: ...}   # keyed by label

event_types:
  acquisition:                            # key = event_type_source_id
    table_metadata_key: trials            # routes into the shared table
    columns:
      <field_source_id>:                  # the payload field this type reads
        column_name: outcome              # matches the central column by name (no separate pointer)
        labels: {0: go, 1: no_go}         # this source's own code -> label
  auxiliary:
    table_metadata_key: trials
    columns:
      <field_source_id>:
        column_name: outcome
        labels: {5: go, 7: no_go}         # a different source: same labels, different codes
```

The metadata follows this syntactically sweeter structure if the user sticks with the happy case of only having one EventTable per event type. 
```yaml
Events:
  my_interface:                       # metadata_key
    event_types:
      trial:                          # event_type_source_id
        columns:
          outcome_field:              # field_source_id
            column_name: outcome
            description: "Trial outcome."
            column_categories:        # a categorical column: possible to have labels and meanings annotated here
              labels:   {0: go, 1: no_go}   # has to be here, it labels the acquisition data
              meanings: {0: "...", 1: "..."}  # remains close to other fields
```

h-mayorquin:
Thanks, I do think the central structure reads more elegantly for the pooling case. Yesterday I was thinking on a hybrid path similalr to what you are proposing. My hesitation is mainly that it gives us a second way to express the same thing, which is more to maintain, and since NWBEP001's intent is that one-table-per-event-type is the norm and merging is the exception, I wasn't sure the trade was worth it. Two questions so I understand exactly what you are proposing:

1. Is this an **additional** way to express `meanings` and the column `description` for a merged column (the inline `column_categories` stays valid and central is just an alternative), or the **blessed / only** path, where for a merged column the semantics can only be written centrally at `EventTables[<table_metadata_key>]["columns"][<column_name>]["meanings"]` (and `["description"]`), and the inline `column_categories.meanings` is no longer read once the column is merged? The two have quite different implications: the additional-path version is purely additive and never drops anything, whereas the only-path version means meanings a user wrote inline for a solo column would silently stop being read once that column becomes part of a merge.

2. What use case are you picturing where someone wants to both merge event types **and** annotate meanings? I am asking so I center my mind in your use case.

pauladkisson (approved):
The code changes look good, and I think we should proceed with this PR, but see #1794

PR #1788, opened 2026-07-15
Add MockFiberPhotometryInterface

h-mayorquin:
I realize we don't have this after working on the Guppy interface PR #965.

Neuroconv has a `Mock<X>Interface` for every other modality but none for fiber photometry, so every test and demo that needs an acquisition. The mock generates its own synthetic data so it can bridge that gap and test other configurations for which we do not have gin data. That's the purpose.


src/neuroconv/tools/testing/mock_interfaces.py
```diff
@@ -370,6 +373,112 @@ def _get_events_data_dict(self) -> dict[str, _EventsData]:
...
+    as ``GuppyInterface``) requires.
+    """
+
+    def __init__(
+        self,
+        *,
+        stream_names: str | list[str] = ("signal", "control"),
+        excitation_wavelengths_in_nm: tuple[float, ...] = (465.0, 405.0),
+        emission_wavelength_in_nm: float = 525.0,
+        location: str = "unknown",
```
pauladkisson: Earlier in the docstring, you say that this interface is supposed to be fully populated with no placeholders, but location equal to unknown is a placeholder.

tests/test_modalities/test_fiber_photometry/test_base_fiber_photometry_interface.py
```diff
@@ -0,0 +1,63 @@
...
+
+import pytest
+from jsonschema.validators import Draft7Validator
+from numpy.testing import assert_array_equal
+from pynwb import NWBHDF5IO
+
+from neuroconv.tools.testing.mock_interfaces import MockFiberPhotometryInterface
+
+
+class TestMockFiberPhotometryInterface:
```
pauladkisson: These tests should inherit from FiberPhotometryInterfaceTestMixin for consistency and succinctness. I'm not sure which of these existing tests are redundant with the mixin tests, but that's also something worth checking.

pauladkisson (changes requested):
I have some comments.

src/neuroconv/tools/testing/mock_interfaces.py
h-mayorquin: Yes. Thanks, I actually removed this functionality. I was thinking that that this would simulate the metadata extraction from source but 1) I don't think that most sources have this info, 2) we can add this when we needed it. 

This PR now adds a more bare-bones interface that simulates the data aspect only and metadata edition remains the blessed path for adding this type of info.

tests/test_modalities/test_fiber_photometry/test_base_fiber_photometry_interface.py
h-mayorquin: Should be done.

pauladkisson (approved):
Looks good to me.

PR #1790, opened 2026-07-17
Add Inscopix GPIO interfaces: channels as TimeSeries and derived events

h-mayorquin:
This adds two interfaces for the Inscopix `.gpio` file, which until now we could not convert at all even though it carries much of a session's auxiliary signal (LED-power and focus monitors, general-purpose inputs, digital lines, BNC sync/trigger, and coded inputs like an odor-concentration line). `InscopixGpioInterface` writes every channel as its own irregular `TimeSeries`, faithful to how the file stores it, with `exclude_channels` to drop any you don't want. `InscopixGpioEventsInterface` derives discrete events from the channels you name and writes each as a native `pynwb.event.EventsTable` in `nwbfile.events`. The two read the same file independently, so you run either or both, and a channel can legitimately appear in both (a coded line's raw trace and its derived events).

There is no automated way of telling which channels carry a continuous signal and which carry discrete events, as the format simply does not record the distinction, so I went with two interfaces instead of guessing. The analog side stores every channel raw by default, and the events side asks you to name the channels you want and how each should be read.

Requested by @austin2118ace.


austin2118ace:
Awesome work, I'll test it out tomorrow and lyk

PR #1793, opened 2026-07-17
[Draft/Discussion] Time alignment improvement

h-mayorquin:
Draft

Here is the model that I am thinking about:
https://neuroconv--1793.org.readthedocs.build/en/1793/user_guide/temporal_alignment.html


pauladkisson:
I was thinking about this, and I think there are two separate axes here to consider. 

**Axis one is the temporal alignment situation/goal**
1. Two clocks have some offset but no drift --> one clock needs to be shifted
2. Two clocks have a synchronization signal --> the timestamps need to be interpolated relative to that signal.
3. One primary clock records the timestamps of a secondary clock --> secondary clock timestamps need to be re-assigned.

**Axis two is the data/interface situation.**
1. Simple case: one interface is responsible for one timestamps vector. All data depend on that one timestamp vector.
2. Multi-timestamps case: one interface is responsible for more than one timestamp vector, but all timestamp vectors lie on the same time basis.
3. The multi-time basis case: one interface is responsible for more than one timestamp vector, and some of those timestamp vectors lie on different time bases.

Another consideration is whether the interface has any non-timestamp data. An interface without any non-timestamp data is basically completely obsolete in case 1.3.

h-mayorquin:
@pauladkisson 

I have updated the documentation to reflect our discussion on Friday, July 17. It now opens with the distinction between gross and fine alignment and is organised around the phases we talked about. I also dropped "timebase" from the vocabulary: it reads as a claim that the objects share a common clock, and we should stay agnostic about that, so the page talks about time-bearing objects carrying an offset instead.

- **Phase 1, gross alignment.** `shift_times` moves all the objects in an interface or converter as a whole, a rigid transformation. This only needs the temporal objects to be *enumerable*, not addressable, which is why it can go first and independently of the rest.
- **Phase 2, fine alignment.** Setting timestamps requires saying *which* object you mean, so it needs an addressing mechanism. The hard part here is not technical but design: a unified way of addressing that works across interfaces and actually makes sense to the user.
- **Phase 3, bulk alignment.** Once addressing is in place, we can switch the timestamps of some interfaces onto those of another, the common case of interpolating between two acquisition systems through TTL (transistor-transistor logic) pulses.

The plan moving forward is in https://github.com/catalystneuro/neuroconv/pull/1795, which implements phase 1 for events. Everything past `shift_times` on the page is still proposed, so this is  just discussion.

docs/user_guide/temporal_alignment.rst
```diff
@@ -1,137 +1,220 @@
...
+never resamples or changes the data values; it only sets the timing of the samples you already have.
 
-There are several ways to synchronize acquisition systems post-hoc. This tutorial will walk you through the 3 methods
-implemented in NeuroConv. The API also allows you to define an entirely custom method for synchronization.
+Time-bearing objects and the offset
+-----------------------------------
 
-Note that NeuroConv does not resample the data, as this requires a resampling method that can change the values of
-the underlying data. Rather, we aim to provide the timing of the samples in a common clock.
+Alignment acts on **time-bearing objects**: the objects an interface writes that carry a time coordinate relative to
```
pauladkisson: The formatting on these neurodata_objects is inconsistent.

docs/user_guide/temporal_alignment.rst
```diff
@@ -1,137 +1,220 @@
...
 
-.. image:: ../../_static/images/time_alignment_1.png
-   :alt: Diagram of the first method of time alignment, where the start time of the secondary system is shifted to match the primary system.
+Every offset starts at ``0.0``, which is the identity: the source times are written exactly as the acquisition system
+recorded them. Aligning is nothing more than changing offsets, and they are applied when the data is written, so the
+times held in memory stay in the source clock.
+
+Note what this deliberately does *not* claim: it says only that these objects have times, not that they share a clock
+with each other or with anything else. Whether two things are co-timed is up to you and your rig, and you express it
+by the offsets you set.
```
pauladkisson: These two paragraphs are confusing because they kind of imply that alignment is just a matter of shifting the offset, and then the next section clearly states that alignment is not always a matter of shifting the offset. 

"Aligning is nothing more than changing offsets" and "Whether two things are co-timed is up to you and your rig, and you express it
by the offsets you set."

docs/user_guide/temporal_alignment.rst
```diff
@@ -1,137 +1,220 @@
...
+
+**Gross alignment.** The samples are correctly spaced relative to one another, and only the placement of the stream as
+a whole on the shared clock is unknown. Fixing it is a rigid translation: every interval inside the stream survives
+untouched. ``shift_times`` and ``set_offset`` do this, and it is what most conversions need.
+
+**Fine alignment.** The internal structure itself is wrong. Two clocks running at slightly different rates drift apart
+over a session, so no single offset can be right for both the first sample and the last; the per-sample times have to
+be rewritten, usually from a synchronization signal. ``set_times`` does this.
+
+A quick test for which one you have: if sliding the stream can make it right, it is gross. If sliding it makes the
```
pauladkisson: This sentence is confusing, I think, in part because "it is fine." is a common expression in English for 'everything is all good.'" 

But also, I don't think it really clarifies anything in particular. The definitions should speak for themselves.

docs/user_guide/temporal_alignment.rst
```diff
@@ -1,137 +1,220 @@
...
+
+``shift_times(delta)`` moves **every time-bearing object in the interface** by ``delta`` seconds. It is a rigid
+translation: the spacing between samples, the gaps between events, and every duration are all preserved, only the
+position on the shared clock changes. It is relative, so repeated calls accumulate.
+
+.. code-block:: python
+
+    events_interface.shift_times(3.0)   # every event now sits 3.0 seconds later on the session clock
+
+This is what most conversions need. The canonical case is a secondary system that sends a single pulse to the primary
```
pauladkisson: "This is what most conversions need." -- That has not been my experience. Have most of your conversions only needed offset alignment?

docs/user_guide/temporal_alignment.rst
```diff
@@ -1,137 +1,220 @@
...
-slight differences in the clock speeds of the two systems. Examples of this include sending TTL pulses from a camera
-used to acquire optical imaging to a NIDQ board every time a frame is captured or every time a volume scan begins. You
-can then align the timestamps of the secondary system by setting them to the pulse times as received by the primary
-system, aligning the times to that system. Once the timestamps are known they can be set in any DataInterface via the
-DataInterface method
-:py:meth:`~neuroconv.basetemporalalignmentinterface.BaseTemporalAlignmentInterface.set_aligned_timestamps`.
-
-.. image:: ../../_static/images/time_alignment_2.png
-   :alt: Diagram of the second method of time alignment, where the timestamps of the secondary system are aligned to the primary system.
+.. image:: ../_static/images/time_alignment_moves_together.png
```
pauladkisson, line 83: I like these visualizations. They're quite nice.

docs/user_guide/temporal_alignment.rst
```diff
@@ -1,137 +1,220 @@
...
+   :alt: An interface's time-bearing objects all shift together by the offset; the gaps between them never change.
    :width: 600px
    :align: center
 
+``shift_times`` is the only method that needs no further information: because it adds the *same* delta to every
+offset, it can safely apply to everything the interface holds. Every other operation changes one object relative to
+the others, so it has to say which one, which means first knowing what there is to address.
```
pauladkisson: This doesn't make much sense to me and is my first real design critique. 

`shift_times` and `set_offsets` seem completely symmetric to me with respect to how much information they need. Both could be addressed to individual neurodata_objects or to the entire set of neurodata objects included in an interface. The only real difference, as far as I can tell, is in the name, whether they shift or set the offset time.

docs/user_guide/temporal_alignment.rst
```diff
@@ -1,137 +1,220 @@
...
-:py:meth:`~neuroconv.basetemporalalignmentinterface.BaseTemporalAlignmentInterface.align_by_interpolation`.
-Note the data values for the series itself is *not* changed during the process, only the timestamp values are
-inferred for common reference time.
+Identifying the times inside an object is per neurodata type, not a single naming rule:
 
-.. image:: ../../_static/images/time_alignment_3.png
-   :alt: Diagram of the third method of time alignment, where the timestamps of the secondary system are aligned to the primary system using a synchronization signal.
-   :width: 600px
-   :align: center
+* Well-known types are handled by type: a ``TimeSeries`` moves its ``timestamps`` or ``starting_time``, a ``Units``
```
pauladkisson: This section seems completely unnecessary. Every interface is going to have to figure out what timestamps are subject to temporal alignment on a case-by-case basis

docs/user_guide/temporal_alignment.rst
```diff
@@ -1,137 +1,220 @@
...
 
+    single_object_interface.set_offset(3.0)      # key optional when there is only one object
 
-Extracting synchronization signal
----------------------------------
+The difference from ``shift_times`` matters when an alignment step runs more than once. ``shift_times`` accumulates, so
+calling it twice adds twice; ``set_offset`` is absolute, so calling it twice with the same value leaves the object
+exactly where the first call put it. Use ``set_offset`` when you know where something belongs and the step may be
+re-run, and ``shift_times`` for a correction you want to *add* on top of an existing alignment, a known cable or
+trigger latency, say. ``set_offset(0.0)`` restores an object's source times.
```
pauladkisson: Wait, I need a little bit of clarification on this one. 

What happens if you have an interface with timestamps that do not start at zero, but you want to set them to zero?

docs/user_guide/temporal_alignment.rst
```diff
@@ -1,137 +1,220 @@
...
+            np.interp(interface.get_times(key=key), pulses_local, pulses_reference),
+            key=key,
+        )
+
+Because every object goes through the *same* mapping, their relationships to one another are preserved just as they
+are under a rigid shift; what changes is that the mapping is no longer a constant offset, so drift is removed as well
+as displacement. Objects that do not support ``set_times`` (an events table, a trials table) cannot be corrected this
+way, and if the interface holds any, the loop above will raise on them rather than silently skip them.
+
+Not every time-bearing object supports this, because it needs a single array of times to replace. It applies to a
```
pauladkisson: Why not? We could just key into each individual timestamps array in an object if it had multiple.

pauladkisson (commented):
Excited to settle on a superior temporal alignment system.

I have several comments.

docs/user_guide/temporal_alignment.rst
h-mayorquin: Yes, I agree this is overarching.

docs/user_guide/temporal_alignment.rst
h-mayorquin: They differ in purpose and in implementation:

- `shift_times` rigidly shifts *all* the time-bearing objects by one amount (a converter can have several time series; a dynamic table can have several `_time` columns). This requires no input, it applies to the whole interface or converter. The use case also does not require you to understand absolute position, if your trials started later, fix them by this amount, if you have an interface with a set of objects (same acquisition system) and you know that system started later just shift the whole interface by that amount. This is useful in trialized conversions when the inter-trial-time is missing.

- `set_offset` sets one object's absolute start relative to `session_start_time`, this is specific to one object (not an interface). This requires the object address an input and requires you to know the absolute time. I expect this to be less used or only use by internal machinery in converters or interfaces.

Implementation-wise, `shift_times` only needs to *enumerate* the time-bearing objects, while `set_offset` needs a system to *address* any one of them. Enumeration is the weaker requirement, which is why shift comes first.

I do think the naming could be better.

docs/user_guide/temporal_alignment.rst
h-mayorquin: Yes, I agree.

docs/user_guide/temporal_alignment.rst
h-mayorquin: Thanks, after looking at the rest of your comments I realize that the structure of the document and the model needs improvement so I will do a big change, I will come back to you with that and describe what changed.

h-mayorquin (commented):
Thanks, I think that most of your comments are on point and this needs a big update. I will come back to you.

h-mayorquin:
Ok, I thought about your feedback and I think this is a better model.

The one that drove the most change was your point that `shift_times` and `set_offset` looked symmetric and differed mainly in the name. I defended them at first, but chasing it down I came around to your read: they were too symmetric to both earn a place, so **`set_offset` is gone entirely**. Once it was out, the distinction that actually matters turned out not to be shift-versus-set but *scope*: some corrections apply to a whole interface at once (one offset, or one clock-to-reference map, for every object it holds), and some apply to a single object. This is an organization principle now.

The document was also overly centered on implementation and on `shift_times`,as you pointed out and that was overeaching. I change this to read like a Diátaxis explanation, and it separates the two big cases by scope: **gross vs fine alignment** (same clock, only misplaced, versus different clocks that drift), by describing on the intro where each applies. 

Some concrete changes:
- Your suggestion that we could just key into an object's individual arrays is exactly how per-object addressing works now.
- The interpolation entry point is `remap_times`, renamed off the legacy `align_by_interpolation` so a new keyed method doesn't collide with the old single-array one.  The ambition here is that the new form will allow us to remap all the relevant times of an interface as you did it with guppy.
- The "which times move" section is gone (it was leaking too many implementation details) and other overeaching claims are gona as well. 
- I added more visualizations / diagrams.

pauladkisson:
I took a look and I like the new system/document a lot. No notes.

PR #1779, opened 2026-07-09
Adapt DoricFiberPhotometryInterface to use BaseFiberPhotometryInterface

alessandratrapani:
- [x] Adapt DoricFiberPhotometryInterface to use BaseFiberPhotometryInterface --> https://github.com/catalystneuro/neuroconv/pull/1754
- [x] Add DoricFiberPhotometryInterface  for csv format
- [x] Add support for legacy format
- [x] Add data on gin for testing --> https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/52

PR #1796, opened 2026-07-20
Remove the duplicated `FiberPhotometryInterfaceTestMixin` definition

h-mayorquin:
`FiberPhotometryInterfaceTestMixin` is defined twice in `data_interface_mixins.py`, both copies from #1779. Python keeps the second, so the first has never run and deleting it changes no behavior. I bothered because the two are not identical and the dead one comes first in the file, so it is what you land on when you go looking for the mixin to write a subclass against, which is how I hit it while reviewing #1788. For whoever reviews: the second copy is the correct one to keep, since the first is missing three `_check_*` helpers that its own `check_read_nwb` calls.


pauladkisson (approved):
Looks good, thank you. Sorry I didn't catch this earlier. I think it was likely from some kind of merge conflict.

PR #1798, opened 2026-07-20
Stop the events interfaces from writing descriptions and labels the source does not carry

h-mayorquin:
The events interfaces were filling description and meaning slots with strings built from the source's own identifiers. `TDTEventsInterface` wrote `"Onset times of the TDT (Tucker-Davis Technologies) epoc 'PtAB'."` as an event type's description and seeded a strobe column with an identity label map (`{16: "16"}`), and the writer then promoted those into a `MeaningsTable`, whose whole job is to answer "what does this value mean". A reader who opens the file and asks what `PtAB` means gets told it is the epoc named `PtAB`. The identity labels are worse than useless: they occupy the slot where a real codebook goes and make an un-annotated column look annotated. None of this is a lie about the data, but a tank records store handles and codes and nowhere records what they stand for, so every one of these strings is NeuroConv authoring prose on the user's behalf and filing it where it reads as provenance from the rig. I took the rule to be that we write a description, label, or meaning only where the source or the user supplied one, and applied it at each site.

That rule pushed in two directions. `get_metadata()` should report what the source holds, so `event_description` and a column's `description` are now optional and an interface that reads neither omits both keys rather than emitting empty ones (`columns` likewise disappears for a timestamps-only event type), following the strict-provenance profile settled in #1557. The empty string belongs at write time instead, where NWB (Neurodata Without Borders) requires a description field whether or not anyone wrote one. On the writer side, a `MeaningsTable` is now built only where there is something to look up: a value column whose meanings are all blank gets none and a partly filled one keeps just the entries that were filled, and a merged table's `event_type` discriminator gets a row per type the user described, with no table at all when nobody described any. For the same reason I dropped the composed description a merged table used to get when no `EventTables` entry named it, since that sentence only restated the `event_type` column back at the reader. The visible consequence is that an un-annotated conversion now trips NWB Inspector's "Description is missing", which I think is correct: the description really is missing, the user is the only one who can supply it, and a generated fallback would suppress exactly the signal that tells them so.


pauladkisson (approved):
Looks good to me.

h-mayorquin:

PR #1800, opened 2026-07-20
Updates for BaseFiberPhotometry

pauladkisson:
This PR implements the updates to the BaseFiberPhotometry interface as [discussed](https://docs.google.com/document/d/1uj8LCz9L5ydBrfF6mTqXRc92MVsuAzOKsTbYVnNyFwM/edit?usp=sharing) with @h-mayorquin.

Of particular note is the `get_example_metadata` method, which provides a fully fleshed-out example of how the fiber photometry metadata ought to look for easy template editing as well as self-documenting code.


src/neuroconv/datainterfaces/fiber_photometry/basefiberphotometryinterface.py
```diff
@@ -125,16 +122,114 @@ def set_aligned_timestamps(self, aligned_timestamps: np.ndarray) -> None:
...
+        record the optical hardware, indicator, and table, add that metadata under ``"FiberPhotometry"``,
+        ``"Devices"``, and ``"DeviceModels"``; :meth:`get_example_metadata` returns a complete template.
         """
         metadata = super().get_metadata()
-        default_fiber_photometry_metadata = get_default_fiber_photometry_metadata(metadata_key=self.metadata_key)
-        return dict_deep_update(metadata, default_fiber_photometry_metadata)
+        series_metadata = dict(
+            name="FiberPhotometryResponseSeries",
+            description="Fiber photometry response series.",
+            unit="a.u.",
```
pauladkisson: Okay, wait, so does this violate the principle of metadata defaults living in the add_to_nwbfile method? 

It's literally just a name and generic description, so I wasn't sure how it interacted with #1801 @h-mayorquin.

h-mayorquin: This is my current thinking.

**`name`**: This is not real metadata and I think it is fine for us to make the default.
**`description`** this is the gray area for me. My current thinking: if we can assemble a description from something the source actually recorded (e.g. a json description of some parametesr that don't really have a home) I think that should be fine. A generic `"Fiber photometry response series."` asserts nothing, so I'd rather have it empty there.
**`unit`** I don't think this should be metadata at all. Unit isn't a user-editable choice, it's a property of what the data is, so it shouldn't sit in the editable dict. If a source ever reports calibrated units we read them and pass them straight into the `FiberPhotometryResponseSeries(...)` call; otherwise we write the right default (`"a.u."`) at that same instantiation. I don't know the fiber photometry extractors as well as you do so maybe I am missing something.

h-mayorquin (changes requested):
This is bundling two things:
1) dropping placeholders from the metadata which is fine
2) Adding a method for scaffolding for which I would like a bit more of discussion.

https://github.com/catalystneuro/neuroconv/issues/1802

We can separate 1 from 2 or agree on #1802 as you prefer.

src/neuroconv/datainterfaces/fiber_photometry/basefiberphotometryinterface.py
pauladkisson: Great, thanks for the clarification. Ironing out all of these edge cases, I think, really helps make the vision more actionable and concrete.

pauladkisson:
> This is bundling two things:
> 
> 1. dropping placeholders from the metadata which is fine
> 2. Adding a method for scaffolding for which I would like a bit more of discussion.
> 
> #1802
> 
> We can separate 1 from 2 or agree on #1802 as you prefer.

Just removed the Get Example Metadata from this PR, so now it should be clean.

h-mayorquin (approved):
LGTM

PR #1801, opened 2026-07-20
Document get_metadata() principles

h-mayorquin:
This writes down what we agreed in #1557: `get_metadata()` returns only what the source recorded, and the values NWB requires regardless are supplied where the object is built. Until now that rule lived in a single two-line bullet inside `ophys_metadata_structure.rst`, scoped to ophys, so there was nothing modality-independent to point a contributor at. The new `metadata_principles.rst` covers what an interface may report, what happens to the required fields no source supplies, and the three flow principles from #1511 that keep write-time placeholders from leaking back into the dictionary.


docs/developer_guide/metadata_principles.rst
```diff
@@ -0,0 +1,116 @@
...
+  optical information should not produce an ``ImagingPlane`` with an ``indicator`` and an
+  ``excitation_lambda``; a recording with no probe information should not produce an
+  ``ElectrodeGroup`` at a named ``location``. The structure implies the source described these
+  things.
+
+A placeholder in the dictionary makes it impossible for anything downstream to tell whether a value
+came from the source, the user, or NeuroConv. Warning about it at write time does not help, since the
+value is written regardless.
+
+An un-annotated conversion therefore produces fields the user still has to fill in, and NWB Inspector
```
pauladkisson: I don't understand what this means. Could you clarify?

docs/developer_guide/metadata_principles.rst
```diff
@@ -0,0 +1,116 @@
...
+like an answer.
+
+See `issue #1557 <https://github.com/catalystneuro/neuroconv/issues/1557>`_ for the discussion.
+
+Staying faithful is not the whole story, though. Some of what a source leaves out is genuinely
+required by NWB, and a file cannot be written without it, so a value has to come from somewhere even
+when the interface reports none.
+
+
+Placeholders for required fields
```
pauladkisson, line 56: I think I understand what this is saying, but it might be easier to express in a step-by-step or a priority list:
- If a field isn't required, then you just omit it.
- If it is required, then you write a placeholder.
- In addition to writing a placeholder, it needs to be in a centralized location.
- There needs to be a check on NWB inspector that flags it.

Something like that.

docs/developer_guide/metadata_principles.rst
```diff
@@ -0,0 +1,116 @@
...
+Checklist for a new interface
+------------------------------
+
+When writing or reviewing an interface:
+
+- Every key ``get_metadata()`` returns corresponds to something read from the source.
+- No key holds ``""``, ``{}``, ``None``, ``np.nan``, ``"unknown"``, or any other sentinel.
+- No object is returned that the source gives no evidence for: no imaging plane without optical
+  information, no electrode group without probe information, no fiber without a fiber.
+- Required NWB fields with no source value are filled in ``add_to_nwbfile``, where the object is built.
```
pauladkisson, line 123: I think adding required fields in the add to NWB file method ad hoc per interface is okay and probably good enough for now. 

But I wanted to flag that I think it would be nice as a potential future improvement to have a dedicated place for this. ex. `prepare_metadata_for_conversion`.

pauladkisson (commented):
I have a couple comments about wording and clarifications, but I'm on board with all of the substantive principles.

docs/developer_guide/metadata_principles.rst
h-mayorquin: Fair enough point, the important point is that they should not be added by default in the metadata. Previously, they were added in `get_metadata` and that meant that in the conversion pipeline the distinction between user intent and library intent was lost. 

But if the user places the placeholder as an opt-in I think that's fine.

docs/developer_guide/metadata_principles.rst
h-mayorquin: The idea was to mention that conversions might produce data that will still be flaggged by the inspector as missing. Before they were not because we were putting placeholders. I don't think this is necessary here so I removed it.

docs/developer_guide/metadata_principles.rst
h-mayorquin: Yes, good suggestion. I add this in bullet point format for clarity.

docs/developer_guide/metadata_principles.rst
pauladkisson: I actually meant as an automatic call inside add_to_nwbfile.

h-mayorquin: I misunderstood, what would the purpose?

pauladkisson: Just organizational clarity, nothing more.

h-mayorquin: Just to make sure I follow, do you mean a common injection point for the placeholders? That's how I'm reading it. If so, I think it still has to happen at the last point of delivery (object construction), because some modality pipelines (e.g. spikeinterface's `add_recording_to_nwbfile`) are public and get called directly on a raw recording with no interface involved (I know that alessio uses them that way). So the interface's `add_to_nwbfile` is only one of two entry points, and a `prepare_metadata_for_conversion` hook at the interface level would fill defaults for the interface path but silently skip everyone using the pipeline functions directly. Object construction is the one seam both paths pass through, so it's the only place the defaults reach every caller.

pauladkisson: I see. Yeah, that's a good point.

h-mayorquin:
After this PR and once we agree on https://github.com/catalystneuro/neuroconv/issues/1802 we can add another slab of documentation regarindg the metadta scafolding.

h-mayorquin:
@pauladkisson Is there anything outstanding here? I would like to point my agents at the docs when developing interfaces.

pauladkisson (approved):
Nope, should be good to go.

PR #1805, opened 2026-07-22
Add `DoricEventsInterface` for Doric digital IO events

h-mayorquin:
Doric Neuroscience Studio `.doric` files record digital IO lines (a camera-exposure TTL, a behavior trigger) as densely sampled `0`/`1` traces, so unlike TDT or CSV events these aren't stored as records, they are derived from a signal. This interface detects each `DigitalIO` line's rising edges and writes one `pynwb.event.EventsTable` per line into `nwbfile.events`, on the same `BaseEventsInterface` that the TDT and CSV events interfaces already use. Every toggling line is converted, a constant line is skipped, and `session_start_time` is read from the file's `Created` attribute.

To keep this first pass simple, I've left a few things for follow-up PRs (if they are necessary): reading the DoricStudio CSV export and the legacy EPConsole layout (both also carry digital events), and writing each pulse's on-duration (its rising-to-falling span) as a durative event.


pauladkisson (approved):
Code looks great, and I confirmed that it works. It matches Guppy's implementation on the Guppy example data exactly, but see https://github.com/LernerLab/GuPPy/issues/394.

h-mayorquin:
> Code looks great, and I confirmed that it works. It matches Guppy's implementation on the Guppy example data exactly, but see [LernerLab/GuPPy#394](https://github.com/LernerLab/GuPPy/issues/394).

Oh, interesting.

Thanks for the review

PR #1806, opened 2026-07-22
Add how-to for annotating  fiber metadata annotation in fiber-photometry interfaces 

h-mayorquin:
As in the title

This bundles #1807 as I would like to separate the part of device model annotation from the rest of the annotations so the complexity is presented gradually.


h-mayorquin:
@pauladkisson @alessandratrapani could you take a look? since you seem to have been thinking about fiber photometry lately it will be great to get your feedback.

Direct link:
https://neuroconv--1806.org.readthedocs.build/en/1806/how_to/annotate_fiber_photometry_metadata.html

h-mayorquin:
The last changes is that I reworked the signal and isosbestic section so it writes one series per excitation wavelength instead of stacking both into one, to comply with what we settled in catalystneuro/ndx-fiber-photometry#55 and the recommendation Paul merged in catalystneuro/ndx-fiber-photometry#56. Tell me if that reads the way you meant it, since it does mean the most common GCaMP setup now takes two interfaces and a converter.

docs/how_to/annotate_fiber_photometry_metadata.rst
```diff
@@ -0,0 +1,718 @@
...
+
+How to Annotate a Signal and Isosbestic Control
+-----------------------------------------------
+
+The near-universal GCaMP setup records one fiber at two excitation wavelengths, the calcium-dependent
+signal (465 nm) and an isosbestic control (405 nm). That is two traces, so the table gets **two rows**,
+one per trace, differing in their excitation wavelength and excitation source.
+
+The two traces do **not** go into one series. `ndx-fiber-photometry
+<https://github.com/catalystneuro/ndx-fiber-photometry>`_ recommends one series per excitation and
```
pauladkisson: We could link the README section to be even more specific here. https://github.com/catalystneuro/ndx-fiber-photometry#recommended-organization-of-response-series

pauladkisson (approved):
This is great. I only had one minor comment that you can take or leave.

docs/how_to/annotate_fiber_photometry_metadata.rst
h-mayorquin: Thanks. Added.

PR #1807, opened 2026-07-22
Relax two over-strict required fields in fiber photometry metadata

h-mayorquin:
I would like to do this before the how-to do #806 so I can present the complexity of metadata annotation gradually.

This PR relaxes the base fiber photometry interface (`BaseFiberPhotometryInterface`) metadata requirements in two ways where neuroconv is more strict than the schema:

- The optical fiber's device model is now optional. This is not really required by the schema, and requiring it forces users to invent catalog specifications (numerical aperture, source type, detector type) that they might not have.
- The response-series table-region description is now defaulted at construction. On this, I am just imitating what we do in ecephys ([permalink](https://github.com/catalystneuro/neuroconv/blob/88f528f1ac4d85870a564436d8b89971d962e183/src/neuroconv/tools/spikeinterface/spikeinterface.py#L464-L466)). I don't think this can ever be extracted from source formats and it is not that informative; a default is more informative and removes a hurdle from the users.

See #1801 (documenting the `get_metadata()` principles) and also #1806 (the fiber photometry how-to), the latter requires and bundles this.


pauladkisson (approved):
LGTM

PR #1808, opened 2026-07-22
Add interleave demultiplexing to CSVFiberPhotometryInterface

h-mayorquin:
I did this to simplify the Neurophotometrics (NPM) interface PR [#1756](https://github.com/catalystneuro/neuroconv/pull/1756) and to keep the demux configs available alongside this interface, so that reader can wrap this shared primitive instead of re-implementing its own de-interleave logic. 

Background:
Some fiber photometry CSVs interleave the excitation channels frame-by-frame down the rows rather than giving each channel its own column, so a reader has to demultiplex one channel out before it can build a series. This adds an optional `demux` config to the interface, in two shapes: a by-column mode that reads the rows whose label column (e.g. a Neurophotometrics `LedState`) equals a value, and a by-stride mode that reads every k-th row from an offset when a header-less file just cycles the channels in a fixed order. The interface stays single-series (one `FiberPhotometryResponseSeries` per interface): you instantiate one interface per channel and compose them in a converter.


src/neuroconv/datainterfaces/fiber_photometry/csv/csvfiberphotometrydatainterface.py
```diff
@@ -64,6 +72,14 @@ def __init__(
         timestamps_column : str or int
             The column holding the timestamps (seconds) for the series' time axis. A column name for a
             CSV with a header row, or a positional index (0-based) for a header-less CSV.
+        demux_config : dict or None, optional
```
pauladkisson: This says `dict` or `None`, but the actual code lists it as a DemuxConfig object.

pauladkisson: It also implies dictionary-based get item in the body of the description, which appears to be outdated as well.

src/neuroconv/datainterfaces/fiber_photometry/csv/csvfiberphotometrydatainterface.py
```diff
@@ -154,8 +174,24 @@ def get_available_columns(cls, file_path: FilePath, read_kwargs: dict | None = N
...
+        # read path without demuxing (MultiFileCSVFiberPhotometryInterface) simply never sets it.
+        demux_config = getattr(self, "_demux_config", None)
+        if demux_config is None:
+            return self._read_csv(file_path, usecols=columns)
+        if isinstance(demux_config, ColumnDemux):
+            read_columns = columns if demux_config.column in columns else [*columns, demux_config.column]
+            dataframe = self._read_csv(file_path, usecols=read_columns)
+            return dataframe[dataframe[demux_config.column] == demux_config.value]
+        dataframe = self._read_csv(file_path, usecols=columns)
+        return dataframe.iloc[demux_config.skip_rows :].iloc[demux_config.index :: demux_config.channels]
```
pauladkisson: I know that this is strictly speaking unnecessary, but I would find it more readable if we had an `elif isinstance(demux_config, StrideDemux):` for this section.

pauladkisson (approved):
This looks great. Very minor comments.

I double-checked that this demux option works for all of my sample npm data and matches Guppy output with StrideDemux.

src/neuroconv/datainterfaces/fiber_photometry/csv/csvfiberphotometrydatainterface.py
h-mayorquin: Why not? added.

src/neuroconv/datainterfaces/fiber_photometry/csv/csvfiberphotometrydatainterface.py
h-mayorquin: Should be fixed.

PR #1809, opened 2026-07-22
Add time_unit to CSVEventsInterface

pauladkisson:
Adds an optional `time_unit` argument (`seconds`, `milliseconds`, or `microseconds`) to `CSVEventsInterface` that scales the timestamp and duration columns to seconds. `value_columns` are left untouched.

This came out of the Neurophotometrics events work (#1757): NPM event onsets are recorded in a non-seconds clock (specifically, milliseconds), and the scaling I'd put on `NPMEventsInterface` is really a general CSV concern — any events CSV can be recorded in ms/µs, and `CSVEventsInterface` already owns a `durations_column` that shares the same time base. Splitting it out so it can stand on its own; #1757 then just forwards `time_unit` through.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 * Fixed the global compression test failing against `hdf5plugin>=7.0.0`. The test asserted on each filter's free-text description string, which 7.0.0 reworded for LZ4 and Zstd when switching to The HDF Group's bundled implementations (the filter IDs are unchanged). The test now asserts only on the stable numeric filter ID, which is what actually proves the requested filter was applied. [PR #1766](https://github.com/catalystneuro/neuroconv/pull/1766)
 
 ## Features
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
LGTM.

I'm glad we delayed this back on [#1755](https://github.com/catalystneuro/neuroconv/pull/1755#discussion_r3508596583), where I asked whether to let users declare a non-seconds unit or force the input to seconds.

Waiting turned out to be the right call, because now I realize this shouldn't be `CSVEventsInterface`-only. I think the following interfaces would benefit from the same knob, and this PR is a good place to set the precedent:

- `CSVFiberPhotometryInterface` / `MultiFileCSVFiberPhotometryInterface`: assume seconds
- `CsvTimeIntervalsInterface` (and `TimeIntervalsInterface`): "all timing columns must be in seconds"
- `FicTracDataInterface`: silently hardcodes `/1000` (milliseconds), no override

I will do those changes later.

PR #1817, opened 2026-07-24
Add DoricCSVEventsInterface for discrete events from Doric Neuroscience Studio CSV exports

h-mayorquin:
This adds `DoricCSVEventsInterface`, which reads the digital I/O lines of a DoricStudio CSV export and writes one `EventsTable` per line. I kept it separate from the existing `DoricEventsInterface` because the CSV export is a genuinely different container: the `.doric` HDF5 stores events as an explicit list, while the CSV is a grouped two-row header over densely sampled `0`/`1` traces on a shared `Time(s)` clock, so the two share no reading code. Each digital line is edge-detected into events, and a line that never toggles is skipped so no empty table reaches the writer.


src/neuroconv/datainterfaces/events/doric_events/doriccsveventsdatainterface.py
```diff
@@ -0,0 +1,192 @@
...
+            Path to the DoricStudio CSV export.
+        event_specs : dict, optional
+            Per digital line, how its transitions become events, keyed by the line's column name
+            (e.g. ``{"DI/O-1": {"detect": "high_period"}}``). ``detect`` is one of ``"rising"`` /
+            ``"falling"`` (a point event at each edge) or ``"high_period"`` / ``"low_period"`` (a
+            durative event, onset at one edge and duration to the next opposite edge), default
+            ``"high_period"`` (lossless for an active-high line; use ``"low_period"`` for an
+            active-low line). If None (default), every digital line in the file is read as a
+            ``high_period``. When given, only the named lines are read (selection by inclusion).
+        metadata_key : str, optional
```
pauladkisson, line 58: Shouldn't this be dynamically generated from the file path or something so that two Doric CSV events data interfaces don't collide?

pauladkisson (approved):
Code looks good, and I confirmed that it matches Guppy's output, but see https://github.com/LernerLab/GuPPy/issues/398

src/neuroconv/datainterfaces/events/doric_events/doriccsveventsdatainterface.py
h-mayorquin: Good point. The tradeoff is readability vs metadata clashes. Do you think that users are likely to instantiate two doric csv interfaces in the same session? if that is the case, I think we should add the stem to avoid the collision, if this is unlikely, then we should keep it like this for readability on the metadata edition.

What do you think?

pauladkisson: I think it's unlikely but possible.

pauladkisson: Across the repo, I feel that we are a little all-over-the-place on the default value of the metadata_key. But that's ok, not everything has to be perfectly standardized.

h-mayorquin: The purist in me would like to make these as unique as possible, but a couple of things push the other way:

- Reliable unique-per-format keys are hard: a few formats have an obvious handle (file stem for generic CSV, channel for multi-channel ophys), but most would need a scheme invented per interface.
- A derived key becomes a contract the moment users write their metadata edits against it, so if the scheme later proves wrong and has to change, they break; a generic constant never sets that trap.
- So a derived default earns its place only when the format inherently yields several instances at once (SpikeGLX streams, multi-channel ophys); for a single-export format like Doric CSV, a fixed constant plus an explicit `metadata_key` for the rare multi-file case is the better trade.


I do agree that things are all over the palce and this prompt to try to document this https://github.com/catalystneuro/neuroconv/pull/1822 so we can start unifying it slowly but surely. Let me know what you think.

PR #1820, opened 2026-07-24
Add time_unit to CSVFiberPhotometryInterface

pauladkisson:
Adds a `time_unit` argument (`seconds`, `milliseconds`, `microseconds`) to `CSVFiberPhotometryInterface` and `MultiFileCSVFiberPhotometryInterface` that scales the timestamps column to seconds on read, mirroring what #1809 did for `CSVEventsInterface`.

This is a blocker for #1756, which has some data in milliseconds.


h-mayorquin (approved):
LGTM

PR #1821, opened 2026-07-24
Add durative event support to DoricEventsInterface

h-mayorquin:
This has #1817 (the Doric CSV events interface) as its base and reuses that PR's machinery, `discretize_trace`, `validate_event_specs`, and the `event_specs` shape.

On top, it teaches the modern `.doric` HDF5 interface to read each digital line through a per-line `event_specs` (`detect` one of `rising`/`falling`/`high_period`/`low_period`), so lines become durative events with a duration. The default is now `high_period` (previously rising-edge onsets only; `.doric` is unreleased, so no deprecation), matching the CSV interface.


pauladkisson (approved):
LGTM!

PR #1823, opened 2026-07-26
Read the legacy "EPConsole" `.doric` layout in `DoricEventsInterface`

h-mayorquin:
Doric has shipped two HDF5 layouts over the years and `DoricEventsInterface` only read the modern one. The older "EPConsole" files put everything under a `Traces` root, with each stream nested as `Traces/<console>/<stream>/<stream>` and the console's clock living in a sibling `Time(s)` group, so there is no `DigitalIO` group to walk and the digital lines are instead the streams whose name starts with `DI--O`. This adds a reader for that shape and picks between the two by looking at the root group, which is enough to tell them apart and costs no data read.


src/neuroconv/tools/events.py
```diff
@@ -1,14 +1,26 @@
 """Tools for the discrete-events data interfaces."""
 
-# A detection spec holds the reading and, optionally, a name for the event type it produces. The
-# signal-conditioning vocabulary (bits, thresholds, binarize, and the deferred hysteresis and debounce)
-# is not here yet: a Doric digital line is already a 0/1 signal, so no conditioning arises for it. Those
-# keys land with the first interface that needs them, alongside the shared mock.
-_SPEC_KEYS = ("detection", "event_name")
+# TODO before the next release: make validate_detection_configuration and resolve_detection_plan
+# private. Nothing here has shipped in a tagged release, so the rename costs no deprecation, and
+# everything in this module is interface-internal rather than a surface anyone should import.
```
pauladkisson: Why are we delaying this?

src/neuroconv/tools/signal_processing.py
```diff
@@ -61,62 +61,241 @@ def get_falling_frames_from_ttl(trace: np.ndarray, threshold: float | None = Non
     return falling_frames
 
 
-# NOTE: copied from the Intan digital PR (#1812) to unblock the Doric CSV events interface; de-duplicate
```
pauladkisson, line 64: This is a lot of complicated signal detection logic. We really need all of this to just detect the events from some TTLs?

src/neuroconv/datainterfaces/events/doric_events/doriccsveventsdatainterface.py
```diff
@@ -1,13 +1,20 @@
...
 
-import numpy as np
 from pydantic import FilePath, validate_call
 
 from neuroconv.utils import DeepDict
 
 from ..baseeventsinterface import BaseEventsInterface, _EventsData
-from ....tools.events import resolve_detection_plan, validate_detection_configuration
-from ....tools.signal_processing import discretize_trace
+from ....tools.events import (
```
pauladkisson, line 9: Why exactly are we updating the Doric CSV Events interface in this PR?

pauladkisson (commented):
No hard blockers, just some clarifications. Might want to split out the CSV stuff for easier review.

src/neuroconv/tools/events.py
h-mayorquin: I am doing the making it private now.

src/neuroconv/tools/signal_processing.py
h-mayorquin: Yes, it does not belong here, I removed it.

src/neuroconv/datainterfaces/events/doric_events/doriccsveventsdatainterface.py
h-mayorquin: The signal processing path is now divided between condition a signaling and detecting edges. The previous method was removed and that's why there needs to be changes here.

h-mayorquin:
Let me decouple this from the pipeline change so the diff is easier.

src/neuroconv/tools/signal_processing.py
h-mayorquin: removed it, only the detect events remains.

src/neuroconv/tools/events.py
h-mayorquin: removed the unnecessary diff

h-mayorquin:
@pauladkisson simplify the diff, delayed the machinery to other PR.

pauladkisson (approved):
Looks good, thanks for the simplifications.

PR #1824, opened 2026-07-26
fix: reject duplicate device registry names

### LINKED ISSUE #1782: Shared device registry silently reuses a same-named device/model instead of erroring on conflicting metadata
While planning the migration of the fiber photometry interfaces onto the top-level device registry added in #1780 (`metadata["Devices"]` / `metadata["DeviceModels"]`, written via `_add_device_to_nwbfile` / `_add_device_model_to_nwbfile`), I noticed the shared helpers drop a safety check that the fiber-photometry-specific device path they're replacing had.

The shared helpers are idempotent purely by NWB `name`: if a device (or model) of that name is already on the file, they return it unchanged and never look at the incoming metadata.

`_add_device_model_to_nwbfile` (`_metadata_and_file_helpers.py:225-227`):
```python
model_name = device_model_metadata["name"]
if model_name in nwbfile.device_models:
    return nwbfile.device_models[model_name]
```
`_add_device_to_nwbfile` (`_metadata_and_file_helpers.py:276-278`):
```python
device_name = device_metadata["name"]
if device_name in nwbfile.devices:
    return nwbfile.devices[device_name]
```

The fiber photometry path being replaced did more. `_assert_metadata_matches_existing` compared the scalar fields of the incoming metadata against the already-added object and raised if they disagreed — so declaring the same-named shared object with *different* contents in two interfaces was a loud error rather than a silent merge. That behavior is even written up as a design principle in the fiber photometry metadata dev guide ("Declaring the same-named shared object with different contents in two interfaces is a loud error rather than a silent merge, so an authoring mistake surfaces immediately").

So in a multi-interface conversion where two interfaces each contribute a device named `"optical_fiber"` but with a different `numerical_aperture` or `manufacturer`, the shared registry now keeps whichever was written first and silently discards the second's metadata. An authoring mistake that used to fail immediately passes quietly.

Repro:
```python
metadata = {"DeviceModels": {
    "a": {"type": "OpticalFiberModel", "name": "fiber_model", "numerical_aperture": 0.48},
    "b": {"type": "OpticalFiberModel", "name": "fiber_model", "numerical_aperture": 0.22},
}}
_add_device_model_to_nwbfile(nwbfile, metadata=metadata, metadata_key="a")
_add_device_model_to_nwbfile(nwbfile, metadata=metadata, metadata_key="b")
# nwbfile.device_models["fiber_model"].numerical_aperture == 0.48 — "b" silently ignored
```

stanbot8:
Different metadata keys can silently merge devices that use the same NWB name. Reject duplicate device and device model names across metadata keys before the converter adds an NWB object. Closes #1782.


h-mayorquin:
Hi, thanks, needs a changelog.

stanbot8:
@h-mayorquin Thanks for the review. I added the changelog entry under v0.9.4.

h-mayorquin:
Hi, thanks for your contribution. 

There are problems with our CI and secrets but I tested this locally and it works. I will merge and re-open some changes on top of this to test the whole machinery to be sure about it.

PR #1828, opened 2026-07-27
Fix the daily dev workflows

h-mayorquin:
Three unrelated failures on the same job:

- **probeinterface was never actually installed from main.** pip skips a git install when the clone declares a version it already has, and probeinterface main declares the released `0.3.2`, so the job kept testing the PyPI wheel while spikeinterface main blew up against it. I put `--force-reinstall --no-deps` on every dev step, not just that one, since the others work only by the accident that their projects bump their dev versions.
- **SpikeInterface made the segment time attributes private**, so the two reads of `_recording_segments[segment_index].t_start` now call the public `get_start_time`, available well before our minimum pin. Invisible on released versions today, a failure across every ecephys test once the rename ships.
- **dandi dropped Python 3.10**, and its dev install was unconditional, so the whole 3.10 leg died before a single test ran. That step is now guarded.

I expect the 3.13 leg to surface real failures rather than go green, since this is the first time in four months that probeinterface main is genuinely installed. Ordinary PR checks cannot validate any of this; the workflow only runs on a schedule or a manual dispatch.


pauladkisson (commented):
This is failing on both Python 3.10 and 3.13. https://github.com/catalystneuro/neuroconv/actions/runs/30406105151/job/90431812924

h-mayorquin:
> This is failing on both Python 3.10 and 3.13. https://github.com/catalystneuro/neuroconv/actions/runs/30406105151/job/90431812924

I tried to fix some others but there might be some which require releases.

pauladkisson:
> > This is failing on both Python 3.10 and 3.13. https://github.com/catalystneuro/neuroconv/actions/runs/30406105151/job/90431812924
> 
> I tried to fix some others but there might be some which require releases.

So I would find this PR more complete if the dev dailies actually pass when it's all said and done, which requires fixing some variety of small bugs. But if you'd rather focus on just the --force-reinstall update, just say the word and I'm happy to approve.

h-mayorquin:
> > > This is failing on both Python 3.10 and 3.13. https://github.com/catalystneuro/neuroconv/actions/runs/30406105151/job/90431812924
> > 
> > 
> > I tried to fix some others but there might be some which require releases.
> 
> So I would find this PR more complete if the dev dailies actually pass when it's all said and done, which requires fixing some variety of small bugs. But if you'd rather focus on just the --force-reinstall update, just say the word and I'm happy to approve.

I am aiming to fix them now. I will report if there is a really hard block or if they are done. Thanks.

h-mayorquin:
@pauladkisson I think this fixes the errors now.

pauladkisson (approved):
Confirmed that the daily tests are now passing. https://github.com/catalystneuro/neuroconv/actions/runs/30638946649

h-mayorquin:

PR #1830, opened 2026-07-28
Extend CSVFiberPhotometryInterface's column demux for multi-value labels and startup rows

pauladkisson:
This PR extends the CSVFiberPhotometryInterface to accommodate a couple new cases in the NPM data format.

<details>

While adding `NPMFiberPhotometryInterface` in #1756, I hit two places where the column demux in `CSVFiberPhotometryInterface` didn't fit the raw Neurophotometrics files. NPM's `LedState`/`Flags` column isn't a plain channel label — it's a packed bit word, where the three lowest bits are the excitation LEDs and higher bits are TTL/digital lines. That breaks both assumptions the column demux was built on:

- **One label value per channel.** Rows that share an excitation LED but differ in a digital-line bit carry different values (`17` and `273` are both 415 nm), so matching a single raw value returns only part of a channel.
- **A startup frame is excluded by not matching any interface's value.** NPM writes an initialization frame with *every* excitation bit set, so it matches whichever wavelength is asked for and lands in every channel.

So this PR changes `ColumnDemux`:

- `value` is renamed `values` and may now be a list, reading the rows whose label column matches any of the listed values. A single value still works.
- `skip_rows` drops leading rows before the label is consulted.

Both are generic to interleaved CSVs, so they're split out here rather than bundled into #1756, which they block. `CSVFiberPhotometryInterface` is unreleased (still under `v0.9.4 (Upcoming)`), so the rename needs no deprecation path.

</details>


pauladkisson:
The tests are failing due to AxonConverter, which is unrelated.

h-mayorquin:
> The tests are failing due to AxonConverter, which is unrelated.

Yes, my bad, I will fix that.

h-mayorquin (commented):
Looks Ok to me but there is no dedicated test to values and the read_kwargs already has the skip_rows functionality.

src/neuroconv/datainterfaces/fiber_photometry/csv/csvfiberphotometrydatainterface.py
h-mayorquin: this docstring is outdated.

docs/conversion_examples_gallery/fiberphotometry/csv_fp.rst
```diff
@@ -28,9 +28,12 @@ Some CSVs **interleave** the excitation channels frame-by-frame down the rows ra
 channel its own column, so one row is one channel at one timepoint. Pass a ``demux_config`` to read a
 single channel out of such a file. There are two shapes, chosen by ``by``:
 
-- ``{"by": "column", "column": ..., "value": ...}`` when a column labels each row's channel (e.g. a
-  Neurophotometrics ``LedState``): reads the rows whose ``column`` equals ``value``. A startup frame is
-  excluded simply by not being any interface's ``value``.
+- ``{"by": "column", "column": ..., "values": ...}`` when a column labels each row's channel (e.g. a
+  Neurophotometrics ``LedState``): reads the rows whose ``column`` equals ``values``. Pass a list for
```
h-mayorquin: whose column is in values?

src/neuroconv/datainterfaces/fiber_photometry/csv/_demux.py
```diff
@@ -11,16 +11,25 @@
...
+    A startup frame carrying a label of its own is excluded for free, by not matching any interface's
+    ``values``. ``skip_rows`` is for the case where that is not enough: it drops leading rows before
+    the label is consulted, for a startup frame whose label *does* match (e.g. an NPM initialization
+    frame written with every excitation bit set, which matches every wavelength).
     """
 
     by: Literal["column"] = "column"
     column: str | int
-    value: str | int
+    values: str | int | list[str | int]
```
h-mayorquin: we should not accept values = []

src/neuroconv/datainterfaces/fiber_photometry/csv/csvfiberphotometrydatainterface.py
pauladkisson: Fixed

docs/conversion_examples_gallery/fiberphotometry/csv_fp.rst
pauladkisson: I tried to clarify this. It still might not be perfect, but I think it's clear enough.

src/neuroconv/datainterfaces/fiber_photometry/csv/_demux.py
pauladkisson: Fixed and tested.

pauladkisson:
> Looks Ok to me but there is no dedicated test to values and the read_kwargs already has the skip_rows functionality.

I added a dedicated test for values.

I guess you could do it with `read_kwargs`, but `skip_rows` is more convenient, and it's already present on the `StrideDemux` model, so I don't see why we would have it on one and not the other.

h-mayorquin:
> > Looks Ok to me but there is no dedicated test to values and the read_kwargs already has the skip_rows functionality.
> 
> I added a dedicated test for values.
> 
> I guess you could do it with `read_kwargs`, but `skip_rows` is more convenient, and it's already present on the `StrideDemux` model, so I don't see why we would have it on one and not the other.

On `read_kwargs`, you can ignore what I said, keep `skip_rows`. I hadn't thought it through: the NPM wrapper works out whether row 0 is a startup frame by reading the state column, and `read_kwargs` is applied to every read the interface makes, so putting the skip there would remove the row before it can be inspected.

tests/test_modalities/test_fiber_photometry/test_csv_fiber_photometry_interface.py
```diff
@@ -149,11 +149,74 @@ def test_column_demux_reads_one_labeled_channel(self, tmp_path):
...
+        pd.DataFrame({"timestamps": TIMESTAMPS, "LedState": 1, "data": SIGNAL_DATA}).to_csv(path, index=False)
+        with pytest.raises(ValidationError, match="at least 1 item"):
+            CSVFiberPhotometryInterface(
+                file_path=path,
+                data_columns="data",
+                timestamps_column="timestamps",
+                demux_config={"by": "column", "column": "LedState", "values": []},
+            )
+
+    def test_column_demux_skips_leading_rows(self, tmp_path):
```
h-mayorquin, line 195: `test_column_demux_skips_leading_rows` doesn't exercise `skip_rows` for the reason it says: `LedState == 3` only occurs on row 0, so `values=1` with no `skip_rows` gives the same data and timestamps back. It goes red if you delete the `.iloc`, but only because the test puts `3` in `values`, which no caller would. If `3` recurs as a real frame in the cycle, both the list and the skip become necessary for the right reasons.

h-mayorquin (changes requested):
One last request to double check on a test and should be ready to go.


One more, and it's my own to fix: I named it `demux_config` in #1808 and it should have been `demux_configuration`, since that's how neuroconv spells its own configuration objects everywhere else, `backend_configuration`, `detection_configuration`, `sorting_configuration`, `dataset_io_configuration`. It's free because the interface is still unreleased, and since you're building the consumers right now it's one rename instead of two. Happy either way: do it here if it's convenient, or I'll do it in a follow-up and you can review. Same for the type alias, `DemuxConfig` to `DemuxConfiguration`.

tests/test_modalities/test_fiber_photometry/test_csv_fiber_photometry_interface.py
pauladkisson: Fixed and simplified. There wasn't any real need to pass in a list of values when a single value exercises the skip-rows functionality perfectly adequately.

pauladkisson:
> One more, and it's my own to fix: I named it `demux_config` in #1808 and it should have been `demux_configuration`, since that's how neuroconv spells its own configuration objects everywhere else, `backend_configuration`, `detection_configuration`, `sorting_configuration`, `dataset_io_configuration`. It's free because the interface is still unreleased, and since you're building the consumers right now it's one rename instead of two. Happy either way: do it here if it's convenient, or I'll do it in a follow-up and you can review. Same for the type alias, `DemuxConfig` to `DemuxConfiguration`.

No problem, just implemented.

h-mayorquin (approved):
Thanks a lot.

PR #1832, opened 2026-07-28
Allow writing units with no electrodes to a units table that already has an electrodes column

h-mayorquin:
Appending units to a units table that already has an `electrodes` column, without giving electrode indices for them, failed with `TypeError: 'NoneType' object is not subscriptable` from deep inside hdmf. The column was simply left out of the `add_unit` call, so pynwb wrote `None` into the `DynamicTableRegion` and the bounds check tripped over it. There was already a `# TODO, fix electrodes` on the line that skipped the null value for that column, and this resolves it: those units now get an empty region, which is valid NWB and round-trips correctly. I kept the writer permissive here instead of raising, because units with no electrode association are ordinary in NWB, and from inside this function there is no way to tell a deliberate case from a naming mistake further upstream.

This came out of AllenNeuralDynamics/aind-ephys-pipeline#121 (@alejoe91, @CodyCBakerPhD), where units from a second probe are appended to a file whose electrodes table was written for the first one, so no indices can be resolved for them and the write dies on the second probe. The root cause there is upstream in how the device is named, so this does not fix their pipeline; it makes our writer produce a file rather than an error that points nowhere near the problem. Inspecting those units in memory before writing hits a separate hdmf bug, which I fixed in hdmf-dev/hdmf#1549, but nothing here depends on that landing. While I was in this branch I also changed the duplicate-unit-name lookup a few lines above to read the stored electrode indices directly, since it was building a pandas DataFrame of every column of the matching rows to compare one list of indices.

Alessio, Cody, if you have an idea on how to surface your error better on this side it would be good.


alejoe91:
Thanks @h-mayorquin 

LGTM! I'm wondering if the same issue would occur for any other `VectorIndex` column? E.g., what happens if I add a batch of units with waveform_mean/sd and then another batch when this is set to None? Is it worth adding these columns to the fix and extending the test?

BTW, the issue with the pipeline should be fixed here: https://github.com/AllenNeuralDynamics/aind-ephys-pipeline/pull/122

h-mayorquin:
Thanks @alejoe91, good instinct, and yes it happens. I went looking and it turned out to be three separate failures rather than one, so I extended the fix here instead of leaving it for later.

Your example does break, but not through the `VectorIndex` path we both assumed. `waveform_mean` and `waveform_sd` are added with `index=False`, so they are plain fixed shape columns, and the null value for an array-valued column was `np.array(np.nan)`, a 0-dimensional array that cannot sit next to rows of shape `(num_samples, num_channels)`. It dies in `np.append` rather than in the region bounds check. 

There is a genuine `VectorIndex` version of the bug though, and `electrodes` was only one instance of it: every ragged column had the same problem, which I confirmed with a ragged unit property and a ragged channel property. 

And there is a third one going the other way, adding units without waveforms and then a batch with them, where the column extending path allocated a flat array and broke on shape.

So rather than skipping `electrodes` by name I made that the general rule: a ragged column gets an empty row, and a fixed shape float column gets an array of the same shape filled with `NaN`. Testing `isinstance(column, VectorIndex)` also means we never read a sample value out of a `DynamicTableRegion`, which is what trips the hdmf bug in the first place, so that workaround now covers every ragged column rather than just this one. Four regression tests added. The diff has changed shape quite a bit since your review, so it is worth another look.

alejoe91 (approved):

PR #1840, opened 2026-07-30
Stop `_GuppyInterface` from deriving a `session_start_time`

### LINKED ISSUE #1839: GuPPy interface reads timeRecStart=0.0 as a real epoch, giving a 1970 session_start_time
Basically, the Guppy interface should not be specifying a session start time since the Guppy processing tool does not provide one.

<details>

While planning CSV acquisition support for `GuppyConverter` — TDT is wired up in #1783 and CSV is the next of GuPPy's four acquisition formats — I checked whether `_GuppyInterface` could read a real GuPPy CSV session unchanged. It can: `sample_data_csv_1` in `ophys_testing_data` constructs, produces metadata, and writes a full NWB file with no warnings. But the session start time comes out as the Unix epoch.

```python
from neuroconv.datainterfaces.fiber_photometry.guppy.guppydatainterface import _GuppyInterface

interface = _GuppyInterface(
    folder_path=OPHYS_DATA_PATH
    / "fiber_photometry_datasets"
    / "CSV"
    / "sample_data_csv_1"
    / "sample_data_csv_1_output_1"
)
interface.get_metadata()["NWBFile"]["session_start_time"]
# datetime.datetime(1970, 1, 1, 0, 0, tzinfo=datetime.timezone.utc)
```

`_read_time_correction` decides whether a session has an absolute start time by asking whether `timeRecStart` is present in `timeCorrection_<recording_site>.hdf5`:

https://github.com/catalystneuro/neuroconv/blob/main/src/neuroconv/datainterfaces/fiber_photometry/guppy/guppydatainterface.py#L410-L412

For this session the key *is* present and holds `0.0`:

```python
import h5py

with h5py.File(".../sample_data_csv_1_output_1/timeCorrection_region.hdf5", "r") as f:
    print("timeRecStart" in f, f["timeRecStart"][:])
# True [0.]
```

so `datetime.fromtimestamp(0.0, tz=timezone.utc)` yields 1970-01-01 rather than the "no absolute start time" path the code intends.

The comment on that line, and the `CSV/README.md` entry in `ophys_testing_data` ("CSV inputs carry no `timeRecStart`, so no embedded session start time"), both state the key is absent for CSV inputs. That is not what GuPPy writes — it writes `0.0` as the sentinel instead. `test_guppy_interface.py` exercises the absent case by deleting the key from a mock output folder, and `mock_guppy` always writes a real recording start, so no fixture in the repo produces the `0.0` case and it has gone unnoticed.

Scope: TDT sessions are unaffected. The tank supplies a real `session_start_time`, and `GuppyConverter` registers the GuPPy interface before the acquisition interfaces so that the acquisition's value wins the metadata merge. The impact is on formats whose acquisition interface embeds no recording start — CSV and NPM both document that `get_metadata` does not populate `session_start_time`, and Doric only supplies one for HDF5 files carrying a `Created` attribute. For those, GuPPy's value is the only one available, so the epoch reaches the written NWB file silently.

</details>

pauladkisson:
Fixes #1839

<details>

`_GuppyInterface.get_metadata()` no longer populates `NWBFile/session_start_time`. GuPPy writes `timeRecStart` unconditionally, as the raw store's first timestamp, so the "key is absent" guard was dead code — and the value is a clock origin rather than a recording start, absolute only when the acquisition's own timestamps happen to be. The acquisition interface or user metadata supplies the session start instead. `_read_time_correction` stops reading the key.

`generate_mock_guppy_output_folder` loses its `session_start_time` argument. It now builds a raw session-relative timebase and derives `timeRecStart`, `correctionIndex`, and `timestampNew` from it the way GuPPy does. It emits `timeRecStart = 0.0`, matching the real `sample_data_csv_1` output; `correctionIndex` now carries the real lights-on trim offset.

The test diff is pure deletion: the two `session_start_time` tests go, along with the parametrized `expected_session_start_time`. Nothing replaces them — the interface no longer has the behavior.

No expected value in the suite moved: `starting_time` is still exactly `1.0` and `rate` still `200.00000000000426`. Verified against real data — the CSV and TDT GuPPy sessions in `ophys_testing_data` both now report `None` (previously 1970-01-01 and 2018-10-30; TDT's real start comes from the acquisition interface).

Two things this PR does not touch: #1783 registers the GuPPy interface first with a rationale that cites GuPPy reporting its own `session_start_time`, which needs revisiting on its next rebase; and `ophys_testing_data`'s `CSV/README.md` states CSV inputs carry no `timeRecStart`, which is wrong and needs its own PR.

</details>


h-mayorquin (approved):
I read the issue title only and that makes sense to me but I have not reviewed this at all. If you read it yourself it is fine to me.

PR #1841, opened 2026-07-30
Fix EDFRecordingInterface silently discarding all subject metadata

AlexN-SFV:
## The bug

`EDFRecordingInterface.get_metadata` reads `subject_id` and `date_of_birth` out of the EDF header and then throws them away. No EDF conversion has ever carried the header's subject information into the NWB file.

```python
subject_metadata = self.extract_subject_metadata()
metadata.get("Subject", dict()).update(subject_metadata)   # <- updates a throwaway dict
```

`metadata` is a `DeepDict`, which subclasses `defaultdict`. `defaultdict` only auto-creates missing keys through `__missing__`, and `dict.get` deliberately does not go through `__missing__` — it returns the default argument instead. So whenever `"Subject"` is not already present, `.get("Subject", dict())` hands back the fresh throwaway `dict()`, `.update()` mutates that, and it is discarded when the expression ends. `DeepDict` overrides neither `get` nor `__missing__`, so nothing intervenes.

Reproduced on `main` with a file whose patient field carries a code:

```
extract_subject_metadata(): {'subject_id': 'MCH-42', 'date_of_birth': '02 may 1951'}
metadata["Subject"]      : {}
```

The extraction was always correct; only the assignment was broken. This is the only occurrence of the pattern in the codebase — elsewhere (e.g. the Inscopix interfaces) `metadata["Subject"]` is assigned directly, which is what this PR adopts.

## The fix

Assign the key instead of updating a copy, and default the schema-required `species`/`sex` fields. That last part is necessary rather than cosmetic: once `"Subject"` exists the metadata schema requires `subject_id`, `species` and `sex`, so simply un-dropping the data would make merely *reading a patient code out of a file* turn valid metadata into invalid metadata.

The header's birthdate is also parsed into a `datetime`, and dropped when it cannot be read. `Subject.date_of_birth` requires a `datetime`, while the readers return that field as a string (`"02 may 1951"` via pyedflib), so un-dropping it without parsing would have started **failing** conversions of every file that carries a birthdate — turning a silent omission into a hard error.

## Tests

`tests/test_modalities/test_ecephys/test_edf_interface.py` is new. The only existing EDF coverage lives in `tests/test_on_data`, which needs the gin dataset, so this bug had nothing guarding it. The fixtures are written with `pyedflib` — already a dependency of the `edf` extra — so the tests need no external data and no network.

**Five of the eight fail on `main` without the fix**, including an end-to-end conversion that reads the `Subject` back out of the written NWB file. The other three cover behaviour the bug did not affect (the `NWBFile` half of `get_metadata`, and a file with no patient information, which must *not* gain a placeholder `Subject`).

One note for reviewers, since it looks like an inconsistency: the metadata schema types `Subject.date_of_birth` as a date-time **string**, while pynwb requires a `datetime` **object**. Those are reconciled by `validate_metadata`, which runs the metadata through `_NWBMetaDataEncoder` before validating — exactly as already happens for `session_start_time`. The schema-validity test therefore uses that helper rather than `jsonschema.validate` directly.

## Behaviour change

`Subject` metadata now appears for EDF files whose header carries a patient code or birthdate, where previously it was absent. Files with no patient information are unaffected and gain no placeholder subject.

`date_of_birth` is a naive `datetime`, because the EDF header carries no timezone. pynwb localises naive datetimes to the converting machine's zone and warns — the same thing it already does with `session_start_time` from an EDF header, for which the documented pattern is that the user attaches `tzinfo` themselves. I have not invented a timezone the file does not contain.

## Notes

- Two commits: the fix, then the tests, so the failing-before/passing-after split is easy to check.
- The changelog entry has no PR link yet; happy to add it once this has a number.
- Found while adding discontinuous-EDF+ support, which is a separate PR. This one is independent of it and stands alone.


h-mayorquin:
Thanks for this, yes, I reproduce the bug so that would be good to fix.  I would like the PR to be just about that fix as there are many things here besides that:

* The sex which could be its own feature
* Addition of placeholders, they should not be added here. See [Metadata Principles](https://neuroconv.readthedocs.io/en/main/developer_guide/metadata_principles.html).
* There is a lot of validation which is the job of the [NWB Inspector](https://nwbinspector.readthedocs.io/en/dev/best_practices/nwbfile_metadata.html). It already owns both of the checks here: `check_session_start_time_old_date` and `check_session_start_time_future_date` do exactly what the 1850 floor and the future-date rejection do, just at the layer where a user can see and act on the result.


On the birthdate I'm undecided, so before anything else: how and where did you hit the failure?

On testing, no EDF on gin carries a patient code or a birthdate today, which is a good part of why this went unnoticed, so I opened [ephy_testing_data#183](https://gin.g-node.org/NeuralEnsemble/ephy_testing_data/pulls/183) adding a file with every header field the format defines filled in. Once it merges we can test against that instead of writing EDFs inside the test file.

h-mayorquin:
This will require this:
https://github.com/catalystneuro/neuroconv/pull/1845

h-mayorquin:
I merged the part of this that did not depend on an open question as #1929:
https://github.com/catalystneuro/neuroconv/pull/1929

It assigns `metadata["Subject"]` instead of updating the throwaway dict, reads the patient field's sex, and writes `experimenter` as a list. You are co-author on the commit. The tests run against `edf/metadata_annotations/full_metadata.edf`, which is on gin now, so there is no need to write EDFs with `pyedflib` inside the test file.

That leaves the birthdate here, and my question from before still stands: how and where did you hit the failure? I know it throws an error, as the readers return `17 mar 1985` and `make_nwbfile_from_metadata` parses ISO 8601 only. What I don't know is which form your own file gave you, and that decides how narrow the parse can be.

h-mayorquin:
@AlexN-SFV the birthdate is parsed now in #1999, so my question above isn't needed anymore.

Everything this PR fixed is released: #1929 went out in v0.10.0 and #1999 in v0.10.2. Closing in favor of those two (thanks for finding the bug!). Feel free to open an issue if an EDF file of yours still loses any of its subject information.

Issue #1797, opened 2026-07-20
[New Format]: Ponemah, Spike2


AudreyMagsig:
### What format would you like to see added to NeuroConv?

Ponemah Files: https://www.datasci.com/products/software/ponemah/

I'd also use the Spike2 add-on, but it doesn't look supported per the note. 

### Does the format have any documentation?

Ponemah Files: https://www.datasci.com/products/software/ponemah/

### Existing APIs for format

Not known. Spike2 has a converter, but I've found that the import using the SonPy library sometimes deletes data drops, so I've been exporting from Spike2 as a spreadsheet text file.

### Do you have any example files you are willing to share?

Yes!

### Do you have any interest in helping implement the feature?

Yes.

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/catalystneuro/neuroconv/blob/main/.github/CODE_OF_CONDUCT.rst)
- [x] Have you ensured this bug was not already [reported](https://github.com/catalystneuro/neuroconv/issues)?

h-mayorquin:
Thanks @AudreyMagsig,

On Ponemah: this would be a from-scratch reader for a proprietary format. We couldn't find an existing open reader (in neo or elsewhere) to build on, so it's substantial work, and honestly we don't have the resources to take on a whole new reader right now. If you can share example files, the place I'd start is python-neo, since that's the layer neuroconv reads formats through. Once a Ponemah `rawio` exists there, wiring it into neuroconv is easier and that would also benefit the community. The neo rawio overview is here: https://neo.readthedocs.io/en/latest/rawio.html, and the template to copy is `examplerawio.py`: https://github.com/NeuralEnsemble/python-neo/blob/master/neo/rawio/examplerawio.py. On the neuroconv side, the guide for building a data interface is here: https://neuroconv.readthedocs.io/en/main/developer_guide/build_data_interface.html, and the general contributing guide is here: https://neuroconv.readthedocs.io/en/main/developer_guide/contributing.html. I maintain neo, so I'm happy to advise both here and there.

On Spike2: the CED reader (`sonpy`) only ships a working wheel for Python 3.14 on Linux/macOS, so we expect to recover Spike2 support once neuroconv supports Python 3.14. That work is tracked in https://github.com/catalystneuro/neuroconv/pull/1542, currently blocked by zarr problems.

AxelNoun:
Status update on both halves of this, since they turned out to be quite different problems.

## Spike2

[...]

PR #1855, opened 2026-08-01
Fix the spike2 extra: the sonpy marker can never be satisfied

AxelNoun:
Refs #1797. Depends on NeuralEnsemble/python-neo#1890 and the PR that closes it.

### The problem

`pyproject.toml` declares the extra as:

```toml
"sonpy>=1.7.1; python_version=='3.9' and platform_system != 'Darwin'",
```

while the project sets `requires-python = ">=3.10"`. The marker can never be satisfied, so
`pip install "neuroconv[spike2]"` installs no sonpy on any supported Python — while
`docs/conversion_examples_gallery/recording/spike2.rst` still recommends exactly that command.

The marker was correct when written: sonpy used to ship one `py3-none-any` wheel per Python
minor version (1.7.1 was `>=3.7,<3.8`, 1.9.5 was `>=3.9,<3.10`), so 3.9 really was the only
target. It became inert when 3.9 was dropped.

### What changed upstream

sonpy 1.9.12 (2026-06-09) moved to per-interpreter binary wheels:

| platform | wheels available |
|---|---|
| Windows `win_amd64` | cp39 – cp314 |
| Linux `manylinux_2_39_x86_64` | cp314 only |
| macOS `macosx universal2` | cp314 only |

The source distribution is not a fallback: it ships a Windows `.pyd`
(`PE32+ executable (DLL) […] for MS Windows`), so building it on Linux produces an
unimportable package.

### The changes

**1. `pyproject.toml`** — target the combinations that actually have a usable wheel:

```toml
"sonpy; platform_system=='Windows' or python_version>='3.14'",
```

Verified against the real wheel matrix: the marker evaluates `True` in exactly the 15
platform/version combinations where a wheel exists, and `False` in every other.

**2. `_test_sonpy_installation()`** — the current exclusion list encodes sonpy's old
distribution shape (`excluded_python_versions=["3.10", "3.11"]` plus a darwin/arm carve-out).
It now excludes 3.10 – 3.13 on Linux and macOS, and nothing on Windows, matching the table
above. The macOS wheel is `universal2`, so the previous arm-specific special case is no
longer needed.

**3. The two assertions in `tests/test_modalities/test_ecephys/test_ecephys_interfaces.py`**
asserted the Python-version error messages, which no longer fire on Windows. Replaced with a
single platform-aware test that asserts the platform-specific message and skips on Windows
and on Python 3.14+.

### A third dead guard, left for a follow-up

`tests/test_on_data/ecephys/test_recording_interfaces.py` guards the real data test with:

```python
platform == "darwin" or this_python_version > version.parse("3.9")
```

With `requires-python = ">=3.10"` the second clause is always true, so
`TestSpike2RecordingInterface` never runs either — the same dead-guard pattern, from the same
cause. That is why the three changes above keep CI green: nothing currently exercises the
reader, on any platform.

Restoring it is a one-line change (`platform != "win32" and this_python_version <
version.parse("3.14")`), but it cannot land yet. Once that test runs on `windows-latest`,
`Spike2RecordingInterface` reaches Neo's `CedRawIO`, which is broken against sonpy >= 1.9.12:
it accesses `sonpy.lib`, a namespace that release removed. See
NeuralEnsemble/python-neo#1890.

I have the change ready and will open it as a follow-up once the Neo fix is released, together
with a bump of `neo>=0.14.5` in `ecephys_minimal` to that version. Happy to fold it in here
instead if you would rather keep it together.

### Verification

- The new marker was evaluated across all 15 platform/version combinations in neuroconv's
  support range and matches sonpy's published wheel matrix exactly.
- CI impact traced per job: on ubuntu and macOS the marker keeps sonpy uninstalled, the
  platform exclusion fires and the rewritten test asserts the resulting message; on Windows
  sonpy installs, the import succeeds and the test skips. No job changes colour.
- The `ModuleNotFoundError` message produced by the new exclusion list was checked
  character-for-character against what the rewritten test asserts.
- `black --line-length 120 --check` clean on both changed Python files.

### Open question

Should `docs/conversion_examples_gallery/recording/spike2.rst` gain a note that Spike2 is
Windows-only below Python 3.14? Happy to add it here or leave it for a docs PR.


h-mayorquin:
Needs a changelog.

h-mayorquin:
Otherwise it looks good to me.

AxelNoun:
Added the changelog entry, thanks for the review.

Still unsure about my open question above: should `spike2.rst` gain a
note that the extra only resolves on Windows, or on Python 3.14+ for
Linux and macOS? I have it written and can push it here in a second, or
leave it for a docs PR whichever you prefer.

As far as I can tell the remaining red checks are the live-service jobs
and the project-board automation, which need secrets unavailable to fork
PRs; the changelog check should go green with this push.

h-mayorquin:
> Added the changelog entry, thanks for the review.
> 
> Still unsure about my open question above: should `spike2.rst` gain a note that the extra only resolves on Windows, or on Python 3.14+ for Linux and macOS? I have it written and can push it here in a second, or leave it for a docs PR whichever you prefer.
> 
> As far as I can tell the remaining red checks are the live-service jobs and the project-board automation, which need secrets unavailable to fork PRs; the changelog check should go green with this push.

Yes, I think a note is good.

AxelNoun:
@h-mayorquin Changelog added, and the branch is now up to date with main.

The four red checks are the live-service tests. they fail at the credential
validation step because the secrets aren't available to PRs from forks. All
required checks are green.

Unrelated heads-up: there's an existing Bug Fixes entry on main (the
MiniscopeConverter legacy-list one) that also links PR #1855 Looks like a wrong
number upstream. Happy to leave it alone, just flagging since my entry now makes
two #1855 links in the same file.

Also still curious about the docs question above: should spike2.rst mention the
platform/Python constraint on sonpy, or keep that out of the format gallery? Same
answer would apply to the Neo follow-up I'm planning.

h-mayorquin:
This is ready to go. 

Regarding your questions: 
1) Separate PR for the Neo follow-up, no need to fold it in here.
2) I answered the docs question a few comments up so it probably got lost, but yes to the note. The one in `spike2.rst` right now says Spike2 is not supported at all as `sonpy` only works on python 3.9, so that is the one to replace.

h-mayorquin:
I did the comment bit on the rst myself, this will merge after the tests pass.

h-mayorquin (approved):
LGTM

AxelNoun:
> This is ready to go.
> 
> Regarding your questions:
> 
>     1. Separate PR for the Neo follow-up, no need to fold it in here.
> 
>     2. I answered the docs question a few comments up so it probably got lost, but yes to the note. The one in `spike2.rst` right now says Spike2 is not supported at all as `sonpy` only works on python 3.9, so that is the one to replace.

Thanks for the review!

I'll open the Neo follow-up once NeuralEnsemble/python-neo#1890 ships in a release, re-enabling `TestSpike2RecordingInterface` and bumping `ecephys_minimal`'s `neo` floor to that version, as noted above. One correction to that note: the dead guard actually appears twice in `test_recording_interfaces.py` (the class decorator and the `spike2` case in `test_metadata_key_is_accepted_and_forwarded`), so I'll fix both together. Will link the PR here and on #1797 once it's ready.

PR #1878, opened 2026-08-05
Preserve extractor property order in the electrodes and units tables

### LINKED ISSUE #792: Preserve order of recording extractor custom properties in `add_electrodes()`

adityasingh2400:
Fixes #792

The issue is filed as "the order of properties is not preserved". Investigating it, the order is not merely different from the extractor's, it is not stable at all: the electrodes and units tables get a different column order in every Python process, so the same script over the same data writes a structurally different NWB file each run.

The cause is that `_add_electrodes_to_nwbfile` and `_add_units_table_to_nwbfile` build `data_to_add` as an insertion-ordered dict in a deliberate order, then convert it to a `set` and iterate the set at the point where the columns are actually created (`properties_to_add_by_columns - {"channel_name", "electrode_name"}` and `properties_to_add_by_columns - {"unit_name"}`). Set iteration order for strings follows their hashes, and Python randomizes string hashing per process, so the column order is randomized per process. Six consecutive runs of a recording with eight custom properties gave six different electrode table layouts, while the recording's own property order stayed constant.

The fix keeps the sets for membership tests and iterates `data_to_add` for the ordering, so the columns come out in the order the extractor reports its properties. The row-adding paths and the null-value lookups had the same problem and are now ordered by the existing table's `colnames`, which is already a tuple. There is no behavior change beyond the ordering, and nothing is added or dropped from either table.

This supersedes the stale draft #793. That PR was closed for inactivity rather than rejected, and CodyCBakerPhD said there that the approach of preserving insertion order was fine. His further suggestion of an explicit priority (NWB schema columns, then SpikeInterface columns, then custom ones) is deliberately not attempted here: it is a policy on top of a defined order, and currently there is no defined order to build it on. It can be layered on this without touching the call sites again.

Tested with `generate_recording` and `generate_sorting`, no acquisition files needed. Three tests were added, two of which assert an exact column order rather than the order-insensitive `assertCountEqual` that `test_default_electrode_column_names` uses, and one of which asserts that two NWBFiles built from the same recording agree. Against the unmodified source they fail on every hash seed:

```
-- PYTHONHASHSEED=0 --
FAILED ...::TestAddElectrodes::test_electrode_column_order_follows_recording_properties
FAILED ...::TestAddUnitsTable::test_units_column_order_follows_sorting_properties
2 failed, 1 passed
-- PYTHONHASHSEED=1 --  (same two failures)
-- PYTHONHASHSEED=2 --  (same two failures)
```

With the fix they pass on every seed, and the five seeds that previously produced five different layouts now produce one:

```
ELECTRODES: ['location', 'group', 'group_name', 'channel_name', 'electrode_name', 'electrode_type',
             'chamber_type', 'chamber_x', 'chamber_y', 'chamber_z', 'acx_x', 'acx_y', 'acx_z', 'rel_x', 'rel_y']
UNITS     : ['unit_name', 'amp_alpha', 'beta_snr', 'gamma_isi', 'delta_fr', 'epsilon_x', 'quality', 'spike_times']
```

Full suites are green: `tests/test_modalities/test_ecephys` gives 194 passed, 2 skipped (191 passed before the three new tests), and `tests/test_minimal` gives 517 passed, 72 skipped. `black`, `ruff` and `codespell` are clean on both changed files.


h-mayorquin:
Thanks for digging into this. The randomization per process is real and it is not what #792 reported (nice catch on the hash seeds!). I checked and I get the same thing: six runs give six different `colnames`.

Is determinism your use case here? If it is, #1888 should cover it, as it writes the properties in natsort order.

If what you want is the extractor's order itself then I am not taking that, as I don't want to couple the writing logic to the order spikeinterface happens to report properties in. My full reasoning is [on #792](https://github.com/catalystneuro/neuroconv/issues/792#issuecomment-5258284501).

adityasingh2400:
Yes, determinism is the use case. The extractor's order was how I got there rather than the goal, and your reasoning against coupling to it is fair. The point that spikeinterface does not offer that order as a contract is the one I would not have an answer to.

I checked #1888 against the other places I had reordered, to be sure the sorting covers all of it. It does. The two loops you sorted are the ones that call `add_column`, so they are where `colnames` actually comes from. The paths I also touched, `properties_requiring_null_values` and `properties_with_data`, only build `nul_values_for_rows` and `data_dict`, and both are consumed as row kwargs, so their iteration order never reaches the file. Sorting the two column loops is sufficient on its own.

Closing this in favor of #1888.

One thing you may want from here, your call. My tests include an order-agnostic case that writes two NWBFiles from the same recording and asserts the two `colnames` match. It catches the randomization without pinning an expected list, so it would survive a later change to the ordering policy. Happy to send it over to #1888 if that is useful. If not, the exact-order assertions you already have do catch the hash seed problem.

h-mayorquin:
No need I have some tests already for regression on this and I don't think that one works. Both calls run in the same process and `PYTHONHASHSEED` is fixed when the interpreter starts, so the two sets iterate the same way with or without the sort. I ran it against main on four seeds and it is green on all of them while the order is moving underneath:

```
seed 0: alpha_property, beta_property, zeta_property, mu_property
seed 2: beta_property, mu_property, alpha_property, zeta_property
seed 3: alpha_property, zeta_property, mu_property, beta_property
```

The exact order assertions do catch it as they fail on any seed once the `natsorted` comes out.

adityasingh2400:
You are right, and my own numbers said so if I had read them properly. The three tests I added ran `2 failed, 1 passed` against unmodified main, and the one that passed was that determinism test. It never caught the bug at all. `PYTHONHASHSEED` being fixed when the interpreter starts is the reason, so both writes in one process see the same iteration order regardless.

Thanks for running it across four seeds instead of taking my word for it. Your exact-order assertions are the right check here.

PR #1879, opened 2026-08-05
Support GuPPy sessions whose events come from several formats

### LINKED ISSUE #1876: GuppyConverter: support sessions whose photometry and events come from different formats
We also might want to support multiple different types of event interfaces so that TDT events can be converted alongside custom CSV events. Multiple different types of fiber photometry acquisition interfaces are probably unnecessary, though...

<details>

While adding the `GuppyConverter` in #1783, I wired the converter's format dispatch to a single
`acquisition_format` argument:

```python
AcquisitionFormat = Literal["tdt", "csv", "doric", "npm"]
```

That one value selects *both* halves of a GuPPy session — which interfaces read the raw photometry
traces out of `fiber_photometry_folder_path`, and which read the raw behavioral events out of
`events_folder_path`. GuPPy itself imposes no such coupling, and the mismatch is not exotic: GuPPy's
custom-event import writes the imported events back out as CSVs, sitting next to photometry that was
recorded in whatever the acquisition system produced. So a session with TDT (or Doric, or NPM) traces
and CSV events is a routine shape, and the converter currently has no way to express it — you must
pick one format and it is applied to both sides.

## Scope

The plumbing is already most of the way there. The two folder paths are separate constructor
arguments, and the three places that branch on the format each handle the two sides independently:

- `_build_acquisition_interfaces` — photometry only
- `_build_events_interfaces` — events only
- `_build_event_source_id_map` — events only

So the change is to split the single argument into an acquisition-side and an events-side format,
rather than to restructure the dispatch. Worth keeping a shorthand for the (still common) matched
case so existing call sites do not all have to name the format twice.

Two details that need care once the two sides can differ:

- The number of events interfaces is format-dependent — CSV fans out to one interface per store
  while TDT/Doric/NPM each build a single one — so interface naming has to stay stable across the
  combinations.
- `get_metadata` joins the event types an interface seeded against the stores listed in
  `storesList.csv` by name. NPM is the one format whose seeded ids differ from the store ids, which
  is what `_build_event_source_id_map` exists to fix up; that map has to follow the events-side
  format, not the acquisition-side one.

The real cost here is test data, not code: this needs a fixture session on gin whose photometry and
events genuinely come from different formats.

</details>

pauladkisson:
Fixes #1876 

<details>

Closes #1876.

The only cross-format case a GuPPy session actually has is the custom-event import, which writes the events it imported back out as one-column `timestamps` CSVs into the session folder. Those files are recognizable on sight — `is_event_csv` already tells them from acquisition CSVs, which is how the Doric and NPM readers decide what to look at — so `events_folder_path` is scanned for them and nothing has to be declared: an event store with a CSV of its own is read from it, and every other store falls to `acquisition_format`.

A store both sides carry would be written twice into the merged `EventsTable`, so whenever the two split a session between them the acquisition format is asked what it holds (`discover_event_store_ids`, one per `<format>_utils`) and an overlap is refused. That question is only asked when the acquisition side actually supplies something, so a session whose every event was imported never touches the acquisition format's event source at all.

Traces still come from a single `acquisition_format` — a series column-stacks one store per recording site onto a single timestamps vector, which stores from two acquisition systems do not share.

No API change and no renames: a mixed session registers the acquisition format's interface as `Events` and each imported store's as `Events_<store>`, exactly as a single-format session already did.

Verification: 6 new tests in `tests/test_on_data/fiber_photometry/guppy/test_mixed_formats.py`, covering a TDT tank plus an imported-event CSV in one folder — routing by discovery, both sources in the merged table, the GuPPy registry linking the imported event, a same-named CSV being refused, and a session whose every store was imported. 265 fiber photometry and 521 `tests/test_minimal` tests pass, pre-commit clean.

</details>


h-mayorquin:
If this is only for GuPPy's custom event importer I don't think we need `events_formats` at all. Can we just always probe for the CSVs on `events_folder_path` internally? I think you do this somewhere else right?

pauladkisson:
> If this is only for GuPPy's custom event importer I don't think we need `events_formats` at all. Can we just always probe for the CSVs on `events_folder_path` internally? I think you do this somewhere else right?

Yes, we can roll with that if you think that would be a cleaner implementation. Implementing...

h-mayorquin:
Yes, it is good to make less promises in the signature (the event folders is the same for both) and less for users to handle.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppyconverter.py
```diff
@@ -271,25 +286,113 @@ def _build_acquisition_interfaces(
...
+        if not event_store_ids:
+            return {}
+        if acquisition_format == "csv":
+            return {"csv": event_store_ids}
+
+        imported = csv_utils.discover_event_store_ids(folder_path=folder_path)
+        imported_store_ids = [store_id for store_id in event_store_ids if store_id in imported]
+        acquisition_store_ids = [store_id for store_id in event_store_ids if store_id not in imported]
+
+        if imported_store_ids and acquisition_store_ids:
```
alessandratrapani: I have a question on this:  each format's discover_event_store_ids seems to answer "what does the raw acquisition source physically contain" rather than "what does storesList.csv actually list for this session." Couldn't that cause a false conflict? For instance, a TDT tank or Doric file holding an unrelated epoc/digital line that just happens to share a name with a genuinely imported CSV store and trip the assert not `carried_by_both `on an otherwise valid conversion?

pauladkisson: Yes, that is. That would cause a conflict and is an edge case. Thank you for taking a look, by the way. I don't think you'd be able to get that kind of session through Guppy in the first place, though, since it collapses everything into a single list of store IDs, independent of the format.

pauladkisson: So really just a question of how defensive we want to be here.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppyconverter.py
```diff
@@ -271,25 +286,113 @@ def _build_acquisition_interfaces(
...
+        signal/control stores gets none at all.
+
+        A session drawing its events from both sides registers the acquisition format's interface as
+        ``Events`` and each imported store's as ``Events_<store>``, so the two never collide.
 
-        Between them the interfaces must cover **every** store in ``event_store_ids``. What they
-        *call* the types they seed is a separate question, answered by
-        :meth:`_build_event_source_id_map`.
+        Between them the interfaces cover **every** store GuPPy listed, which
+        :meth:`_build_event_store_ownership` is what guarantees. What they *call* the types they seed
```
h-mayorquin: Is this still true?

h-mayorquin (changes requested):
Two points

The three `discover_event_store_ids` on the non-CSV formats don't seem to have a use now, do they?  Doric's can't catch anything as `build_doric_events_interface` only reads the lines it is handed, so a store routed to `csv` never reaches it. TDT's and NPM's do guard a real double write, their interfaces seed every epoc and every label whatever we assign them. But `get_metadata` already builds `covered_stores` across the events interfaces, so a store covered by two of them is the same collision, found from what this session seeded instead of from what the files physically hold. Codecov is telling the same story, the patch is at 76% with those two files at 14% and 25% as only the TDT path is exercised.

One more, `csv_utils.discover_event_store_ids` should skip the `event*` files GuPPy derives for NPM. The scan is unconditional now so it runs on every NPM session, and it claims any single-column `timestamps` CSV. `npm_source_files` already excludes that same pattern for the acquisition side. I couldn't reproduce a live case, the current GuPPy demuxes in memory and neither NPM session I have here holds those files, so this one is insurance.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppyconverter.py
pauladkisson: Okay, so I ended up dropping the discover event store IDs function for all of the formats except for CSV, which makes the question considerably simpler. 

Essentially, if it shows up in CSV, that indicates that it is an imported CSV, which has priority. Otherwise, it defaults to the store IDs from the specified acquisition format.

src/neuroconv/datainterfaces/fiber_photometry/guppy/guppyconverter.py
pauladkisson: I think this comment is out of date.

pauladkisson:
> The three discover_event_store_ids on the non-CSV formats don't seem to have a use now, do they?

Indeed, removed.

> One more, csv_utils.discover_event_store_ids should skip the event* files GuPPy derives for NPM.

I believe `is_event_csv()` already skips any CSV that comes from Doric or NPM by construction.

pauladkisson:
I made a minor refactor to kind of clean up the events logic. It should be easier to follow now.

h-mayorquin (approved):
Thanks for addressing my points, this is good to go.

PR #1880, opened 2026-08-05
Support Standalone GuppyInterface for NWB-in, NWB-out in Guppy

### LINKED ISSUE #1877: GuppyConverter: support NWB as a GuPPy input data format
Alternatively, we could sidestep the interfaces and just add some utils for NWB, but one way or the other, we need to be able to pair it with Guppy processing output. 

<details>

The `GuppyConverter` added in #1783 covers four of the input data formats GuPPy reads — `tdt`,
`csv`, `doric`, and `npm` — each with a `<format>_utils` module and a branch in the converter's
format dispatch. GuPPy also reads NWB, and that one is not covered.

It is not simply a fifth branch. For the other four, the raw acquisition lives on disk in some
non-NWB format and the converter's job is well defined: read the traces and the events, write them
into a fresh NWB file alongside the GuPPy-derived outputs. When the source is *already* NWB, that
operation collapses, and it is genuinely unclear what the converter should do. At least three
readings, and they lead to quite different implementations:

1. **Re-read and re-write.** Read the series back out of the source NWB file and write them into the
   new one. Keeps the converter's contract identical to the other four formats, at the cost of
   duplicating the acquisition data.
2. **Append to the existing file.** Skip the acquisition interfaces entirely and add only the
   GuPPy-derived outputs and the two GuPPy registries into the source file. Avoids the duplication,
   but this is no longer a converter shape at all — there is no "conversion", and the
   `ConverterPipe` scaffolding does not fit it.
3. **Link rather than copy.** Reference the source file's series from the new file instead of
   copying the data.

A second question sits underneath all three: what layout the source NWB file must have for GuPPy's
own NWB reader to accept it, and correspondingly how the series in that file map onto the store ids
in `storesList.csv`. Whatever we build should also be checked to round-trip — a file neuroconv
writes should be readable back by that reader.

Deliberately leaving the choice open here; it wants a decision before any code.

Wherever it lands, the touch points are the same ones every format goes through: the
`AcquisitionFormat` literal, `_build_acquisition_interfaces`, `_build_events_interfaces`, and
`_build_event_source_id_map` — unless reading (2) wins, in which case this belongs somewhere other
than the converter.

</details>

pauladkisson:
Fixes #1877 

This PR makes one major design decision that is worth some discussion. The main question is how to combine input data (Fiber photometry + events) that is already in NWB with Guppy output data that is not yet in NWB.

So far I have considered 2 major options:
1. Creating a fresh NWB file with Fiber Photometry + events + Guppy output
2. Adding Guppy output to an existing nwbfile

This PR chooses option 2 for the following reasons:

1. Any data coming from the NWB file unrelated to guppy input (ex. trials table, ephys, etc.) is automatically included
2. Analyzing data from DANDI and then exporting it to NWB is DANDI-ready
3. Extending the existing `_GuppyInterface` into a public `GuppyInterface` requires minimal code changes

In addition, this PR adds an option for Guppy to write its analyzed events to the events table directly rather than going through a separate events interface in the case that the events registry and EventsTable are not already written. This allows the interface to support case and accurately link cases where NWB files already have events, but they're not formatted in the way that Guppy needs them to be, as well as the Guppy + CSV case.


h-mayorquin (changes requested):
Making `GuppyInterface` public as a way of achieving this seems OK to me but I think we need some changes.

We made the converter merge every event source into a single events table because we needed a single object to link to. That was fine as the converter was the only place that handled GuPPy objects publicly. Now we get two entry points for different purposes, the converter as the aggregator and the interface to add only GuPPy's output. I would like the interface to be consistent about what it writes, and some of your code here already hints at how.

Instead of merging the events from the acquisition sources, I think we should write GuPPy's own internal events as the table we link. That makes the interface do the same thing standalone and inside the converter, and gives us a stable surface for linking the PSTHs and the other objects that reference events into a single consistent structure. It also eliminates a lot of code (`resolve_event_store_rows`, `select_analyzed_rows`, `_resolve_event_rows` and the timestamp matching in `_build_guppy_registries`) and some assertions.

In principle this is a small duplication of the events info in the NWB file but I don't think the case holds. GuPPy drops the start, does its own alignment and writes onsets only, and that is fine. The acquisition tables are just the raw sources, so they come out the same whether people go through the converter or add the events interfaces themselves (which is an additional gain from this approach). The GuPPy onsets are a processing output and stay their own table, a stable scaffold that will always be available.

pauladkisson:
> Making `GuppyInterface` public as a way of achieving this seems OK to me but I think we need some changes.
> 
> We made the converter merge every event source into a single events table because we needed a single object to link to. That was fine as the converter was the only place that handled GuPPy objects publicly. Now we get two entry points for different purposes, the converter as the aggregator and the interface to add only GuPPy's output. I would like the interface to be consistent about what it writes, and some of your code here already hints at how.
> 
> Instead of merging the events from the acquisition sources, I think we should write GuPPy's own internal events as the table we link. That makes the interface do the same thing standalone and inside the converter, and gives us a stable surface for linking the PSTHs and the other objects that reference events into a single consistent structure. It also eliminates a lot of code (`resolve_event_store_rows`, `select_analyzed_rows`, `_resolve_event_rows` and the timestamp matching in `_build_guppy_registries`) and some assertions.
> 
> In principle this is a small duplication of the events info in the NWB file but I don't think the case holds. GuPPy drops the start, does its own alignment and writes onsets only, and that is fine. The acquisition tables are just the raw sources, so they come out the same whether people go through the converter or add the events interfaces themselves (which is an additional gain from this approach). The GuPPy onsets are a processing output and stay their own table, a stable scaffold that will always be available.

Thank you for the review and the suggestion. I think you are right. Having the GuppyInterface write its own events every time seems like the more robust approach. One question: do you think that the converter should still write the raw events into a merged events table? Or now that it's separated it could probably be written into separate `EventsTable`s, which would reduce the converter's responsibilities.

h-mayorquin:
The latter. Reduce the converter responsabilities.

pauladkisson:
> The latter. Reduce the converter responsibilities.

Okay, I double-checked the diff, and now the GuppyInterface writes its own events, as you suggested.

h-mayorquin (changes requested):
Two requests and one question before this is ready to go.

1) `GuppyInterface` is public now but `docs/api/interfaces.fiber_photometry.rst` still carries only `GuppyConverter`. The gallery links to the class with `:py:class:` so that reference resolves to nothing.

2) `converted_nwbfile_path` is function-scoped, so the `read_nwb` plus `add_to_nwbfile` plus write cycle runs once per test for the five tests in `TestGuppyInterfaceLinksIntoExistingTables` that only read the result. I ran the module and those five setups are 4.8s of an 11.6s total. `scope="module"` with `tmp_path_factory` fixes it, and returning the read `NWBFile` alongside the path collapses the five `read_nwb` calls in the test bodies as well. Should bring the module to around 6.5s.

And the new warning in `_resolve_recording_site_rows` says the store names no `FiberPhotometryResponseSeries`, but a store also drops out of the map when the series is right there and carries no `fiber_photometry_table_region`. Is this a real case or is the `region is None` branch just defensive? Maybe each specific case deserves a specific warning.

pauladkisson:
> Two requests and one question before this is ready to go.
> 
> 1. `GuppyInterface` is public now but `docs/api/interfaces.fiber_photometry.rst` still carries only `GuppyConverter`. The gallery links to the class with `:py:class:` so that reference resolves to nothing.
> 2. `converted_nwbfile_path` is function-scoped, so the `read_nwb` plus `add_to_nwbfile` plus write cycle runs once per test for the five tests in `TestGuppyInterfaceLinksIntoExistingTables` that only read the result. I ran the module and those five setups are 4.8s of an 11.6s total. `scope="module"` with `tmp_path_factory` fixes it, and returning the read `NWBFile` alongside the path collapses the five `read_nwb` calls in the test bodies as well. Should bring the module to around 6.5s.
> 
> And the new warning in `_resolve_recording_site_rows` says the store names no `FiberPhotometryResponseSeries`, but a store also drops out of the map when the series is right there and carries no `fiber_photometry_table_region`. Is this a real case or is the `region is None` branch just defensive? Maybe each specific case deserves a specific warning.

All of these should be addressed. I added an extra warning for store names with no fiber photometry table region, which is technically possible but probably shouldn't happen under normal circumstances.

h-mayorquin (approved):
This is ready to go.

PR #1883, opened 2026-08-10
Only write the fiber photometry devices a table row references

### LINKED ISSUE #1881: `add_fiber_photometry_devices` silently downgrades other interfaces' devices to plain `Device`
### Summary

`add_fiber_photometry_devices()` (`src/neuroconv/tools/fiber_photometry.py`) is called from
`BaseFiberPhotometryInterface.add_to_nwbfile()` and iterates **every** entry in the shared
`metadata["Devices"]` / `metadata["DeviceModels"]` registries, not just the devices that actually
belong to fiber photometry:

```python
def add_fiber_photometry_devices(*, nwbfile: NWBFile, metadata: dict) -> None:
    for device_model_metadata_key in metadata.get("DeviceModels", {}):
        _add_device_model_to_nwbfile(nwbfile=nwbfile, metadata=metadata, metadata_key=device_model_metadata_key)

    for device_metadata_key in metadata.get("Devices", {}):
        _add_device_to_nwbfile(nwbfile=nwbfile, metadata=metadata, metadata_key=device_metadata_key)
```

Because `NWBConverter` merges every sub-interface's `get_metadata()` output into **one shared
`metadata` dict**, `metadata["Devices"]` is not scoped to fiber photometry — it also contains
whatever other interfaces in the same converter have registered there (e.g. `DANNCEInterface`
registers one entry per camera, keyed by camera name).

`_add_device_to_nwbfile` falls back to a plain `pynwb.device.Device` whenever a registry entry has
no `"type"` field (see its docstring: *"the entry may carry a `type` field naming the concrete
class... when omitted, a plain `pynwb.device.Device` is created"*). Camera entries written by
`DANNCEInterface.get_metadata()` have no `"type"` field — `DANNCEInterface.create_camera_devices()`
builds the real `ndx_pose.CalibratedCamera` directly in Python instead of going through the
`"type"`-dispatch helper.

### Reproduction

Use a `NWBConverter` that combines:
- any interface built on `BaseFiberPhotometryInterface` (e.g. `DoricFiberPhotometryInterface`), and
- `DANNCEConverter` (or any interface that pre-registers its own non-`Device` device subtype by name
  in `metadata["Devices"]` without a `"type"` key),

with the fiber-photometry interface listed **before** the camera interface in
`data_interface_classes` (`NWBConverter.add_to_nwbfile` calls interfaces in that dict's order).

Steps:
1. `metadata = converter.get_metadata()` merges both interfaces' `Devices` entries into one dict —
   `metadata["Devices"]` now has `Camera1..CameraN` (no `"type"`) alongside the FP devices.
2. `converter.run_conversion(...)` calls `add_to_nwbfile` for each interface in order.
3. The fiber-photometry interface runs first and calls `add_fiber_photometry_devices(nwbfile,
   metadata)`, which loops over **all** of `metadata["Devices"]`, including `Camera1..CameraN`, and
   creates each with `_add_device_to_nwbfile` — since they carry no `"type"`, they come out as plain
   `Device`.
4. `DANNCEInterface.create_camera_devices()` runs later, sees `device_name in nwbfile.devices` is
   already `True` (idempotent-by-name), and just returns the existing plain `Device` instead of
   building the intended `CalibratedCamera`.

Net effect: cameras silently lose their calibration data (`intrinsic_matrix`, `rotation_matrix`,
`translation_vector`, `distortion_coefficients`) in the written NWB file, with no error or warning —
`nwbfile.devices["Camera1"]` is a bare `Device`.

### Expected behavior

`add_fiber_photometry_devices()` should only create/add the devices that are actually referenced by
the fiber photometry configuration (the `FiberPhotometryTable` rows' `optical_fiber`,
`excitation_source`, `photodetector`, and any `band_optical_filter` / `edge_optical_filter` /
`dichroic_mirror` fields, plus their `device_model_metadata_key` models) — not every key present in
the shared top-level `Devices`/`DeviceModels` registries. `add_fiber_photometry_lab_metadata()`
(same module) already does this resolution correctly for wiring rows to device *names*; the device
*creation* step should walk the same reachable set instead of `metadata.get("Devices", {})`
wholesale.

### Impact

Any `NWBConverter` that combines a fiber-photometry interface with another interface that
pre-registers non-generic devices in the shared `Devices` registry is affected, silently and without
any error — the plain `Device` still validates fine, it just isn't the intended subtype. Currently
observed with `DANNCEConverter`'s `CalibratedCamera` cameras, but the same failure mode applies to
any interface relying on a `"type"`-less, code-constructed device entry in the shared registry
(e.g. ecephys/ophys interfaces that build their device object directly rather than through
`_add_device_to_nwbfile`'s `"type"` dispatch).

### Workaround

Order the affected interface (e.g. `DANNCE`) *before* the fiber-photometry interface(s) in
`data_interface_classes`, so the correct device subtype is created first — `_add_device_to_nwbfile`
is idempotent on name and returns the existing device unchanged if one is already present. This
avoids the bug but is fragile: it depends on interface ordering that has nothing to do with the
actual bug, and silently breaks again if the ordering ever changes.

h-mayorquin:
Alessandra found in #1881 that a fiber photometry interface inside a converter was writing every other interface's devices as well, as `add_fiber_photometry_devices` walked the whole shared `metadata["Devices"]` and `metadata["DeviceModels"]` registries instead of its own entries. An entry naming no `type` came out of that as a plain `Device`, and since device creation is idempotent on name, the interface that owned the entry reused the plain one instead of building the subtype it meant to, silently and only when it happened to run second. This PR scopes the write to the devices a `FiberPhotometryTable` row actually references, so a device or model that no row points at is no longer written here.

This should close #1881.


alessandratrapani (approved):
Looks good! Thanks!

PR #1888, opened 2026-08-11
Sort the electrodes and units table columns

### LINKED ISSUE #792: Preserve order of recording extractor custom properties in `add_electrodes()`

h-mayorquin:
Back in #792 @CodyCBakerPhD asked for the columns of a family like `acx_x`, `acx_y` and `acx_z` to sit together so the electrodes table reads better in Neurosift. #793 and #1878 both give him that by making the written order follow the extractor. I have argued before that [the complexity is not worth the effort](https://github.com/catalystneuro/neuroconv/issues/792#issuecomment-2028441254) and gone through [what the metadata list version would cost](https://github.com/catalystneuro/neuroconv/issues/792#issuecomment-2028447834).

What I [proposed then instead](https://github.com/catalystneuro/neuroconv/issues/792#issuecomment-2028335759) was natsort and this PR implements it. The properties are written in natsort order in the electrodes and the units tables, so the families group and `shank_2` comes before `shank_10`. Determinism comes for free. The two loops that create the columns iterate a set right now, so the order follows string hashing and python randomizes that per process, and the same recording writes a different `colnames` on every run. I also moved `natsort` into the `ecephys_minimal` extra as `tools/spikeinterface/spikeinterface.py` imports it unconditionally now.
My full reasoning for not taking the extractor order will be on #792.
This should close #792.


h-mayorquin:
@adityasingh2400 tested this. @alejoe91 could you do a quick check as you are on of the main users here and also you have a lot of experience with this.

src/neuroconv/tools/spikeinterface/spikeinterface.py
alejoe91, line 1485: Should this be `natsorted` too to make sure it's deterministic? 

In case of an existing table, these are added by row so it will follow the order of existing columns. But for newly created tables, could this result in a randomized order of schema properties?

alejoe91 (approved):
@h-mayorquin generally looks good to me! Keeps schema columns first and adds the rest as natsorted, which I think is a good behavior.

I just have one comment, but I might be missing something

src/neuroconv/tools/spikeinterface/spikeinterface.py
h-mayorquin: No, I don't think it is needed.

`properties_with_data` only fills columns that already exist: on an existing table they are its previous columns and on a new table they are the ones pynwb's `ElectrodesTable` defines up front (`location`, `group` and `group_name`). pynwb creates those in its own declared order so the order of the kwargs we pass to `add_electrode` never reaches `colnames`.

PR #1904, opened 2026-08-12
Build a read NWBFile with the manager that read it

### LINKED ISSUE #1903: `configure_and_write_nwbfile` fails on a read NWBFile whose extension was not imported
`configure_and_write_nwbfile` fails whenever the NWB file is read from disk and contains an extension that was not imported.

<details>

`configure_and_write_nwbfile` raises `ValueError: Could not find location ... in builder` when the
`NWBFile` was read from disk by a process that has not imported the extension supplying one of its
containers.

### Where I hit it

I'm adding NWB export to GuPPy ([LernerLab/GuPPy#357](https://github.com/LernerLab/GuPPy/pull/357)).
GuPPy reads NWB as an input format, so a session can be analyzed straight out of an existing NWB
file — which is exactly what the public `GuppyInterface` from
[#1880](https://github.com/catalystneuro/neuroconv/pull/1880) is for. I followed the recipe from the
conversion gallery (`docs/conversion_examples_gallery/fiberphotometry/guppy_fp.rst`) verbatim:

```python
interface = GuppyInterface(folder_path=guppy_folder)

nwbfile = read_nwb(path=source_nwbfile_path)
interface.add_to_nwbfile(nwbfile=nwbfile, metadata=interface.get_metadata())
configure_and_write_nwbfile(nwbfile=nwbfile, nwbfile_path=nwbfile_path, backend="hdf5")
```

and it failed on our own test session with:

```
ValueError: Could not find location 'lab_meta_data/fiber_photometry/fiber_photometry_table/location/data'
            in builder (fiber_photometry_table is missing).
```

Dropping the GuPPy part entirely and just reading the file and writing it back out fails the same
way, so nothing about GuPPy is involved.

### What actually decides it

Whether the extension package was imported **before** `read_nwb`. Not the file's contents, not the
cached schema versions, not which extension it is.

`read_nwb` loads the extension's cached spec off the file either way, so reading and inspecting the
data behaves identically. But without the import, pynwb constructs a dynamically generated container
class rather than the one the package registers:

```python
# without `import ndx_fiber_photometry` before the read
type(nwbfile.lab_meta_data["fiber_photometry"]) is ndx_fiber_photometry.FiberPhotometry   # False

# with it
type(nwbfile.lab_meta_data["fiber_photometry"]) is ndx_fiber_photometry.FiberPhotometry   # True
```

In the first case the builder that `get_default_dataset_io_configurations` walks
(`manager.build(nwbfile, export=True)`, `_dataset_configuration.py:111`) comes back without that
container's children, so `get_dataset_builder` cannot resolve the dataset location and raises.

This matters in ordinary use because a caller has no particular reason to import an extension it
never names — pynwb's cached-namespace loading is precisely what makes that unnecessary everywhere
else.

### Minimal reproduction

Uses only `ndx-events`, which neuroconv already depends on. Each variant runs in its own subprocess,
since a single import anywhere in the process decides the outcome — run them together and the
failure disappears.

```python
"""Reproduction: configure_and_write_nwbfile on a read NWBFile whose extension was not imported."""
import subprocess
import sys
from pathlib import Path

PATH = Path(__file__).parent / "events_source.nwb"


def write():
    from datetime import datetime, timezone
    import numpy as np
    from ndx_events import LabeledEvents
    from pynwb import NWBHDF5IO, NWBFile

    nwbfile = NWBFile(
        session_description="session",
        identifier="id",
        session_start_time=datetime(2020, 1, 1, tzinfo=timezone.utc),
    )
    nwbfile.add_acquisition(
        LabeledEvents(
            name="labeled_events",
            description="labeled events",
            timestamps=np.arange(10, dtype="float64"),
            data=np.zeros(10, dtype="uint8"),
            labels=["a"],
        )
    )
    with NWBHDF5IO(str(PATH), "w") as io:
        io.write(nwbfile)


def export(import_extension):
    if import_extension:
        import ndx_events  # noqa: F401
    from pynwb import read_nwb
    from neuroconv.tools.nwb_helpers import configure_and_write_nwbfile

    nwbfile = read_nwb(path=str(PATH))
    try:
        configure_and_write_nwbfile(
            nwbfile=nwbfile, nwbfile_path=str(PATH.parent / "events_out.nwb"), backend="hdf5"
        )
        print("OK")
    except Exception as exception:
        print(f"{type(exception).__name__}: {exception}")
    finally:
        nwbfile.read_io.close()


if __name__ == "__main__":
    if len(sys.argv) > 1:
        {"write": write, "no": lambda: export(False), "yes": lambda: export(True)}[sys.argv[1]]()
    else:
        subprocess.run([sys.executable, __file__, "write"], check=True)
        for flag, label in (("no", "without import"), ("yes", "with import   ")):
            print(f"{label}: ", end="", flush=True)
            subprocess.run([sys.executable, __file__, flag], check=True)
```

Output:

```
without import: ValueError: Could not find location 'acquisition/labeled_events/data' in builder (data is missing).
with import   : OK
```

### Scope

Four extensions tried, three of them affected — in three different places in the file, so this is
not specific to `lab_meta_data` or to fiber photometry:

| extension | container | location in file | no import | with import |
|---|---|---|---|---|
| ndx-fiber-photometry | `FiberPhotometryTable` | `lab_meta_data` | **ValueError** | OK |
| ndx-events | `LabeledEvents` | `acquisition` | **ValueError** | OK |
| ndx-pose | `PoseEstimationSeries` | `processing` | **ValueError** | OK |
| ndx-miniscope | `Miniscope` (Device) | `devices` | OK | OK |

ndx-miniscope is unaffected because its container carries attributes only — there is no dataset to
size, so the lookup never happens. The rule appears to be: any un-imported extension container that
neuroconv needs to size a dataset on.

`pynwb`'s own `io.export` handles all of these without complaint, since it needs no such lookup.

### Why the test suite does not catch this

The fiber photometry tests build their source file in-process through `MockFiberPhotometryInterface`
(e.g. `tests/test_modalities/test_guppy/test_guppy_interface_nwb_linking.py`, which runs
`read_nwb` → `add_to_nwbfile` → `configure_and_write_nwbfile` and passes). Constructing the source
imports the extension, so the read that follows gets the registered class and the builder is
complete. A user opening a file they did not write in that same process gets the other path.

Worth noting for anyone reproducing: writing and reading in one process masks this completely. It
cost me several rounds of false negatives before I spotted it.

### Environment

- neuroconv 0.10.0rc1; also reproduced from a source checkout at `cd2bbb609`, whose
  `tools/hdmf.py` and `tools/nwb_helpers/` are identical to `main` (`aec83a6e7`)
- pynwb 4.0.0, hdmf 6.1.1.dev22+g4a8967fe8
- ndx-fiber-photometry 0.2.3, ndx-events 0.2.2, ndx-ophys-devices 0.3.1, ndx-pose 0.3.0, ndx-miniscope 0.5.1
- Python 3.13, macOS

</details>

pauladkisson:
Fixes #1903, confirmed working for ndx-fiber-photometry, and looks pretty clean to me.

<details>

Closes #1903.

`get_default_dataset_io_configurations` and `configure_backend` built the NWBFile with the global
manager from `pynwb.get_manager()`. Both now build it with `nwbfile.read_io.manager` when the file
was read from disk, falling back to the global manager otherwise. This is the same manager
`HDMFIO.export` itself builds with, so no behavior changes for a file constructed in-process.

Adds a regression test covering both backends. It writes its source file through an extension
namespace registered in a local type map only, so the global type map stays ignorant of it no matter
what else the test session imported.

Verified: `pytest tests/test_minimal` (547 passed, 68 skipped) and `pytest tests/test_modalities`
match `main` exactly.

</details>


src/neuroconv/tools/hdmf.py
```diff
@@ -232,6 +233,28 @@ def _get_data(self, selection: tuple[slice]) -> np.ndarray:
...
+
+    Returns
+    -------
+    hdmf.build.builders.BaseBuilder
+        The builder object for the NWBFile.
+    """
+    # A file read from disk is built by the manager that read it, whose type map carries the namespaces cached in
+    # that file. The global manager only knows the extensions this process imported, and builds a container of an
+    # unimported extension without any of its datasets.
+    manager = nwbfile.read_io.manager if nwbfile.read_io is not None else get_manager()
```
h-mayorquin:
```suggestion
    manager = BuildManager(nwbfile.read_io.manager.type_map) if nwbfile.read_io is not None else get_manager()
```

The current approach fails in append mode.

`read_io.manager` is also the manager that writes the file in append mode, and it caches every builder it makes here. `configure_backend` wraps the data in a `DataIO` after this build, so `io.write` reuses the cached builder as it is and the compression never reaches the file. I checked `main` against this branch, an appended `TimeSeries` goes from `gzip` to no compression and nothing throws an error.

A fresh `BuildManager` keeps the type map, as that is the part that fixes the unimported extension, and leaves the cache out of it. It needs `from hdmf.build import BuildManager` up top.

src/neuroconv/tools/hdmf.py
```diff
@@ -232,6 +233,28 @@ def _get_data(self, selection: tuple[slice]) -> np.ndarray:
         return self.data[selection]
 
 
+def get_nwbfile_builder(nwbfile: NWBFile) -> BaseBuilder:
```
h-mayorquin: can we make this private? keep the public API to a minimum for fast changes later and all of that.

h-mayorquin (changes requested):
The approach looks sensible to me. The fix as it is breaks append mode but is easy to solve (see the comment).

We should add a test for append mode so this does not break in the future.

src/neuroconv/tools/hdmf.py
pauladkisson: Done!

src/neuroconv/tools/hdmf.py
pauladkisson: Fixed and tested.

h-mayorquin (approved):
Thanks for addressing my points. This is ready to go.

PR #1905, opened 2026-08-12
Add a pyPhotometry interface

### LINKED ISSUE #1602: [New Format]: PyPhotometry

The PyPhotometry system is used by Peter Magill at Oxford as part of their ASAP-funded work. It would be great to support automatic conversion of this data. We don't have a project with them (yet), but I just wanted to put this on our radar.

https://pyphotometry.readthedocs.io/en/latest/

### Existing APIs for format

_No response_

I don't have any example data, but this page describes the two supported data formats: https://pyphotometry.readthedocs.io/en/latest/user-guide/importing-data/

No.

### Code of Conduct

h-mayorquin:
This should close #1602.

The reader dispatches on the header's `mode` field rather than on `n_analog_signals`, since the vocabulary changed twice and an unknown mode has to raise instead of falling back. One interface reads one signal, because the board samples its inputs one after another and each signal gets the start time its slot implies. 

The test data is already merged on gin:
https://gin.g-node.org/CatalystNeuro/ophys_testing_data/pulls/57.


h-mayorquin:
@pauladkisson, this is ready for a review which I would like here. Three things to highlight about the format and one design decision of how we represent the data:

1. In this format two regions never appear together, so there is nothing to aggregate.
2. As in the other formats, no metadata is stored in the source. Everything has to be user provided.
3. Streams correspond to photodetectors and (optionally) to a strobing lamp. They are named as such, `detector_1_excitation_1` and `detector_1_excitation_2` for a recording that strobes two lamps onto one detector, `detector_1_excitation_1` and `detector_2_excitation_2` for one that gives each lamp its own. And each of them has a different start time that accounts for the staggered reading of the signals.

The representation issue: In the strobed modes each point is a lit reading minus a dark one taken beside it. The dark reading is ambient light plus the detector's own offset. Older firmware subtracted on the board and wrote one number. From version 1.1 it writes both and the subtraction moves into the reader.

We write all three: the difference in `acquisition`, plus both raw traces beside it and this is how we link to the fiber photometry table

* `FiberPhotometryResponseSeries`, the difference, references its `FiberPhotometryTable` row, the one naming the fiber, the excitation source and the photodetector that produced it. This is what every pipeline expects and what the older firmware wrote itself.
* `FiberPhotometryResponseSeriesRawLEDOn`, the lit reading before the subtraction, references that same row. Every field of it is true of the lit reading too.
* `FiberPhotometryResponseSeriesRawBaseline`, the dark reading, references no row at all. My decision here is to not put a reference as some columns of the table do not make sense: excitation wavelength for example.

I brought this up before here: https://github.com/catalystneuro/ndx-fiber-photometry/issues/54

docs/conversion_examples_gallery/fiberphotometry/tdt_fp.rst
```diff
@@ -549,6 +549,14 @@ This metadata can then be easily incorporated into the conversion by updating th
     >>> # t1 and t2 are optional arguments to specify the start and end times for the conversion
     >>> interface.run_conversion(nwbfile_path=nwbfile_path, metadata=metadata, t1=0.0, t2=1.0, overwrite=True)
 
+NeuroConv aims to automatically add all the metadata annotations that are present in the source format.
```
pauladkisson, line 552: This note seems reasonable to me, but I'm a little confused why it's landing in this PR and only TDT?

pauladkisson (approved):

pauladkisson:
> @pauladkisson, this is ready for a review which I would like here. Three things to highlight about the format and one design decision of how we represent the data:
> 
> 1. In this format two regions never appear together, so there is nothing to aggregate.
> 2. As in the other formats, no metadata is stored in the source. Everything has to be user provided.
> 3. Streams correspond to photodetectors and (optionally) to a strobing lamp. They are named as such, `detector_1_excitation_1` and `detector_1_excitation_2` for a recording that strobes two lamps onto one detector, `detector_1_excitation_1` and `detector_2_excitation_2` for one that gives each lamp its own. And each of them has a different start time that accounts for the staggered reading of the signals.
> 
> The representation issue: In the strobed modes each point is a lit reading minus a dark one taken beside it. The dark reading is ambient light plus the detector's own offset. Older firmware subtracted on the board and wrote one number. From version 1.1 it writes both and the subtraction moves into the reader.
> 
> We write all three: the difference in `acquisition`, plus both raw traces beside it and this is how we link to the fiber photometry table
> 
> * `FiberPhotometryResponseSeries`, the difference, references its `FiberPhotometryTable` row, the one naming the fiber, the excitation source and the photodetector that produced it. This is what every pipeline expects and what the older firmware wrote itself.
> * `FiberPhotometryResponseSeriesRawLEDOn`, the lit reading before the subtraction, references that same row. Every field of it is true of the lit reading too.
> * `FiberPhotometryResponseSeriesRawBaseline`, the dark reading, references no row at all. My decision here is to not put a reference as some columns of the table do not make sense: excitation wavelength for example.
> 
> I brought this up before here: [catalystneuro/ndx-fiber-photometry#54](https://github.com/catalystneuro/ndx-fiber-photometry/issues/54)


These design decisions sound eminently reasonable to me.

pauladkisson:
Looks like the tests are failing for some unrelated reason.

docs/conversion_examples_gallery/fiberphotometry/tdt_fp.rst
h-mayorquin: All the others already follow the same template, TDT was missing.

h-mayorquin:
> Looks like the tests are failing for some unrelated reason.

Yes, is yet another flaky bug with the ndx-extensions global namespace behavior we have encountered before. I am trying to catch that one for some days now.

PR #1910, opened 2026-08-12
Walk the NWB file for its objects instead of reading a cached dict

### LINKED ISSUE #1909: Datasets added after `nwbfile.objects` is first read are never configured
`nwb_file.objects` yields a stale cache that doesn't update when new objects are added to the NWB file. We should use `nwb_file.all_children()` instead.

It might be worth raising a separate issue on the pynwb side for this as well.

<details>

`nwbfile.objects` is a cache that pynwb builds on first read and never invalidates. NeuroConv both
reads it in the middle of a write and reads it again to decide what to configure, so any dataset
added after the first read is written with no compression and no chunking.

### Where I hit it

I'm adding NWB export to GuPPy ([LernerLab/GuPPy#357](https://github.com/LernerLab/GuPPy/pull/357)).
Now that [#1904](https://github.com/catalystneuro/neuroconv/pull/1904) fixed the builder failure from
[#1903](https://github.com/catalystneuro/neuroconv/issues/1903), I swapped GuPPy's NWB-sourced export
route off `pynwb`'s own `io.export` and back onto the gallery recipe:

```python
interface = GuppyInterface(folder_path=guppy_folder)
nwbfile = read_nwb(path=source_nwbfile_path)
interface.add_to_nwbfile(nwbfile=nwbfile, metadata=interface.get_metadata())
configure_and_write_nwbfile(nwbfile=nwbfile, nwbfile_path=nwbfile_path, backend="hdf5")
```

It writes cleanly, and the source file's own datasets come out `gzip` where the plain `io.export`
left them raw. But every dataset the interface added — the derived traces, the transients tables,
the events table, the recording sites registry — comes out with `compression=None` and `chunks=None`.
The backend configuration for that file lists 11 datasets, all of them from the source, and not one
under `processing/guppy`.

### What is happening

`NWBFile.objects` is populated once, lazily, and nothing ever resets it
([`pynwb/file.py:589`](https://github.com/NeurodataWithoutBorders/pynwb/blob/dev/src/pynwb/file.py)):

```python
@property
def objects(self):
    if self.__obj is None:
        self.all_children()
    return self.__obj
```

So the first read of `.objects` decides its contents permanently:

```python
nwbfile = mock_NWBFile()
objects = nwbfile.objects                                     # 1 object
nwbfile.add_acquisition(mock_TimeSeries(name="added", data=np.arange(10)))
nwbfile.objects is objects                                    # True — the same dict, not a rebuild
any(o.name == "added" for o in nwbfile.objects.values())      # False
any(c.name == "added" for c in nwbfile.all_children())        # True
```

NeuroConv reads `.objects` in two different roles, and they collide:

| role | call site |
|---|---|
| **inspects the file mid-write**, freezing the cache | `datainterfaces/fiber_photometry/guppy/nwb_linking.py:24` (`_series_by_name`), `datainterfaces/behavior/vame/vamedatainterface.py:616,633` |
| **walks the file to decide what to configure** | `tools/nwb_helpers/_dataset_configuration.py:112,243`, `tools/nwb_helpers/_configure_backend.py:53` |

In GuPPy's case the freeze happens early in `GuppyInterface.add_to_nwbfile` — only the processing
module and the session parameters exist yet — when it resolves GuPPy's store ids against the file's
`FiberPhotometryResponseSeries`:

```
add_to_nwbfile (guppydatainterface.py:771)
  _get_or_add_guppy_recording_sites_table (guppydatainterface.py:1154)
    _resolve_recording_site_rows (guppydatainterface.py:1186)
      resolve_acquisition_store_rows (nwb_linking.py:65)
        _series_by_name (nwb_linking.py:24)   ->  nwbfile.objects
```

The cache freezes at 42 objects. The finished file has 85. The 43 containers added after that point
are invisible to `get_default_dataset_io_configurations`, so nothing wraps them in an `H5DataIO` and
they are written with backend defaults.

`all_children()` recomputes the walk on every call (and repopulates `__obj` as a side effect), so
the distinction between the two is entirely whether the answer is allowed to be stale.

### Reproduction

No extensions, no GuPPy — the only variable is whether anything read `.objects` before the add.

```python
"""Reproduction: a dataset added after `nwbfile.objects` is first read is never configured."""

import h5py
import numpy as np
from pynwb import NWBHDF5IO, read_nwb
from pynwb.testing.mock.base import mock_TimeSeries

from neuroconv.tools.nwb_helpers import configure_and_write_nwbfile

SOURCE_PATH = "source.nwb"


def write_source():
    nwbfile = mock_NWBFile()
    nwbfile.add_acquisition(mock_TimeSeries(name="FromSource", data=np.arange(1000, dtype="float64")))
    with NWBHDF5IO(SOURCE_PATH, "w") as io:
        io.write(nwbfile)


def export(read_objects_first: bool):
    nwbfile = read_nwb(path=SOURCE_PATH)

    if read_objects_first:
        nwbfile.objects  # what any mid-write inspection does

    nwbfile.add_acquisition(mock_TimeSeries(name="AddedAfter", data=np.arange(1000, dtype="float64")))

    out_path = f"out_{read_objects_first}.nwb"
    configure_and_write_nwbfile(nwbfile=nwbfile, nwbfile_path=out_path, backend="hdf5")
    nwbfile.read_io.close()

    with h5py.File(out_path, "r") as file:
        return {name: file[f"acquisition/{name}/data"].compression for name in ("FromSource", "AddedAfter")}


if __name__ == "__main__":
    write_source()
    for read_objects_first in (False, True):
        print(f"objects read before the add: {str(read_objects_first):<5} -> {export(read_objects_first)}")
```

```
objects read before the add: False -> {'FromSource': 'gzip', 'AddedAfter': 'gzip'}
objects read before the add: True  -> {'FromSource': 'gzip', 'AddedAfter': None}
```

### Scope

This is not specific to files read from disk. The same two lines against an in-memory
`mock_NWBFile()` — read `.objects`, then `add_acquisition`, then `configure_and_write_nwbfile` —
produce `gzip` and `None` respectively, so an ordinary conversion is exposed too: in any converter
whose earlier interface reads `.objects`, every dataset a later interface adds silently loses its
configuration. `NWBConverter` runs its interfaces in `data_interface_objects` order, so which
datasets are affected depends on interface ordering rather than on anything about the data.

`get_existing_dataset_io_configurations` (`_dataset_configuration.py:243`) walks the same dict, so
`repack_nwbfile` and the append paths read from the same cache.

### Impact

Silent. Nothing raises, the file is valid, and the data is correct — it is simply uncompressed and
unchunked, which is only visible by inspecting the written file's filters. A conversion that looks
like it applied the default backend configuration may have applied it to a fraction of the file.

For scale, GuPPy's NWB-sourced export writes three derived traces per recording site, plus the
transients, events and recording-sites tables, uncompressed; on a real multi-hour session those
traces are among the largest objects in the file.

### Upstream note

The invalidation hook exists on the pynwb/hdmf side: attaching a child appends to the parent's
`__children` and calls `set_modified()` up the chain
([`hdmf/container.py:469-477`](https://github.com/hdmf-dev/hdmf/blob/dev/src/hdmf/container.py)),
which `NWBFile` could use to drop the cached dict. Worth a separate report there; NeuroConv's own
exposure is decided by which of the two accessors each call site uses.

### Environment

- neuroconv `main` @ `ae9bbbfc2`
- pynwb 4.0.0, hdmf 6.1.0
- Python 3.13.13, macOS

</details>

pauladkisson:
Code looks good, and I confirmed that no `nwbfile.objects` references are left in this repo (besides the regression tests, of course). 

<details>

Closes #1909.

Replaces every read of `nwbfile.objects` with a fresh `nwbfile.all_children()` walk: the two
configuration walks in `_dataset_configuration.py`, the object-id lookup in `configure_backend`
(built once before the loop), and the three mid-write inspections in `nwb_linking.py` and
`vamedatainterface.py` that were freezing the cache.

One behavior change comes with it. Building a `backend_configuration`, then adding a container, then
calling `configure_backend` used to work by accident, because both walks read the same stale cache.
`find_locations_requiring_remapping` now sees the extra container and raises `ValueError: The number
of default configurations (N) does not match the number of specified configurations (M)!`.

Tests: a new `test_stale_objects_cache.py` covering the configuration, the `DataIO` wrapping, the
written file's filters, and that `ValueError`, plus one case in the GuPPy linking module. All eight
fail without the source change. `tests/test_minimal` and `tests/test_modalities` show no failures
beyond the pre-existing ones in this environment.

</details>


h-mayorquin (changes requested):
As you said, probably worth opening and issue about the behavior of `nwbfile.objects`

This looks good to me but reviewing this uncovered a design decision that I would like to discuss with you. Check this script.


```python
nwbfile = converter.create_nwbfile(metadata=metadata)

# 1. build the configuration
backend_configuration = converter.get_default_backend_configuration(nwbfile=nwbfile, backend="hdf5")
backend_configuration.dataset_configurations["acquisition/ElectricalSeriesAP/data"].compression_method = "Blosc"

# 2. remember something else belongs in the file, and add it
nwbfile.add_trial(start_time=0.0, stop_time=1.0)
another_interface.add_to_nwbfile(nwbfile=nwbfile)

# 3. write
configure_and_write_nwbfile(nwbfile=nwbfile, nwbfile_path=path, backend_configuration=backend_configuration)
```

Before this PR this worked perfectly but it did not compress or chunk the steps that were not configured (everything on step 2 above). This can be called "configure_and_write is permissive and only checks for what is included" but now after this PR the workflow above fails because `configure_and_write_nwbfile` expects the backend configuration to be total. I am honestly unsure about what is the best behavior, I think the new behavior is better but I also think that maybe is even better to automatically chunk chunk and add compression to the things in 2 so they are not skipped (as before) but also do not rise and stop a user on its tracks.

What behavior do you think is best?

pauladkisson:
> As you said, probably worth opening and issue about the behavior of `nwbfile.objects`
> 
> This looks good to me but reviewing this uncovered a design decision that I would like to discuss with you. Check this script.
> 
> ```python
> nwbfile = converter.create_nwbfile(metadata=metadata)
> 
> # 1. build the configuration
> backend_configuration = converter.get_default_backend_configuration(nwbfile=nwbfile, backend="hdf5")
> backend_configuration.dataset_configurations["acquisition/ElectricalSeriesAP/data"].compression_method = "Blosc"
> 
> # 2. remember something else belongs in the file, and add it
> nwbfile.add_trial(start_time=0.0, stop_time=1.0)
> another_interface.add_to_nwbfile(nwbfile=nwbfile)
> 
> # 3. write
> configure_and_write_nwbfile(nwbfile=nwbfile, nwbfile_path=path, backend_configuration=backend_configuration)
> ```
> 
> Before this PR this worked perfectly but it did not compress or chunk the steps that were not configured (everything on step 2 above). This can be called "configure_and_write is permissive and only checks for what is included" but now after this PR the workflow above fails because `configure_and_write_nwbfile` expects the backend configuration to be total. I am honestly unsure about what is the best behavior, I think the new behavior is better but I also think that maybe is even better to automatically chunk chunk and add compression to the things in 2 so they are not skipped (as before) but also do not rise and stop a user on its tracks.
> 
> What behavior do you think is best?

I did notice this change. I'm happy to discuss, but I do think that the way that Claude did it is right.

Basically, my thought is that building the backend configuration should happen just before write. Doing it before, like adding something to the NWB file in between, is not a recommended use case that we shouldn't worry about supporting. 

It is a subtle difference from the way that the backend system was working before, but technically, the way the backend system was working before was just like a silent bug. 

And also, it just doesn't seem to me like there would ever be a good reason to mess around with the NWB file in between getting the backend configuration and writing the NWB file.

h-mayorquin:
OK, if I read you correctly you saw that this implies the strict interpretation and you are fine with it.

I am OK with going with your choice of the strict interpretation but that needs two things.

1) The documentation uses the pattern the policy forbids and has to go:
https://github.com/catalystneuro/neuroconv/blob/a3b930eeed1794cd00ccc15439a9d5df9c685517/docs/user_guide/backend_configuration.rst#L200-L218

That is our main worked example for customizing a backend configuration. It builds the file, derives the configuration from it, then passes both to `run_conversion`. We should move it now to the pattern that we are supporting, using `configure_and_write_nwbfile` like the other examples below.

2) We should block the combination that gets you here. `run_conversion(nwbfile=..., backend_configuration=...)` adds the interfaces to the file you passed and configures afterwards, so a configuration derived from that file is stale before the write, always. After this PR that combination can only raise the count mismatch, and there is no configuration a caller can build from that file that passes. `append_on_disk_nwbfile=True` with a `backend_configuration` has the same problem.

h-mayorquin:
OK, if I read you correctly you saw that this implies the strict interpretation and are fine with it.

I am OK with going with your choice of the strict interpretation but that needs two things.

1) The documentation uses the pattern the policy forbids and has to go:
https://github.com/catalystneuro/neuroconv/blob/a3b930eeed1794cd00ccc15439a9d5df9c685517/docs/user_guide/backend_configuration.rst#L200-L218

That is our main worked example for customizing a backend configuration. It builds the file, derives the configuration from it, then passes both to `run_conversion`.

We should move it now to the pattern that we are supporting, using `configure_and_write_nwbfile` like the other examples below.

2) We should block the combination that gets you here. `run_conversion(nwbfile=..., backend_configuration=...)` adds the interfaces to the file you passed and configures afterwards, so a configuration derived from that file is stale before the write, always. After this PR that combination can only raise the count mismatch, and there is no configuration a caller can build from that file that passes. `append_on_disk_nwbfile=True` with a `backend_configuration` has the same problem.

pauladkisson:
> OK, if I read you correctly you saw that this implies the strict interpretation and are fine with it.
> 
> I am OK with going with your choice of the strict interpretation but that needs two things.
> 
> 1. The documentation uses the pattern the policy forbids and has to go:
>    https://github.com/catalystneuro/neuroconv/blob/a3b930eeed1794cd00ccc15439a9d5df9c685517/docs/user_guide/backend_configuration.rst#L200-L218
> 
> That is our main worked example for customizing a backend configuration. It builds the file, derives the configuration from it, then passes both to `run_conversion`.
> 
> We should move it now to the pattern that we are supporting, using `configure_and_write_nwbfile` like the other examples below.
> 
> 2. We should block the combination that gets you here. `run_conversion(nwbfile=..., backend_configuration=...)` adds the interfaces to the file you passed and configures afterwards, so a configuration derived from that file is stale before the write, always. After this PR that combination can only raise the count mismatch, and there is no configuration a caller can build from that file that passes. `append_on_disk_nwbfile=True` with a `backend_configuration` has the same problem.

Makes sense to me. Both points should now be addressed.

h-mayorquin (approved):
This is great. Thanks for going not only through my request but beyond it.

Issue #1802, opened 2026-07-21
Discussion - Metadata scaffolding: template or example?


h-mayorquin:
On [#1557](https://github.com/catalystneuro/neuroconv/issues/1557) we discussed that `get_metadata()` should return only data available in the source file and we have been slowly moving in that direction ([#1800](https://github.com/catalystneuro/neuroconv/pull/1800) for fiber photometry, [#1798](https://github.com/catalystneuro/neuroconv/pull/1798) for events). However, there we also discussed providing metadata scaffolding that would make it easier for the users to fill and discover what metadata they should fill as input. We did not agree on the specifics and that is the purpose of this issue. 

The way I see it, there is only one big decision here:
* **return an example**: hand back the full chain filled with realistic, illustrative values. Its advantage is comprehension speed: a user who knows the modality reads `location="VTA"`, `excitation_wavelength_in_nm=470.0`, and `label="GCaMP6s"` and immediately recognizes what each field is for, which a blank never conveys on its own. `get_example_metadata()` is a good name, because "example" sets the expectation that the values are illustrative and yours to replace. 
* **return a template**: hand back the same structure and the same cross-references, but with the fields only the experimenter can supply left blank (`None`) and everything the interface genuinely knows already filled in. The discovery benefit is identical, because the hard part to reconstruct by hand are the available keys, the level of nesting and the `*_metadata_key` wiring, not the leaf values, and what stays blank doubles as a checklist of exactly what the source could not tell us. `get_metadata_template()` is a good name, because "template" carries the obligation to fill it in.

I think that `get_metadata_template()` is better, `get_example_metadata()` reproduces the risk in #1557 where the library would enable writing placeholder values. A user might only change some of the fields, or worse, paste the example wholesale to silence an inspector that is flagging genuinely missing metadata. `get_metadata_template()` has most of the advantages but puts the None placeholder which provides the discoverability without the danger of writing wrong data. 

To make this concrete, here are examples. Each modality below shows both returns side by side, with `NWBFile` and `Subject` omitted and `# from source` marking the fields the interface fills in both.

**Ophys**

```python
# get_example_metadata()
{
    "Devices": {
        "microscope": {"name": "microscope", "description": "Two-photon microscope.", "manufacturer": "Bruker"},
    },
    "Ophys": {
        "ImagingPlanes": {
            "imaging_plane": {
                "name": "imaging_plane",
                "device_metadata_key": "microscope",
                "imaging_rate": 30.0,            # from source
[...]

PR #1911, opened 2026-08-12
Implement `get_metadata_template()` for fiber photometry

h-mayorquin:
In #1802 we settled on a template of blanks instead of an example of values, and @pauladkisson's answer there on which fields earn their place is what I built this against. This PR adds `get_metadata_template()` to `BaseFiberPhotometryInterface`, returning the whole provenance chain sized at one `FiberPhotometryTable` row per trace, with every `*_metadata_key` already resolved and every field only the experimenter can supply left `None`. I did not put it on `BaseDataInterface` yet, as there is one implementation to generalize from and adding it there later is just as additive as adding it now.

The user guide gets a `Metadata Templates` page carrying the same structure as a YAML and a JSON file, as `load_dict_from_file` takes either one for a conversion specification's `metadata` block. Those files are checked in and the tests load them, so the page and the method can't drift apart. I also fixed three things I only found by writing that block out and filling it in: the template named a device type `OpticalFilter` that does not exist in ndx-ophys-devices, a row's `notes` never reached the file, and an optional `*_metadata_key` left at `None` was handed to the device writer as a real key.

The two pages this adds, on the build for this PR:

https://neuroconv--1911.org.readthedocs.build/en/1911/user_guide/metadata_templates.html

https://neuroconv--1911.org.readthedocs.build/en/1911/how_to/annotate_fiber_photometry_metadata.html#how-to-annotate-from-a-template


src/neuroconv/datainterfaces/fiber_photometry/basefiberphotometryinterface.py
```diff
@@ -288,7 +399,9 @@ def stub(array: np.ndarray) -> np.ndarray:
 
         response_series = FiberPhotometryResponseSeries(
             name=series_metadata["name"],
-            description=series_metadata.get("description", ""),
+            # ``or ""`` rather than a ``get`` default: the template hands back ``description=None`` for
```
pauladkisson: Do we really want to be so defensive here? I thought the point of the template was that it _couldn't_ be passed directly to `add_to_nwbfile`.

src/neuroconv/tools/fiber_photometry.py
```diff
@@ -156,52 +156,54 @@ def add_fiber_photometry_devices(*, nwbfile: NWBFile, metadata: dict) -> None:
 
     # One device type at a time, each collected across every row. Rows share hardware (two channels
     # commonly read one fiber), so a key repeats; the helper is idempotent on name and ignores it.
+    # A key set to ``None`` counts as absent: that is what ``get_metadata_template`` returns for an
```
pauladkisson: Similarly here, I'm a little wary about editing our ingestion to match the template's `None`s.

pauladkisson (commented):
The docs pages look great!

src/neuroconv/datainterfaces/fiber_photometry/basefiberphotometryinterface.py
h-mayorquin: Yes. I agree with you, those changes should not be here. They are removed now.

The story is that at first I was thinking on having specific errors when someone passed an incomplete template. But then I saw the result and it coupled too much the code to the template none sentinels (like here) and I reverted.

src/neuroconv/tools/fiber_photometry.py
h-mayorquin: Changed.

pauladkisson (approved):

PR #1916, opened 2026-08-13
Add a pyPhotometry events interface

h-mayorquin:
The digital lines of a `.ppd` are not a stream of their own: each one rides in the low bit of the same words that carry the fluorescence, so they come out of the reader #1905 adds rather than needing one of their own. This adds `PyPhotometryEventsInterface`, which reads them through the shared `detection_configuration` grammar, defaults to `high_period` so both edges of a pulse survive, and gives each line the rate and start time of the analog input it shares a slot with.

Based on #1905, so that one should be reviewed and merged first. One unrelated line rides along: `sphinx-design==0.7.0` needs Python 3.11 while the project supports 3.10, which makes `uv run` fail to resolve on main today, so it gets the same `python_version>='3.11'` marker the `Sphinx` pin beside it already carries.


pauladkisson (approved):
This one also looks good.

PR #1919, opened 2026-08-13
Add `shift_times` to the fiber photometry interfaces

h-mayorquin:
#1795 added `interface.alignment.shift_times` to the events interfaces and no other modality has it yet. This PR brings it to `BaseFiberPhotometryInterface`, and every interface built on it gets it at once, one rigid offset placing a whole recording on the session clock. GuPPy is left out as its derived products need an enumeration of their own.


h-mayorquin:
@pauladkisson 
I am taking your on the offer so can you review this?

I did fiber photometry on purpose, that way you review one concrete design in an area you know very well. Your feedback is very valuable so whatever comes out of it we can propagate to the canges of #1795  or revert the PR if you find a fatal flaw in the design. Once our thinking is aligned the other modalities will be much easier to move.

pauladkisson:
The main thing I'd like to discuss is how we want to reconcile the old temporal alignment system (`get_timestamps`, `get_original_timestamps`, `set_aligned_timestamps`, etc.) with the new temporal alignment system (`interface.alignment.shift_times`, etc.). This is also the major difference between this PR and #1795, since events didn't have any pre-existing temporal alignment.

The way that it currently works in the code seems very awkward and for example after you set an offset, `self._aligned_timestamps` aren't actually the aligned timestamps that get returned by get timestamps.

I think a better approach here would be to deprecate the old alignment methods and disallow using both the old system and the new system, so users have to pick. Open to alternatives, though.

h-mayorquin:
Thanks for the quick review.

 I would like to deprecate the old methods once we have full coverage with the new ones but I agree that we should not mix them. What do you think of this:

* Shift only: `get_timestamps()` returns `native + offset` and the file matches it.
* Set only: `get_timestamps()` returns the array you set and the file matches it.
* Shift and then set: raises in `set_aligned_timestamps`.
* Set and then shift: raises at write. The interface can't intercept `shift_times` as it is a call on the component.

This drops the offset cancellation that is in there right now. It only exists to make the composition safe, and with the composition gone there is nothing for it to do.

It also fixes what you pointed at. `_aligned_timestamps` can no longer differ from what `get_timestamps` returns, as the two can never be set at the same time.

The one thing it costs is fine alignment followed by placement on the session clock. That becomes `set_aligned_timestamps(interpolated + delta)` instead of a shift on top of the array. One addition, and I think that is acceptable while the interpolation half is not designed yet.

What do you think?

Another option is to hammer on and add full support for the new methods here, `set_times` and `remap_times` as #1793 describes them, if you are willing to review the full thing and make this the case study for that. But I was thinking about adding gradual support.

pauladkisson:
> Another option is to hammer on and add full support for the new methods here, set_times and remap_times as https://github.com/catalystneuro/neuroconv/pull/1793 describes them, if you are willing to review the full thing and make this the case study for that. But I was thinking about adding gradual support.

I do think that adding full support might end up being less work because it will require fewer passes through the codebase and less intermediate handling.

pauladkisson:
> I would like to deprecate the old methods once we have full coverage with the new ones but I agree that we should not mix them. What do you think of this:
> 
> * Shift only: `get_timestamps()` returns `native + offset` and the file matches it.
> * Set only: `get_timestamps()` returns the array you set and the file matches it.
> * Shift and then set: raises in `set_aligned_timestamps`.
> * Set and then shift: raises at write. The interface can't intercept `shift_times` as it is a call on the component.

This could work. To work out the implications here, this would mean that `get_timestamps` is our bridging method that is aware of both temporal alignment strategies and adjudicates between them. And we will keep using this method until we get something like `interface.alignment.times`. Okay, I'm on board.

It would be better to raise at shift time instead of at write time. Is there no way to do that? We could do a temporary shim: 
```python
interface._alignment = _TemporalAlignment()

@property
def alignment(self):
    # error if _aligned_timestamps
    return self._alignment
```

src/neuroconv/datainterfaces/fiber_photometry/basefiberphotometryinterface.py
```diff
@@ -370,7 +384,12 @@ def stub(array: np.ndarray) -> np.ndarray:
                 index = commanded_voltage_metadata.get("index")
                 if index is not None and commanded_voltage_data.ndim == 2:
                     commanded_voltage_data = commanded_voltage_data[:, index]
-                commanded_voltage_timestamps = self._get_stream_timestamps(stream_name=commanded_voltage_stream_name)
```
pauladkisson, line 379: OK, upon re-reading this module, I realized that the commanded voltage timestamps are actually a latent bug. Currently, there's no way to align them with the response series timestamps. Whoops.

pauladkisson: With that being said, the commanded voltage series should usually be on the same time basis as the fiber photometry response series. So maybe we could just drop the separate set of timestamps.

h-mayorquin: I was thinking about the commanded voltage series too: it should be its own time-bearing object in the full timing model PR. The problem is that right now we use the metadata to decide whether there is a voltage series at all, which I think is backwards. The TDT interface, which is where it shows up, should declare it at init, and both consumers should read that declaration: the metadata to know whether to build the entry for the user to fill in, and the alignment so the series can be indexed, since it has its own sampling frequency (Dv1A at 1017 Hz against Fi1d's 6103, on the same clock). That last part is also why I don't think we can just drop the separate timestamps: what they share is a clock and an origin, not an array.

pauladkisson: Do you want to address this in this PR or defer it to a separate one?

pauladkisson (approved):

src/neuroconv/datainterfaces/fiber_photometry/basefiberphotometryinterface.py
h-mayorquin: I am unsure how to move forward. The above was just a suggestion, let me think on it a bit.

PR #1922, opened 2026-08-14
Add `set_times` and `remap_times` to the fiber photometry interfaces

h-mayorquin:
On #1919 @pauladkisson made the case that adding full support in one go is less work than a gradual rollout, fewer passes through the codebase and less intermediate handling. This PR does that for fiber photometry: `alignment` becomes the mapping #1793 describes, with `keys()` and per-object access, and it gains `set_times` and `remap_times` beside `shift_times`. The three older methods route into them behind a `FutureWarning` now that each of them has a successor, and the exclusivity shim from #1919 goes away here since there is nothing left to exclude.

It is based on #1919 rather than on main, so the diff is only what full support adds.


src/neuroconv/_temporal_alignment.py
```diff
@@ -1,29 +1,160 @@
...
+        self._offset = 0.0
+
+    def shift_times(self, delta: float) -> None:
+        """Shift this object alone by ``delta`` seconds (relative, accumulates).
+
+        This breaks the relationship between this object and its siblings, which the interface-wide shift is
+        there to protect, so it is for a deliberate single-object correction (a cable latency on one stream)
+        rather than for placing an interface.
+        """
+        self._offset += float(delta)
```
pauladkisson: I think this is right, but it's worth perhaps documenting somewhere. 

These two operations do not commute, so if you set times and then shift times, it's not the same as if you shift times and then set times.

src/neuroconv/_temporal_alignment.py
```diff
@@ -1,29 +1,160 @@
...
-    time. ``shift_times`` is relative and accumulates.
+    Carries the interface-wide offset, ``output = native + offset``, default ``0.0`` (identity), and names the
+    objects the interface registered. ``shift_times`` and ``remap_times`` need no key, as one correction
+    applies to every object; ``set_times`` is per-object literal values, so it needs one unless the interface
+    has exactly one object to mean.
     """
 
     def __init__(self):
         self._offset = 0.0
+        self._objects: dict[str, _TimeSeriesAlignment] = {}
```
pauladkisson: This could use a better name. Perhaps `name_to_timeseries_alignment`?

src/neuroconv/_temporal_alignment.py
```diff
@@ -1,29 +1,160 @@
...
+        """
+        if len(self._objects) != 1:
+            named = ", ".join(repr(name) for name in self._objects) or "none"
+            raise ValueError(
+                "`set_times` writes the times of one object, so it can only be called without a key on an "
+                f"interface that has exactly one. This one names {named}. Reach the object you mean with "
+                "`alignment[key].set_times(times)`."
+            )
+        (single_object,) = self._objects.values()
+        single_object.set_times(times)
```
pauladkisson: I don't know if this convenience method is really worth the trouble. Seems more confusing than helpful. 

"There should be one, and preferably only one, obvious way to do it."

src/neuroconv/_temporal_alignment.py
```diff
@@ -1,29 +1,160 @@
...
+    def remap_times(self, *, stream_sync_times, reference_sync_times) -> None:
+        """Re-express every time-bearing object on a reference clock through synchronization pulses.
+
+        One acquisition clock means one correction, so this needs no key. See
+        :meth:`_TimeSeriesAlignment.remap_times`.
+        """
+        for single_object in self._objects.values():
+            single_object.remap_times(stream_sync_times=stream_sync_times, reference_sync_times=reference_sync_times)
+
+    def _get_times(self, key: str) -> np.ndarray:
```
pauladkisson: Okay, this is similarly a little confusing. Is the user supposed to be using `interface.alignment._get_times[key]` or `interface.alignment[key].get_times()`?

src/neuroconv/_temporal_alignment.py
```diff
@@ -1,29 +1,160 @@
...
+    its own. The source times are never mutated, so what is written is
+    ``(replacement or native) + own offset + the interface's offset``, the last of which is added by the
+    container.
+    """
+
+    def __init__(self, *, get_native_times):
+        # A callable rather than an array, so an interface registers its objects without reading its source.
+        self._get_native_times = get_native_times
+        self._times: np.ndarray | None = None
+        self._offset = 0.0
```
pauladkisson: After reflecting on this a bit, I think that storing both `_times` and `_offset` as attributes causes some problems. 

- Whenever you set the times, you have to reset the offset, which is a non-obvious side effect. 
- The local time series offsets have to coordinate with the global interface offset, which makes the whole temporal alignment system a bit more complicated.

Plus, a local shift could be easily represented by simply shifting the `_times` array. The one downside I see is that you lose lazy shifting, but I don't think that is really that bad.

src/neuroconv/datainterfaces/fiber_photometry/basefiberphotometryinterface.py
```diff
@@ -109,51 +110,62 @@ def _get_stream_timestamps(self, *, stream_name: str) -> np.ndarray:
...
-            )
-        return self._alignment
-
     def get_original_timestamps(self) -> np.ndarray:
         """Return the original (unaligned) timestamps of this interface's primary stream."""
         return self._get_stream_timestamps(stream_name=self.stream_names[0])
 
     def get_timestamps(self) -> np.ndarray:
-        """Return this interface's times: the aligned ones if set, otherwise the original ones, shifted.
+        """Return the times this interface's response series will be written on.
```
pauladkisson, line 120: This should also be deprecated, maybe, in favor of a public alignment method.

src/neuroconv/datainterfaces/fiber_photometry/basefiberphotometryinterface.py
```diff
@@ -109,51 +110,62 @@ def _get_stream_timestamps(self, *, stream_name: str) -> np.ndarray:
...
+        .. deprecated::
+            Use ``interface.alignment.remap_times(stream_sync_times=..., reference_sync_times=...)``, whose
+            argument names say which clock each set of pulses came off. Removed on or after August 2027.
+        """
+        warnings.warn(
+            "`align_by_interpolation` is deprecated and will be removed on or after August 2027. Use "
+            "`interface.alignment.remap_times(stream_sync_times=..., reference_sync_times=...)` instead.",
+            FutureWarning,
+            stacklevel=2,
+        )
```
pauladkisson, line 176: Aren't these both new methods? Why are we adding them just to deprecate them?

pauladkisson (commented):
Glad we are doing the full model because this revealed some rough edges that I did not anticipate.

src/neuroconv/_temporal_alignment.py
h-mayorquin: I agree with this. Now the per-object offset is gone and so is the per-object `shift_times` that was the only thing writing to it. In fact this was not in  #1793 and I think my agent implementd this under a misguided sense of symmetry.

src/neuroconv/_temporal_alignment.py
h-mayorquin: After the changes describe on https://github.com/catalystneuro/neuroconv/pull/1922#discussion_r3786606467 I think this disolves away.

The interface offset applies on top of whatever the object holds, so shift-then-set and set-then-shift both land on `times + offset`

src/neuroconv/_temporal_alignment.py
h-mayorquin: Changed to`_name_to_time_bearing_object` as that is what I am using on the docs in https://github.com/catalystneuro/neuroconv/pull/1793.

src/neuroconv/datainterfaces/fiber_photometry/basefiberphotometryinterface.py
h-mayorquin: Because they are inherited from the parent class so we need to surface the migration to the user at this level as they keep working in other modalities. Similar to that cycles of migration in roiextractors simplifications.

src/neuroconv/_temporal_alignment.py
h-mayorquin: The second, and the first is gone. The object now holds a reference to the alignment it was registered on, so `get_times()` returns the times that will actually be written, interface offset included, and there is one read rather than two that differ by an offset.

src/neuroconv/_temporal_alignment.py
h-mayorquin: I re-read #1793 and I was wrong, the per-object shift_times is there (sorry to my agent). That said, I can't think of a use for it, and for the sake of simplicity I would rather delay having bot the per-interface offset and a per-object offset.

One possible use is lazy shifting of an internal time-bearing object but can't think on a concrete case.

src/neuroconv/_temporal_alignment.py
h-mayorquin: This and [your other review comment](https://github.com/catalystneuro/neuroconv/pull/1922#discussion_r3785393988) are related. A keyless `set_times` and a keyless `get_timestamp` both assume an interface has one time-bearing object, and for an interface that has several the notion is incoherent.

They are  syntactic sugar for the single-object interfaces or for objects like pose estimation where everything is aligned. This is most of them so I went with that first. 

Thinking this over I would rather take your instinct, that there should be one obvious way, so the keyless setter is deleted and times are read through the object, `alignment[key].get_times()`. We can add the sugar later if anyone asks for it, or if it turns out we want it once the full design has landed, and it is easier to add then than to take back.

src/neuroconv/datainterfaces/fiber_photometry/basefiberphotometryinterface.py
h-mayorquin: see https://github.com/catalystneuro/neuroconv/pull/1922#discussion_r3786835758 but the caveats of https://github.com/catalystneuro/neuroconv/pull/1922#discussion_r3786625012 apply.

src/neuroconv/_temporal_alignment.py
```diff
@@ -1,29 +1,129 @@
...
+is a single sample axis and takes ``set_times`` while a table is timestamps plus durations and cannot, there
+being no one array to hand it. Only the series-shaped one exists here so far.
 """
 
+from collections.abc import Iterator
+
+import numpy as np
+
+
+class _TimeBearingSeries:
```
pauladkisson, line 24: This is a bit of an aside, but I don't think that this object necessarily needs to be a TmeSeries, any time-bearing object should qualify here.

src/neuroconv/datainterfaces/fiber_photometry/basefiberphotometryinterface.py
h-mayorquin: I changed this to a future warning that points the public alignment method.

src/neuroconv/_temporal_alignment.py
```diff
@@ -1,29 +1,129 @@
...
+        reference_sync_times = np.asarray(reference_sync_times)
+        if local_sync_times.shape != reference_sync_times.shape:
+            raise ValueError(
+                "The synchronization pulses have to pair up: `local_sync_times` has "
+                f"{local_sync_times.size} of them and `reference_sync_times` has {reference_sync_times.size}."
+            )
+        remapped_times = np.interp(self.get_times(), local_sync_times, reference_sync_times)
+        # These are the times to be written, and ``get_times`` adds the interface's offset on the way back
+        # out, so what is stored is the remapped times less that offset.
+        self._times = remapped_times - self._alignment.offset
```
pauladkisson: I want to consider this case in a little bit of detail. 

The way that this system works is that there are essentially three different time bases. 

1. Unaligned object time - the raw timestamps that you get from self.get_times
2. Aligned interface time - this object's timestamps are aligned to the rest of the timestamps in this interface (`self._times`).
3. Aligned NWB time - this object's timestamps are aligned to the global session clock (`remapped_times == self.get_times()`).

Maybe this is okay, but it was a little bit confusing to me. And it wasn't clear until I really thought it through which time basis `reference_sync_times` were supposed to be on. Now that I have thought it through, it's clear that reference sync times ought to be on a time basis 3, which seems appropriate.

 I'm wondering if we can do better, but I'm not sure that we can. At the very least, this nuance should be documented.

pauladkisson (commented):
Definitely moving in the right direction here. A couple more comments.

src/neuroconv/_temporal_alignment.py
h-mayorquin: The name is mostly in opposition to `DynamicTable`, which has different needs when remapping (e.g. it carries a duration) rather than to one array. The abstraction really is "a thing with a single time axis", and `TimeSeries` is the object that instantiates it, so I named it after that. 

I can make the class docstring say the criterion is the single time axis rather than the NWB type, if that reads better.

pauladkisson: I think that the structure defined in `_TimeBearingSeries` will suffice for all cases, including `DynamicTable`, But it is decidedly out of scope for this PR, which focuses on the fiber photometry interfaces, so we can cross that bridge when we get to it.

src/neuroconv/_temporal_alignment.py
h-mayorquin: I would call it two bases rather than three. Times are always read as `_times + offset`, so a remap writes `_times` with the offset already subtracted, purely so that the read adds it back and lands on the remapped values. That is a storage detail rather than a basis of its own, and it only exists at all if someone shifted before remapping: with no shift the two are the same code and `_times` holds the reference-clock times outright. I think that in most cases users will not shift and then remap as well.

The rule is that every operation acts on the times as they currently stand, so the one thing a caller has to get right is that `local_sync_times` is on the timeline the object currently reports, carrying a shift if one was applied. `reference_sync_times` cannot vary this way, since it is on the clock being aligned to.

I did improve the docstring of this method, though, so check it out.

src/neuroconv/_temporal_alignment.py
```diff
@@ -1,29 +1,156 @@
...
+        - ``local_sync_times``: the times the camera's own log gives for the same pulses. Those are on the
+          (local) camera clock.
+
+        So in this case, remap is called on the camera interface to remap its times towards the
+        electrophysiology board.
+
+        Times falling between two local pulses are interpolated proportionally, and nothing is resampled:
+        the data is left alone and only the times move. The two arrays pair up positionally, so a pulse only
+        one of the systems recorded has to be dropped from the other array as well, and a mismatched count
+        raises rather than guessing an offset. Samples outside the first and last local pulse are clamped
```
pauladkisson: This reminds me we should probably expose the numpy.interp kwargs as an option for this method. 

That way, users can do whatever they want in terms of extrapolation.

src/neuroconv/_temporal_alignment.py
pauladkisson: > I think that in most cases users will not shift and then remap as well.

I agree, these are niche details. In most cases, I would expect a user will either shift or set or remap and likely will not mix these methods.

> That is a storage detail rather than a basis of its own

Okay, so if we want to avoid exposing the interface time basis to the user because it's a storage detail and doesn't represent a legitimate time basis, then `set_times` should also subtract the offset. Currently, `set_times` requires the times that the user passes in to be on the interface time basis rather than the final global NWB time basis. 

If we make this change, set_times and remap_times become symmetrical. Importantly, these methods will not commute with shift times, because the user is expected to set the times that they want in the final NWB file.

I realize that this walks back a change that I suggested earlier. Thanks for bearing with me. I am coming to understand the requirements as I am reviewing.

h-mayorquin: No, thanks to you. I had not noticed that `set_times` was not doing what I had in my head: I assumed the values you pass are the ones the file gets, and they were not, the interface offset was still being added on top of them. So your suggestion fixes this. The change is pushed now along with some tests.

There are two readings of `set_times` and they were not as clear to me before this discussion:

**`shift_times` and `set_times` commute.** The argument is the object's times *inside* the interface, so the file gets `times + offset`. It composes with a shift and commutes with it, but `set_times(get_times())` is not a no-op, it moves the object by the offset a second time.

**`set_times` and `get_times` round-trip.** The argument is what the file will contain, so the file gets `times`. It supersedes an earlier shift for that object, it does not commute with one, and it round-trips with `get_times`.

While both are internally consistent views, I think the second is better because the surprise shows up at the level of the objects the user interacts with. Under the first, the file quietly disagrees with the numbers you passed, and you only find out on reading the file back. If `get_times()` always returns the timestamps the file will carry, the user has immediate access to what will be written at all times.

The second principle also means the getter and the setter finally speak the same language, which is how most people will learn the API.

I have written this into the user guide on #1793 so it is settled as we move forward: https://neuroconv--1793.org.readthedocs.build/en/1793/user_guide/temporal_alignment.html#fine-alignment

src/neuroconv/_temporal_alignment.py
h-mayorquin: Done, in 37c04c877cdf906268a30b5670cb603184a94b60. Rather than a kwargs passthrough, `remap_times` takes an `interpolation_function` that is called as `f(times, local_sync_times, reference_sync_times)`, which is `numpy.interp`'s own signature, so `partial(np.interp, left=np.nan, right=np.nan)` covers your case and any other scheme is a short adapter.gs.

Documented here, with both cases: https://neuroconv--1793.org.readthedocs.build/en/1793/user_guide/temporal_alignment.html#choosing-how-the-map-is-built

src/neuroconv/_temporal_alignment.py
h-mayorquin: Let's hope so, this will make it simpler, but I expect that remapping would require internals for the events case when we also need to remap the duration if we remap the events. 

Let's keep it private, and remember to have a better name if it never splits by case.

pauladkisson (approved):
Looks good to me. Thank you for addressing the feedback.

h-mayorquin:
> Looks good to me. Thank you for addressing the feedback.

Thanks, can you approve the base #1919 as well?

pauladkisson:
> > Looks good to me. Thank you for addressing the feedback.
> 
> Thanks, can you approve the base #1919 as well?

I can, but there's still the outstanding issue of the commanded voltage time stamps.

PR #1983, opened 2026-08-25
Fix Dallies

h-mayorquin:
The dailies have been red since #1973 moved `--numprocesses` into `addopts`, and the two docker jobs install a bare `pytest` so they fail at argument parsing. This PR installs `pytest-xdist` in both and moves them to the weeklies. The images they exercise are built on release and weekly, so a daily run was re-testing an unchanged artifact.


h-mayorquin:
I dispatched both docker workflows on this branch and they pass, 3 tests and 1 test:

https://github.com/catalystneuro/neuroconv/actions/runs/32908379120
https://github.com/catalystneuro/neuroconv/actions/runs/32908385244

The `[gw0]` prefix on every test is the part to look at, as it means the xdist workers started, which is what was failing at argument parsing. The move to the weeklies is not exercised by those runs and only shows up once `weeklies.yml` runs on schedule after the merge.

PR #2007, opened 2026-09-05
Keep nested `DeepDict` values as `DeepDict`s through `deep_update`

### LINKED ISSUE #2006: DeepDict.deep_update raises AttributeError on a second update into the same key
`DeepDict.deep_update` stores a new key's value as whatever it was handed, so a plain dict goes in as a plain dict. A later `deep_update` that reaches the same key sees a dict on both sides and calls `self[key].deep_update(value)`, which a plain dict does not have.

```python
from neuroconv.utils import DeepDict

d = DeepDict()
d.deep_update({"a": {"b": 1}})
type(d["a"])  # dict, not DeepDict
d.deep_update({"a": {"c": 2}})
# AttributeError: 'dict' object has no attribute 'deep_update'
```

The same thing happens for a value assigned directly (`d["a"] = {"b": 1}`) and then updated into. The constructor converts nested dicts into `DeepDict`s, so the class already intends nested values to be `DeepDict`s; `deep_update` and `__setitem__` just do not keep that invariant. The fix is to convert a dict value to a `DeepDict` when it is stored, and to convert an existing plain dict before recursing into it.

bendichter:
Closes #2006.

`DeepDict.deep_update` stored a new key's dict value as the plain dict it was handed, so a second update reaching that key called `self[key].deep_update(value)` on something that has no such method. The constructor already converts nested dicts into `DeepDict`s, so this makes `deep_update` keep the same invariant: a dict value is stored as a `DeepDict`, and a plain dict already sitting under the key (assigned through `__setitem__`) is converted before the update recurses into it.

Two regression tests cover the two ways a plain dict could end up there. The code and tests were drafted by the AI.


h-mayorquin (changes requested):
Discussion:

Should this not go in `__setitem__` and `update` instead, so the class keeps the invariant itself and `deep_update` needs no branches at all? `dict.update` is C-implemented and does not go through `__setitem__`, so it needs its own override.

I think that is better for three reasons. It fixes something already shipping, since `AudioInterface.get_metadata` does `metadata.update(Behavior=behavior_metadata)` at `audiointerface.py:134`, so `type(metadata["Behavior"])` is `dict` on main and this PR does not reach it. It makes the code smaller instead of bigger, because with the invariant held at the door both branches added here become unreachable. And it closes the class of bug instead of the instance: after this PR, `metadata["Subject"] = {...}` followed by `metadata["Subject"]["x"]["y"] = 1` still raises `KeyError`.

PR #2009, opened 2026-09-05
Forward `compare_key` and `list_dict_deep_update` through `dict_deep_update`

### LINKED ISSUE #2008: dict_deep_update drops compare_key and list_dict_deep_update below the top level
`dict_deep_update` takes `compare_key` and `list_dict_deep_update`, but the recursive call it makes for a nested mapping forwards only `append_list` and `remove_repeats`, so both fall back to their defaults one level down. The same happens in `append_replace_dict_in_list`, which calls `dict_deep_update(ls[idx], d)` with no arguments at all when it merges two list entries.

The visible symptom is a `KeyError` on the default `compare_key`:

```python
from neuroconv.utils import dict_deep_update

d = {"Ecephys": {"Electrodes": [{"id": 1, "desc": "old"}]}}
u = {"Ecephys": {"Electrodes": [{"id": 1, "desc": "new"}]}}
dict_deep_update(d, u, compare_key="id")
# KeyError: 'name'
```

The same call on `d["Ecephys"]` and `u["Ecephys"]` works, since that is the top level. `list_dict_deep_update=False` is ignored the same way, silently, so a nested list entry is merged where the caller asked for it to be replaced. Every in-repo caller uses the default `compare_key="name"`, which is why this has not come up. The fix is to forward the arguments in both calls.

bendichter:
Closes #2008.

`dict_deep_update` forwarded only `append_list` and `remove_repeats` into the recursive call it makes for a nested mapping, and `append_replace_dict_in_list` forwarded nothing when it merged two matched list entries, so `compare_key` and `list_dict_deep_update` fell back to their defaults below the top level. A `compare_key` other than `name` raised `KeyError` there, and `list_dict_deep_update=False` was ignored. Both calls now carry the arguments.

`copy` is passed as `False` in the recursion. The top-level call has already deep-copied everything below it when asked to, so the inner copies were redundant, and when the caller asked for no copy the nested dicts are now updated in place like the top one instead of being replaced by copies. A test pins that.

The code and tests were drafted by the AI.


h-mayorquin (approved):
This looks good to me. Two things:

The changelog says that a `compare_key` other than `name` "no longer raises `KeyError` below the top level". It still does when an existing entry in the list has no such key, as the comparison is asymmetric:

```python
[d.get(compare_key, None) == i[compare_key] for i in ls if isinstance(i, collections.abc.Mapping)]
```

With this branch applied, `dict_deep_update(a, b, compare_key="id")` where one existing entry has no `id` throws `KeyError: 'id'`. That is not something you introduced, but the entry should not say it is gone.

The other one is `copy`. The new test pins that with `copy=False` a nested dict is updated in place, and a list entry is not, as `append_replace_dict_in_list` calls `dict_deep_update(ls[idx], d)` without forwarding it and so gets the default:

```python
entry = {"name": "x", "desc": "old"}
a = {"items": [entry]}
dict_deep_update(a, {"items": [{"name": "x", "desc": "new"}]}, copy=False)
a["items"][0] is entry   # False, and entry is left unchanged
```

I don't think these should be blocking because now that the metadata migration to dict is almost over few things that hit this remain, #2001 takes `Ecephys.Electrodes` off the list, and besides I am not sure what the user has to be doing to hit those. So, I think fixing only this is Ok if you want to merge as it is.

PR #2011, opened 2026-09-06
Fix the YAML CLI's default output folder and `nwbfile_name` suffix handling

### LINKED ISSUE #2010: YAML CLI fails without --output-folder-path, and nwbfile_name loses leading or trailing n, w, b characters
Two problems in `run_conversion_from_yaml`, both in how it handles paths and names.

The `neuroconv` command line entry point passes the specification path as a `str`, and `run_conversion_from_yaml` calls `.parent` on it when `--output-folder-path` is not given:

```
$ neuroconv spec.yml
  File ".../_yaml_conversion_specification.py", line 83, in run_conversion_from_yaml
    output_folder_path = specification_file_path.parent
AttributeError: 'str' object has no attribute 'parent'
```

The `data_folder_path` branch a few lines above wraps its argument in `Path`, and the output branch should do the same. The existing CLI tests always pass `--output-folder-path`, which is why this went unnoticed.

Separately, the output file name is derived with `session["nwbfile_name"].strip(".nwb")`. `str.strip` removes any of the characters `.`, `n`, `w`, `b` from both ends rather than the suffix, so `brain_waves.nwb` is written as `rain_waves.nwb`, `nwb_session_1.nwb` as `_session_1.nwb`, and `session_web.nwb` as `session_we.nwb`. `removesuffix(".nwb")` is what was meant.

bendichter:
Closes #2010.

The `neuroconv` entry point hands the specification path over as a string, and `run_conversion_from_yaml` called `.parent` on it to derive the default output folder, so the command failed with `AttributeError: 'str' object has no attribute 'parent'` whenever `--output-folder-path` was left out. The path is now normalized to a `Path` once at the top of the function, which is also what the `data_folder_path` branch already did for its own argument.

In the same function, the `.nwb` suffix of a session's `nwbfile_name` was removed with `str.strip`, which strips any of the characters `.`, `n`, `w`, `b` from both ends rather than the suffix, so `brain_waves.nwb` came out as `rain_waves.nwb`. It now uses `removesuffix`.

The new test drives the real `neuroconv` command with no `--output-folder-path` against a specification that declares no data interfaces, so it needs no test data. It fails on `main` with the error above and passes here. It lives with the other YAML tests under `test_on_data` because the function imports `dandi`, which the minimal test jobs do not install.

The code and test were drafted by the AI.


h-mayorquin (approved):
Clean fix.

tests/test_on_data/test_yaml/test_yaml_conversion_specification_cli.py
```diff
@@ -121,3 +121,38 @@ def test_run_conversion_from_yaml_default_nwbfile_name(self):
...
 if __name__ == "__main__":
     unittest.main()
+
+
+def test_run_conversion_from_yaml_cli_without_output_folder_path(tmp_path):
+    # No data interfaces, so this needs no test data: the point is the path handling around the conversion.
+    # The output folder defaults to the specification's own folder, and the file name is the session's
+    # ``nwbfile_name`` with its ``.nwb`` suffix removed as a suffix, so the leading ``b`` survives.
+    yaml_file_path = tmp_path / "specification.yml"
+    yaml_file_path.write_text(
```
h-mayorquin, line 131: This is missing the encoding and that's why the test is failing.

PR #2013, opened 2026-09-06
Let `get_default_dataset_io_configurations` auto-detect the backend

### LINKED ISSUE #2012: get_default_dataset_io_configurations(backend=None) raises KeyError instead of auto-detecting the backend
`get_default_dataset_io_configurations` documents `backend=None` as "auto-detect from append mode", and its error messages tell the user to "set `backend=None` or remove the keyword argument to allow it to auto-detect". That path cannot be reached: the first line of the function is

```python
DatasetIOConfigurationClass = DATASET_IO_CONFIGURATIONS[backend]
```

so `backend=None` raises `KeyError: None` before any of the detection code runs, whatever the file's state.

```python
from neuroconv.tools.nwb_helpers._dataset_configuration import get_default_dataset_io_configurations

list(get_default_dataset_io_configurations(nwbfile=mock_NWBFile(), backend=None))
# KeyError: None   (the intended error is the ValueError asking for a backend)
```

Two smaller things in the same guard block. The check for a file opened read-only uses `nwbfile.read_io.mode`, which `NWBZarrIO` does not expose (the module has `_get_io_mode` for exactly this reason), so a Zarr file read in mode `r` would raise `AttributeError` there. And that check's message says the backend must be specified "if the `nwbfile` is being appended", when it fires precisely because the file is not open for appending.

The lookup should move below the detection block, and the guard should use `_get_io_mode` with a message that describes the case it catches.

bendichter:
Closes #2012.

`get_default_dataset_io_configurations` documents `backend=None` as auto-detect from an appended file, and its own error messages tell the user to set it that way, but the configuration class was looked up from `backend` on the function's first line, before the detection ran. `None` therefore raised `KeyError` in every case. The lookup now follows the detection block, where `backend` has been resolved.

In the same guard block, the check for a file read without appending used `nwbfile.read_io.mode`, which `NWBZarrIO` does not expose; it now goes through `_get_io_mode`, which the module already has for that purpose. Its message also said the backend must be given "if the nwbfile is being appended", when it fires because the file is not open for appending, so it now says that.

Five tests cover auto-detection from an appended HDF5 and Zarr file, the in-memory case, and the read-only case for both backends. All five fail on `main` and pass here. The code and tests were drafted by the AI.


src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
```diff
@@ -93,16 +93,15 @@ def get_default_dataset_io_configurations(
...
         raise ValueError(
             "Keyword argument `backend` (either 'hdf5' or 'zarr') must be specified if the `nwbfile` was not "
             "read from an existing file!"
         )
-    if backend is None and nwbfile.read_io is not None and nwbfile.read_io.mode not in ("r+", "a"):
+    if backend is None and nwbfile.read_io is not None and _get_io_mode(io=nwbfile.read_io) not in ("r+", "a"):
         raise ValueError(
-            "Keyword argument `backend` (either 'hdf5' or 'zarr') must be specified if the `nwbfile` is being appended."
+            "Keyword argument `backend` (either 'hdf5' or 'zarr') must be specified if the `nwbfile` was read "
+            "from an existing file without opening it for appending (mode 'r+' or 'a')!"
```
bendichter, line 104:
```suggestion
            "from an existing file without opening it for appending (mode 'r+' or 'a')."
```

src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
```diff
@@ -93,16 +93,15 @@ def get_default_dataset_io_configurations(
...
 
-    DatasetIOConfigurationClass = DATASET_IO_CONFIGURATIONS[backend]
-
     if backend is None and nwbfile.read_io is None:
         raise ValueError(
             "Keyword argument `backend` (either 'hdf5' or 'zarr') must be specified if the `nwbfile` was not "
             "read from an existing file!"
         )
-    if backend is None and nwbfile.read_io is not None and nwbfile.read_io.mode not in ("r+", "a"):
+    if backend is None and nwbfile.read_io is not None and _get_io_mode(io=nwbfile.read_io) not in ("r+", "a"):
```
h-mayorquin, line 101: `NWBZarrIO` does have a mode attribute, why does the changelog says this is not longer available?

h-mayorquin (changes requested):
Looks fine to me but just a request of something that the changelog claims.

src/neuroconv/tools/nwb_helpers/_dataset_configuration.py
bendichter: You're right, ZarrIO has had a public `mode` since 0.11.0. Dropped the helper and use `.mode` directly; the changelog line is fixed.

h-mayorquin (approved):
Thanks.

h-mayorquin:
I will merge here and fix the thing that is failing separately so take this out of your plate.

PR #2015, opened 2026-09-06
Write a masks-only segmentation under default metadata

### LINKED ISSUE #2014: add_segmentation_to_nwbfile with default metadata refuses an extractor that has ROIs but no traces
`add_segmentation_to_nwbfile` accepts an extractor with ROIs and no traces when the caller passes metadata that names no `RoiResponses` block (there is a test for that, `test_no_roi_responses_no_traces`). It refuses the same extractor when the caller passes no metadata at all:

```python
from roiextractors.testing import generate_dummy_segmentation_extractor
from neuroconv.tools.roiextractors import add_segmentation_to_nwbfile

segmentation = generate_dummy_segmentation_extractor(
    num_rois=3, has_raw_signal=False, has_dff_signal=False, has_deconvolved_signal=False, has_neuropil_signal=False
)
add_segmentation_to_nwbfile(segmentation_extractor=segmentation, nwbfile=mock_NWBFile())
# ValueError: RoiResponses metadata was provided but the segmentation extractor has no trace data.
```

With `metadata=None` the function fills in `_get_ophys_metadata_placeholders()`, and that template carries a `RoiResponses` entry under `default_metadata_key`. `_add_roi_response_traces_to_nwbfile` then sees the key present and raises the "metadata was provided" error, although nobody provided anything. A few lines lower the same function already knows the difference and computes `user_provided_roi_responses_metadata`, which excludes the placeholder key; the raise just runs before that and uses the wider test. The `_segmentation_extractor_has_data` check at the entry point deliberately admits a masks-only extractor, so this is the one path where it is then rejected anyway.

The raise should apply only to metadata the caller wrote.

bendichter:
Closes #2014.

`add_segmentation_to_nwbfile` fills in `_get_ophys_metadata_placeholders()` when the caller passes no metadata, and that template carries a `RoiResponses` entry under `default_metadata_key`. `_add_roi_response_traces_to_nwbfile` saw the key present and raised "RoiResponses metadata was provided but the segmentation extractor has no trace data" for an extractor with ROIs and no traces, although nobody had provided anything. The same function already distinguished caller-written metadata from the placeholder a few lines lower, in `user_provided_roi_responses_metadata`; the raise now uses that distinction, so a masks-only segmentation converts under default metadata the way it already did under caller metadata that named no traces.

The new test sits beside `test_no_roi_responses_no_traces`, which covers the caller-metadata case, and fails on `main` with the error above. The code and test were drafted by the AI.


src/neuroconv/tools/roiextractors/roiextractors.py
h-mayorquin: This should be changed as well but this is minor.

h-mayorquin (approved):
LGTM, only one docstring suggestions so it does not reflect the old behavior.

src/neuroconv/tools/roiextractors/roiextractors.py
bendichter: Good catch, updated. The docstring now says the placeholder entry under default_metadata_key does not count as caller metadata, and that an extractor with no traces writes nothing in that case. Also rebased onto main to clear the changelog conflict.

PR #2019, opened 2026-09-06
Key the imaging base's default series entry the way the writer reads it

### LINKED ISSUE #2018: BaseImagingExtractorInterface.get_metadata keys the series by None when no metadata_key is set, and the writer cannot find it
`BaseImagingExtractorInterface.__init__` accepts `metadata_key=None`, and in the dict-based format `get_metadata` emits

```python
metadata["Ophys"] = {"MicroscopySeries": {self.metadata_key: dict(name="MicroscopySeries")}}
```

with that `None` as the key. `add_imaging_to_nwbfile`, on the other side, falls back to `"default_metadata_key"` when handed `metadata_key=None`, so the entry the interface produced is never looked up:

```python
from neuroconv.datainterfaces.ophys.baseimagingextractorinterface import BaseImagingExtractorInterface

class MinimalImagingInterface(BaseImagingExtractorInterface):
    @classmethod
    def get_extractor_class(cls):
        from roiextractors.testing import generate_dummy_imaging_extractor
        return generate_dummy_imaging_extractor

interface = MinimalImagingInterface(num_samples=5, num_rows=4, num_columns=4)
list(metadata["Ophys"]["MicroscopySeries"])   # [None]
interface.add_to_nwbfile(mock_NWBFile(), metadata=metadata)
# KeyError: 'default_metadata_key'
```

Every imaging interface in the repository picks its own key when none is given, which is why this has not come up, but the base class is the documented way to write a new imaging interface and a downstream one that leaves the key alone breaks on the first write. `get_metadata_template` in the same class already resolves the key as `self.metadata_key or "default_metadata_key"`; `get_metadata` should do the same, so the two sides agree.

bendichter:
Closes #2018.

`BaseImagingExtractorInterface.get_metadata` filed its `MicroscopySeries` entry under `self.metadata_key`, which is `None` for an interface constructed without one, while `add_imaging_to_nwbfile` falls back to `"default_metadata_key"` for that same interface. The two sides never met, so a direct subclass that left the key alone raised `KeyError: 'default_metadata_key'` on its first write. The entry now goes under the same default the writer uses, which is also what `get_metadata_template` in this class already did.

Every imaging interface in the repository picks its own key, so nothing in-tree changes behaviour; the case is a downstream subclass of the base. The two tests build such a subclass over the dummy extractor and check both the key and the write. Both fail on `main`. The code and tests were drafted by the AI.


h-mayorquin (changes requested):
I think this only surfaces when someone subclasses the base, since the interfaces that would otherwise hit it, the Bruker ones, still return old-format metadata and never reach that line, which is fine. The [pattern](https://github.com/catalystneuro/neuroconv/blob/main/docs/developer_guide/metadata_principles.rst#the-metadata_key-parameter) is to resolve this at init and we already do it on most interfaces. I think we should do the same here as well.

PR #2021, opened 2026-09-06
Stop two helpers from writing into their callers' inputs

### LINKED ISSUE #2020: get_json_schema_from_method_signature and convert_df_to_time_intervals modify their callers' inputs
Two helpers write into arguments they were handed.

`get_json_schema_from_method_signature(method, exclude=...)` does `exclude += ["self", "cls"]`, which extends the caller's list in place:

```python
exclude = ["b"]
get_json_schema_from_method_signature(f, exclude=exclude)
exclude  # ['b', 'self', 'cls']
```

A list reused across calls, or a module-level constant, accumulates the two names on every call.

`convert_df_to_time_intervals(df, column_name_mapping=...)` renames the columns of the caller's DataFrame with `inplace=True`, and when the frame has no `stop_time` column it adds one to the caller's frame as well:

```python
df = pd.DataFrame({"start": [0.0, 1.0], "cond": [1, 2]})
convert_df_to_time_intervals(df, column_name_mapping={"start": "start_time"})
list(df.columns)  # ['start_time', 'cond', 'stop_time']
```

`TimeIntervalsInterface.add_to_nwbfile` passes `self.dataframe` to it, so after one write the interface's own frame carries the renamed columns and a generated `stop_time`, and a second call with the same mapping finds nothing to rename and a `stop_time` that was not in the file. Both functions should work on a copy.

bendichter:
Closes #2020.

`get_json_schema_from_method_signature` extended the caller's `exclude` list in place with `"self"` and `"cls"`, so a list reused across calls, or a module-level constant, grew by two names on every call. It now builds a new list.

`convert_df_to_time_intervals` renamed the columns of the caller's DataFrame with `inplace=True` and, when the frame had no `stop_time`, added the generated one to the caller's frame too. `TimeIntervalsInterface.add_to_nwbfile` passes the interface's own frame, so after one write the frame carried the renamed columns and a `stop_time` that was not in the file, and a second write with the same `column_name_mapping` found nothing to rename. The function now works on a copy.

Three tests cover the untouched `exclude` list, the untouched DataFrame, and two consecutive writes through the CSV interface with a `column_name_mapping`. All three fail on `main`. The code and tests were drafted by the AI.


tests/test_modalities/test_text/test_time_intervals.py
```diff
@@ -34,6 +34,25 @@ def test_convert_df_to_time_intervals_no_start_time():
     assert time_intervals.colnames == ("start_time", "stop_time", "condition")
 
 
+def test_convert_df_to_time_intervals_leaves_the_dataframe_alone():
```
h-mayorquin, line 37: I think this might be redundant with some other tests already present but we can do a clean up later if and when suite test time becomes a problem (I recently did a speed up).

h-mayorquin (approved):
Makes sense. Thanks.

bendichter:
@copilot resolve the merge conflicts on this branch.

PR #2023, opened 2026-09-06
Let a session's YAML `conversion_options` override the global block

### LINKED ISSUE #2022: YAML specification: global conversion_options override a session's own conversion_options
In `run_conversion_from_yaml`, the options for each interface are assembled as

```python
conversion_options[key] = dict(session_conversion_options.get(key, dict()), **global_conversion_options)
```

The right-hand mapping wins in that construction, so a value in the top-level `conversion_options` block overrides the same option stated for a single session. A specification that sets `stub_test: True` globally to keep the run fast and then turns it off for one session, or that names a global `iterator_options` and refines it for one interface in one session, silently gets the global value for that session.

Metadata in the same function merges the other way: the global block, then the experiment's, then the session's, with the more specific one winning. The conversion options should follow the same order, so a session's own options override the global ones.

```yaml
conversion_options:
  stub_test: true

experiments:
  exp:
    sessions:
      - nwbfile_name: full_session
        source_data: {...}
        conversion_options:
          ap:
            stub_test: false   # currently ignored in favor of the global True
```

The specification schema and the docs say nothing about precedence, so this is a behaviour change only for a specification that states the same option at both levels, which is the case where the current result is the surprising one.

bendichter:
Closes #2022.

`run_conversion_from_yaml` built each interface's options as `dict(session_options, **global_options)`, so the top-level `conversion_options` block won wherever a session stated the same option. A session could not turn off a global `stub_test`, or refine a global option for one of its interfaces, and got the global value silently. The merge now runs the other way, with the session's values on top, which is the order the metadata blocks already merge in (global, then experiment, then session).

The merge moved into a small helper, `_resolve_conversion_options`, so it can be tested without a data-bearing specification. Three unit tests cover the global block reaching every interface, a session overriding and extending it for one interface while the other keeps the global value, and session options for an interface the specification does not declare being ignored. They live under `test_minimal`, since the module imports `dandi` only inside the run function.

Neither the schema nor the docs stated a precedence, so this changes behaviour only for a specification that names the same option at both levels. The code and tests were drafted by the AI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Removed the `hdf5` and `sbx` installation extras, aliases of `hdf5imaging` and `scanbox` kept for gallery pages that named the old spellings and marked for removal at the start of 2026. Use `neuroconv[hdf5imaging]` and `neuroconv[scanbox]`.
 
 ## Bug Fixes
+* In a YAML conversion specification, a session's own `conversion_options` now override the top-level `conversion_options` block where both state the same option, in the same order the metadata blocks merge. The global values used to win, so a session could not turn off a global `stub_test` or refine a global option for one of its interfaces. [PR #2023](https://github.com/catalystneuro/neuroconv/pull/2023)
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
I was not aware we had this functionality. This looks good to me.

PR #2027, opened 2026-09-06
Drop three `get_conversion_options_schema` overrides that repeat the parent

### LINKED ISSUE #2026: Three get_conversion_options_schema overrides repeat what the parent already does
`ConverterPipe.get_conversion_options_schema` is a line-for-line copy of `NWBConverter.get_conversion_options_schema`, differing only in its docstring.

`SpikeGLXConverterPipe` and `OpenEphysBinaryConverter` each override the method as

```python
conversion_options_schema = super().get_conversion_options_schema()
conversion_options_schema["properties"].update(
    {name: interface.get_conversion_options_schema() for name, interface in self.data_interface_objects.items()}
)
```

which re-adds, per interface, the entry the parent has just added. The one difference is that the parent passes each interface schema through `unroot_schema`, which keeps `required`, `properties`, `type`, `additionalProperties`, `title` and `description`, while the override keeps the raw schema. For every interface those two converters hold (`SpikeGLXRecordingInterface`, `SpikeGLXNIDQInterface`, `SpikeGLXSyncChannelInterface`, `OpenEphysBinaryRecordingInterface`, `OpenEphysBinaryAnalogInterface`) the generated schema holds only `additionalProperties`, `properties` and `type`, so `unroot_schema` returns it unchanged and the overrides produce the same dictionary the parent does.

All three can go. Two copies of the same logic are a place for the next change to land in only one of them.

bendichter:
Closes #2026.

`ConverterPipe.get_conversion_options_schema` was a line-for-line copy of the `NWBConverter` method. `SpikeGLXConverterPipe` and `OpenEphysBinaryConverter` each called the parent and then re-added every interface's schema on top of the entry the parent had just added. The one thing the overrides did differently was skip `unroot_schema`, and for the five interfaces those converters hold the generated schema carries only `additionalProperties`, `properties` and `type`, which `unroot_schema` keeps, so the override and the parent returned the same dictionary. I checked that by building each interface's schema from its `add_to_nwbfile` signature and comparing it against its unrooted form.

No behaviour changes; 42 lines go. The change was drafted by the AI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * `HDF5DatasetIOConfiguration` and `ZarrDatasetIOConfiguration` now describe a dataset's codec pipeline with an ordered `compressors` list and a matching `compressor_options`, which makes the HDF5 `shuffle` and `fletcher32` filters reachable for the first time and gives both backends the same vocabulary. [PR #1979](https://github.com/catalystneuro/neuroconv/pull/1979)
 
 ## Improvements
+* Removed the `get_conversion_options_schema` overrides on `ConverterPipe`, `SpikeGLXConverterPipe` and `OpenEphysBinaryConverter`, which repeated what `NWBConverter` already does and returned the same schema. [PR #2027](https://github.com/catalystneuro/neuroconv/pull/2027)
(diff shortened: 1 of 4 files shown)


h-mayorquin (approved):
LGTM

PR #2029, opened 2026-09-06
Make the `figshare` download helpers work outside a notebook

### LINKED ISSUE #2028: tools/figshare.py fails outside a notebook and relies on an undeclared dependency
`src/neuroconv/tools/figshare.py` holds two helpers, `download_article` and `download_collection`, for pulling a Figshare collection to disk. Nothing in the repository references the module: it is not imported by `neuroconv.tools`, not in the API docs, and has no tests. Its history is a 2022 import from the previous repository and one repository-wide encoding sweep.

It also cannot be relied on as it stands. It imports `requests` at module level, which is not a declared dependency of neuroconv and is only present when some other package happens to pull it in (`probeinterface` and `nwbinspector` do today). And it takes its progress bar from `tqdm.notebook`, which renders only inside Jupyter with `ipywidgets` installed; from a script it raises on first use:

```
ImportError: IProgress not found. Please update jupyter and ipywidgets.
```

so `download_collection` fails before downloading anything. Two smaller things: it creates the destination with `os.mkdir`, which fails when the parent folder does not exist, and the docstring says it writes `metadata.yaml` where the code writes `metadata.json`.

I would remove it. It is not part of the documented surface, it has no dependency footprint the package can honor, and a Figshare download is a generic task that `dandi`, `pooch` or a few lines of `requests` cover. If someone is using it, the alternative is to declare `requests`, use `tqdm.auto`, and give it a test.

bendichter:
Closes #2028.

This started as a removal, since nothing in the repository referenced `neuroconv.tools.figshare`; the helpers are useful for pulling a Figshare collection to disk, so it keeps them and makes them work instead.

The progress bar came from `tqdm.notebook`, which renders only inside Jupyter with `ipywidgets` present and otherwise raises `ImportError: IProgress not found`, so `download_collection` failed from a script before transferring anything. It now uses plain `tqdm`, which works in both places. `requests` is not a dependency of neuroconv and reached the module only through other packages; it is now imported on first use through `get_package`, so an environment without it gets "pip install requests" at the call rather than an `ImportError` at import time. A destination whose parent folder does not exist is created with `os.makedirs` rather than refused by `os.mkdir`, and the docstring now says `metadata.json`, which is what the code writes. The API responses also go through `raise_for_status`, so a failed request fails there rather than as a JSON error on an HTML body.

Two tests stand in for the API and the file transfer so nothing touches the network: one downloads a two-article collection into a folder with a missing parent, the other checks that a file already present at the right size is skipped. The code and tests were drafted by the AI.


h-mayorquin (approved):
I would prefer to remove this functionality as the original issue suggested. This has not been used.

That said, the code looks OK so I am fine with it either way.

bendichter:
I created this to help work with figshare data, and I used it in that instance, though I don't think it's been used since. It's sort of tangential to this library. I'd be OK with removing it.

PR #2031, opened 2026-09-06
Drop a dead constructor argument and a duplicate import

### LINKED ISSUE #2030: DatasetIOConfiguration is passed an object_name it does not declare; utils imports one name twice
Two small leftovers.

`DatasetIOConfiguration.from_neurodata_object_with_defaults` and `DatasetIOConfiguration.get_kwargs_from_neurodata_object` both pass `object_name=neurodata_object.name` into the model constructor, but no configuration model declares an `object_name` field. Pydantic's default `extra` policy is `ignore`, so the value is dropped without a word. Either the field was removed at some point and the two call sites were missed, or it was never added. Nothing reads it, and `location_in_file` already carries the name, so the argument should go rather than the field be added.

`neuroconv/utils/__init__.py` imports `get_json_schema_from_method_signature` twice in the same `from .json_schema import (...)` block.

bendichter:
Closes #2030.

`DatasetIOConfiguration.from_neurodata_object_with_defaults` and `DatasetIOConfiguration.get_kwargs_from_neurodata_object` both passed `object_name=neurodata_object.name` to the model constructor. No configuration model declares that field, and pydantic's default `extra` policy is `ignore`, so the value was dropped without a word; nothing reads it, and `location_in_file` already carries the name. Both call sites lose the argument.

`neuroconv/utils/__init__.py` imported `get_json_schema_from_method_signature` twice in one block; the second line goes.

No behaviour changes. The backend configuration and utils suites pass. The change was drafted by the AI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * `HDF5DatasetIOConfiguration` and `ZarrDatasetIOConfiguration` now describe a dataset's codec pipeline with an ordered `compressors` list and a matching `compressor_options`, which makes the HDF5 `shuffle` and `fletcher32` filters reachable for the first time and gives both backends the same vocabulary. [PR #1979](https://github.com/catalystneuro/neuroconv/pull/1979)
 
 ## Improvements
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
LGTM

bendichter:
@copilot resolve the merge conflicts on this branch.

PR #2033, opened 2026-09-06
Expose `always_write_timestamps` and `data_representation` on LFP interfaces

### LINKED ISSUE #2032: BaseLFPExtractorInterface.add_to_nwbfile does not expose always_write_timestamps or data_representation
`BaseRecordingExtractorInterface.add_to_nwbfile` takes `always_write_timestamps` and `data_representation`, and both reach `add_recording_to_nwbfile`. `BaseLFPExtractorInterface.add_to_nwbfile` overrides the method to change the default `parent_container`, but its signature stops at `iterator_options`, so neither option can be passed through an LFP interface:

```python
interface = NeuroScopeLFPInterface(file_path=...)
interface.add_to_nwbfile(nwbfile, always_write_timestamps=True)
# TypeError: add_to_nwbfile() got an unexpected keyword argument 'always_write_timestamps'
```

The same holds for `run_conversion(..., always_write_timestamps=True)` and for a converter's `conversion_options`, since both route through this method, and the conversion-options schema derived from the signature omits the two options as well. An LFP band with an irregular clock, or one whose channels carry heterogeneous offsets, has the same need for these as the raw band does. The override should carry both arguments and forward them.

bendichter:
Closes #2032.

`BaseLFPExtractorInterface.add_to_nwbfile` overrides the recording base's method to make `processing/LFP` the default `parent_container`, but its signature stopped at `iterator_options`. `always_write_timestamps` and `data_representation`, which the parent takes and hands to `add_recording_to_nwbfile`, could therefore not reach the writer through an LFP interface, whether by `add_to_nwbfile`, `run_conversion` or a converter's `conversion_options`, and the options schema derived from the signature left them out. Both are now taken and forwarded. An LFP band with an irregular clock, or with heterogeneous channel offsets, has the same need for them as the raw band.

Nothing tested the LFP base directly, so the new module builds a minimal LFP interface over spikeinterface's generated recording, the way `MockRecordingInterface` is built, and checks the default container, the explicit timestamps, the physical-unit write over heterogeneous offsets, and that the schema names both options. The three option tests fail on `main`. The code and tests were drafted by the AI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Removed the `hdf5` and `sbx` installation extras, aliases of `hdf5imaging` and `scanbox` kept for gallery pages that named the old spellings and marked for removal at the start of 2026. Use `neuroconv[hdf5imaging]` and `neuroconv[scanbox]`.
 
 ## Bug Fixes
+* `BaseLFPExtractorInterface.add_to_nwbfile` now accepts and forwards `always_write_timestamps` and `data_representation`, which the recording base already took, so an LFP interface can write an irregular clock as explicit timestamps or heterogeneous offsets as physical units the same way a raw interface does, through `add_to_nwbfile`, `run_conversion` and a converter's `conversion_options`. [PR #2033](https://github.com/catalystneuro/neuroconv/pull/2033)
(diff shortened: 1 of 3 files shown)


h-mayorquin (approved):
LGTM

PR #2037, opened 2026-09-08
Adopt `interface.alignment` in `ExternalVideoInterface`

h-mayorquin:
@Akseli-Ilmanen asked for a guide to aligning behavior video, and writing it showed `interface.alignment` had no way to place one file without building its times. This PR adopts the surface in `ExternalVideoInterface` with one object per file, adds `alignment[key].start_at(starting_time)` to the component, and ships the how-to.

This is the how-to which is the most important part to review:
https://neuroconv--2037.org.readthedocs.build/en/2037/how_to/align_external_video.html


Akseli-Ilmanen:
Hi @h-mayorquin,

This guide looks amazing so far, I really like the table below "Which recipe applies" it and the linking back to `nwb_trials` at the end!

Below are a few feedback points:

### 1) Legend for figures
The consistent blue/gray colour scheme for video/recording system is very good. But people unfamiliar with `.avi` files, and may not realize that these are video files. Or they may be confused what is meant with `part_01` or `trial_01`. Maybe it would be helpful to use `.mp4` instead, and include something like a legend (e.g. below).  Also, at the very top of the page, you could already colour code this section: Most rigs record behavior video [BLUE] next to something else: extracellular electrophysiology, fiber photometry, optical imaging, an operant box [GREY].
image


### 2) Channel names in `detection_configuration`

```python
from neuroconv.datainterfaces import IntanDigitalInterface, IntanRecordingInterface

recording_interface = IntanRecordingInterface(file_path="session.rhd")
digital_interface = IntanDigitalInterface(
    file_path="session.rhd",
    detection_configuration={
        "DIGITAL-IN-02": [
            {
                "signal_conditioning": {"binarize": "midpoint"},
                "detection": "rising",
                "event_name": "camera_frame",
            }
        ]
    },
)

frame_pulse_times = digital_interface.get_event_times("camera_frame")
```

For the example above, I think a key point is that in the `detection_configuration` dictionary, the key (e.g. "DIGITAL-IN-02") has to correspond to the channel name in the recording system where video frames are recorded. I think this could be made more explicitly in the text. 

The good thing is that when you specify a wrong key, one gets this useful error "ValueError: detection_configuration names 'DIGITAL-IN-02', which is not one of the file's signals: ['DIGITAL-IN-01'].", but still making this more explicit may be helpful.

### 3) `video_interface`

I would call the variable `video_interface` not just `interface`, so it's more distinguishable from `digital_interface`. At least this initially confused me.

```python
from neuroconv.datainterfaces import ExternalVideoInterface
interface = ExternalVideoInterface(file_paths=["session.avi"], video_name="BehaviorCamera")
interface.alignment.keys()
```

### 4) Add optional `trial_onsets` line

So, I tested the combination of `triggered camera` with `a pulse per frame` with our data, and it works great!

I would add one little change: In the cell below, add an optional maybe commented out line to add trial onsets from the first rise of the first frame in the trial.

```python
import numpy as np

frame_pulse_times = digital_interface.get_event_times("camera_frame")

# The gap between trials is far larger than the frame interval, so the split is unambiguous.
frame_interval = np.median(np.diff(frame_pulse_times))
gap_indices = np.flatnonzero(np.diff(frame_pulse_times) > 10 * frame_interval) + 1
bursts = np.split(frame_pulse_times, gap_indices)

segment_keys = interface.alignment.keys()

assert len(bursts) == len(segment_keys), f"{len(bursts)} bursts for {len(segment_keys)} files."
for segment_key, burst in zip(segment_keys, bursts):
    interface.alignment[segment_key].set_times(burst)

# ADDED: Define trial onset as the first frame in the trial
# trial_onsets = np.array([burst[0] for burst in bursts])
```

### 5) Ability to drop bursts from `digital_interface` for individual videos/trials

In the documentation, you mentioned how dropped frames can be resolved via `set_times`. This step is a bit tedious but for the free-running camera case is probably what one has to do, to not lose the entire recording session. But for the `triggered camera` case, if dropped frames only occur in a few videos, it may be easier just to discard the entire video. E.g. for our data, we typically have 100-200 videos, with typically 1-2 videos having dropped frames. In this case, we find it easier just to exclude those few videos. 

From our Arduino video rig, we can already determine which video files had missing frames, and remove those videos. The problem (see error below) is that Intan rig still received the pulses for the frames of the dropped videos.
```python
import numpy as np

frame_pulse_times# The gap between trials is far larger than the frame interval, so the split is unambiguous.
frame_interval = np.median(np.diff(frame_pulse_times))
gap_indices = np.flatnonzero(np.diff(frame_pulse_times) > 10 * frame_interval) + 1
bursts = np.split(frame_pulse_times, gap_indices)

segment_keys = interface.alignment.keys()

assert len(bursts) == len(segment_keys), f"{len(bursts)} bursts for {len(segment_keys)} files."
for segment_key, burst in zip(segment_keys, bursts):
    interface.alignment[segment_key].set_times(burst)
```
OUTPUT:
```python
AssertionError                            Traceback (most recent call last)
Cell In[44], [line 10](vscode-notebook-cell:?execution_count=44&line=10)
      6 bursts = np.split(frame_pulse_times, gap_indices)
      7 
      8 segment_keys = interface.alignment.keys()
      9 
---> [10](vscode-notebook-cell:?execution_count=44&line=10) assert len(bursts) == len(segment_keys), f"{len(bursts)} bursts for {len(segment_keys)} files."
     11 for segment_key, burst in zip(segment_keys, bursts):
     12     interface.alignment[segment_key].set_times(burst)

AssertionError: 144 bursts for 138 files.
```

 So in this case, we would want to remove a few bursts from `digital_interface` matching the indices where we removed the video files. It would be great if there was a function where we can specify which indices in `video_interface.alignment.keys()` are missing (i.e. have been removed) and then one can remove the matching idxs in `bursts`. The solution could look something like this:

```python

# From your Arduino log: 0-based indices, into the full 144-trial sequence,
# of trials whose video never reached disk.
dropped_trial_indices = {12, 47, ...}  # whatever your Arduino rig reports

kept_bursts = [burst for i, burst in enumerate(bursts) if i not in dropped_trial_indices]

segment_keys = video_interface.alignment.keys()  # 138 — only the files that exist

assert len(kept_bursts) == len(segment_keys), f"{len(kept_bursts)} bursts for {len(segment_keys)} files."
for segment_key, burst in zip(segment_keys, kept_bursts):
    video_interface.alignment[segment_key].set_times(burst)
```

### 6) `set_times` accepts a burst whose length doesn't match the video's frame count

Testing point 5 further, I found `set_times` silently accepts a burst that's the wrong length for its video — no error, whether the mismatch is a truncated array (`burst[:-1]`) or a shifted one:

```python
burst = burst[:-1]  # one fewer time than the video has frames
video_interface.alignment[segment_key].set_times(burst)  # no error
```

The how-to says "the interface throws an error when the number of times does not match the number of frames," but that doesn't seem to fire here — [either it's not checked at `set_times`, or it's deferred to `create_nwbfile`/write time — I haven't isolated which]. If it's deferred, our preference would be to check it eagerly at `set_times`, since with 138 segments in a loop it's much easier to debug a failure with the offending `segment_key` still in scope than a generic mismatch after the whole loop completes.

For our use case (drops only mean whole videos are missing — see point 5 — never a partial drop within a kept video), we'd like this to be opt-in and strict:

```python
video_interface.alignment[segment_key].set_times(burst, validate_frame_count=True)
```

Fine for this to be slower, since it has to read the actual frame count off the video file to check.


I hope this input is helpful!

Best
Akseli

h-mayorquin:
Hi, @Akseli-Ilmanen
Thanks a bunch for testing this with your data, this is very useful!

1) The figures now use `.mp4` names and carry a legend. I didn't color the opening sentence as the legend already does that job.
2) Added a sentence saying the key is the name of the input the camera is plugged into.
3) Done, it is `video_interface` everywhere now.
4) Added `trial_onsets = np.array([burst[0] for burst in bursts])` after the loop. The trials table at the end uses it.
5) I didn't add a function for this as dropping bursts by position is a one-line list comprehension. The guide now shows that line for a case like your 144 bursts and 138 files.
6) You were right. The check only ran on the total across all the files. It now runs per file and the error names the file. I kept it at write time instead of in `set_times` as `set_times` is the alignment API shared by every interface.

Akseli-Ilmanen:
Nice work!

